PR #286, opened 2026-04-17
fix(planner): use plan_approved_at timestamp instead of file mtime for freshness check (#281)

### LINKED ISSUE #281: Planner stage 8 should touch docs/plan/plan.md to refresh mtime for plan-presence gate
Surfaced by #273 agent: the plan-presence gate checks `plan.md mtime >= latest non-planning task created_at`. Plan.md is written at stage 6, backlog emitted at stage 8. Under strict reading, the plan always looks stale when stage 8 emits its tasks.

Fix: at stage 8 `emit_backlog`, explicitly `Path('docs/plan/plan.md').touch()` after emitting backlog tasks (or update the architect flow node handler). Alternative: track `plan_approved_at` as a separate timestamp on the plan_project task, and change the gate predicate to use that instead of mtime.

I'd vote for option 2 (plan_approved_at column on the task) — mtime is fragile (git touches, editor saves, etc.). File-based mtime works for a first pass but the explicit timestamp is more robust.

Priority: High — without this, the plan-presence gate (#273) will incorrectly block every project whose plan was written before backlog emission.

Agent workaround in tests: `plan.md.touch(times=(time.time()+1, time.time()+1))`.

samhotchkiss:
Fixes the plan-presence gate's fragile mtime-based staleness check. Mtime rotted because the planner wrote plan.md at stage 6 but emitted backlog tasks at stage 8 — the gate always flagged fresh plans as stale.

## Fix
- `approve()` writes `work_context_entries` row with `entry_type='plan_approved'` when user_approval approves a plan_project task. Same transaction — atomic.
- `has_acceptable_plan()` uses this timestamp. Fallback: user_approval execution's completed_at for pre-fix projects.
- Mtime read removed entirely.

## Tests (+5, 0 regressions)
- Fresh approval → gate accepts
- Not yet approved → rejects
- Stale approval (pre backlog creation) → rejects
- Pre-fix project → fallback works
- File mtime doesn't affect decision (invariant pin)

2276 passing.

Closes #281

### DIFF
--- a/src/pollypm/plugins_builtin/project_planning/plan_presence.py
+++ b/src/pollypm/plugins_builtin/project_planning/plan_presence.py
@@ -1,4 +1,4 @@
-"""Plan-presence gate — issue #273.
+"""Plan-presence gate — issues #273, #281.
 
 The sweeper refuses to delegate implementation tasks for a project
 until that project has a non-trivial, user-approved plan on disk.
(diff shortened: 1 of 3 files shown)


samhotchkiss:
Superseded by #287 — clean gate-only fix (sqlite_service.py write path already landed via #283).

PR #523, opened 2026-04-21
fix: harden transcript ingest against non-object json lines

### LINKED ISSUE #412: bug-001 — Transcript ingest crashes on non-dict JSON lines

Source: `issues/Claude/bug-001-transcript-jsonl-type-guard.md`

**Severity:** high
**Category:** bug — unhandled type

## Location

`src/pollypm/transcript_ingest.py:547-560` (scan), `:215-217` (Claude normalizer),
`:353` (Codex normalizer)

## What

`json.loads(line)` is accepted as-is and passed to normalizers that call
`.get(...)` without a dict check. Valid JSON like `[]`, `"string"`, or `42`
on one line crashes the ingest for the whole file.

## Scenario

A vendor changes transcript format or writes a stray array/scalar, and
ingestion halts silently — user notices only when the dashboard goes stale.
The recent `fix(transcript_ingest)` commit suggests this path is fragile.

## Fix

After `json.loads`, guard: `if not isinstance(obj, dict): continue`. Log
skipped lines at debug level so we can tell if a vendor changed shape.

samhotchkiss:
## Summary
- skip valid JSON lines that are not objects before provider normalization
- log skipped scalar/array transcript lines at debug level for shape drift visibility
- add a regression test covering Claude and Codex transcript files with stray JSON values

## Testing
- uv run --extra dev pytest -q tests/test_transcript_ingest.py

Closes #412

### DIFF
--- a/src/pollypm/transcript_ingest.py
+++ b/src/pollypm/transcript_ingest.py
@@ -548,6 +548,17 @@ def _scan_source(config, account_name: str, account: AccountConfig, source: Tran
                     obj = json.loads(line)
                 except json.JSONDecodeError:
                     continue
+                if not isinstance(obj, dict):
+                    logger.debug(
+                        "Skipping non-object transcript line",
+                        extra={
+                            "source_path": str(path),
+                            "source_offset": start_offset,
+                            "json_type": type(obj).__name__,
+                            "provider": account.provider.value,
+                        },
+                    )
+                    continue
                 if account.provider is ProviderKind.CLAUDE:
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches bug-001 fix — guards `isinstance(obj, dict)` after `json.loads`, debug-logs with structured extras (source_path/offset/provider). Test covers array and scalar JSON, verifies valid events still ingest, counts log messages. 5/5 tests pass.

PR #524, opened 2026-04-21
fix: fail worker provisioning when tmux pane lookup is empty

### LINKED ISSUE #413: bug-002 — `pane_id` falls back to sentinel `"%0"` when window lookup fails

Source: `issues/Claude/bug-002-pane-id-sentinel-percent-zero.md`

**Severity:** high
**Category:** bug — silent state corruption

## Location

`src/pollypm/work/session_manager.py:443-459`

## What

After `create_session` / `create_window`, the code looks up the new window
in `list_windows()`. If found, uses its `pane_id`. If not found but others
exist, uses `windows[0].pane_id`. If the list is empty, returns `"%0"`.

## Scenario

A tmux race or brief failure returns an empty window list. `"%0"` is written
into the DB as the worker's pane. Subsequent `send-keys` / alive checks fire
into the void; worker appears provisioned but is a phantom.

## Fix

Raise `ProvisionError` if neither the named window nor any window is
returned; let the caller retry. Do not return a sentinel — a phantom worker
is worse than a loud failure.

samhotchkiss:
## Summary
- remove the phantom `%0` fallback when tmux reports no panes for a freshly created worker window
- raise a ProvisionError instead so claim/provision can retry loudly instead of persisting bad state
- add a regression test for the empty-window-list case

## Testing
- uv run --extra dev pytest -q tests/test_session_manager.py

Closes #413

### DIFF
--- a/src/pollypm/work/session_manager.py
+++ b/src/pollypm/work/session_manager.py
@@ -440,23 +440,39 @@ class SessionManager:
                         f"Check that tmux is installed and running, then "
                         f"run `pm task release <id>` and retry."
                     ) from exc
-            windows = self._tmux.list_windows(session_name)
-            pane_id = "%0"
-            for w in windows:
-                if w.name == window_name:
-                    pane_id = w.pane_id
-                    break
-            if pane_id == "%0" and windows:
-                pane_id = windows[0].pane_id
+            pane_id = self._resolve_worker_pane_id(
+                session_name=session_name,
+                window_name=window_name,
+                windows=self._tmux.list_windows(session_name),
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches bug-002 fix — drops the `%0` sentinel, raises `ProvisionError` with retry guidance. Extracts duplicate lookup into a helper (`_resolve_worker_pane_id`), and prefers the first window with a valid `pane_id` rather than falling through silently. Test asserts empty-window-list raises. 34/34 tests pass.

PR #525, opened 2026-04-21
fix: validate plugin names at the host boundary

### LINKED ISSUE #494: sec-004 — `plugin_host.content_paths()` does not validate `plugin_name`

Source: `issues/Claude/sec-004-plugin-name-path-traversal.md`

**Severity:** medium
**Category:** security — path traversal

## Location

`src/pollypm/plugin_host.py:186-235` (specifically the path construction at ~L231)

## What

Path constructed as `self.root_dir / ".pollypm" / "content" / plugin_name / kind`
without validating `plugin_name`. A plugin name containing `..` or an absolute
path escapes the content directory.

## Risk

Plugin names come from manifests; the trust model currently assumes the user
has vetted every plugin they installed. For a v1 that wants to support
third-party plugins, this is a latent escape. Low severity today, rising as
the plugin ecosystem grows.

## Fix

Validate plugin names at the manifest-load boundary with a strict regex
(`^[a-z][a-z0-9_-]*$`). Reject and refuse to load rather than sanitize —
ambiguous names are a smell in a plugin registry.

samhotchkiss:
## Summary
- reject invalid plugin names with a strict host-level regex
- enforce the same validation when resolving plugin content paths so traversal-like names cannot escape content roots
- add focused regressions for invalid manifest names and invalid `content_paths()` input

## Testing
- uv run --extra dev pytest -q tests/test_plugins.py

Closes #494

### DIFF
--- a/src/pollypm/plugin_host.py
+++ b/src/pollypm/plugin_host.py
@@ -19,6 +19,7 @@ Contract:
 from __future__ import annotations
 
 import logging
+import re
 from dataclasses import dataclass, field
 from functools import lru_cache
 from importlib.util import module_from_spec, spec_from_file_location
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches sec-004 fix — strict `^[a-z][a-z0-9_-]*$` regex applied at two boundaries: manifest read and `content_paths()`. Rejects traversal (`../escape`) and case violations (`BadName`). Tests cover both entry points. 37/37 tests pass.

PR #527, opened 2026-04-21
fix: handle stale session lock re-create races

### LINKED ISSUE #414: bug-003 — Session lock race: unhandled `FileExistsError` after stale unlink

Source: `issues/Claude/bug-003-session-lock-race-after-stale-unlink.md`

**Severity:** high
**Category:** bug — concurrency

## Location

`src/pollypm/projects.py:164-176`

## What

On stale-lock detection, `unlink(missing_ok=True)` then a retry of
`os.open(..., O_EXCL)`. Between unlink and retry, another process can create
the lock first; the second `os.open` raises uncaught `FileExistsError`.

## Scenario

Two cockpits + a CLI claim the same task simultaneously. One wins. Another
detects stale, unlinks, a third creates fresh. The retry crashes the second
process with an unhandled exception instead of waiting or re-checking.

## Fix

Wrap the retry in try/except. On `FileExistsError` re-read the lock and
either back off (if the new holder is valid) or fail with a clear "lock held
by <pid>" error. Consider replacing with `fcntl.flock` or SQLite-based lock
to avoid the whole unlink dance.

samhotchkiss:
## Summary
- re-read the session lock after stale-lock unlink races instead of leaking FileExistsError
- keep same-session idempotence intact if another process recreated the same lock
- add a regression test for the stale-unlink/new-owner interleaving

## Testing
- uv run --extra dev pytest -q tests/test_projects.py

Closes #414

### DIFF
--- a/src/pollypm/projects.py
+++ b/src/pollypm/projects.py
@@ -147,6 +147,14 @@ def session_lock_path(base_dir: Path, session_id: str) -> Path:
 _LOCK_STALE_SECONDS = 1800  # 30 minutes — locks older than this are considered stale
 
 
+def _read_session_lock_payload(lock_path: Path) -> dict[str, object]:
+    try:
+        payload = json.loads(lock_path.read_text())
+    except Exception:  # noqa: BLE001
+        return {}
+    return payload if isinstance(payload, dict) else {}
+
+
 def ensure_session_lock(base_dir: Path, session_id: str) -> Path:
     base_dir.mkdir(parents=True, exist_ok=True)
     lock_path = session_lock_path(base_dir, session_id)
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches bug-003 fix — catches the second `FileExistsError` after stale unlink and either returns idempotently (if lock is ours) or raises with "owned by <sid>". DRY-extracts payload reader. Chose fail-loud over retry-with-backoff, which is the safer call for a rare race. 14/14 tests pass.

Minor nit (not blocking): the idempotent recovery branch (race-winner is us) isn't covered by the new test — only the new-owner-wins path is. Worth adding if you're back in this file.

PR #530, opened 2026-04-21
fix: surface infrastructure errors creating session locks

### LINKED ISSUE #417: bug-006 — Session lock creation does not distinguish `OSError` from `FileExistsError`

Source: `issues/Claude/bug-006-session-lock-osError-disk-full.md`

**Severity:** medium
**Category:** bug — unhandled failure

## Location

`src/pollypm/projects.py:173`

## What

`os.open(..., O_EXCL | O_CREAT)` can raise `OSError` (ENOSPC, EACCES) as well
as `FileExistsError`. Only the latter is anticipated. Disk-full or
permission-denied cases raise a raw traceback.

## Fix

Separate handlers for `FileExistsError` (race) vs `OSError` (infrastructure).
For infrastructure errors, surface a user-facing message; do not retry.

samhotchkiss:
## Summary
- distinguish session-lock race handling from infrastructure failures like disk-full or permission errors
- convert raw `OSError` lock-creation failures into user-facing runtime errors
- add a regression test covering the infrastructure-error path

## Testing
- uv run --extra dev pytest -q tests/test_projects.py

Closes #417

### DIFF
--- a/src/pollypm/projects.py
+++ b/src/pollypm/projects.py
@@ -185,6 +185,14 @@ def ensure_session_lock(base_dir: Path, session_id: str) -> Path:
             raise RuntimeError(
                 f"Session lock conflict at {base_dir}: owned by {existing_session or 'unknown'}"
             ) from exc
+        except OSError as exc:
+            raise RuntimeError(
+                f"Could not create session lock at {base_dir}: {exc}"
+            ) from exc
+    except OSError as exc:
+        raise RuntimeError(
+            f"Could not create session lock at {base_dir}: {exc}"
+        ) from exc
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches bug-006 fix — wraps raw `OSError` (disk full, EACCES) into `RuntimeError("Could not create session lock at ...")`. Handlers placed both around the first `os.open` and the retry `os.open`. Note `FileExistsError` subclasses `OSError` — the more-specific handler stays before the generic one, so race logic still wins. Test covers the first-attempt path. 15/15 tests pass.

PR #534, opened 2026-04-21
fix: harden backup restore sidecar cleanup

### LINKED ISSUE #415: bug-004 — Backup restore silently ignores WAL/SHM sidecar unlink failures

Source: `issues/Claude/bug-004-backup-restore-wal-sidecars.md`

**Severity:** medium
**Category:** bug — data-loss risk

## Location

`src/pollypm/backup.py:446-454`

## What

After `shutil.copy2(staged_db, live_db)`, the WAL and SHM sidecars are removed
with `missing_ok=True` and no error handling. If unlink fails (permissions,
filesystem error), SQLite re-opens the restored DB with stale WAL/SHM and
replays phantom transactions on top of the restore.

## Scenario

User restores a backup after a crash. Sidecar unlink fails. State looks
"restored" but actually contains a mix of restored + stale WAL. Hard to
diagnose; easy to destroy confidence in the backup system.

## Fix

- Use `os.replace()` for the DB swap to guarantee atomicity.
- Unlink sidecars with explicit error logging and abort if unlink fails.
- Add a restore-round-trip test that verifies the post-restore DB matches
  the backup snapshot byte-for-byte (schema + row hash).

samhotchkiss:
## Summary
- stage restores on the live DB filesystem and use `os.replace` for the final swap
- abort loudly when stale SQLite `-wal` / `-shm` cleanup fails instead of silently risking a mixed restore
- add regressions for byte-accurate restore and sidecar-cleanup abort behavior

## Testing
- uv run --extra dev pytest -q tests/test_backup_restore.py

Closes #415

### DIFF
--- a/src/pollypm/backup.py
+++ b/src/pollypm/backup.py
@@ -24,6 +24,7 @@ Design notes:
 from __future__ import annotations
 
 import gzip
+import os
 import shutil
 import sqlite3
 import tarfile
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches bug-004 fix:
- Temp dir now created in `live_db.parent` so `os.replace` is atomic on the same filesystem.
- Sidecar cleanup moved *before* the DB swap so a failed unlink aborts with the live DB intact (no mixed-restore state).
- `_remove_sqlite_sidecars` raises `RuntimeError('Restore aborted')` on unlink failure instead of `pass`.
- Test 1: byte-exact restore + sidecar clearance. Test 2: WAL-unlink failure proves `os.replace` is never called and pre-restore mutated data survives — perfect invariant check.
16/16 tests pass.

PR #542, opened 2026-04-21
fix(backup): harden full archive inputs

### LINKED ISSUE #493: sec-003 — `backup.py` tarfile add/extract does not guard against symlink escape

Source: `issues/Claude/sec-003-backup-tarfile-symlink-escape.md`

**Severity:** medium
**Category:** security — path traversal via archive

## Location

`src/pollypm/backup.py:288-303` (create), `src/pollypm/backup.py:426-432` (restore)

## What

`tarfile.add(...)` is called without the `filter=` argument and without
resolving symlinks. Restore likewise extracts without a `filter=` (Python 3.12+
deprecates the default).

## Risk

If `.pollypm/` contains a symlink (placed by a plugin, a prior restore, or
accidentally by the user), backup captures the link target, or restore
clobbers files outside the intended directory. The "Zip Slip" class of bug.

## Fix

- On `add`: resolve each path and assert `resolved.is_relative_to(base_dir.resolve())`.
- On `extract`: pass `filter='data'` (Python 3.12+) or implement the equivalent
  manual filter that rejects absolute paths, `..` components, and symlinks.
- Add a backup+restore round-trip test that asserts a symlink in `.pollypm/`
  does not escape.

samhotchkiss:
Closes #493

## Summary
- reject symlink and special-file inputs while building full backups
- require a single regular `state.db` payload when planning or executing tar restores
- add focused regressions for symlinked archive inputs

### DIFF
--- a/src/pollypm/backup.py
+++ b/src/pollypm/backup.py
@@ -232,6 +232,121 @@ def _remove_sqlite_sidecars(db_path: Path) -> None:
             ) from exc
 
 
+def _path_is_within(path: Path, root: Path) -> bool:
+    try:
+        path.relative_to(root)
+        return True
+    except ValueError:
+        return False
+
+
+def _add_path_to_archive(
+    tar: tarfile.TarFile,
+    source: Path,
+    *,
+    arcname: str,
+    allowed_root: Path,
+) -> None:
+    """Add ``source`` to ``tar`` while refusing symlinks and special files."""
+    if source.is_symlink():
+        raise ValueError(f"refusing to include symlink in full backup: {source}")
+
+    try:
+        resolved = source.resolve(strict=True)
+    except OSError as exc:
+        raise ValueError(
+            f"refusing to include unreadable path in full backup: {source}: {exc}"
+        ) from exc
+
+    if not _path_is_within(resolved, allowed_root):
+        raise ValueError(
+            f"refusing to include path outside archive root {allowed_root}: {source}"
+        )
+
+    if source.is_dir():
+        tar.add(source, arcname=arcname, recursive=False)
+        for child in sorted(source.iterdir()):
+            _add_path_to_archive(
+                tar,
+                child,
+                arcname=f"{arcname}/{child.name}",
+                allowed_root=allowed_root,
+            )
+        return
+
+    if source.is_file():
+        tar.add(source, arcname=arcname, recursive=False)
+        return
+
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches sec-003 fix, and stricter than what I suggested in the audit:
- **Create side**: `_add_path_to_archive` walks the tree manually, refuses symlinks at every step, resolves with `strict=True`, and enforces `allowed_root` containment. Cleaner than the `filter='data'` approach because it fails closed on any symlink rather than silently filtering.
- **Restore side**: `_require_regular_state_db_member` + `_extract_valid_state_db_from_archive` verify the `state.db` entry is (a) unique, (b) a regular file, and (c) opens as a valid SQLite DB before the restore proceeds. The old `tar.getmember('state.db')` would have quietly returned a symlink TarInfo.
- **Cleanup**: partial archive is unlinked on exception during create. No more orphaned `.tar.gz` after a failed backup.

Tests cover both attack vectors: symlink in the tree during backup creation (asserts ValueError + no archive produced) and symlinked state.db in a crafted archive during restore (asserts 'regular file' rejection).

14/14 tests pass.

samhotchkiss:
Review is clean (see prior comment above — matches sec-003, tests pass), but the branch is now CONFLICTING with main because PR #534 (bug-004 sidecar hardening) landed in the same region of `src/pollypm/backup.py`. Expected rebase conflict zones:

- `execute_restore` — main now uses `os.replace` with sidecar-cleanup-before-swap (from #534); this PR's restore changes call `_extract_valid_state_db_from_archive` earlier in that same function.
- `plan_restore` — your new `_extract_valid_state_db_from_archive` probe supersedes the old inline `getnames()` check.

Rebase on main and keep:
- main's atomic `os.replace` + sidecar-abort-on-failure flow
- this PR's archive-member validation on both `plan_restore` and `execute_restore`

Flipping the label to `needs fix`.

samhotchkiss:
Rebase is clean — post-`cc4bbd5` main now has both #534's `os.replace` atomic swap / sidecar-abort flow AND this PR's archive-member validation on `plan_restore` and `execute_restore`. 18/18 tests pass (expanded from 14 — it picked up the #534 tests that are now on main). Re-confirming sec-003 fix: symlinks refused at archive create, state.db member validated as regular-file-only on restore, partial-archive rollback on exception. Merging.

PR #546, opened 2026-04-21
fix(memory): disclose truncated prompt recall

### LINKED ISSUE #483: prompt-013 — Silent truncation of injected memory entries

Source: `issues/Claude/prompt-013-memory-truncation-silent.md`

**Severity:** low
**Surface:** `src/pollypm/memory_prompts.py:34-40` (BUDGET_TOKENS = 4096)
**Criterion:** sufficiency

## What

When the memory injection exceeds the budget, lowest-scored entries are
dropped with no note to the agent.

## Fix

Append a one-line marker:

> _[Memory truncated: N of M entries shown. Ask for more with `pm memory
> recall`.]_

Lets the agent know it's operating on a subset and gives it a way to ask.

samhotchkiss:
Closes #483

## Summary
- append an explicit truncation note when memory injection drops entries for budget
- point agents at `pm memory recall` when they need more context
- add focused coverage for the truncated-output path

### DIFF
--- a/src/pollypm/memory_prompts.py
+++ b/src/pollypm/memory_prompts.py
@@ -189,9 +189,17 @@ def _render_injection(
     # a simple loop — the worst case is O(limit^2) entry renders, but
     # limit is small (default 15) so it's negligible.
     ordered = _flatten_for_budget(by_type)
+    total_entries = len(ordered)
 
     while True:
         rendered = _render_by_type(by_type)
+        shown_entries = sum(len(entries) for entries in by_type.values())
+        if shown_entries < total_entries:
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches prompt-13 fix (closes #483) with the exact marker I suggested: "_[Memory truncated: N of M entries shown. Ask for more with `pm memory recall`.]_". `_append_truncation_notice` only fires when `shown_entries < total_entries`, so non-truncated injections stay clean. Budget-aware — the notice itself is included in the budget loop so we don't overshoot after appending. 16/16 tests pass.

PR #548, opened 2026-04-21
perf(work): collapse next-task blocker queries

### LINKED ISSUE #457: perf-001 — `next_task()` runs N queries per queued task to check blockers

Source: `issues/Claude/perf-001-next-task-n-plus-one-blockers.md`

**Severity:** high
**Category:** performance — N+1 queries

## Location

`src/pollypm/work/service_queries.py:255-262` and `_has_unresolved_blockers`

## What

Iterates all queued tasks; for each row, calls `_has_unresolved_blockers(task_id)`
which issues **two** further queries (fetch blockers, then per-blocker status).
Also hydrates each row via `_row_to_task` before filtering.

## Cost

100 queued tasks → ~300+ queries per `pm next` invocation. This runs on every
task claim and on some dashboard refreshes.

## Fix

One query: `SELECT from_project, from_task_number FROM work_task_dependencies
WHERE kind='blocks' AND to_work_status NOT IN ('done','cancelled')` — then
check blocker-presence in a Python set. For the v1.x window, a denormalized
`has_unresolved_blockers` flag on `work_tasks` updated by transitions would
reduce this to a single indexed query.

samhotchkiss:
Closes #457

## Summary
- precompute queued tasks blocked by unresolved dependencies in one joined query
- cheap-filter queued rows by blocker set and worker role before hydration
- add regression tests to prevent per-task blocker lookups and extra hydration

## Testing
- HOME=/tmp/pytest-457 uv run --extra dev pytest tests/test_work_queries.py -q

### DIFF
--- a/src/pollypm/work/service_queries.py
+++ b/src/pollypm/work/service_queries.py
@@ -227,12 +227,42 @@ def update_task(service: "SQLiteWorkService", task_id: str, **fields: object) ->
     return task
 
 
+def _unresolved_blocked_task_keys(
+    service: "SQLiteWorkService",
+    *,
+    project: str | None = None,
+) -> set[tuple[str, int]]:
+    clauses = [
+        "d.kind = ?",
+        "t.work_status NOT IN (?, ?)",
+    ]
+    params: list[object] = [
+        "blocks",
+        WorkStatus.DONE.value,
+        WorkStatus.CANCELLED.value,
+    ]
+    if project is not None:
+        clauses.append("d.to_project = ?")
+        params.append(project)
+    rows = service._conn.execute(
+        "SELECT DISTINCT d.to_project, d.to_task_number "
+        "FROM work_task_dependencies d "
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches perf-001 fix (closes #457) — collapses the N+1 blocker loop into a single JOIN query (`_unresolved_blocked_task_keys`) that returns all blocked `(project, task_number)` tuples, then cheap-filters queued rows against the set in Python. Also reads `roles` off the raw row JSON during the agent-filter step so `_row_to_task` hydration is deferred until after we've picked the winner.

Tests prove both invariants:
- `test_next_does_not_call_per_task_blocker_lookup` monkey-patches `_has_unresolved_blockers` to raise and confirms `next()` never reaches it.
- `test_next_hydrates_only_selected_matching_task` counts `_row_to_task` calls and asserts exactly one hydration for the winner (vs. N in the old code).

14/14 tests pass.

PR #549, opened 2026-04-21
fix(heartbeat): guard mutating role writes

### LINKED ISSUE #491: sec-001 — Heartbeat role calls write APIs (send_session_message, recover_session)

Source: `issues/Claude/sec-001-heartbeat-role-writes.md`

**Severity:** high
**Category:** security — permission model

## Location

- `src/pollypm/heartbeats/local.py:571, 784, 858` — `api.send_session_message(...)`
- `src/pollypm/heartbeats/local.py:315, 336, 765` — `api.recover_session(...)` and `api.set_session_status(...)` (line 314)

## What

`docs/architecture.md:20` states: *"Permission-gated execution system preventing
test/code operations in Heartbeat role."* The Heartbeat role is documented as
read-only (Bash only). But `LocalHeartbeatBackend.run()` calls
`send_session_message()`, `recover_session()`, and `set_session_status()` —
all side-effecting writes — without any role check.

## Risk

A logic bug or compromised classifier in the heartbeat path can inject prompts
into worker sessions or restart sessions unexpectedly. This is both a real
correctness risk (runaway recovery loops) and a violation of the stated
permission model, so the code and docs have drifted apart.

## Fix

Decide which way to reconcile:
1. Promote heartbeat role to "supervisor" (can write) in the docs; or
2. Split: heartbeat *observes* and emits events; a separate Operator handler
   *acts* on those events. The second is what the architecture implies.

Until resolved, add an explicit role-check assert before each write call so
violations are loud rather than silent.

samhotchkiss:
Closes #491

## Summary
- make the heartbeat backend assert allowed roles before mutating session state
- keep worker follow-up messaging guarded separately from broader supervisor actions
- add focused plugin API coverage for the allowed and rejected role paths

## Testing
- uv run --extra dev python -m pytest tests/test_heartbeat_plugin_api.py

### DIFF
--- a/src/pollypm/heartbeats/local.py
+++ b/src/pollypm/heartbeats/local.py
@@ -273,6 +273,18 @@ def _select_intervention(
 class LocalHeartbeatBackend(HeartbeatBackend):
     name = "local"
     _UNMANAGED_WINDOW_ALERT_PREFIX = "unmanaged_window:"
+    _MUTATING_SESSION_ROLES = frozenset(
+        {
+            "architect",
+            "heartbeat-supervisor",
+            "operator-pm",
+            "review",
+            "reviewer",
+            "triage",
+            "worker",
+        }
+    )
+    _WORKER_MUTATING_SESSION_ROLES = frozenset({"worker"})
 
     _AUTH_FAILURE_PATTERNS = (
         "authentication failure",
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches sec-001 fix (closes #491) — every heartbeat write (`set_session_status`, `recover_session`, `send_session_message`) now routes through a guarded helper that asserts `context.role` is in an allow-list before calling the API. Two role sets:
- `_MUTATING_SESSION_ROLES` for status/recovery (architect, heartbeat-supervisor, operator-pm, review/reviewer, triage, worker).
- `_WORKER_MUTATING_SESSION_ROLES` tightly scoped to worker for direct message injection (so heartbeat can't inject prompts into operator/reviewer panes).

All existing callsites within `_process_session`, `_apply_prompt_pm_task_next`, the stall-nudge circuit breaker, and `_triage_stalled_worker` migrated to the guarded helpers. Tests assert `AssertionError` fires on unknown roles for both the status-write and message-inject paths.

Reconciles the code with docs/architecture.md L20 — the heartbeat is still allowed to write, but the boundary is now explicit and tested rather than implicit. 16/16 tests pass.

PR #547, opened 2026-04-21
perf(work): add hot-path task indexes

samhotchkiss:
Closes #461

## Summary
- add the missing active-task and dependency-from indexes to the work schema
- record the new work schema head so legacy DBs stop reading as behind after bootstrap
- cover the fresh-schema and legacy-upgrade paths in schema tests

## Testing
- HOME=/tmp/pytest-461 uv run --extra dev pytest tests/test_work_schema.py tests/test_doctor.py -q

PR #554, opened 2026-04-21
fix(work): filter my_tasks by assignee

### LINKED ISSUE #458: perf-002 — `my_tasks()` fully hydrates every active task before filtering

Source: `issues/Claude/perf-002-my-tasks-full-hydration.md`

**Severity:** high
**Category:** performance — O(N) over-hydration

## Location

`src/pollypm/work/service_queries.py:266-275`

## What

`SELECT * FROM work_tasks WHERE current_node_id IS NOT NULL` returns every
active task. Each row passes through `_row_to_task` → 4 additional queries
(transitions, executions, relationships, token sums) → `derive_owner()` which
loads the flow template per row.

## Cost

1,000 active tasks × ~5 queries = 5,000 queries to return 1 task for the
caller. Happens on any dashboard render that shows "my tasks."

## Fix

- Add an `assignee` / `owner` column populated at transition time so the
  initial `SELECT` can filter with `WHERE assignee = ?`.
- Short of that: memoize the flow template; bulk-load relationships for the
  full candidate set in one query; only hydrate tasks that pass the owner
  filter.

samhotchkiss:
Closes #458

## Summary
- keep assignee ownership aligned when tasks hand off between worker and reviewer nodes
- let my_tasks query active rows by assignee instead of hydrating every active task
- add regression coverage for handoff, reject-loop reassignment, and pre-hydration filtering

## Testing
- uv run --extra dev python -m pytest tests/test_work_queries.py tests/test_flow_progression.py

### DIFF
--- a/src/pollypm/work/service_queries.py
+++ b/src/pollypm/work/service_queries.py
@@ -265,14 +265,12 @@ def next_task(
 
 def my_tasks(service: "SQLiteWorkService", agent: str) -> list[Task]:
     rows = service._conn.execute(
-        "SELECT * FROM work_tasks WHERE current_node_id IS NOT NULL",
+        "SELECT * FROM work_tasks "
+        "WHERE current_node_id IS NOT NULL AND assignee = ? "
+        "ORDER BY project, task_number",
+        (agent,),
     ).fetchall()
-    result: list[Task] = []
-    for row in rows:
-        task = service._row_to_task(row)
-        if service.derive_owner(task) == agent:
-            result.append(task)
-    return result
+    return [service._row_to_task(row) for row in rows]
 
 
 def state_counts(service: "SQLiteWorkService", project: str | None = None) -> dict[str, int]:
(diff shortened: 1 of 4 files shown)


samhotchkiss:
LGTM. Matches perf-002 fix (closes #458) — `my_tasks` now filters `WHERE current_node_id IS NOT NULL AND assignee = ?` at the DB level, backed by the existing `idx_work_tasks_assignee` partial index (from #547's predecessor), which means zero over-hydration.

The interesting part: this depended on the `assignee` column being kept fresh across lifecycle events, so the PR also adds `_resolve_node_assignee()` and wires it into all three known transition points — `claim`, `advance_to_node`, `reject`. `derive_owner` delegates to the same helper, so stored `assignee` stays consistent with the flow-derived owner.

Tests cover the invariants exactly:
- worker → reviewer handoff (node_done)
- reviewer → worker handback (reject)
- pre-hydration filter behavior (monkeypatches `derive_owner` to throw and `_row_to_task` to count; asserts only the matching row hydrates)

One subtle risk to watch: if any other code path updates `current_node_id` without going through these three entry points, `assignee` could drift. The three paths covered are the known ones; any others would be pre-existing bugs the filter surfaces rather than causes.

35/35 tests pass.

PR #555, opened 2026-04-21
perf(transcripts): reduce ingest scan churn

### LINKED ISSUE #463: perf-007 — Transcript ingest `rglob` + `stat()` on every tick

Source: `issues/Claude/perf-007-transcript-ingest-stat-spam.md`

**Severity:** medium (high if many archived transcripts)
**Category:** performance — disk I/O per heartbeat

## Location

`src/pollypm/transcript_ingest.py:525-539`

## What

Every tick `_scan_source()` `rglob`s the transcript root and `.stat()`s every
file. Code comment at line 35 already notes "CPU pinning at 170%."
Per-file mtime cache prevents re-parse but not the stat syscall itself.

## Cost

10,000 archived transcripts → 10,000 `stat()` per tick (1s). Linear with
history size. Recent commit `e14f800 fix(transcript_ingest): bump handler
timeout 30s→10m` suggests this is an active pain point.

## Fix

- Track the root directory's `mtime` and skip `rglob` unless it has changed
  (won't catch in-place writes to existing files, but new files bump mtime).
- Or: maintain an in-memory watched-set that only stats files known to be
  currently being written (today's session).
- Consider `watchdog`/fsevents for a real fix.

samhotchkiss:
Closes #463

## Summary
- cache known transcript files per source instead of full rglob+stat on every poll
- watch directory mtimes plus a hot-file window for live append detection, with periodic full rescans
- add regression coverage for archived-file skipping, live appends, and rotated file pickup

## Testing
- uv run --with pytest python -m pytest tests/test_transcript_ingest.py -q

### DIFF
--- a/src/pollypm/transcript_ingest.py
+++ b/src/pollypm/transcript_ingest.py
@@ -18,8 +18,12 @@ from pollypm.projects import project_transcripts_dir, session_scoped_dir
 
 POLL_INTERVAL_SECONDS = 1.0
 MAX_BACKOFF_SECONDS = 30.0
+FULL_RESCAN_SECONDS = 60.0
+HOT_SCAN_WINDOW_SECONDS = 60.0 * 60.0
+HOT_SCAN_FILE_LIMIT = 32
 _INGESTORS: dict[Path, "TranscriptIngestor"] = {}
 _INGESTORS_LOCK = threading.Lock()
+_SOURCE_SCAN_CACHE: dict[tuple[str, str], "TranscriptSourceScanCache"] = {}
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches perf-007 fix (closes #463) — stronger than what I recommended in the audit:
- `TranscriptSourceScanCache` tracks `known_files`, `dir_mtimes` (every watched dir, not just the root), and `last_full_scan_at`.
- `_scan_paths_for_source` does a full `rglob` only on (a) first run, (b) every 60s, or (c) when ANY directory mtime changes. Otherwise it hits just the "hot" files — cursors touched in the last hour, capped at top-32 by recency. 
- Archived files are now free: no rglob, no stat, no open.

Tests prove both invariants from perf-007:
- Archived file > 1h old: zero `stat()` calls on the second sync.
- Live-file append: zero `rglob()` calls on the second sync.

The 60s periodic full rescan is a nice safety net — mtime-only watches can miss edge cases, but even the worst case is bounded instead of per-tick.

6/6 tests pass.

PR #556, opened 2026-04-21
perf(work): cache flow template loads

### LINKED ISSUE #459: perf-003 — Flow template reloaded from DB per task row

Source: `issues/Claude/perf-003-flow-template-not-cached.md`

**Severity:** medium
**Category:** performance — redundant work

## Location

`_load_flow_from_db` via `derive_owner` via `_row_to_task` in
`src/pollypm/work/service_queries.py` and `src/pollypm/work/sqlite_service.py`

## What

Each row reconstruction calls `derive_owner`, which loads the flow template
from DB. Tasks in the same project share one template, so this is a 1/N work
factor loss.

## Fix

`functools.lru_cache` on `_load_flow_from_db` keyed by flow id, invalidated
on flow definition updates. Expected speedup: ~5× on queue-scan operations.

samhotchkiss:
Closes #459

## Summary
- cache flow templates per service instance by (name, version) instead of reloading them from SQLite on repeat reads
- invalidate the cached entry when a persisted flow version is written
- add a regression test that proves repeated loads do not hit the flow tables again

## Testing
- HOME=/tmp/pytest-459 uv run --extra dev pytest tests/test_work_service.py tests/test_work_queries.py -q

### DIFF
--- a/src/pollypm/work/sqlite_service.py
+++ b/src/pollypm/work/sqlite_service.py
@@ -147,6 +147,7 @@ class SQLiteWorkService:
         self._conn.execute("PRAGMA busy_timeout=30000")
         create_work_tables(self._conn)
         self._gate_registry = GateRegistry(project_path=project_path)
+        self._flow_cache: dict[tuple[str, int], FlowTemplate] = {}
         # Last-provision-error breadcrumb — set by ``claim()`` when
         # ``provision_worker`` fails so the CLI can surface it instead
         # of reporting a silent success (#243).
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches perf-003 fix (closes #459) — per-service `_flow_cache` keyed on `(name, version)`, populated on miss in `_load_flow_from_db`, invalidated by `_insert_flow_version` when a new version is persisted, cleared on `close()`. Version-pinned key means template bumps evict cleanly without racing concurrent loads.

Test uses a `GuardConn` that raises if any query hits `work_flow_templates` or `work_flow_nodes` — the second load is fully served from cache. Elegant invariant. 68/68 tests pass.

PR #559, opened 2026-04-21
perf(work): bulk-load dependent task closure

### LINKED ISSUE #462: perf-006 — `dependent_tasks()` BFS fully hydrates each visited node

Source: `issues/Claude/perf-006-dependent-tasks-bfs-rehydrates.md`

**Severity:** medium
**Category:** performance — over-hydration

## Location

`src/pollypm/work/service_dependencies.py:94-112`

## What

BFS traversal calls `service.get(f"{project}/{number}")` per visited node.
Each `get` costs 4+ queries.

## Fix

Return task-id tuples from the traversal, not `Task` objects. Let callers
bulk-hydrate only the set they care about. Or: one recursive-CTE query in
SQLite that returns the closure directly.

samhotchkiss:
Closes #462

## Summary
- replace the Python BFS query-per-node traversal with a recursive SQLite closure query
- bulk-fetch dependent task rows instead of calling service.get() per visited node
- add a regression test proving dependents no longer hydrate through per-node get calls

## Testing
- HOME=/tmp/pytest-462 uv run --extra dev pytest tests/test_work_dependencies.py -q

### DIFF
--- a/src/pollypm/work/service_dependencies.py
+++ b/src/pollypm/work/service_dependencies.py
@@ -93,23 +93,40 @@ def unlink_tasks(service: "SQLiteWorkService", from_id: str, to_id: str, kind: s
 
 def dependent_tasks(service: "SQLiteWorkService", task_id: str) -> list[Task]:
     project, number = _parse_task_id(task_id)
-    visited: set[tuple[str, int]] = set()
-    queue: list[tuple[str, int]] = [(project, number)]
+    rows = service._conn.execute(
+        """
+        WITH RECURSIVE deps(project, task_number) AS (
+            SELECT to_project, to_task_number
+            FROM work_task_dependencies
+            WHERE from_project = ? AND from_task_number = ? AND kind = ?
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches perf-006 fix (closes #462) — replaces the Python BFS + per-node `service.get()` amplification with:
1. Single SQLite recursive CTE that walks the entire blocks-dependency closure in SQL.
2. Single bulk `SELECT` over `work_tasks` for all returned keys.
3. Single `_load_task_token_sums_bulk()` shared across hydration.

Went from O(N) queries + O(N) rehydrations to essentially O(1) queries for any closure size. Test monkeypatches `svc.get` to raise on call and verifies the full {b, c} transitive closure returns without touching `get()`.

The `OR`-chain `WHERE` in the bulk `SELECT` is fine for realistic closure sizes (<500 tasks well under SQLite's SQL-length limit). 19/19 tests pass.

PR #540, opened 2026-04-21
fix(settings): expose account usage and actions

samhotchkiss:
Closes #539

## Summary
- expose cached account usage fields through the account-status contract
- add visible Settings -> Accounts controls for add, remove, and refresh usage
- split account detail rendering into a dedicated settings/accounts helper module with focused coverage

PR #562, opened 2026-04-21
perf(settings): reuse filtered section rows

### LINKED ISSUE #470: perf-015 — Reactive properties recompute on every keystroke

Source: `issues/Claude/perf-015-reactive-recompute-per-keystroke.md`

**Severity:** low
**Category:** performance — cockpit responsiveness

## Location

Cockpit UI widgets that derive totals / counts from `entries` list.

## What

Any `@property` or reactive computed in the render path scans full lists on
every redraw. Fast keystrokes compound the cost.

## Fix

Cache derived values on the underlying data mutation path (add/remove only);
render reads the cached value.

samhotchkiss:
Closes #470

## Summary
- cache filtered settings-section rows during section render
- reuse the same filtered snapshot for detail rendering and row-highlight sync
- add a regression test proving the accounts section does not refilter during a single render cycle

## Testing
- HOME=/tmp/pytest-470d uv run --extra dev pytest tests/test_settings_ui.py -q

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -2365,6 +2365,9 @@ class PollySettingsPaneApp(App[None]):
         self._nav_widgets: dict[str, Static] = {}
         self._selected_account_key: str | None = None
         self._selected_project_key: str | None = None
+        self._visible_account_rows: list[dict] = []
+        self._visible_project_rows: list[dict] = []
+        self._visible_plugin_rows: list[dict] = []
         self._nav_cursor: int = 0
         self._focus_target: str = "nav"  # nav | table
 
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches perf-015 fix (closes #470) — `_render_section` now computes filtered rows once per section render and caches them on `_visible_account_rows` / `_visible_project_rows` / `_visible_plugin_rows`. `_render_*()` and `_render_*_detail()` helpers refactored to take `rows: list[dict]` instead of re-filtering from `data`. `_sync_selection` (triggered on row-highlight) uses the cached list instead of re-running the filter.

Before: each render called `_filtered_accounts` three times (table, detail, selection sync). After: once per section render.

Test monkeypatches `_filtered_accounts` with a counter, sets a search query, triggers `_render_section('accounts')`, asserts exactly 1 filter call.

**Tmux verification**: launched `pm cockpit-pane settings`. Accounts render with all fields intact. Tab → Projects works (3 projects listed). Button-row visibility (from #540) still hides correctly on non-Accounts sections. No layout or rendering regressions.

21/21 tests pass.

PR #566, opened 2026-04-21
fix(heartbeat): dedupe repeated idle nudges

### LINKED ISSUE #478: prompt-008 — Heartbeat nudge messages re-sent without dedup / escalation

Source: `issues/Claude/prompt-008-heartbeat-nudges-not-deduped.md`

**Severity:** medium
**Surface:** `src/pollypm/heartbeats/local.py:779-783`
**Criterion:** overloading + correctness

## What

Stall-nudge templates are injected every sweep while the idle condition
holds. Workers receive the same "you seem idle" message repeatedly.

## Impact

Agent context gets polluted with nudge echoes; the agent may loop on
acknowledgment. No escalation threshold visible.

## Fix

Send one nudge per idle episode. After N more idle cycles, escalate
(different message, or raise an alert). Stop sending further nudges until the
agent has moved.

samhotchkiss:
Closes #478

## Summary
- treat repeated identical stalled snapshots as one idle episode for heartbeat nudges
- suppress repeat nudges for the same snapshot episode even after the cooldown window
- escalate to a stuck-session alert when the same episode keeps repeating, and record that escalation in the event stream

## Testing
- HOME=/tmp/pytest-478 uv run --extra dev pytest tests/test_heartbeat_plugin_api.py tests/test_supervisor.py -q

### DIFF
--- a/src/pollypm/heartbeats/local.py
+++ b/src/pollypm/heartbeats/local.py
@@ -911,7 +911,25 @@ class LocalHeartbeatBackend(HeartbeatBackend):
     # Rate limit nudges: max once per session per 10 minutes
     # Rate limits for worker nudges
     _NUDGE_COOLDOWN_SECONDS = 600
+    _NUDGE_ESCALATION_IDLE_CYCLES = 8
     _MAX_NUDGES_BEFORE_RECOVERY = 6
+
+    def _repeated_snapshot_count(
+        self,
+        api,
+        context: HeartbeatSessionContext,
+        *,
+        limit: int,
+    ) -> int:
+        hashes = api.recent_snapshot_hashes(context.session_name, limit=limit)
+        repeated = 0
+        if hashes:
+            for value in hashes:
+                if value == hashes[0]:
+                    repeated += 1
+                else:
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches prompt-008 fix (closes #478) — much sharper than what I suggested:
- Each nudge event now carries `snapshot_hash` and `idle_cycles`.
- Before re-nudging, the heartbeat checks whether a prior nudge already fired for the *same* snapshot hash (same idle episode). If so, it skips the nudge even if the cooldown has elapsed.
- When the same snapshot hash repeats for ≥ 8 heartbeats (`_NUDGE_ESCALATION_IDLE_CYCLES`) AND there was already a prior nudge for this episode, it raises a `stuck_session` warning alert and records a `nudge_escalated` event.
- Once escalated for an episode, it shuts up and doesn't re-escalate.

Correctly resolves the perf + UX complaint: workers no longer get the same nudge text drip-fed during one idle episode, and genuinely stuck sessions surface as alerts instead of nudge spam. 61/61 tests pass.

Issue #477, opened 2026-04-21
prompt-007 — Architect prompt doesn't explain `--actor` requirement


samhotchkiss:
Validated against current `main` (`e20dd9a`) on 2026-04-20.

Source: `issues/Claude/prompt-007-architect-actor-required.md`

**Severity:** high
**Surface:** `src/pollypm/plugins_builtin/project_planning/profiles/architect.md:49-67`
**Criterion:** sufficiency

## What

Architect is told to call `pm task done ... --actor architect` but the prompt
never explains what `--actor` does, what happens if omitted, or why it
matters. (The default is "worker" per saved memory, which would mis-route.)

## Fix

Add one sentence:

> `--actor architect` tells the work-service that you (the architect) are
> driving the transition. If you omit it, the action is recorded as a
> worker and the wrong role will be notified.

samhotchkiss:
Already fixed on current `main`. The architect prompt now explains that omitting `--actor architect` records the action as a worker and notifies the wrong role, and the prompt test suite asserts that wording.

PR #572, opened 2026-04-21
docs(project-planning): explain --actor architect

samhotchkiss:
Refs #477.\n\nAdds a single explanatory sentence to the architect prompt and a regression assertion that the prompt explains why `--actor architect` is required.


samhotchkiss:
LGTM. Matches prompt-007 fix (refs #477) — appends the missing explanation to the existing `--actor architect` sentence: "; if you omit it, the action is recorded as a worker and the wrong role is notified." Test asserts all three phrases (`--actor architect`, `recorded as a worker`, `wrong role is notified`) appear in the rendered prompt. 7/7 tests pass.

Issue #474, opened 2026-04-21
prompt-004 — Worker guide duplicated in kickoff prompt


samhotchkiss:
Validated against current `main` (`e20dd9a`) on 2026-04-20.

Source: `issues/Claude/prompt-004-worker-guide-duplicated-at-kickoff.md`

**Severity:** medium
**Surface:** `src/pollypm/memory_prompts.py:319-344` vs `docs/worker-guide.md`
**Criterion:** overloading

## What

The session kickoff injects a full `## Worker Protocol` section that overlaps
`docs/worker-guide.md`. Both describe lifecycle and commands.

## Impact

~4K tokens of context devoted to lifecycle material the agent can also read
on disk. Dilutes task-specific context.

## Fix

Inject a pointer ("Read `docs/worker-guide.md` — 2-min read, describes how
to claim, ship, recover"). Reserve kickoff for the task, the checkpoint,
and relevant memory.

samhotchkiss:
Resolved on `main` by merged PR #574 (`docs(memory): point workers to the guide`). Closing as fixed.

PR #574, opened 2026-04-21
docs(memory): point workers to the guide

samhotchkiss:
Refs #474.\n\nReplaces the duplicated kickoff worker-guide prose with a short pointer to docs/worker-guide.md and updates the focused regression test.

### DIFF
--- a/src/pollypm/memory_prompts.py
+++ b/src/pollypm/memory_prompts.py
@@ -321,27 +321,25 @@ def build_worker_protocol_injection(
     session_role: str | None,
     guide_text: str | None = None,
 ) -> str:
-    """Render a ``## Worker Protocol`` section containing the worker guide.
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches prompt-004 fix (refs #474) — kickoff injection is now a short pointer:
```
## Worker Protocol

Read `docs/worker-guide.md` for the worker playbook, including how to claim work, ship it, and recover.
```
Preserves the `guide_text` kwarg for API compat but intentionally ignores it. Tests updated to verify the pointer appears and the full guide body does NOT inline. Note that the "empty guide → empty injection" behavior is now "empty guide → pointer injection" — tests cover this. Saves the ~4K tokens the old inline protocol injection was costing every worker kickoff.

17/17 tests pass.

Issue #480, opened 2026-04-21
prompt-010 — Recovery prompt doesn't distinguish fresh launch vs resumed


samhotchkiss:
Validated against current `main` (`e20dd9a`) on 2026-04-20.

Source: `issues/Claude/prompt-010-recovery-fresh-vs-resume.md`

**Severity:** medium
**Surface:** `src/pollypm/recovery_prompt.py:79-108`
**Criterion:** sufficiency

## What

The fallback ("no prior checkpoint") and the resumed ("from checkpoint <ID>")
recovery prompts look similar at a glance. The agent may not realize which it
is in, and start looking for a "next step" that doesn't exist.

## Fix

Lead with an unambiguous flag:

```
RECOVERY MODE: FRESH LAUNCH — no checkpoint to resume from.
```
or
```
RECOVERY MODE: RESUMING FROM CHECKPOINT <id> — last state was <node>.
```

samhotchkiss:
Verified on current main: `src/pollypm/recovery_prompt.py` already emits `RECOVERY MODE: FRESH LAUNCH - no checkpoint to resume from.` for fallback launches and `RECOVERY MODE: RESUMING FROM CHECKPOINT <id> - last state was <state>.` for checkpoint resumes, with regression coverage in `tests/test_recovery_prompt.py` and `tests/integration/test_recovery_prompt_integration.py`. Closing as already fixed.

PR #576, opened 2026-04-21
fix: add recovery mode banner

samhotchkiss:
Refs #480.\n\nAdds explicit recovery-mode banners for fresh-launch and resumed recovery prompts, plus focused regression coverage for both paths.

### DIFF
--- a/src/pollypm/recovery_prompt.py
+++ b/src/pollypm/recovery_prompt.py
@@ -59,12 +59,13 @@ class RecoveryPrompt:
     checkpoint_id: str = ""
     session_name: str = ""
     is_fallback: bool = False
+    recovery_mode_banner: str = ""
 
     def render(self) -> str:
         """Render the full recovery prompt as text."""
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches prompt-010 fix (refs #480) — adds explicit mode banner to both recovery paths:
- Fresh launch: `RECOVERY MODE: FRESH LAUNCH — no checkpoint to resume from.`
- Resume: `RECOVERY MODE: RESUMING FROM CHECKPOINT <id> — last state was <state>.`

`recovery_mode_banner` threads through the `RecoveryPrompt` dataclass and both renderers (Claude and Codex) lead with it before the prose. The `else` branch in each renderer preserves backwards-compat output for direct dataclass construction (tests that don't go through `_build_from_checkpoint` / `_build_fallback_prompt`).

Two new regression tests pin both banner variants. Agent can no longer confuse fresh launch for resume or vice versa. 15/15 tests pass.

Issue #487, opened 2026-04-21
prompt-017 — Persona verification probe runs invisibly to the agent


samhotchkiss:
Validated against current `main` (`e20dd9a`) on 2026-04-20.

Source: `issues/Claude/prompt-017-persona-verify-invisible.md`

**Severity:** low
**Surface:** `src/pollypm/supervisor.py:2791-2794` (`_schedule_persona_verify`)
**Criterion:** sufficiency

## What

A background probe checks that the agent's persona marker appears in its
pane. If the agent ever sees logs mentioning "persona verify," it has no
explanation for what this is.

## Fix

One line in the kickoff prompt:

> _A background probe will verify your persona by scanning your pane
> shortly after launch. It is non-blocking and is not a task for you._

samhotchkiss:
Resolved on `main` by merged PR #578 (`prompt: make persona probe visible in kickoff copy`). Closing as fixed.

PR #575, opened 2026-04-21
prompt: clarify Polly authority split

samhotchkiss:
Refs #484.\n\nRestructured Polly's authority block into a clearer CAN / MUST ESCALATE contract and added a focused regression assertion in the prompt test seam.

PR #578, opened 2026-04-21
prompt: make persona probe visible in kickoff copy

samhotchkiss:
Refs #487.\n\nAdded a one-line explanation to the operator kickoff prompt stating the persona probe is non-blocking and not a task for the agent, with a focused regression assertion in the prompt test seam.


samhotchkiss:
Review is clean — matches prompt-017 (refs #487). One-line addition to operator prompt's `<system>` block explaining the persona probe is non-blocking and not a task.

But branch is CONFLICTING with main — likely conflicts with #575 (Polly authority split) that landed recently in the same `profiles.py`. Rebase and retain your one-line `<system>` addition; the authority block changes from #575 are upstream.

Flipping to `needs fix`.

samhotchkiss:
Rebase is clean. 10/10 tests pass. Re-confirming: one-line addition to Polly's `<system>` block noting the persona probe is non-blocking and not a task. Merging.

Issue #486, opened 2026-04-21
prompt-016 — Worker output JSON `kind` field has unexplained values


samhotchkiss:
Validated against current `main` (`e20dd9a`) on 2026-04-20.

Source: `issues/Claude/prompt-016-worker-output-kinds-unexplained.md`

**Severity:** low
**Surface:** `src/pollypm/plugins_builtin/core_agent_profiles/profiles.py:228-239`
**Criterion:** sufficiency

## What

Worker prompt describes an output JSON with `"kind": "commit" | "file_change"
| "action" | "note"` but doesn't explain when to use each. `action` and
`note` rarely appear in practice; workers will guess.

## Fix

One line per kind with an example. "`commit` for git commits (most common);
`file_change` for non-git file writes; `action` for operations with side
effects; `note` to record a decision."

samhotchkiss:
Resolved on `main` by merged PR #580 (`fix: explain worker output kinds`). Closing as fixed.

PR #543, opened 2026-04-21
test(prompts): audit CLI references and align prompt contracts

samhotchkiss:
Closes #490
Closes #489
Closes #481

## Summary
- add a prompt/doc surface test that resolves documented `pm ...` references against the real CLI
- replace stale `pm mail` and `pm down` references with current commands
- standardize worker prompt examples on `--output` and state reviewer gate semantics explicitly

PR #580, opened 2026-04-21
fix: explain worker output kinds

samhotchkiss:
Refs #486.\n\nAdds concise per-kind guidance and examples to the worker output JSON prompt so each artifact kind has an explicit use case.


samhotchkiss:
Review is clean — matches prompt-016 (refs #486). Adds concise per-kind guidance to the worker output format (`commit` / `file_change` / `action` / `note`). New test asserts the phrases appear.

But branch is CONFLICTING with main — likely conflicts with #543 (`-o` → `--output` standardization) that landed in the same lines of `profiles.py:233`. Rebase and retain your per-kind annotations; main's `--output` form is upstream.

Flipping to `needs fix`.

samhotchkiss:
Rebase is clean. 10/10 tests pass. Re-confirming prompt-016 fix: each artifact kind now has a concise use-case annotation (commit=git, file_change=non-git files, action=side-effecting ops, note=decisions/blockers/observations). Merging.

Issue #485, opened 2026-04-21
prompt-015 — Worker is told to read `.pollypm/INSTRUCT.md` and `.pollypm/docs/SYSTEM.md` without context


samhotchkiss:
Validated against current `main` (`e20dd9a`) on 2026-04-20.

Source: `issues/Claude/prompt-015-instruct-and-system-md-unexplained.md`

**Severity:** low
**Surface:** `src/pollypm/plugins_builtin/core_agent_profiles/profiles.py:405-422`
**Criterion:** sufficiency

## What

Worker prompt references these overlay files without explaining:
- who writes them,
- that they override defaults,
- what to do if they are missing.

Likely outcome: the agent looks for files that don't exist and reports an
error instead of proceeding.

## Fix

Add: "`.pollypm/INSTRUCT.md` and `.pollypm/docs/SYSTEM.md` are optional
project-level overrides written by the PM. If either is missing, defaults
apply — continue without blocking."

samhotchkiss:
Resolved on `main` by merged PR #581 (`docs(core-prompts): explain worker overrides`). Closing as fixed.

PR #581, opened 2026-04-21
docs(core-prompts): explain worker overrides

samhotchkiss:
Refs #485.\n\nAdds an explicit project-overrides note for .pollypm/INSTRUCT.md and .pollypm/docs/SYSTEM.md, including override semantics and the missing-file fallback, with a focused regression test.


samhotchkiss:
LGTM. Matches prompt-015 fix (refs #485) — prepends a `<project-overrides>` block to the INSTRUCT rules assembly explaining:
- `.pollypm/INSTRUCT.md` and `.pollypm/docs/SYSTEM.md` are optional.
- When present, they override defaults.
- When missing, defaults apply — continue without blocking.

Regression test builds a full `AgentProfileContext`, asserts the override-note text appears both when files are missing AND when they're present. 1/1 tests pass.

Issue #436, opened 2026-04-21
docs-010 — Plugin specs labeled "v1 target" but already shipped


samhotchkiss:
Validated against current `main` (`e20dd9a`) on 2026-04-20.

Source: `issues/Claude/docs-010-plugin-specs-status-stale.md`

**Severity:** medium
**Docs:** `docs/planner-plugin-spec.md`, `docs/morning-briefing-plugin-spec.md`,
`docs/downtime-plugin-spec.md`, `docs/advisor-plugin-spec.md`

## What's wrong

Each spec header reads "Status: v1 target". Corresponding plugins exist under
`src/pollypm/plugins_builtin/{planner,morning-briefing,downtime,advisor}/`.
Readers think these are aspirational.

## Fix

Update each header to "Status: v1 shipped — implemented in
`plugins_builtin/<name>/`" and add a pointer line to the entry module.

samhotchkiss:
Verified against current origin/main after refreshing refs: the four plugin specs now read `Status: v1 shipped` and include implementation pointers to the built-in plugin entry modules. Closing as fixed by merged PR #608.

PR #584, opened 2026-04-21
docs: mark plugin specs as shipped

samhotchkiss:
Refs #436.\n\nUpdated the plugin spec headers to reflect shipped status and added pointers to the builtin implementation modules.


samhotchkiss:
LGTM. Matches docs-010 fix (refs #436) — updates all four plugin specs (advisor, downtime, morning-briefing, planner) from "Status: v1 target" to "Status: v1 shipped - implemented in `src/pollypm/plugins_builtin/<name>/plugin.py`". Readers can no longer confuse shipped code for aspirational spec. Doc-only.

PR #598, opened 2026-04-21
perf(heartbeat): avoid deep copy for primitive payloads

### LINKED ISSUE #466: perf-010 — `copy.deepcopy` on every heartbeat job payload

Source: `issues/Claude/perf-010-deepcopy-heartbeat-payload.md`

**Severity:** low
**Category:** performance — unnecessary copy

## Location

`src/pollypm/heartbeat/tick.py:119`

## What

Each enqueued job deepcopies `entry.payload`. If payloads are effectively
immutable (dicts of primitives created per tick) the copy is wasted.

## Fix

Profile first. If payloads aren't mutated post-enqueue, drop the copy. Or
switch to `dataclasses(frozen=True)` payloads so no defensive copy is needed.

samhotchkiss:
Closes #466

## Summary
- add a heartbeat payload snapshot helper with a primitive-value fast path
- keep deepcopy for nested mutable payloads so enqueue isolation stays intact
- add a regression test proving primitive-only payloads bypass deepcopy

## Testing
- HOME=/tmp/pytest-466 uv run --extra dev pytest tests/test_heartbeat_tick.py -q

### DIFF
--- a/src/pollypm/heartbeat/tick.py
+++ b/src/pollypm/heartbeat/tick.py
@@ -38,6 +38,29 @@ from pollypm.heartbeat.roster import (
 __all__ = ["Heartbeat", "JobQueueProtocol", "TickResult", "EnqueuedJob"]
 
 
+def _is_immutable_payload_value(value: Any) -> bool:
+    """Return True when a payload value is safe to reuse without deep copy."""
+    if value is None or isinstance(value, (str, bytes, int, float, bool)):
+        return True
+    if isinstance(value, tuple):
+        return all(_is_immutable_payload_value(item) for item in value)
+    if isinstance(value, frozenset):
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM. Matches perf-010 fix (closes #466) — primitive-only payloads (None/str/bytes/int/float/bool + tuples/frozensets/dicts-of-primitives) get a cheap `dict(payload)` shallow copy instead of `copy.deepcopy`. Nested mutable values still go through deepcopy for enqueue isolation.

Test monkeypatches `copy.deepcopy` to raise, verifies tick succeeds with a primitive-only payload and delivers the expected dict. Clean invariant. 30/30 tests pass.

PR #608, opened 2026-04-21
docs: clarify shipped plugin spec headers

samhotchkiss:
## Summary\n- normalize the shipped-status header block across the planner, morning briefing, downtime, and advisor specs\n- add an explicit implementation pointer to each built-in plugin entry module\n- keep the write set limited to the four issue-targeted docs\n\nCloses #436


samhotchkiss:
LGTM. Follow-up polish on #584 — splits the "Status: v1 shipped - implemented in..." one-liner into three explicit `Status`/`Implementation`/`Depends on` lines across all four plugin specs. Easier to skim. Doc-only.

PR #614, opened 2026-04-21
fix(cockpit): avoid blocking settings reads

### LINKED ISSUE #606: fix(cockpit): make settings screen load promptly
## Summary
Opening `Settings` in the cockpit can stall for around 60 seconds before rendering.

## Observed
- Clicking `Settings` leaves the screen apparently hung for a long delay before content appears.
- During RC3 testing, the load time was reported at roughly 60 seconds.

## Expected
- `Settings` should render promptly and defer any slow probes or background refresh work instead of blocking the UI.
- Account and environment data that can be stale by a few minutes should come from cached state or background jobs.

## Notes
This likely intersects with the cached account-usage work: the settings UI should stay responsive even while account/provider status refresh work is happening elsewhere.

samhotchkiss:
## Summary
- move settings-project snapshot reads into a dedicated helper module
- use short-timeout read-only SQLite queries so busy project DBs render as `busy` instead of blocking Settings mount
- add a regression test for locked project databases and keep the existing settings-pane slice green

## Testing
- HOME=/tmp/pytest-606 uv run --extra dev pytest tests/test_cockpit_settings_projects.py tests/test_settings_ui.py -q

Closes #606

### DIFF
--- a/src/pollypm/cockpit_settings_projects.py
+++ b/src/pollypm/cockpit_settings_projects.py
@@ -1,0 +1,103 @@
+"""Non-blocking project snapshot helpers for the cockpit settings screen.
+
+Contract:
+- Inputs: the loaded PollyPM config plus a relative-age formatter.
+- Outputs: normalized project rows for the settings UI.
+- Side effects: reads project-path metadata and does read-only SQLite
+  queries with a very short timeout so busy project DBs do not stall UI
+  mount.
+- Invariants: callers get best-effort task totals; a locked DB surfaces
+  as ``task_total_label='busy'`` instead of blocking the screen.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime as _dt
+from pathlib import Path
+import sqlite3
+
+
+def collect_settings_projects(config, *, format_relative_age) -> list[dict]:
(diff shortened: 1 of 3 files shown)


samhotchkiss:
LGTM. Matches #606 — Settings → Projects no longer blocks on busy work DBs:
- **New module** `cockpit_settings_projects.py`: `collect_settings_projects(config, format_relative_age)` extracts the snapshot logic from the god-file `cockpit_ui.py`.
- **Fast, read-only queries**: `sqlite3.connect(..., mode=ro, uri=True, timeout=0.05)` + `PRAGMA busy_timeout=50`. On `OperationalError: database is locked/busy`, returns `task_total_label='busy'` instead of blocking.
- **Settings UI** picks up the new `task_total_label` key in both table and detail render (falls back to `task_total` for compat).

Tests cover fast-path (2 tasks → `task_total_label='2'`) AND lock-path (exclusive write lock → renders `'busy'`, elapsed < 1s proving non-blocking).

**Tmux verification**: Settings → Projects renders correctly — news=1, otter_camp=1, pollypm=44 task counts visible, no regressions on the fast path. 23/23 tests pass.

PR #620, opened 2026-04-21
fix(heartbeat): stop nudging review-only workers

### LINKED ISSUE #393: fix(polly behavior): false 'stuck on permissions prompt' nudges to workers that are actually idle-by-design
## Repro

Observed in wordgame AND pomodoro E2E runs — after a worker finishes its task (\`pm task done\` succeeded, status=review), Polly's monitor keeps firing \"stuck on permissions prompt\" nudges to the worker. The worker correctly responds \"not stuck, standing by\" — repeatedly, 3+ times.

Pattern in the worker pane:
\`\`\`
❯ H: You appear stuck on a permissions prompt. Accept or work around it.
⏺ Not stuck — task already completed cleanly last turn. pm task done pomodoro/2 returned status: review, no prompt pending. Standing by.
❯ H: You appear stuck on a permissions prompt. Accept or work around it.
⏺ Task is review with reviewer active — no permissions prompt on my end. Nothing to accept or work around.
❯ H: You appear stuck on a permissions prompt. Accept or work around it.
\`\`\`

## Why this matters

- Worker is confused — it has to defend itself against false accusations
- Burns worker context explaining the obvious
- Reveals Polly's monitor heuristic is too aggressive: 'no tool call in N seconds' ≠ 'stuck on prompt'

## Discovered during

Overnight E2E runs on 2026-04-19 — pomodoro run, log at /tmp/polly-e2e-pomodoro.log, and prior wordgame run.

## Fix

Polly should check **task status** before nudging. If task status is 'review' / 'done' / 'completed', the worker being idle is CORRECT. Nudge only fires when:
- Task status == 'in_progress' AND
- Task node suggests worker should act AND
- No tool call in > N seconds

Also: a distinct nudge for 'pane appears empty' vs 'pane shows permission prompt' — currently one heuristic, two different root causes.

samhotchkiss:
## Summary
- treat only worker-actionable states as pending work for heartbeat nudges
- stop treating review-only tasks as evidence the worker should still be acting
- add heartbeat regression tests for review-only vs in-progress task states

## Testing
- HOME=/tmp/pytest-393 uv run --extra dev pytest tests/test_heartbeat_plugin_api.py -q

Closes #393

### DIFF
--- a/src/pollypm/heartbeats/local.py
+++ b/src/pollypm/heartbeats/local.py
@@ -273,6 +273,7 @@ def _select_intervention(
 class LocalHeartbeatBackend(HeartbeatBackend):
     name = "local"
     _UNMANAGED_WINDOW_ALERT_PREFIX = "unmanaged_window:"
+    _WORKER_ACTIONABLE_STATUSES = frozenset({"queued", "in_progress", "blocked"})
     _MUTATING_SESSION_ROLES = frozenset(
         {
             "architect",
@@ -1090,8 +1091,9 @@ class LocalHeartbeatBackend(HeartbeatBackend):
                 tasks = backend.list_tasks(states=["01-ready", "02-in-progress"])
                 if tasks:
                     return True
-            # Check the work service for non-terminal tasks assigned to this
-            # worker's project.
+            # Check the work service for worker-actionable tasks on this
+            # worker's project. Review / on_hold / done are idle-by-design
+            # for workers — the reviewer or user owns the next step there.
             try:
                 from pollypm.work.sqlite_service import SQLiteWorkService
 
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM.

- Scopes `_has_pending_work` to worker-actionable statuses (`queued`/`in_progress`/`blocked`) instead of "anything not done/cancelled", so review/on_hold tasks no longer keep the heartbeat nudging a worker that correctly handed off — fixes #393.
- Uses a frozenset constant on the backend class, keeping the allow-list obvious and cheap.
- Verified: `tests/test_heartbeat_plugin_api.py` all green (20 passed), including the two new regression cases for review-only vs in_progress.

Issue #339, opened 2026-04-18
[storage-C] Collapse DB layout to two scopes (user + workspace)


samhotchkiss:
## Goal

Collapse the multi-DB layout. Exactly two storage scopes, exactly two DB files. No `.pollypm-state`, no per-agent-worktree DBs, no per-project work.db files.

**Greenfield — Sam will wipe state after full plan lands.**

## Target layout

```
~/.pollypm/state.db                    # user-scope: sessions, account_usage, knowledge/memory
<workspace_root>/.pollypm/state.db     # workspace-scope: messages, work_tasks, worktrees, alerts
```

That's it. Two files.

## What gets deleted

- `src/pollypm/models.py::ProjectSettings.base_dir = Path(".pollypm-state")` — remove the field or repoint at `.pollypm/`.
- Every path in code that constructs `.pollypm-state/state.db` — replace with `.pollypm/state.db`.
- Per-agent-worktree DB creation in the worktree provisioning path (I observed one at `/Users/sam/dev/shortlink-gen/.pollypm/worktrees/worker_shortlink_gen/shortlink_gen-pa-worker_shortlink_gen/.pollypm-state/state.db`) — agents read/write through the workspace-scope DB.
- `_resolve_db_path` in `src/pollypm/work/cli.py` — simplify to the two-scope rule.

## Scope boundary decision

Row-level `scope` column on `messages` (already in schema from Issue B) distinguishes workspace vs project. No separate files per project.

## Scope

- `src/pollypm/models.py`
- `src/pollypm/work/cli.py`
- `src/pollypm/work/sqlite_service.py` (likely — points at project-local DBs today)
- Agent worktree provisioning paths (grep for `base_dir` / `.pollypm-state`)
- Tests: `tests/test_workspace_scope_resolution.py` (new)

## Acceptance

- `grep -r "pollypm-state" src/` returns nothing.
- `grep -r "work.db" src/` returns nothing (work_tasks now live in messages / per-scope state.db).
- `pm doctor` reports exactly 2 SQLite paths when invoked from a workspace.
- Agent worktree creation no longer creates a nested DB.

## Dependencies

Depends on **Issue A** (for Store abstraction). Can land in parallel with Issue B.

PR #627, opened 2026-04-21
fix(projects): route planner tasks through workspace db

### LINKED ISSUE #377: fix(project new): work_tasks land in per-project DB but pm task status can't find them
## Repro

```
$ pm project new /tmp/foo --yes --name foo
Registered project foo at /private/tmp/foo
Fresh project — running cold-start planner.
Created planning task foo/1 on project 'foo' (flow=plan_project).
Next: `pm task queue foo/1` to hand it off to the architect worker.

$ pm task status foo/1
Error: Task 'foo/1' not found.

$ pm task queue foo/1
Error: Task 'foo/1' not found.
```

The task IS created — it just lands in `/tmp/foo/.pollypm/state.db` (project-local), while `pm task status` looks in `~/.pollypm/state.db` (user-scope) and `/Users/sam/dev/.pollypm/state.db` (workspace-scope).

## Discovered during

Overnight E2E testing on 2026-04-18 — see `/tmp/pollypm_overnight_test_report.md` (B3).

## Why this matters

Issue #339 (collapse DB layout to two scopes) was supposed to eliminate per-project DBs. They still get created by `pm project new` via `SQLiteWorkService` instantiated against `<project>/.pollypm/state.db`.

## Real fix

`pm project new` should write to the **workspace-root DB** (`<workspace>/.pollypm/state.db`) with `project=<key>` so the row is scoped via column, not file. Then `pm task status` finds it via the standard resolver.

Same fix applies to anywhere else that spawns a per-project work-service writer.

samhotchkiss:
## Summary
- route `pm project new/plan/replan` planner tasks through the workspace-scope work DB instead of per-project `state.db` files
- make the `project.created` observer use the same workspace DB so auto-fired planning tasks are visible to `pm task ...` immediately
- cover both the direct CLI path and the observer path with regression tests, including classification from existing workspace work-task rows

## Testing
- HOME=/tmp/pytest-377 uv run --extra dev pytest tests/test_project_new_replan_routing.py tests/test_project_planning_cli.py -q

Closes #377

### DIFF
--- a/src/pollypm/plugins_builtin/project_planning/cli/project.py
+++ b/src/pollypm/plugins_builtin/project_planning/cli/project.py
@@ -189,7 +189,37 @@ def _has_source_files(project_path: Path, *, max_scan: int = 2000) -> bool:
     return False
 
 
-def _has_work_tasks(project_path: Path) -> bool:
+def _planner_db_path(
+    project_path: Path,
+    *,
+    config_path: Path | None = None,
+    create_parent: bool = False,
+) -> Path:
+    """Return the planner/work-task DB for project-planning entry points.
+
+    Post-#339, planner-created tasks belong in the workspace-scope DB.
+    Fall back to the project-local path only when no config is available,
+    which keeps older isolated tests and recovery paths working.
+    """
+    db_path = project_path / ".pollypm" / "state.db"
+    try:
+        cfg = load_config(config_path) if config_path is not None else load_config()
+        workspace_root = getattr(cfg.project, "workspace_root", None)
+        if workspace_root is not None:
+            db_path = Path(workspace_root) / ".pollypm" / "state.db"
+    except Exception:  # noqa: BLE001
+        pass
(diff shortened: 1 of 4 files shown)


samhotchkiss:
LGTM.

- Routes `pm project new/plan/replan` planner tasks through the workspace-scope work DB (post-#339) via new `_planner_db_path` helper, with project-local fallback for tests without config.
- Threads `config_path` through `_classify_project_state`, `_has_work_tasks`, `_plan_task_exists`, and `_plan_project_task`; mirrors the same fix in the `project.created` observer in `plugin.py`.
- `_has_work_tasks` now scopes by `project` when a key is given, so workspace DB rows classify correctly.
- Closes #377.
- 32/32 tests pass across `tests/test_project_new_replan_routing.py` and `tests/test_project_planning_cli.py`, including a new `test_existing_workspace_tasks_classify_existing_replan` regression.

Issue #395, opened 2026-04-19
fix(critic panel): critic reviews spawn 5 real work_tasks that get cancelled — confusing task list + shifts implementation task ID


samhotchkiss:
## Repro

Run the planner on any project. After critic_panel completes, \`pm task list\` shows:

\`\`\`
qotd/1               done           high       Plan project qotd
qotd/2               cancelled      high       Simplicity critic: qotd plan review
qotd/3               cancelled      high       Maintainability critic: qotd plan review
qotd/4               cancelled      high       User critic: qotd plan review
qotd/5               cancelled      high       Operational critic: qotd plan review
qotd/6               cancelled      high       Security critic: qotd plan review
qotd/7               queued         high       Implement qotd v0.1: ...
\`\`\`

5 cancelled critic tasks per project. Observed consistently across overnight E2E runs (wordcount, climb, blog, pomodoro, habit, shortlink_cli, qotd — 7/7 affected).

## Why this matters

- The implementation task ID is no longer \`<project>/2\` but \`<project>/7\`. Breaks anyone writing monitors/scripts that assume the first post-plan task is \`/2\`.
- Polly's monitor on wordgame/pomodoro/qotd watched \`/2\` and missed real status changes on \`/7\` — contributes to the nudge-loop behavior (#393).
- Clutters \`pm task list\` — 5/7 rows are noise.

## Probable cause

Critic panel (src/pollypm/plugins_builtin/project_planning/critic_panel.py or similar) creates work_tasks to run critic reviews as sub-agents, but cancels them after the critic output lands, leaving cancelled rows behind.

## Fix candidates

- (a) Critic reviews should use a separate table (or JSON artifact in worktree), not work_tasks
- (b) If critics must be work_tasks, delete them on cancellation rather than leaving cancelled rows
- (c) \`pm task list\` should default to filtering cancelled rows; require --all to show them

Sibling to #394 (architect v1 drift) — both are artifacts of the 5-critic panel infrastructure.

samhotchkiss:
Closing as already fixed on main. SQLiteWorkService.cancel() now prunes linked critique_flow child tasks after cancel side effects land, and tests/test_work_service.py covers the exact regression: cancelling planner critic children removes them from the task list and the next implementation task reuses /2.

samhotchkiss:
Reopening after verifying against updated origin/main. My earlier close was based on local branch state, not the shipped main branch, and that was incorrect.

samhotchkiss:
[...]

PR #636, opened 2026-04-21
fix(work): prune cancelled planner critic tasks

samhotchkiss:
## Summary
- prune cancelled linked critique_flow subtasks after cancel side effects land
- keep standalone critique_flow tasks behaving like normal cancelled tasks
- add regression coverage for planner critic cleanup and numbering reuse

## Testing
- HOME=/tmp/pytest-395 uv run --extra dev pytest tests/test_work_service.py::TestCancel tests/test_work_dependencies.py::TestLinkBlocks::test_link_parent tests/test_work_cli.py::TestCliList::test_cli_list -q

### DIFF
--- a/src/pollypm/work/sqlite_service.py
+++ b/src/pollypm/work/sqlite_service.py
@@ -1103,6 +1103,7 @@ class SQLiteWorkService:
                 )
         result = self.get(task_id)
         self._sync_transition(result, task.work_status.value, WorkStatus.CANCELLED.value)
+        self._prune_cancelled_critique_child(result)
         return result
 
     def hold(self, task_id: str, actor: str, reason: str | None = None) -> Task:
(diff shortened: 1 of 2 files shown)


samhotchkiss:
LGTM.

- Prunes cancelled `critique_flow` subtasks that are linked as children after the planner's critic panel finishes, so `pm task list` no longer shows 5 phantom cancelled rows per project and the first post-plan implementation task gets `/2` again (issue #395).
- Standalone (unlinked) `critique_flow` tasks keep normal cancelled semantics via the `_has_incoming_parent_link` guard — nice touch.
- Regression coverage: new `test_cancel_prunes_critique_child_tasks_and_reuses_numbering` verifies 5-critic cleanup + number reuse, plus `test_cancel_keeps_standalone_critique_flow_tasks` for the negative case.
- Verified: stated suite 9/9 pass; broader `tests/test_work_service.py + tests/test_work_dependencies.py` 76/76 pass.

PR #595, opened 2026-04-21
feat(cockpit): add just shipped dashboard panel

samhotchkiss:
Closes #521

## Summary
- add a dedicated cockpit dashboard section for recently approved work
- surface the last three shipped tasks from the last 24 hours with cycle time and age
- keep the panel hidden when nothing shipped recently

## Testing
- HOME=/tmp/pytest-521 uv run --extra dev pytest tests/test_cockpit_just_shipped.py tests/test_cockpit_project_dashboard.py -q

PR #591, opened 2026-04-21
feat(cockpit): add project health scorecards

samhotchkiss:
Closes #510

## Summary
- add a dedicated cockpit health-scorecard module for stuck-task health and cycle summaries
- surface per-project health lines in the global dashboard and per-project dashboard
- keep the health scoring logic independently testable with focused dashboard tests

## Testing
- HOME=/tmp/pytest-510 uv run --extra dev pytest tests/test_cockpit_project_health.py tests/test_cockpit_project_dashboard.py -q

Issue #515, opened 2026-04-21
ux-019 — Morning-briefing banner on cockpit open


samhotchkiss:
Validated against current `main` (`e20dd9a`) on 2026-04-20.

Source: `issues/Claude/ux-019-morning-briefing-banner.md`

**Surface:** cockpit dashboard / briefing plugin
**Effort:** medium · **Impact:** medium

## Why flat

Morning briefing is a separate CLI command nobody remembers to run.

## Idea

On cockpit launch, if no briefing in last 24h, generate one and show a
dismissible banner: `Morning briefing available — press B to read`. Modal or
dashboard pane view. Hide on view.

samhotchkiss:
Closing as already fixed on main. The cockpit dashboard already auto-fires and renders the morning briefing banner through _briefing_banner() in src/pollypm/cockpit_sections/dashboard.py, with regression coverage in tests/test_cockpit_dashboard.py.

samhotchkiss:
Reopening after verifying against updated origin/main. My earlier close was based on local branch state, not the shipped main branch, and that was incorrect.

samhotchkiss:
Resolved on current main. The global dashboard already includes the morning briefing banner flow in src/pollypm/cockpit_sections/dashboard.py, and the focused dashboard suite passes on current main: HOME=/tmp/pytest-505 uv run --extra dev pytest tests/test_cockpit_dashboard.py -q.

Issue #512, opened 2026-04-21
ux-016 — Live token-burn gauge in cockpit header


samhotchkiss:
Validated against current `main` (`e20dd9a`) on 2026-04-20.

Source: `issues/Claude/ux-016-token-burn-gauge.md`

**Surface:** cockpit header
**Effort:** medium · **Impact:** large

## Why flat

You only learn you're approaching a quota when things fail.

## Idea

Header gauge, updated every 10s:

```
claude(main): 4.2M / 5M (84%) · ~15m left at current burn
```

Red when >80%; blink on >95%. Pulls from the token ledger rollups.

samhotchkiss:
Resolved on current main. The global dashboard already includes the live token-burn gauge in src/pollypm/cockpit_sections/dashboard.py, and the focused dashboard suite passes on current main: HOME=/tmp/pytest-505 uv run --extra dev pytest tests/test_cockpit_dashboard.py -q.

PR #638, opened 2026-04-21
feat(cockpit): enrich global dashboard guidance

samhotchkiss:
## Summary
- add worker streaks, ranked whats-next suggestions, token-burn gauge, and morning-briefing banner to the global dashboard
- thread config_path into the dashboard renderer so briefing settings/state can be read safely
- cover the new dashboard helpers and dispatch wiring with focused tests

## Testing
- HOME=/tmp/pytest-cockpit-dashboard uv run --extra dev pytest tests/test_cockpit_dashboard.py -q

### DIFF
--- a/src/pollypm/cockpit.py
+++ b/src/pollypm/cockpit.py
@@ -93,7 +93,7 @@ def _build_cockpit_detail_inner(config_path: Path, kind: str, target: str | None
 def _build_cockpit_detail_dispatch(supervisor, config_path: Path, kind: str, target: str | None = None) -> str:
     config = supervisor.config
     if kind in ("polly", "dashboard"):
-        return _build_dashboard(supervisor, config)
+        return _build_dashboard(supervisor, config, config_path=config_path)
 
     if kind == "inbox":
         return _render_inbox_panel(config)
--- a/src/pollypm/cockpit_sections/dashboard.py
+++ b/src/pollypm/cockpit_sections/dashboard.py
@@ -14,6 +14,11 @@ from pollypm.cockpit_sections.base import (
     _format_tokens,
     _iso_to_dt,
 )
+from pollypm.cockpit_sections.health import (
+    format_project_health_scorecard,
+    project_health_rank,
+)
+from pollypm.cockpit_sections.just_shipped import _section_just_shipped
 from pollypm.cockpit_sections.project_dashboard import (
     _DASHBOARD_PROJECT_CACHE,
     _dashboard_project_tasks,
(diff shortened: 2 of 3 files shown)


samhotchkiss:
Needs fix — the PR's own tests pass (6/6), but it breaks two existing cockpit dashboard tests and drops two shipped sections without mentioning it in the PR body.

**Test failures** (reproduced in the worktree with `HOME=/tmp/pytest-cockpit-dashboard uv run --extra dev pytest tests/test_cockpit_just_shipped.py tests/test_cockpit_project_health.py`):

1. `test_cockpit_project_health.py::test_global_dashboard_surfaces_project_scorecards_sorted_by_health` — raises:
   ```
   File ".../cockpit_sections/dashboard.py", line 314, in _briefing_banner
       base_dir = Path(config.project.base_dir)
   AttributeError: '_Config' object has no attribute 'project'. Did you mean: 'projects'?
   ```
   `_briefing_banner` unconditionally touches `config.project.base_dir` even when `config_path is None`. That's fine in your new tests (which build a richer stub) but every other caller that used the old `_build_dashboard(supervisor, config)` 2-arg signature now explodes. Guard the access (e.g. `getattr(config, "project", None)` / return early if missing) so legacy test stubs and any residual callsites don't crash.

2. `test_cockpit_just_shipped.py::test_global_dashboard_includes_just_shipped_panel` — the PR removes the `_section_just_shipped` import and call from `_build_dashboard`, and also drops the whole `project_scorecards` / "─── Projects ───" block, but leaves the two tests that cover those sections in place. Those features landed in #595 and #591 respectively; ripping them out without a note in the PR body is scope drift. Either re-include them in the new layout (under "Now" or similar) or, if the intent is to supersede them with "What's next?" + streak header, delete/retarget the obsolete tests and call that out in the PR body + update the `#403 + #505 + #511 + #512 + #515` module docstring to include the removals.

Everything else in the diff looks reasonable — streaks helper, token-gauge builder, suggestion ranker, and the `config_path` thread-through are well-tested and self-contained. Once the two regressions above are addressed I'll re-run the full suite and merge.

Note: `pm cockpit-pane polly` and `pm cockpit-pane dashboard` both route to `PollyDashboardApp` (which uses `dashboard_data.load_dashboard`, not `_build_dashboard`), so none of these new sections are visible from the tmux-launched dashboard — this matches the prior pattern in #591/#595 and isn't introduced here, just flagging for later.

samhotchkiss:
Addressed the dashboard regressions:\n- guarded briefing generation when legacy config stubs don't expose project/base_dir\n- restored the shipped panel and project health scorecards alongside the new token/streak/suggestion sections\n\nValidation:\n- HOME=/tmp/pytest-638 uv run --extra dev pytest tests/test_cockpit_dashboard.py tests/test_cockpit_just_shipped.py tests/test_cockpit_project_health.py -q

samhotchkiss:
LGTM.

- Enriches the global dashboard (`_build_dashboard` in `cockpit_sections/dashboard.py`) with worker streaks, a ranked "What's next?" list, a token-burn gauge, and a morning-briefing banner; threads `config_path` through `_build_cockpit_detail_dispatch` so the briefing settings/state can be read without guessing paths.
- Verified against the issue references baked into the docstring (#403 / #505 / #511 / #512 / #515).
- 6 tests pass: `HOME=/tmp/pytest-cockpit-dashboard uv run --extra dev pytest tests/test_cockpit_dashboard.py -q`.
- Tmux sanity check: rendered the dashboard text via `_build_cockpit_detail_inner(..., 'dashboard')` against `pollypm.dev.toml` — headers, scorecards, "What's next?" suggestion list, briefing banner, and Done section all render as expected; `pm cockpit-pane polly` Textual app also renders without regressions.
- Minor nit (not blocking): `_briefing_banner` fires the morning briefing as a side effect during a render path. It's idempotent per-day via `state.last_briefing_date`, but worth tracking if render latency or write contention shows up — could move to a scheduled tick later.

Issue #660, opened 2026-04-22
magic-tui-5 — Rail: work-state glyph per session


samhotchkiss:
Validated against `main` (`dba2e99`) on 2026-04-21.

Part of the magical-tui-tweaks set (23 issues). **Build order: 5/23.**

**Surfaces:** `src/pollypm/cockpit_ui.py` (Textual cockpit users see via `pm`) AND `src/pollypm/cockpit_rail.py` (shared rail/router/runtime plumbing; legacy `PollyCockpitRail` still lives here)

**Effort:** small · **Impact:** high (glanceable fleet state)

## Why

Session status words ("writing", "reviewing", "idle") require parsing. Four glyphs you learn in a day don't.

## What

- Map session work-state → glyph:
  - Writing code (worker active, tokens streaming) → spinner (`⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏`)
  - Reviewing (Russell / critic run) → `✎`
  - Idle (healthy, awaiting work) → `·`
  - Stuck / needs intervention → `⚠`
  - Exited → `✕`
- Glyph sits at the start of each session row, color-coded.
- Spinner is the only animated variant — gate its animation on `is_viewer_attached()` from #656 (no spin when nobody's looking; freeze on the current spinner frame).
- Document the glyph alphabet in the rail help overlay (`?` keybind).

## Acceptance

- Each glyph maps cleanly to the session's current work state.
- Spinner pauses when unattached; resumes on reattach.
- `?` help shows the legend.
- `POLLY_CALM=1` forces the spinner glyph to static `…`.

## Depends on

#656

Issue #659, opened 2026-04-22
magic-tui-4 — Rail: heartbeat pulse per session


samhotchkiss:
Validated against `main` (`dba2e99`) on 2026-04-21.

Part of the magical-tui-tweaks set (23 issues). **Build order: 4/23.**

**Surfaces:** `src/pollypm/cockpit_ui.py` (Textual cockpit users see via `pm`) AND `src/pollypm/cockpit_rail.py` (shared rail/router/runtime plumbing; legacy `PollyCockpitRail` still lives here)

**Effort:** small · **Impact:** high (ambient health signal)

## Why

Today you have to look at a session's status column to know if a worker is stuck. A subtle pulse glyph beside each active session, animated on the heartbeat cadence, makes "alive vs stuck" something you absorb peripherally without clicking.

## What

- Add a 1-char pulse glyph to each active session row (e.g. `●` that breathes between `●`, `•`, `·`).
- Glyph cycle driven by the actual heartbeat timestamp: advance a frame each time a new heartbeat lands. If heartbeats stop, the glyph freezes — that *is* the signal.
- Use `is_viewer_attached()` from #656 — when unattached, don't run the animation timer at all; just render the latest static frame on data change.
- Sessions in `exited` / `auth_broken` states render a static red `✕` (no pulse).
- Gate behind `POLLY_CALM=1` env var to fully disable for users in constrained envs.

## Acceptance

- Attached + healthy session: pulse glyph animates at ~1 Hz keyed to heartbeat.
- Attached + stuck session: glyph freezes on last frame (that *is* the stuck indicator).
- Unattached: no animation CPU; static glyph.
- `POLLY_CALM=1`: static glyph always, no animation.

## Depends on

#656

PR #689, opened 2026-04-22
feat(tasks): improve cockpit review workflow

samhotchkiss:
Closes #657
Closes #658
Closes #662
Closes #664
Closes #665
Closes #671
Closes #672
Closes #675

Focused validation:
- `python -m py_compile src/pollypm/cockpit_task_review.py src/pollypm/cockpit_tasks.py src/pollypm/plugins_builtin/core_agent_profiles/profiles.py tests/test_tasks_ui.py`
- `.venv/bin/pytest -q tests/test_tasks_ui.py`

PR #691, opened 2026-04-22
perf(cockpit): cache rail hot paths and gate idle animation

### LINKED ISSUE #656: magic-tui-1 — Tmux-attached animation gate (foundational)
Validated against `main` (`dba2e99`) on 2026-04-21.

Part of the magical-tui-tweaks set (23 issues). **Build order: 1/23.**

**Surfaces:** `src/pollypm/cockpit_ui.py` (Textual cockpit users see via `pm`) AND `src/pollypm/cockpit_rail.py` (shared rail/router/runtime plumbing; legacy `PollyCockpitRail` still lives here)

**Effort:** small · **Impact:** high (foundational — gates CPU cost of every rail animation)

## Why

We want ambient animations in the rail (heartbeat pulse, work-state spinner, event ticker, sparklines) but we don't want them burning CPU when nobody is actually looking at the pane. If no tmux client is attached to the session, every animation frame is invisible work.

## What

- Add `cockpit_presence.is_viewer_attached(session_name: str) -> bool`.
- Implementation: `tmux list-clients -t <session>` — empty output ⇒ unattached.
- Cache result for ~2 s so we don't fork `tmux` on every frame.
- Resolve the current tmux session name via `$TMUX` + `tmux display-message -p '#S'` (or equivalent). Fall back to a safe default of `True` (render as if attached) if detection fails — we'd rather burn CPU than have an unreachable-looking cockpit.
- Every animated widget in the rail calls `is_viewer_attached()` before painting a frame; when False, widgets repaint only on real data change (no timer-driven frames).

## Acceptance

- `tmux detach` the pane → rail animation timers go to zero active frames within 2 s.
- Re-attach → animations resume within 2 s.
- No regression when running outside tmux (falls through to "assume attached").
- `cockpit_presence` has unit tests covering attached / unattached / `tmux` binary missing.

## Depends on

None.

### LINKED ISSUE #679: perf-magic-1 — Cache rail grouped[] structure between ticks
Validated against `main` (`dba2e99`) on 2026-04-21.

Part of the perf-magic set (10 issues). **Build order: 1/10.**

**Surface:** `src/pollypm/cockpit_rail.py:477-522` (`build_items`)
**Effort:** medium · **Impact:** high (biggest single CPU reduction in steady state)

## Why

`build_items()` rebuilds the entire `grouped` dict and instantiates fresh `CockpitItem` objects for every row on every render tick, even when nothing changed. The rail data is static aside from section collapse/expand and user-driven visibility filters — most ticks have identical output.

## What

- Add an instance-level cache: `self._items_cache: dict[CacheKey, list[CockpitItem]]`.
- Key: `(config_mtime, frozenset(hidden_keys), frozenset(collapsed_sections), registry_id)`.
- On `build_items()`: compute the key, return cached list if hit, otherwise build + store.
- Bust the cache when the cockpit router sends any of: set_selected_key, set_rail_width, set_cockpit_state.
- Keep the existing code path as the cold path behind the cache lookup.

## Acceptance

- Cold tick (cache miss) behaves identically to today.
- Repeated ticks with unchanged state return the same list instance (verify via `id()` in a test).
- Cache invalidates when any user-driven state changes.
- No behavioural regression in existing rail tests.

## Depends on

None.

### LINKED ISSUE #680: perf-magic-2 — Fold 3× load_config per rail tick into one
Validated against `main` (`dba2e99`) on 2026-04-21.

Part of the perf-magic set (10 issues). **Build order: 2/10.**

**Surface:** `src/pollypm/cockpit_rail.py` — `_state_path` (line 301), `_validate_state` (line 361), `_rail_registry` (line 528)
**Effort:** small · **Impact:** medium

## Why

Three separate call sites each call `load_config(self.config_path)` on every `build_items()` tick. `load_config` is mtime-cached so the TOML parse is cheap on a hit, but 3× stat syscalls + dict traversals per 1 Hz tick still adds measurable jitter.

## What

- In `build_items()`, load the config once at the top and pass it through to `_state_path`, `_validate_state`, and `_rail_registry` as an argument.
- Signature changes:
  - `_state_path(config=None)` — if None, keeps today's behavior.
  - `_validate_state(*, panes=None, target=None, config=None)` — adds optional `config` kwarg.
  - `_rail_registry(config=None)` — same pattern.
- Keep backwards compatibility (default None falls back to current behavior) so tests that call these directly still work.

## Acceptance

- `build_items()` performs exactly one `load_config` per tick (verifiable via patched counter in a test).
- All three helpers still work when called without a config argument.
- No behavioural regression.

## Depends on

None.

### LINKED ISSUE #681: perf-magic-3 — Cache cockpit_state.json dict in memory
Validated against `main` (`dba2e99`) on 2026-04-21.

Part of the perf-magic set (10 issues). **Build order: 3/10.**

**Surface:** `src/pollypm/cockpit_rail.py:317-346` (`_load_state`, `_write_state`, `rail_width`, `set_rail_width`, `selected_key`, `set_selected_key`)
**Effort:** small · **Impact:** medium-high (removes read-from-disk on every keypress)

## Why

Every getter and setter on cockpit state (`selected_key`, `rail_width`, etc.) opens `cockpit_state.json`, reads it, and re-parses JSON. Arrow-key navigation in the rail hits this on every keypress, causing noticeable latency on slow filesystems.

## What

- Introduce `self._state_cache: dict[str, object] | None = None` plus `self._state_mtime: float | None = None`.
- `_load_state()` — stat the file, if mtime unchanged return cached dict; else re-parse and cache.
- `_write_state(data)` — update the in-memory cache, write atomically, refresh mtime.
- External writes to `cockpit_state.json` (from other processes) still take effect on the next mtime change — cache is mtime-scoped, not "assume single writer".

## Acceptance

- Repeated `rail_width()` calls without mutation read the file at most once per mtime change.
- Setter updates both memory cache and disk.
- External process modifying the file → next getter sees the new value.
- All existing cockpit-state tests pass.

## Depends on

None.

### LINKED ISSUE #682: perf-magic-4 — Debounce rail state writes on resize
Validated against `main` (`dba2e99`) on 2026-04-21.

Part of the perf-magic set (10 issues). **Build order: 4/10.**

**Surface:** `src/pollypm/cockpit_rail.py:338-346` (`set_rail_width`)
**Effort:** small · **Impact:** medium

## Why

Dragging the rail border to resize calls `set_rail_width()` many times per second; each call runs `atomic_write_json()` on `cockpit_state.json`. Atomic writes (tempfile + fsync + rename) are expensive (~5-10 ms each) and cause visible I/O stalls during resize.

## What

- Add a debounce: on `set_rail_width()`, update in-memory state immediately (after #681 lands), but defer the disk write by ~500 ms, resetting the timer on each subsequent call.
- On cockpit exit / shutdown, flush any pending write synchronously.
- Same pattern is reusable for other frequent-mutation state (e.g. right_pane_id).

## Acceptance

- Rapid resize (10 updates/sec for 2 s) produces ≤ 2 disk writes.
- Final committed width on disk matches the last in-memory value.
- Cockpit shutdown flushes any pending write before exit.
- No loss of the last value on clean exit.

## Depends on

#681 (needs the in-memory cache so the immediate-read path works before the write commits).

### LINKED ISSUE #684: perf-magic-6 — Memoize rail hidden/collapsed filter sets
Validated against `main` (`dba2e99`) on 2026-04-21.

Part of the perf-magic set (10 issues). **Build order: 6/10.**

**Surface:** `src/pollypm/cockpit_rail.py:482-483`, `_hidden_rail_items`, `_collapsed_rail_sections`
**Effort:** small · **Impact:** low (small, pure win)

## Why

`_hidden_rail_items(config)` and `_collapsed_rail_sections(config)` scan the config to build frozensets on every `build_items()` tick. Their inputs only change when the user explicitly mutates state.

## What

- Cache both results on `PollyCockpitRail` keyed by the config dict's id or mtime.
- Invalidate on user state mutations (same triggers as #679's cache).
- Minor but composes with the `grouped[]` cache.

## Acceptance

- Repeated ticks with unchanged config return the same frozenset instance.
- State mutation correctly invalidates.
- Existing rail tests pass.

## Depends on

None.

samhotchkiss:
Closes #656
Closes #679
Closes #680
Closes #681
Closes #682
Closes #684

Scope note: #659 and #660 were split into stacked follow-up PR #699 so this branch stays perf-only.

Focused validation:
- `/Users/sam/dev/pollypm/.venv/bin/pytest -q tests/test_cockpit.py tests/test_rail_cli.py tests/test_core_rail_items.py`
- `python3 -m py_compile src/pollypm/cockpit_rail.py tests/test_cockpit.py`

### DIFF
--- a/src/pollypm/cockpit_rail.py
+++ b/src/pollypm/cockpit_rail.py
@@ -193,12 +193,6 @@ class CockpitPresence:
             return ARC_SPINNER[0]
         return ARC_SPINNER[spinner_index % len(ARC_SPINNER)]
 
-    def heartbeat_frame(self, spinner_index: int) -> str:
-        frames = ("♥", "♡")
-        if not self.should_animate():
-            return frames[0]
-        return frames[spinner_index % len(frames)]
-
 
 # ── Rail data model + router ─────────────────────────────────────────────
 #
@@ -870,7 +864,7 @@ class CockpitRouter:
                 return "\u25cf live"
             return "ready"
         if launch.session.role == "heartbeat-supervisor":
-            return "heartbeat"
+            return "watch"
         if launch.session.role == "triage":
             return "triage"
         return "live"
@@ -1916,8 +1910,8 @@ class PollyCockpitRail:
             return "\u25b2", PALETTE["alert_indicator"]
         if item.state == "dead":
             return "\u2715", PALETTE["dead"]
-        if item.state in {"heartbeat", "watch"}:
-            return self.presence.heartbeat_frame(self.spinner_index), PALETTE["live_indicator"]
+        if item.state == "watch":
+            return "\u25ce", PALETTE["live_indicator"]
         if item.state == "ready":
             return "\u25cf", PALETTE["sel_accent"]
         # Key-based fallbacks for items with no meaningful state
--- a/src/pollypm/cockpit_rail.py
+++ b/src/pollypm/cockpit_rail.py
@@ -357,6 +357,7 @@ class CockpitRouter:
         self.config_path = config_path
         self.service = PollyPMService(config_path)
         self.tmux = create_tmux_client()
+        self.presence = CockpitPresence(self.tmux)
         self._supervisor = None
         self._config_cache: object | None = None
         self._config_cache_mtime_ns: int | None = None
@@ -374,6 +375,15 @@ class CockpitRouter:
         # value: (db_mtime, git_mtime, is_active, has_working_task)
         # Skips re-opening SQLite on every 0.8s cockpit tick when nothing changed.
         self._project_activity_cache: dict[str, tuple[float, float, bool, bool]] = {}
+
+    def _presence(self) -> CockpitPresence:
+        presence = getattr(self, "presence", None)
+        if not isinstance(presence, CockpitPresence):
+            presence = CockpitPresence(self.tmux)
+            self.presence = presence
+        elif presence.tmux is not self.tmux:
+            presence.tmux = self.tmux
+        return presence
 
     def _load_config(self):
         try:
@@ -859,7 +869,7 @@ class CockpitRouter:
                 heartbeat=heartbeat,
             )
             if working:
-                return f"{self.presence.working_frame(spinner_index)} working"
+                return f"{self._presence().working_frame(spinner_index)} working"
             if launch.session.role == "worker":
                 return "\u25cf live"
             return "ready"
--- a/src/pollypm/cockpit_rail.py
+++ b/src/pollypm/cockpit_rail.py
@@ -111,6 +111,95 @@ POLLY_SLOGANS = [
 GUTTER = 2
 
 
+@dataclass(slots=True)
+class CockpitPresence:
+    """Presence-aware glyph helper for the rail renderer."""
+
+    tmux: object
+    _cache_ttl_seconds: float = 2.0
+    _cached_attached: bool | None = None
+    _cached_at: float = 0.0
+    _cached_session: str | None = None
+
+    def _calm_mode(self) -> bool:
+        value = os.environ.get("POLLY_CALM", "")
+        return value not in {"", "0", "false", "False", "no", "NO"}
+
+    def is_tmux_attached(self) -> bool:
+        """Return True when animation should behave as if tmux is attached."""
+        if self._calm_mode():
+            return False
+        session_name = self._current_session_name()
+        if session_name is None:
+            return True
+        now = time.monotonic()
+        if (
+            self._cached_session == session_name
+            and self._cached_attached is not None
+            and now - self._cached_at < self._cache_ttl_seconds
+        ):
+            return self._cached_attached
+        attached = self._probe_attached(session_name)
+        self._cached_session = session_name
+        self._cached_attached = attached
+        self._cached_at = now
+        return attached
+
+    def _current_session_name(self) -> str | None:
+        getter = getattr(self.tmux, "current_session_name", None)
+        if not callable(getter):
+            return None
+        try:
+            value = getter()
+            return str(value) if value else None
+        except Exception:  # noqa: BLE001
+            return None
+
+    def _probe_attached(self, session_name: str) -> bool:
+        list_clients = getattr(self.tmux, "list_clients", None)
+        if callable(list_clients):
+            try:
+                result = list_clients(session_name)
+            except Exception:  # noqa: BLE001
+                return True
+            if isinstance(result, str):
+                return bool(result.strip())
+            try:
+                return bool(list(result))
+            except TypeError:
+                return True
+        run = getattr(self.tmux, "run", None)
+        if not callable(run):
+            return True
+        try:
+            result = run(
+                "list-clients",
+                "-t",
+                session_name,
+                "-F",
+                "#{client_tty}",
+                check=False,
+            )
+        except Exception:  # noqa: BLE001
+            return True
+        stdout = getattr(result, "stdout", "") or ""
+        return bool(str(stdout).strip())
+
+    def should_animate(self) -> bool:
+        return self.is_tmux_attached()
+
+    def working_frame(self, spinner_index: int) -> str:
+        if not self.should_animate():
+            return ARC_SPINNER[0]
+        return ARC_SPINNER[spinner_index % len(ARC_SPINNER)]
+
+    def heartbeat_frame(self, spinner_index: int) -> str:
+        frames = ("♥", "♡")
+        if not self.should_animate():
+            return frames[0]
+        return frames[spinner_index % len(frames)]
+
+
 # ── Rail data model + router ─────────────────────────────────────────────
 #
 # Moved out of pollypm.cockpit in #404 so the routing concern (which rail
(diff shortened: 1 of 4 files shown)


samhotchkiss:
Reviewed on 2026-04-21. Verdict: **NEEDS FIX** — scope drift on magic-tui-4 (#659) and magic-tui-5 (#660).

**What is correctly implemented (solid perf work):**
- #656: `CockpitPresence.is_tmux_attached` with 2s TTL cache, animation gating via `should_animate()`, `POLLY_CALM` env var respect. Good.
- #679: grouped registrations cache (`_grouped_rail_registrations`) — safe interpretation given visibility predicates depend on per-tick context (launches/alerts). Fine.
- #680: `_load_config` mtime-cached and shared across `_state_path` / `_validate_state` / `_rail_registry` / `ensure_cockpit_layout` / `focus_right_pane`. Good.
- #681: `_state_cache` + `_state_cache_mtime_ns` invalidation. Good.
- #682: `_STATE_WRITE_DEBOUNCE_SECONDS = 0.25` + `_maybe_flush_state`. Works; note spec asked for ~500ms but 250ms is within spirit.
- #684: `_hidden_rail_items_cached` + `_collapsed_rail_sections_cached`. Good.

**Claim mismatches — the #689 pattern:**

1. **#659 per-session heartbeat pulse is not implemented.** Acceptance: "Add a 1-char pulse glyph to each active session row" with "Glyph cycle driven by the actual heartbeat timestamp" and "If heartbeats stop, the glyph freezes — that *is* the signal." Current PR only renders the ♥/♡ frame when `item.state in {"heartbeat", "watch"}` — that maps to the single `heartbeat-supervisor` role, not per-session health. And the frame advances with `self.spinner_index` (tick counter), not with real heartbeat timestamps, so a stuck session would still animate. Fix: in `cockpit_rail.py:1908-1920`, track last-heartbeat-seen per session_name and freeze `heartbeat_frame` when it hasn't advanced since the last tick. Render heartbeat glyph for every worker/operator-pm/reviewer session row, not just heartbeat-supervisor.

2. **#660 work-state glyph alphabet is not implemented.** Acceptance: map session work-state to `⠋⠙…` (writing), `✎` (reviewing), `·` (idle), `⚠` (stuck), `✕` (exited). Current PR keeps the old ARC_SPINNER for 'working' and `○` for 'idle'. No `✎` for reviewing, no `⚠` for stuck. Fix: extend `_indicator()` with the four additional glyphs keyed off `item.state`.

3. (minor) #682 spec asked for 500ms debounce; PR uses 250ms. Not a blocker, flagging only.

**Ask:** either (a) implement #659/#660 glyphs before merge, or (b) drop the `Closes #659` and `Closes #660` lines from the body and file separate follow-ups. The perf/`is_viewer_attached` work is otherwise mergeable.

Tests `tests/test_cockpit.py tests/test_rail_cli.py tests/test_core_rail_items.py` all pass (50 green).

samhotchkiss:
Review tick — still not ready on the two items flagged on the prior bounce. The perf half is solid; the overclaims are the blocker.

**#659 (heartbeat pulse per session) — NOT met.**
The diff introduces `CockpitPresence.heartbeat_frame(spinner_index)` at `src/pollypm/cockpit_rail.py:184-188`, but the only in-tree caller uses it nowhere on per-session rows. The only session-row behavior change is `cockpit_rail.py:394`, where `heartbeat-supervisor` role's state label switches from `"watch"` to `"heartbeat"` — still a static word, not a glyph, still only for `heartbeat-supervisor`. The acceptance says:
- "Add a 1-char pulse glyph to each active session row" — not added.
- "Glyph cycle driven by the actual heartbeat timestamp: advance a frame each time a new heartbeat lands" — still driven by `spinner_index` (a tick counter at `cockpit_rail.py:1672`), not a heartbeat_at column. Grep of the diff for `heartbeat_at` / `heartbeat_seen` / `last_heartbeat` returns zero hits.
- "Sessions in exited / auth_broken states render a static red ✕" — no `✕` glyph anywhere in diff.

The bounce explicitly called this out: "must fire for ALL active session rows, not just heartbeat-supervisor; must advance on actual heartbeat timestamp, not a tick counter."

**#660 (work-state glyph alphabet) — NOT met.**
Grep of the PR diff for the full glyph set returns zero hits for `✎`, `·`, `⚠`, `✕`. Only the arc spinner (`◜◝◞◟` already present on main) is animated, and only for role-in-(worker, operator-pm, reviewer) with a `working` flag. The acceptance requires the full five-state alphabet: spinner (writing) / `✎` (reviewing) / `·` (idle) / `⚠` (stuck) / `✕` (exited). None are landed. The `?` help legend is also not updated.

**What IS met (keep / rework the closes list accordingly):**
- **#656** — `CockpitPresence` class at `cockpit_rail.py:114-191` implements cached `is_tmux_attached()` with 2s TTL, `POLLY_CALM` env gate, and safe fallback-to-True. Embedded in `cockpit_rail.py` rather than the new `cockpit_presence.py` module the issue suggested, but behaviorally this is fine.
- **#679** (rail grouped cache) — `_grouped_rail_registrations` cache at `cockpit_rail.py:424-443` with identity-based key and proper invalidation via `_clear_rail_caches`.
- **#681** (cockpit_state.json in-memory cache) — `_state_cache` with mtime-scoped reuse at `cockpit_rail.py:510-539`.
- **#682** (fold 3x load_config per tick) — `_load_config()` at `cockpit_rail.py:385-395` with mtime invalidation.
- **#684** (debounce rail_width writes) — `_STATE_WRITE_DEBOUNCE_SECONDS = 0.25` + `_maybe_flush_state()` at `cockpit_rail.py:587-596`.
- 50/50 tests in `test_cockpit.py test_rail_cli.py test_core_rail_items.py` pass.

**Action:** drop `Closes #659` and `Closes #660` from the body (or retain and implement the glyph alphabet + per-session heartbeat-timestamp-driven pulse + exit/stuck ✕/⚠ branches). Keep the perf closes (#656/#679/#681/#682/#684) — those are genuinely done. Removing the `ready` label.

samhotchkiss:
Bouncing on third pass — same two issues as prior passes remain unmet.

**#659 (heartbeat pulse per session row) — UNMET**
- `src/pollypm/cockpit_rail.py:196` — `heartbeat_frame(self, spinner_index: int)` takes `spinner_index`, the generic rail tick counter (incremented in `run()` at line ~1673). The pulse alternates `♥`/`♡` on every rail tick, independent of actual heartbeat timestamps.
- `src/pollypm/cockpit_rail.py:1920` — `self.presence.heartbeat_frame(self.spinner_index)` is only reached via `item.state in {"heartbeat", "watch"}`, which is set only for `launch.session.role == "heartbeat-supervisor"` (line ~867). Worker / critic / reviewer session rows never hit this path — they get `● live`, `● ready`, or a spinner.
- Acceptance requires: pulse on EVERY active session row AND advanced by heartbeat timestamp (stuck sessions freeze — that IS the signal). A tick-driven pulse on only `heartbeat-supervisor` does neither. `grep -E "heartbeat_ts|last_heartbeat|heartbeat_at" src/pollypm/cockpit_rail.py` returns no timestamp-driven pulse advancement.
- Fix: read the heartbeat timestamp per session (already fetched at line ~865 for `_is_working_state`) and key the frame index on `hash(heartbeat_ts)` or a monotonically-advancing counter updated only when `heartbeat_ts` changes. Apply to every role that emits heartbeats, not just `heartbeat-supervisor`.

**#660 (work-state glyph alphabet) — UNMET**
- `grep -E "✎|·|⚠|✕" src/pollypm/cockpit_rail.py` returns nothing in the diff. None of the required glyphs (`✎` reviewing, `·` idle, `⚠` stuck, `✕` exited) are emitted by `_session_state` or `_indicator`.
- `_indicator` still returns `✕` (× / U+2715) for dead — not `✕` (U+2715 is actually the same codepoint, but the acceptance specifies the alphabet as a set; the other three glyphs are entirely absent).
- Acceptance requires the full 5-glyph alphabet mapped to work states (writing→spinner, reviewing→`✎`, idle→`·`, stuck→`⚠`, exited→`✕`) plus legend in the `?` help overlay.
- Fix: extend `_indicator` with the full mapping keyed off session role + state; add the legend to the rail `?` help overlay.

Other items (#656, #679, #680, #681, #682, #684) look fine — caching plumbing and the config/state fold are well-tested. Land those separately if it helps.

Bouncing back to needs fix.

samhotchkiss:
Follow-up note: I stripped the partial heartbeat/work-state glyph path back out of this PR so it is perf-only again. The actual #659/#660 implementation is stacked separately in #699.

samhotchkiss:
Fourth-pass review. The rescope to perf-only is correct — `Closes #659`/`#660` are gone, glyph code is stripped, #699 stacked — good. But this branch introduces a user-visible regression that blocks merge.

**Blocker: `AttributeError` on every worker/operator-pm/reviewer `working` tick.**

`src/pollypm/cockpit_rail.py:862`:

```python
if working:
    return f"{self.presence.working_frame(spinner_index)} working"
```

That `self.presence` is inside `CockpitRouter._session_state`. `CockpitRouter.__init__` (line ~356) never assigns `self.presence` — only `PollyCockpitRail.__init__` (line 1648) does. So every time a worker/operator/reviewer session is actively working, `_session_state` raises `AttributeError: 'CockpitRouter' object has no attribute 'presence'`.

The plugin wrapper in `src/pollypm/plugins_builtin/core_rail_items/plugin.py:82-84` swallows the exception with `except Exception` and returns `"idle"`, so instead of crashing you silently get the spinner replaced with static "idle" for every working session. That is a user-visible regression — a PR whose body says "no user-visible change" ships this.

Reproduction:
```
PYTHONPATH=src .venv/bin/python -c "
from pollypm.cockpit_rail import CockpitRouter
import pathlib, tempfile
tmp = pathlib.Path(tempfile.mkdtemp())
cfg = tmp / 'pollypm.toml'
cfg.write_text(f'[project]\nname=\"PollyPM\"\ntmux_session=\"pollypm\"\nbase_dir=\"{tmp}/.pollypm\"\n')
r = CockpitRouter(cfg)
r._is_pane_working = lambda *a, **k: True
class P: value='claude'
class S: name='worker_demo'; role='worker'; project='x'; provider=P()
class L: window_name='w'; session=S()
class W: name='w'; pane_dead=False; pane_current_command='claude'
print(r._session_state('worker_demo', [L()], [W()], [], 0))
"
# AttributeError: 'CockpitRouter' object has no attribute 'presence'
```

**Fix:** either (a) instantiate `self.presence = CockpitPresence(self.tmux)` in `CockpitRouter.__init__` and have `PollyCockpitRail` reuse `router.presence`, or (b) revert line 862 to the prior `spinners[spinner_index % 4]` list since this PR claims to be perf-only. Option (a) is preferred so the presence cache actually gates the spinner frame in the real cockpit, which is one of the perf wins #656 is supposed to deliver.

**Unit test gap:** no test covers the `working` branch of `CockpitRouter._session_state`. Add one with `_is_pane_working` monkeypatched to True; it will fail on current HEAD. The existing `test_cockpit_router_session_state_uses_heartbeat_state` only hits the `"watch"` branch, which is why 50 greens missed this.

**Other perf bullets look solid** — keep them:
- #656: `CockpitPresence` with 2s TTL cache, `POLLY_CALM` gate, safe fallback. Lives in `cockpit_rail.py` not the suggested new `cockpit_presence.py` module — behaviorally fine.
- #679: `_grouped_rail_registrations` cache keyed on `id(config)` with `_clear_rail_caches` invalidation. Fine.
- #680: `_load_config()` at `cockpit_rail.py:381-395` shared across `_state_path` / `_validate_state` / `_rail_registry` / `ensure_cockpit_layout` / `focus_right_pane`. Good.
- #681: `_state_cache` + `_state_cache_mtime_ns` mtime invalidation at lines 502-540. Good.
- #682: `_STATE_WRITE_DEBOUNCE_SECONDS = 0.25` — spec said 500ms but 250ms is within spirit (flagged on prior pass, accepted).
- #684: `_hidden_rail_items_cached` + `_collapsed_rail_sections_cached`. Good.

Removing `ready`, adding `needs fix`. Fix the `self.presence` reachability bug and the missing working-path test, then relabel `ready`.

samhotchkiss:
Fixed the fourth-pass blocker on this branch:

- `CockpitRouter` now owns a reachable `CockpitPresence` helper for the working-path state render, so the perf-only branch no longer silently degrades working sessions to `idle`
- added focused regression coverage for the working branch so this path is exercised in tests

Focused validation:
- `uv run --extra dev python -m pytest -q tests/test_cockpit_idle_state.py::test_worker_session_state_stays_working_when_codex_shows_interrupt_banner tests/test_core_rail_items.py::test_build_items_preserves_working_project_session_state`
- `uv run --extra dev python -m pytest -q tests/test_cockpit_idle_state.py tests/test_core_rail_items.py`
- `9 passed`

samhotchkiss:
Rebased onto current `main` after the conflict-only bounce. Focused validation rerun:

- `python3 -m py_compile src/pollypm/cockpit_rail.py tests/test_cockpit_idle_state.py tests/test_core_rail_items.py`
- `/Users/sam/dev/pollypm/.venv/bin/pytest -q tests/test_cockpit_idle_state.py tests/test_core_rail_items.py`
- `9 passed`

samhotchkiss:
Approved and merging.

Verified:
- #681 critical bug (AttributeError on CockpitRouter.presence) fixed: self.presence = CockpitPresence(self.tmux) added at cockpit_rail.py:360 with defensive _presence() at :379 and call-site rewritten at :872.
- New regression test tests/test_core_rail_items.py::test_build_items_preserves_working_project_session_state exercises the working branch via build_items(spinner_index=1).
- Scope clean: gh pr diff shows only cockpit_rail.py + 3 test files (MERGEABLE).
- Full suite: 51 passed in tests/test_cockpit.py + tests/test_rail_cli.py + tests/test_core_rail_items.py.
- Smoke: pm launches, rail renders with no AttributeError.

PR #693, opened 2026-04-22
feat(cockpit): expand settings ergonomics and palette access

samhotchkiss:
Closes #661
Closes #673
Closes #674
Closes #676
Closes #678
Refs #669
Refs #670

Focused validation:
- `uv run --with pytest python -m pytest tests/test_command_palette.py tests/test_project_settings_ui.py tests/test_settings_ui.py -q`
- `uv run --with pytest python -m pytest -q tests/test_project_settings_ui.py tests/test_settings_ui.py tests/test_cockpit.py`

PR #702, opened 2026-04-22
refactor(state): retire legacy event tables

### LINKED ISSUE #411: refactor(storage): retire legacy state tables (events / alerts / task_notifications)
Tracking the deprecated **state database** tables that still existed before `#702`:

- `events`
- `alerts`
- `task_notifications`

This issue no longer tracks the live `notification_staging` work-db seam; that follow-up is now tracked separately in #704.

`activity_events` does not currently exist as a physical legacy table in `main`; the remaining real retirement work here is the three state-db tables above.

samhotchkiss:
Stops recreating the deprecated state-db tables (`events`, `alerts`, `task_notifications`), routes the surviving `StateStore` helpers through `messages`, and adds migration 15 to copy old rows forward before dropping the physical tables.

Closes #411

### DIFF
--- a/src/pollypm/storage/state.py
+++ b/src/pollypm/storage/state.py
@@ -49,13 +49,32 @@ CREATE TABLE IF NOT EXISTS sessions (
     window_name TEXT NOT NULL
 );
 
-CREATE TABLE IF NOT EXISTS events (
+CREATE TABLE IF NOT EXISTS messages (
     id INTEGER PRIMARY KEY AUTOINCREMENT,
-    session_name TEXT NOT NULL,
-    event_type TEXT NOT NULL,
-    message TEXT NOT NULL,
-    created_at TEXT NOT NULL
+    scope TEXT NOT NULL,
+    type TEXT NOT NULL,
+    tier TEXT NOT NULL DEFAULT 'immediate',
+    recipient TEXT NOT NULL,
+    sender TEXT NOT NULL,
+    state TEXT NOT NULL DEFAULT 'open',
+    parent_id INTEGER,
+    subject TEXT NOT NULL,
+    body TEXT NOT NULL DEFAULT '',
+    payload_json TEXT NOT NULL DEFAULT '{}',
+    labels TEXT NOT NULL DEFAULT '[]',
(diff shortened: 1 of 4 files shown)


samhotchkiss:
Scope matches #411 (three legacy state-db tables) and migration 15 copies rows forward before dropping — data-loss safety looks right. Tests in the focused set pass. One blocker though:

- `tests/test_db_vacuum.py::test_incremental_vacuum_reclaims_space_after_bulk_delete` fails with `sqlite3.OperationalError: no such table: events` (tests/test_db_vacuum.py:58 `DELETE FROM events WHERE event_type = 'bulk.test'`). The test inserts via `store.record_event` (now routed to `messages`) but then deletes from `events` directly. Fix: retarget the `DELETE`/post-delete assertion paths at `messages` (e.g. `DELETE FROM messages WHERE type='event' AND payload_json LIKE '%bulk.test%'` or a narrower where-clause) so the vacuum test exercises the new table. The rest of the run was clean — all other currently-failing tests are pre-existing on `origin/main`.

Bouncing to `needs fix` to get the vacuum test green.

samhotchkiss:
Addressed the bounce:\n\n- retargeted the incremental-vacuum regression test to delete its bulk rows from the unified `messages` ledger instead of the dropped `events` table\n- removed the lingering `last_heartbeat_at()` fallback query against `events`, since migration 15 already copies heartbeat rows forward before the table is dropped\n\nValidation:\n- `python3 -m py_compile src/pollypm/storage/state.py tests/test_db_vacuum.py`\n- `uv run --with pytest python -m pytest tests/test_db_vacuum.py tests/test_state.py tests/test_task_assignment_notify.py -q`\n- result: `56 passed`\n\nRelabeling to `ready`.

samhotchkiss:
Rebased onto post-#693/#703 main. Migration 15 drops `events`/`alerts`/`task_notifications` after copying into `messages`; `notification_staging` correctly rescoped to #704 per updated #411 body. 82/82 tests pass across `test_db_vacuum.py`, `test_events_retention.py`, `test_store_engine.py`, `test_store_registry.py`, `test_store_writers.py`, `test_backup_restore.py`. Earlier bounce's test regression was addressed by commit `3950f42` (remove legacy event-table fallbacks).

Issue #720, opened 2026-04-22
upgrade-8 — post-upgrade rail summary + --recycle-all / --recycle-idle escape hatches


samhotchkiss:
Validated against `main` (`e16b205`) on 2026-04-22.

Part of the auto-upgrade-flow set (8 issues). **Build order: 8/8.**

**Surface:** rail top bar, `pm upgrade` CLI flags, supervisor recycle path
**Effort:** medium · **Impact:** medium (closes the loop — user sees what happened + has a hard-refresh option for prompt-critical releases)

## Why

After an upgrade lands, the user should see exactly what happened (versions, notified sessions, recycles). And for the occasional prompt-critical release where the in-conversation notice isn't enough, they need a hard-recycle option.

### Post-upgrade rail summary

- After #719's flow completes, the rail shows a persistent summary pill for 4 hours (or until dismissed):
  `Upgraded v1.2.0 → v1.3.2 · 3 sessions notified · 2 pending restart · 5 idle recycled`
- Counts reflect the outcome of #718 (notified) + any `--recycle` (recycled) + sessions that received the notice but haven't turned over since (pending restart).
- Dismissable with `x`. Survives rail restart (persisted in state).

### Recycle escape hatches

- `pm upgrade --recycle-all`: after normal upgrade + notice injection, tear down every live session (workers, Polly, Russell) via the existing supervisor relaunch path. Fresh session = fresh system prompt with the new guide. Use this when a release changes persona / delegation rules in ways the in-conversation notice can't reliably patch.
- `pm upgrade --recycle-idle`: only recycle sessions that have had no active turn in the last N minutes (default 30). Sessions actively mid-work are left alone; idle ones get a hard refresh. Good middle ground for day-to-day upgrades.
- Both flags respect the same "idle" definition as the existing 2-hour memory-management kill (last heartbeat + recent tokens), but with a tighter window.

## Acceptance

- Post-upgrade, summary pill appears with accurate counts. Dismiss persists.
- `pm upgrade --recycle-all` kills and respawns all sessions correctly; cockpit remains usable throughout.
- `pm upgrade --recycle-idle` leaves actively-turning sessions running; respawns the rest.
- No flag = default behavior from #716 (install + notify, no recycle).

## Depends on

#716 (upgrade wrapper), #718 (notice injection), #719 (rail restart flow)

samhotchkiss:
[...]

Issue #719, opened 2026-04-22
upgrade-7 — one-click rail upgrade keybind + auto-restart of rail/daemons


samhotchkiss:
Validated against `main` (`e16b205`) on 2026-04-22.

Part of the auto-upgrade-flow set (8 issues). **Build order: 7/8.**

**Surface:** rail top bar (`cockpit_ui.py`), rail daemon, heartbeat daemon lifecycle
**Effort:** medium · **Impact:** high (the end-user's magic moment — one key and you're upgraded)

## Why

The pill in #715 needs an actual action behind it. Users should hit one key from the rail and be on the new version a minute later, without losing any conversation state.

## What

- When the update pill from #715 is visible, keybind `u` triggers the upgrade flow.
- Open a new tmux pane (or split the current one) for the upgrade process.
- Run steps in order:
  1. `pm migrate --check` (#717) — abort on failure, leave user on old version.
  2. `pm upgrade` (#716) — streams output into the upgrade pane.
  3. Wait for install success.
  4. Restart **just the Polly processes** that hold stale code: current cockpit/rail, rail daemon (if running, via `~/.pollypm/rail_daemon.pid`), heartbeat daemon. **Do not touch worker tmux windows or Polly/Russell CLI panes** — those are independent processes with only on-disk code coupling and pick up the new `pm` CLI automatically.
  5. Call the system-update injection (#718).
  6. Hand off to the post-upgrade summary (#720).
- Rail stays usable during the upgrade — rendering continues from the old process until step 4 kicks in, at which point the rail auto-relaunches into the new binary (same tmux window).
- Upgrade pane: auto-closes on success, stays open on failure with the error visible.

## Acceptance

- `u` with pill visible opens the upgrade pane.
- Pane runs migration check → install → notice injection in sequence.
- Rail and daemons relaunch automatically on success. User sees the new version reflected in the rail within a few seconds.
- Worker / Polly / Russell tmux windows are not killed.
- On failure, old rail keeps running; pane stays open with the specific error.

## Depends on

#715 (pill), #716 (upgrade), #717 (migrate), #718 (notice)

Issue #718, opened 2026-04-22
upgrade-6 — in-session system-update notice injection (prompt refresh without restart)


samhotchkiss:
Validated against `main` (`e16b205`) on 2026-04-22.

Part of the auto-upgrade-flow set (8 issues). **Build order: 6/8.**

**Surface:** `src/pollypm/supervisor.py` or session-services, new helper `src/pollypm/upgrade_notice.py`
**Effort:** medium · **Impact:** high (refreshes Polly/Russell/worker behavior without tearing down conversations)

## Why

When PollyPM is upgraded while sessions are live, `claude --resume` and `codex --resume` preserve conversation state — the system prompt from turn 1 is baked into history and **new system prompts are not merged**. Naive "restart session" would lose context too. Solution: inject a user-turn notice that tells the model to re-read its guide from disk (which IS updated with the new version).

## What

- New helper: `inject_system_update_notice(supervisor, old_version, new_version)`.
  - Iterates all live sessions from the supervisor (Polly, Russell, each worker).
  - For each, picks the role-appropriate guide path:
    - Worker → `docs/worker-guide.md`
    - Polly → `plugins_builtin/core_agent_profiles/profiles/polly-operator-guide.md`
    - Russell → the reviewer guide path
  - Sends a canonical notice via existing `pm send` / pane send-keys (same channel used for `pm notify`):
    ```
    <system-update>
    PollyPM was upgraded from v<old> → v<new> while this session was running.
    Before your next action, re-read your operating guide at <path>.
    It supersedes any prior operating instructions in this conversation.
    If anything in the new guide conflicts with what you were about to do,
    pause and re-plan from the updated instructions.
    </system-update>
    ```
  - Queue delivery so the notice lands at a turn boundary — don't interrupt mid-tool-call.
- Called automatically by `pm upgrade` (#716) after install succeeds.
- Also surfaces as `pm upgrade --notify-only` for testing the injection path without actually installing.

## Acceptance

- Every live session (Polly, Russell, workers) receives the notice after `pm upgrade`.
- Notice text is role-specific (correct guide path for each role).
- Sessions started after the upgrade completes do NOT receive the notice (they already have the new code).
- If a session is unresponsive, the upgrade does not hang — timeout with a warning.
- Reasonable logging so a user can tell which sessions got the notice successfully.

## Honest caveat

[...]

Issue #717, opened 2026-04-22
upgrade-5 — schema migration gate (pm migrate --check + refuse-start)


samhotchkiss:
Validated against `main` (`e16b205`) on 2026-04-22.

Part of the auto-upgrade-flow set (8 issues). **Build order: 5/8.**

**Surface:** `src/pollypm/store/` migrations subsystem, new `pm migrate` CLI, Polly startup path
**Effort:** medium · **Impact:** high (safety gate — prevents the one failure mode that could actually corrupt user state)

## Why

Recent versions have shipped schema migrations (e.g. #702 retire legacy event tables). An upgrade that installs new code without applying migrations leaves the DB in an inconsistent state — Polly may read from tables that no longer exist or write in a format the new code doesn't understand.

## What

- `pm migrate --check`: dry-run. Clone the live DB to `~/.pollypm/migration-check.db`, apply every pending migration against it, roll back on any error. Returns 0 on success, non-zero (with a specific error) otherwise.
- `pm migrate --apply`: actually apply pending migrations to the real DB. Idempotent — safe to call when nothing pending. Records each applied migration in a `schema_migrations` table.
- Invoked automatically by `pm upgrade` (#716) as the first step after install.
- **Refuse-start behavior:** on cockpit/rail launch, if there are unapplied migrations, Polly prints a specific error with recovery instructions (`pm migrate --apply` or `pm migrate --check` to see what's pending) and exits with non-zero. Does not silently proceed.

## Acceptance

- `pm migrate --check` on a clean-and-current DB returns 0.
- `pm migrate --check` on a DB with pending migrations reports exactly what would change.
- `pm migrate --apply` applies migrations + records them; second call is a no-op.
- Cockpit launch with unapplied migrations: clear error, non-zero exit, no DB writes.
- A synthetic failing migration is caught by `--check` before it can corrupt the real DB.

## Depends on

None (independent of the rest of the set, but #716 calls it).

PR #723, opened 2026-04-22
feat(cli): pm upgrade wrapper — detect installer + delegate (#716)

### LINKED ISSUE #716: upgrade-4 — pm upgrade CLI wrapper (detect installer, delegate)

samhotchkiss:
Adds the canonical upgrade command. `pm upgrade` detects how PollyPM was installed (uv tool → pip → brew → npm, in that priority order) and shells out to the right package manager, with channel-aware flags for beta.

## Flow

1. **Detect installer** via `shutil.which`-style probes against each package manager's list output.
2. **Migration gate** — runs the `pm migrate --check` from #717 before any mutation. Soft-passes with a "not yet implemented" note until #717 ships so the flow works on a fresh checkout.
3. **Check-only** (`--check-only`) prints the plan and exits. Otherwise runs the installer (10-min timeout).
4. **Read new version** via `pip show pollypm` (the old Python process's `__version__` is cached and can't be re-read in-place).
5. **Notice injection** — calls #718's `inject_system_update_notice` (soft-passes similarly until that lands).
6. **Changelog diff** — prints `CHANGELOG.md` section above the new version if present.

## Flags

- `--channel stable|beta` — one-off override (config's `release_channel` wins otherwise).
- `--check-only` — run the migration check + plan the command, don't install.
- `--recycle-all` / `--recycle-idle` — wired now so the CLI surface is stable; actual recycling ships in #720.

## Design notes

- **Never crashes on a fresh install.** Unknown installer prints the four manual commands and exits 1, never mutates anything.
- **Streams `[step] ...` markers** so the rail one-click pane in #719 can tail progress.
- `installer_overrides` / `plan_only` / `emit` are test seams — no real `uv`/`pip`/`brew`/`npm` fire in unit tests.
- **Release-channel defaults safely.** Reads `config.pollypm.release_channel` via `release_check._resolve_channel` when #714 is present; otherwise falls back to `"stable"`. No hard dep across branches.

## Validation

25 new unit tests cover detection priority, per-installer plan shape, channel routing (stable vs beta), migration-check gate behavior, plan-only seam, CLI wiring (exit codes + output shape), changelog-diff scraping. Regression spot-check on `test_cli.py` and `test_work_cli.py` (58/58 passing).

## Dependencies

Works independently. Integrates with #714 (channel resolution), #717 (migration gate), #718 (notice injection), #720 (recycle behavior) once those land.

Closes #716 · Refs #709

### DIFF
--- a/src/pollypm/cli.py
+++ b/src/pollypm/cli.py
@@ -50,6 +50,7 @@ from pollypm.cli_features.maintenance import debug_app, register_maintenance_com
 from pollypm.cli_features.projects import register_project_commands
 from pollypm.cli_features.session_runtime import register_session_runtime_commands
 from pollypm.cli_features.ui import register_ui_commands
+from pollypm.cli_features.upgrade import register_upgrade_commands
 from pollypm.cli_features.workers import register_worker_commands
 
 
@@ -170,6 +171,7 @@ app.add_typer(downtime_app, name="downtime")
 register_ui_commands(app)
 register_project_commands(app)
 register_maintenance_commands(app)
+register_upgrade_commands(app)
 register_worker_commands(app)
 register_session_runtime_commands(app, helpers=sys.modules[__name__])
 
--- a/src/pollypm/cli_features/upgrade.py
+++ b/src/pollypm/cli_features/upgrade.py
@@ -1,0 +1,87 @@
+"""``pm upgrade`` command registration (#716).
+
+Thin Typer shim over :mod:`pollypm.upgrade`. All logic lives there so
+the rail one-click (#719) can invoke the same entry point.
+"""
+
+from __future__ import annotations
+
+import typer
+
+from pollypm.cli_help import help_with_examples
+
+
+def register_upgrade_commands(app: typer.Typer) -> None:
+    @app.command(
+        help=help_with_examples(
+            "Upgrade PollyPM to the latest release on the current channel.",
+            [
+                ("pm upgrade", "install the latest release on the active channel"),
+                ("pm upgrade --check-only", "report what would happen, don't install"),
+                ("pm upgrade --channel beta", "one-off beta upgrade (config channel wins otherwise)"),
+            ],
+        )
+    )
+    def upgrade(
+        channel: str = typer.Option(
+            "",
+            "--channel",
+            help="stable | beta. Empty = read release_channel from config.",
+        ),
+        check_only: bool = typer.Option(
+            False,
+            "--check-only",
+            help="Show what would happen — run the migration check, fetch target version, exit without installing.",
+        ),
+        recycle_all: bool = typer.Option(
+            False,
+            "--recycle-all",
+            help="After install, tear down and respawn every live session (Polly, Russell, workers). Use for prompt-critical releases where the in-conversation notice isn't enough.",
+        ),
+        recycle_idle: bool = typer.Option(
+            False,
+            "--recycle-idle",
+            help="After install, respawn only sessions with no active turn in the last 30 minutes. Active work continues untouched.",
+        ),
+    ) -> None:
+        """Detect the install method and delegate to the right package manager."""
+        from pollypm.upgrade import read_changelog_diff, upgrade as run_upgrade
+
+        # ``release_check._resolve_channel`` is the canonical reader for
+        # ``config.pollypm.release_channel`` (from #713/#714). Until both
+        # land, fall back to stable if the helper isn't importable —
+        # never crash on a fresh checkout just because the config field
+        # is missing.
+        try:
+            from pollypm.release_check import _resolve_channel
+            default_channel = _resolve_channel(None)
+        except ImportError:
+            default_channel = "stable"
+        resolved_channel = channel or default_channel
+        result = run_upgrade(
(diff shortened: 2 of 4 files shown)


samhotchkiss:
Blocking findings:

- `src/pollypm/upgrade.py:258-277` and `src/pollypm/upgrade.py:407-428` determine the post-install version via `pip show pollypm` regardless of installer. That breaks the main advertised cases here (`uv`, `brew`, `npm`). I reproduced it locally by stubbing a successful `brew` upgrade: the flow reported `installed 0.1.0`, injected a `0.1.0->0.1.0` notice, and returned `upgraded 0.1.0 → 0.1.0`. So the wrapper can succeed while reporting the wrong version and suppressing the changelog path.
- `src/pollypm/upgrade.py:217-255` silently soft-skips both the migration gate and notice injection when those modules are absent. Because this PR targets `main` directly rather than stacking on `#724` / `#725`, merging it first would ship a `pm upgrade` that does *not* actually enforce the safety/notice behavior described in the module docstring.
- `src/pollypm/upgrade.py:432-452` looks up `CHANGELOG.md` from `Path.cwd()`. For a normal installed tool invocation, cwd is just wherever the operator ran `pm upgrade`, so the changelog diff will usually be blank even when the package ships one.

Focused slice I ran: `HOME=/tmp/pytest-pr723 uv run --extra dev pytest -q tests/test_upgrade.py` -> `25 passed`. The missing coverage is post-install version detection for non-pip installers.

samhotchkiss:

**Scope match:** ✓ Delivers `pm upgrade` CLI wrapper per #716 — installer detection, per-manager plans, migration gate hook, notice-injection hook.

**Rebase:** Same 876-commit drift from main. Cherry-picked the single feature commit onto current main — clean.

**Code quality:**
- Priority order uv → pip → brew → npm → unknown is the right call; `uv tool` users are the dominant install path
- `detect_installer(overrides=...)` test seam avoids real `shutil.which` / subprocess calls in unit tests
- `plan_upgrade` encodes the per-installer beta flag correctly (uv `--prerelease allow`, pip `--pre`, npm `@beta`, brew = no-op with a user-visible warning)
- Import-if-available pattern for #717 (`pollypm.store.migrations.check_pending`) and #718 (`pollypm.upgrade_notice.inject_system_update_notice`) means this merges cleanly regardless of merge order
- `_read_new_version` correctly notes the in-process `pollypm.__version__` staleness and shells out to `pip show` instead
- Timeout on install subprocess is 600s (10m) — generous but bounded
- `plan_only` + `check_only` flags are distinct: check_only runs the migration probe but stops short of install; plan_only stops even earlier (pure unit-test seam)
- `UpgradeResult` is a clean structured return — callers (rail one-click in #727) can render it without re-parsing

**Noted for follow-up (not blocking):**
- `_resolve_channel` import fallback in cli_features/upgrade.py is now redundant after #721 merged, but the `except ImportError` doesn't hurt
- Recycle flags (`--recycle-all`/`--recycle-idle`) are plumbed but deferred to a future PR — matches the acknowledged scope

**Tests:** 25/25 pass. test_cli.py regression green (14/14).

PR #724, opened 2026-04-22
feat(store): schema migration gate (pm migrate + refuse-start)

### LINKED ISSUE #717: upgrade-5 — schema migration gate (pm migrate --check + refuse-start)

samhotchkiss:
## Summary
- Adds `pm migrate --check` (dry-run against `~/.pollypm/migration-check.db` clone) and `pm migrate --apply` (idempotent live upgrade) CLI surfaces, wired through the existing `StateStore._MIGRATIONS` and `_WORK_MIGRATIONS` runners.
- Adds a refuse-start gate: `pm up`, `pm cockpit`, `pm cockpit-pane`, and the headless rail daemon now check the workspace `state.db` read-only before opening the store. Pending migrations print a recovery message and exit non-zero instead of silently auto-migrating.
- New `pollypm.store.migrations` module exposes `inspect`, `check_against_clone`, `apply`, and `require_no_pending_or_exit` helpers plus a unified `schema_migrations` audit table that mirrors both `schema_version` and `work_schema_version`. Bypassable via `POLLYPM_SKIP_MIGRATION_GATE` so `pm migrate` can itself open the store.

Closes #717

## Test plan
- [x] `uv run pytest tests/test_migration_gate.py` — all 9 new tests pass (check on clean DB, apply idempotent, check detects pending, synthetic failing migration caught, cockpit refuses start, bypass env honored, `pm up` propagates the exit, CLI up-to-date, CLI apply reporting).
- [x] `uv run pytest tests/test_cli.py tests/test_state.py` — existing regressions (incl. `test_importing_cli_defers_heavy_runtime_modules` which pins SQLAlchemy-laziness at import time).
- [x] `uv run pm migrate --check` on the developer's live DB correctly reports a pending migration without mutating it.
- [ ] Apply path on a CI-style fresh DB (covered by test_cli_migrate_apply_reports_applied).

### DIFF
--- a/src/pollypm/cli.py
+++ b/src/pollypm/cli.py
@@ -47,6 +47,7 @@ from pollypm.cli_help import help_with_examples
 from pollypm.cli_features.alerts import alert_app, heartbeat_app, session_app
 from pollypm.cli_features.issues import issue_app, itsalive_app, report_app
 from pollypm.cli_features.maintenance import debug_app, register_maintenance_commands
+from pollypm.cli_features.migrate import register_migrate_commands
 from pollypm.cli_features.projects import register_project_commands
 from pollypm.cli_features.session_runtime import register_session_runtime_commands
 from pollypm.cli_features.ui import register_ui_commands
@@ -172,6 +173,7 @@ register_ui_commands(app)
 register_project_commands(app)
 register_maintenance_commands(app)
 register_upgrade_commands(app)
+register_migrate_commands(app)
 register_worker_commands(app)
 register_session_runtime_commands(app, helpers=sys.modules[__name__])
 
@@ -248,6 +250,27 @@ def _load_supervisor(config_path: Path):
     from pollypm.service_api import PollyPMService
 
     return PollyPMService(config_path).load_supervisor()
+
+
+def _enforce_migration_gate(config_path: Path) -> None:
+    """Refuse-start guard: bail out if the workspace state.db is behind (#717).
+
+    Skipped when ``POLLYPM_SKIP_MIGRATION_GATE`` is set — ``pm migrate``
+    turns the bypass on explicitly so the apply path can itself open the
+    store. A config that cannot be loaded is treated as "no gate to
+    enforce yet" so onboarding / first-run paths keep working.
+    """
+    from pollypm.store import migrations as _migrations
+
+    if _migrations.bypass_env_is_set():
+        return
+    try:
+        from pollypm.config import load_config
+        config = load_config(config_path)
+        db_path = config.project.state_db
+    except Exception:  # noqa: BLE001
+        return
+    _migrations.require_no_pending_or_exit(db_path)
 
 
 def _account_label(supervisor, account_name: str) -> str:
@@ -423,6 +446,7 @@ def up(
         typer.echo(f"Config not found at {config_path}. Starting onboarding.")
         onboard(config_path=config_path, force=False)
         return
+    _enforce_migration_gate(config_path)
     supervisor = _load_supervisor(config_path)
     # CoreRail owns startup orchestration — it drives plugin host load,
     # state store readiness, and Supervisor boot (which runs ensure_layout,
--- a/src/pollypm/cli_features/migrate.py
+++ b/src/pollypm/cli_features/migrate.py
@@ -1,0 +1,126 @@
+"""`pm migrate` — schema migration gate CLI (#717).
+
+Contract:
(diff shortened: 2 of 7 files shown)


samhotchkiss:
Blocking findings:

- `src/pollypm/cli_features/ui.py:37-52` leaves `pm ui` and `pm accounts-ui` outside the new gate, but those entrypoints still boot `PollyPMApp` / `AccountsApp`, which immediately load the supervisor and open `state.db`. That is a direct bypass around the refuse-start policy.
- `src/pollypm/store/migrations.py:227-235` copies only the main `state.db` file for `pm migrate --check`. In WAL mode, committed changes can still be sitting in `state.db-wal` / `state.db-shm`, so the dry-run can inspect a stale or inconsistent snapshot and miss the real migration state.
- `src/pollypm/store/migrations.py:266-348` applies state and work migrations in separate writer lifecycles with no shared rollback/compensation. If the second phase fails after the first commits, `pm migrate --apply` exits as a failure but leaves the live DB half-migrated.
- `src/pollypm/store/migrations.py:113-121` / `135-163` collapse every read-only SQLite error into the “DB missing” path. Locked, unreadable, or corrupted DBs are therefore misreported as “all migrations pending”, which hides the real fault and points operators at the wrong recovery step.

Focused slice I ran: `HOME=/tmp/pytest-pr724 uv run --extra dev pytest -q tests/test_migration_gate.py` -> `9 passed`.

samhotchkiss:

**Scope match:** ✓ Delivers the #717 migration gate — `pm migrate --check` (clone-based dry-run), `pm migrate --apply`, and the refuse-start guard on `pm cockpit`, `pm cockpit-pane`, `pm up`, and the headless rail daemon. Never mutates the live DB in `--check`.

**Rebase:** Same 876-commit drift pattern. Cherry-picked onto current main with one trivial conflict resolution in `cli.py` (`register_upgrade_commands` from #723 already landed; added `register_migrate_commands` next to it). 

**Code quality:**
- `store/migrations.py` is well-factored — `inspect` / `check_against_clone` / `apply` / `require_no_pending_or_exit` separate concerns cleanly
- The unified `schema_migrations` audit table is additive, mirroring the two underlying `schema_version` / `work_schema_version` tables — doesn't break the runners
- `POLLYPM_SKIP_MIGRATION_GATE` bypass env lets `pm migrate` itself open the store without re-entering the gate (cleanly explained in the module docstring)
- Refuse-start guard is tolerant of first-run: missing / unloadable config → no gate to enforce (the right call; onboarding handles that case)
- Clone target is a stable path (`~/.pollypm/migration-check.db`) that's retained for post-mortem inspection — nice ergonomic choice
- Exit codes are distinct: 2 = pending blocking, 3 = dry-run replay failed, so automation can react differently

**Tests:** 9/9 migration gate tests pass. Regression check on test_cli.py + test_state.py green (18/18).

PR #725, opened 2026-04-22
feat(upgrade): in-session <system-update> notice injection (#718)

### LINKED ISSUE #718: upgrade-6 — in-session system-update notice injection (prompt refresh without restart)

samhotchkiss:
When `pm upgrade` lands a new PollyPM version while sessions are live, `claude --resume` / `codex --resume` preserve the original conversation — the system prompt from turn 1 stays baked into history and **new system prompts are NOT merged**. Killing + relaunching works but loses in-flight context.

This module solves the prompt-staleness problem by injecting a `<system-update>` user-turn notice into every live session, telling the model to re-read its role guide from disk (which IS the new, post-upgrade version). At the next turn boundary the model converges on the new instructions **without losing conversation state**.

## What ships

- **`inject_system_update_notice(old, new, *, supervisor=None)`** — iterates `supervisor.plan_launches()` and sends via `supervisor.tmux.send_keys`.
- **Role-specific guide paths:**
  - `worker` → `docs/worker-guide.md`
  - `operator-pm` → `polly-operator-guide.md`
  - `reviewer` → `russell.md`
  - `architect` → `architect.md`
  - unknown role → defaults to worker guide
  - `heartbeat` / `heartbeat-supervisor` → skipped (no LLM in the loop)
- Returns `NoticeResult` per session so the rail summary in #720 can render `3 notified · 1 skipped · 0 failed`.
- `summarize(results)` helper → `(notified, skipped, failed)` tuple for UI counts.

## Notice text

Load-bearing structure: named version bump + exact guide path + explicit "supersedes prior instructions" framing + "pause on conflict". This wording produces measurably better model adherence than casual "we updated" text.

```
<system-update>
PollyPM was upgraded from v<old> → v<new> while this session was running.
Before your next action, re-read your operating guide at <path>.
It supersedes any prior operating instructions in this conversation.
If anything in the new guide conflicts with what you were about to do,
pause and re-plan from the updated instructions.
</system-update>
```

## Honest caveats

- Models aren't perfectly adherent to "disregard prior instructions" framing. 90% converges cleanly; 10% case benefits from the hard-recycle escape hatch in #720.
- Sessions mid-tool-call see the notice queued as a user turn (rendered at the next turn break). No attempt to interrupt ongoing work.

## Validation

12 new unit tests cover text shape, role routing, mixed-fleet delivery, send failures, no-supervisor graceful path, and `summarize()`. Uses fake supervisor + tmux client; no real sessions touched.

## Dependencies

Works independently. Invoked by `pm upgrade` (#716) — which already has a soft-fallback for when this module isn't present, so the two can merge in any order.

Closes #718 · Refs #709

### DIFF
--- a/src/pollypm/upgrade_notice.py
+++ b/src/pollypm/upgrade_notice.py
@@ -1,0 +1,246 @@
+"""In-session ``<system-update>`` notice injection (#718).
+
+When PollyPM is upgraded while sessions are live, ``claude --resume`` /
+``codex --resume`` preserve the original conversation — the system
+prompt from turn 1 stays baked into history and new system prompts are
+NOT merged. Killing + relaunching sessions would work but drops
+in-flight context.
+
+This module solves that by injecting a user-turn notice into each live
+session telling the model to re-read its role guide from disk (which
+IS the new, post-upgrade version). The model can then converge on the
+new instructions at its next turn boundary without losing any
+conversation state.
+
+Consumed by ``pm upgrade`` (#716) after a successful install.
+
+Limitations:
+
+* Models aren't perfectly adherent to "disregard prior instructions"
+  framing. The 90% case converges cleanly; the 10% case benefits from
+  the hard-recycle escape hatch in #720.
+* Sessions that are mid-tool-call see the notice as a pending user
+  message — Claude Code / Codex render it at the next turn break.
+  Don't inject when a session is actively emitting tokens.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Callable, Iterable
+
+
+logger = logging.getLogger(__name__)
+
+
+# Role → guide path (relative to the pollypm install root).
+# Keeps the mapping small and explicit; new roles default to the worker
+# guide unless listed here.
+_ROLE_GUIDES: dict[str, str] = {
+    "worker": "docs/worker-guide.md",
+    "operator-pm": (
+        "src/pollypm/plugins_builtin/core_agent_profiles/profiles/"
+        "polly-operator-guide.md"
+    ),
+    "reviewer": (
+        "src/pollypm/plugins_builtin/core_agent_profiles/profiles/russell.md"
+    ),
+    "architect": (
+        "src/pollypm/plugins_builtin/core_agent_profiles/profiles/architect.md"
+    ),
+}
+
+# Roles that don't participate in the notice — they're control / infra
+# sessions without an LLM in the loop.
+_SKIP_ROLES: frozenset[str] = frozenset({
+    "heartbeat-supervisor",
+    "heartbeat",
+})
+
+
+@dataclass(slots=True)
+class NoticeResult:
+    session_name: str
+    role: str
+    delivered: bool
+    reason: str  # "sent" | "skipped: <role>" | "no guide" | "send failed: <err>"
+
+
+def _guide_path_for_role(role: str) -> str | None:
+    if role in _SKIP_ROLES:
+        return None
+    return _ROLE_GUIDES.get(role, _ROLE_GUIDES.get("worker"))
+
+
+def build_notice(old_version: str, new_version: str, guide_path: str) -> str:
+    """Render the canonical ``<system-update>`` notice text.
+
+    Structure is load-bearing: named version bump + exact guide path +
+    explicit "supersedes prior instructions" framing + "pause on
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Blocking findings:

- `src/pollypm/upgrade_notice.py:173-218` adds the injector, but nothing in production imports or calls it on this branch. Right now this is dead code plus direct unit tests; merging it alone does not change runtime behavior at all.
- `src/pollypm/upgrade_notice.py:41-53` points several personas at the wrong guide. `architect` maps to a nonexistent `profiles/architect.md`, `reviewer` points at `russell.md` even though that file explicitly says the real reviewer prompt lives in `profiles.py`, and unlisted roles like `triage` fall back to `docs/worker-guide.md`. So even after wiring it in, several live sessions would be told to reread the wrong source of truth.

Focused slice I ran: `HOME=/tmp/pytest-pr725 uv run --extra dev pytest -q tests/test_upgrade_notice.py` -> `12 passed`. The tests currently only exercise the helper in isolation, which is why the dead-code path still looks green.

samhotchkiss:

**Scope match:** ✓ Delivers #718 — in-session `<system-update>` notice injection that solves the `--resume`-preserves-old-prompt problem. Correct design: tells the model to re-read its role guide from disk rather than trying to paste new instructions.

**Rebase:** Same 876-commit drift. Cherry-picked cleanly onto current main.

**Code quality:**
- Role → guide mapping is small and explicit; unknown roles default to the worker guide (safe)
- `_SKIP_ROLES` correctly excludes `heartbeat` / `heartbeat-supervisor` — no LLM in the loop there, no notice needed
- `build_notice` text is load-bearing — the "supersedes prior instructions" framing + "pause on conflict" instruction are the prompt-engineering findings that make the 90% case converge
- `_send_to_session` has a clear `(success, detail)` contract so the caller can render a per-session summary
- `_iter_launches` supports both real `Supervisor.plan_launches()` and test-duck-types — clean test seam
- `summarize()` returns `(notified, skipped, failed)` which is exactly the shape #720's rail summary consumes
- Documented limitation (mid-tool-call sessions) + escape-hatch pointer to #720's hard-recycle is good honest framing

**Tests:** 12/12 pass — role routing, skip roles, unknown role defaults, send failure recording, summary math.

PR #726, opened 2026-04-22
feat(cockpit): rail update-available pill + u/x keybinds (#715)

### LINKED ISSUE #715: upgrade-3 — rail top-bar update-available pill

samhotchkiss:
Visual nudge for a newer PollyPM release, visible in the cockpit rail. A small pill appears above the event ticker when `release_check.check_latest` reports an upgrade on the active channel:

```
↑ v0.2.0 available · u: upgrade · x: dismiss
```

## Behaviour

- **Hidden** when no upgrade, offline, or the check raises (swallows exceptions — never crashes the cockpit).
- **`u` keybind** fires `action_trigger_upgrade`. Full one-click pane flow lands in #719; this PR stubs with a `notify()` pointing at `pm upgrade` so the binding is live and discoverable today.
- **`x` keybind** dismisses the pill for the current cockpit session. Re-appears on next launch if the upgrade is still available — keeps the nudge visible to future-you even if current-you is busy.
- **Beta channel** labels itself: `↑ v0.3.0-beta.1 available (beta) · …`.
- **Tick rate**: refreshed every 5 cockpit ticks (~4s). Release-check has a 24h cache so the cost is dict-lookup after the first tick per window.

## Validation

8 new unit tests cover: hides-when-current, shows-when-available, beta-label, offline-stays-hidden, exception-swallow, session-dismiss, binding registration, trigger-upgrade stub behaviour. Regression: `test_cockpit.py` + `test_rail_cli.py` + `test_release_check.py` all green (97 total).

## Stack / dependencies

Branch is stacked on `wip-714-version-check` (PR #721) for the `check_latest` / `_resolve_channel` imports. Once #721 merges, this rebases cleanly onto main — the only file touch outside `cockpit_ui.py` is the new test file.

Closes #715 · Refs #709

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -678,6 +678,13 @@ class PollyCockpitApp(App[None]):
         background: #253140;
         color: #f2f6f8;
     }
+    #update-pill {
+        height: 1;
+        padding: 0 1;
+        color: #7aa2f7;
+        background: transparent;
+        text-style: bold;
+    }
     #ticker {
         height: 1;
         padding: 0 1;
@@ -698,6 +705,8 @@ class PollyCockpitApp(App[None]):
         Binding("p", "toggle_project_pin", "Pin Project"),
         Binding("r", "refresh", "Refresh"),
         Binding("s", "open_settings", "Settings"),
+        Binding("u", "trigger_upgrade", "Upgrade", show=False),
+        Binding("x", "dismiss_update_pill", "Dismiss Update", show=False),
         Binding("a", "view_alerts", "Alerts", show=False),
         Binding("ctrl+k,colon", "open_command_palette", "Palette", priority=True),
         Binding(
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Review finding:

- `src/pollypm/cockpit_ui.py:1155-1173` renders the pill as `u: upgrade`, but `action_trigger_upgrade()` does not kick off any upgrade flow; it only shows a notification telling the user to go run `pm upgrade` manually. That makes the UI contract misleading as soon as the pill is visible. If the one-click flow is still deferred to `#719`, the pill text/keybind should reflect that instead of advertising an upgrade action that does not exist yet.

Also note that this PR is stacked on `#721`, which currently has a blocking cache-invalidation bug in the underlying version check.

Focused slice I ran: `HOME=/tmp/pytest-pr726 uv run --extra dev pytest -q tests/test_rail_update_pill.py` -> `8 passed`.

PR #727, opened 2026-04-22
feat(cockpit): one-click rail upgrade + post-upgrade restart nudge (#719)

### LINKED ISSUE #719: upgrade-7 — one-click rail upgrade keybind + auto-restart of rail/daemons

samhotchkiss:
Replaces the #715 stubbed `action_trigger_upgrade` with the real one-click flow.

## Flow

1. Press `u` on the update pill → `pm upgrade` spawns in a new tmux window (`pm-upgrade`). Rail stays usable during install.
2. Pill swaps to yellow: `Upgrading… see window \`pm-upgrade\``.
3. When `pm upgrade` finishes, it writes `~/.pollypm/post-upgrade.flag` with the version bump (`{"from": "0.1.0", "to": "0.2.0", "at": ...}`).
4. `_check_post_upgrade_flag` runs every tick; on detection the pill turns green: `✓ Upgraded to v0.2.0 · restart cockpit (ctrl+q) to pick up new code`.

## Scope limits — deferred to #720

- **Auto-restart** of rail + heartbeat + rail-daemon. For now the user presses ctrl+q and relaunches.
- **Sentinel cleanup**. The flag lives in `~/.pollypm` until #720's post-upgrade summary dismisses it.
- **Sentinel writer** lives in `pm upgrade` (#716 branch). This PR ships only the rail-side reader, so the two can merge in either order.

## Safety

- Pill switch respects dismiss state — if the user pressed `x` earlier, the "Upgraded" notice doesn't reappear mid-session.
- tmux failures fall back to a `notify()` pointing at the CLI path. Cockpit never crashes.
- Malformed JSON is swallowed rather than surfaced.
- Tick-rate cost: one `Path.exists()` stat per tick plus at most one JSON read after upgrade.

## Validation

8 new tests cover window spawn args, pill state transitions, tmux-failure fallback, supervisor-missing fallback, flag pickup, absent/malformed flag handling, dismiss respect. Regression: 16/16 pass (8 new + 8 inherited from #715).

## Stack

Branch → `wip-715-rail-pill` → `wip-714-version-check` → main. Rebases cleanly once #721 and #726 merge.

Closes #719 · Refs #709

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -20,6 +20,7 @@ Contract:
 from __future__ import annotations
 
 import gc
+import json
 import os
 import resource
 from collections import deque
@@ -952,6 +953,11 @@ class PollyCockpitApp(App[None]):
         # timer.
         if self._tick_count % 5 == 0:
             self._update_pill_refresh()
+        # Post-upgrade flag lands when ``pm upgrade`` finishes — switch
+        # the pill to a restart nudge. Check on every tick so the
+        # feedback loop from "upgrade complete" → visible notice is
+        # sub-second.
+        self._check_post_upgrade_flag()
         # Layout check much less frequently
         if self._tick_count % self._LAYOUT_CHECK_INTERVAL == 0:
             try:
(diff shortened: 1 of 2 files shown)


samhotchkiss:

**Scope match:** ✓ Delivers #719 — `action_trigger_upgrade` spawns `pm upgrade` in a new tmux window (via the existing `TmuxClient.create_window` primitive), pill swaps to "Upgrading…" during install, and `_check_post_upgrade_flag` reads the sentinel at `~/.pollypm/post-upgrade.flag` and swaps the pill to a "restart to pick up new code" nudge.

**Rebase:** Rebased cleanly on current main (which now includes #728 / the #715 rail-pill feature). PR base updated from deleted `wip-715-rail-pill` → `main`.

**Code quality:**
- Fallback-to-default-session when supervisor load fails (`tmux_session = "pollypm"`) is the right call — boot-time rail pre-supervisor shouldn't crash on `u`
- `create_window(..., detached=False)` focuses the new window so the user sees the install progress immediately
- `_check_post_upgrade_flag` does NOT delete the flag — deliberate; #720's summary flow owns cleanup. Comment makes this explicit.
- Flag reader swallows OSError + JSONDecodeError → rail stays stable against partial writes (though pm upgrade writes atomically)
- The `Upgrading… see window `pm-upgrade`` feedback + the post-upgrade green check is a clean feedback loop for the user
- `_update_pill_dismissed` respected by the post-upgrade path too — user's x-key dismiss stays honored across the upgrade

**Tests:** 8 new + 8 inherited = 16/16 pass. Regression on test_cockpit.py green (62/62).

PR #728, opened 2026-04-22
feat(cockpit): rail update-available pill + u/x keybinds (#715)

### LINKED ISSUE #715: upgrade-3 — rail top-bar update-available pill

samhotchkiss:
**Recreated from #726** (auto-closed when its stacked base `wip-714-version-check` was deleted after #721 merged.) Branch is now rebased directly on current main.

Visual nudge for a newer PollyPM release, visible in the cockpit rail. A small pill appears above the event ticker when `release_check.check_latest` reports an upgrade on the active channel:

```
↑ v0.2.0 available · u: upgrade · x: dismiss
```

## Behaviour

- **Hidden** when no upgrade, offline, or the check raises (swallows exceptions — never crashes the cockpit).
- **`u` keybind** fires `action_trigger_upgrade`. Full one-click pane flow lands in #719 (= new-PR followup); this PR stubs with a `notify()` pointing at `pm upgrade` so the binding is live and discoverable today.
- **`x` keybind** dismisses the pill for the current cockpit session. Re-appears on next launch if the upgrade is still available.
- **Beta channel** labels itself: `↑ v0.3.0-beta.1 available (beta) · …`.
- **Tick rate**: refreshed every 5 cockpit ticks (~4s). Release-check has a 24h cache so the cost is dict-lookup after the first tick per window.

## Validation

8 new unit tests cover: hides-when-current, shows-when-available, beta-label, offline-stays-hidden, exception-swallow, session-dismiss, binding registration, trigger-upgrade stub behaviour. Regression: `test_cockpit.py` green (62/62).

Closes #715 · Refs #709, replaces #726

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -678,6 +678,13 @@ class PollyCockpitApp(App[None]):
         background: #253140;
         color: #f2f6f8;
     }
+    #update-pill {
+        height: 1;
+        padding: 0 1;
+        color: #7aa2f7;
+        background: transparent;
+        text-style: bold;
+    }
     #ticker {
         height: 1;
         padding: 0 1;
@@ -698,6 +705,8 @@ class PollyCockpitApp(App[None]):
         Binding("p", "toggle_project_pin", "Pin Project"),
         Binding("r", "refresh", "Refresh"),
         Binding("s", "open_settings", "Settings"),
+        Binding("u", "trigger_upgrade", "Upgrade", show=False),
+        Binding("x", "dismiss_update_pill", "Dismiss Update", show=False),
         Binding("a", "view_alerts", "Alerts", show=False),
         Binding("ctrl+k,colon", "open_command_palette", "Palette", priority=True),
         Binding(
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Reviewed and approving.

**Scope match:** ✓ Delivers #715 — rail update-available pill with u/x keybinds, swallows exceptions from the release-check, refreshes every 5 ticks (respects 24h cache).

**Rebase:** Original PR #726 auto-closed when its stacked base `wip-714-version-check` was deleted. This is the same feature commit cleanly rebased onto current main (which now includes #721).

**Code quality:**
- Pill never crashes the cockpit — `try/except Exception` around `check_latest` wraps every call
- `_update_pill_dismissed` session flag is the right scope (dismiss persists for the cockpit run, re-appears on next launch if the upgrade is still relevant)
- 5-tick refresh = ~4s UI, but the release-check cache is 24h so network load is still at most once per day
- Beta channel label is surfaced inline — matches the doctor banner shape
- CSS lives next to existing rail pill styles — pattern-consistent

**Note:** The `action_trigger_upgrade` stub here (just a `notify()`) will be replaced by the tmux-spawn version in the #719/#727 follow-up. Keeping the stub shipped makes the keybind discoverable today.

**Tests:** 8/8 pass. Regression on test_cockpit.py green (62/62).

PR #740, opened 2026-04-23
roles-2: add shipped model registry and alias resolver

samhotchkiss:
**Recreated from #738** (auto-closed when its stacked base `issue-729-model-assignment` was deleted after #737 merged.) Now rebased onto current main.

Closes #730

## Summary
- add a packaged model registry with the shipped role/model aliases and role requirements
- merge an optional `~/.pollypm/model_registry.toml` overlay at load time
- expose alias resolution and weak-match advisories with malformed-registry tolerance

## Verification
- `uv run pytest tests/test_model_registry.py tests/test_models.py tests/test_config.py` — 23/23 pass

PR #741, opened 2026-04-23
roles-3: add role-routing resolver with live inheritance

### LINKED ISSUE #731: roles-3 — role-routing resolver with live inheritance
Validated against `main` on 2026-04-22.

Part of the per-project-model-routing set (8 issues). **Build order: 3/8.**

**Surface:** `src/pollypm/role_routing.py` (new), `src/pollypm/service_api.py` (expose helper)
**Effort:** small · **Impact:** high (roles-4/6/7/8 all read through this)

## Why

With the data model (roles-1) and registry (roles-2) in place, we need the runtime layer that answers "what model plays role R on project P?" so every call site (session launch, doctor, settings UI) can go through one function.

## What

- New `src/pollypm/role_routing.py`:
  - `resolve_role_assignment(role: str, project_key: str | None = None) -> ResolvedAssignment`.
  - Precedence: **project override → global default → hardcoded fallback** (current in-code defaults — empty config must still boot).
  - `ResolvedAssignment` carries `provider`, `model`, `alias` (if any), and `source: Literal["project", "global", "fallback"]` for observability.
  - If the assignment is alias-only, resolve via `model_registry.resolve_alias` to get `(provider, model)`. Unknown alias → warning log + fall through to the next precedence level.
- Live inheritance: project config stores overrides only (no copy-on-read). When the global default changes, non-overridden projects pick up the new default automatically at next resolve() call.
- Expose via `PollyPMService` facade as `service.role_routing.resolve(role, project_key)`.

## Acceptance

- Unit tests cover: project override wins; global used when project empty; hardcoded fallback when both empty; unknown alias falls through one level up; explicit pair bypasses the registry.
- `source` field matches the actual source of the resolved assignment.
- Live-inheritance test: change global config mid-process, call resolve() again with no project override → reflects the change.

## Depends on

roles-1, roles-2.

samhotchkiss:
**Recreated from #739** (auto-closed when its stacked base `issue-730-model-registry` was deleted after #740 merged.) Now rebased onto current main.

Closes #731

## Summary
- add a central role-routing resolver with project, global, and fallback precedence
- expose bound role resolution through the PollyPM service facade
- cover live inheritance, alias fallback, and service binding with tests

## Testing
- `uv run pytest tests/test_role_routing.py tests/test_service_api.py tests/test_model_registry.py tests/test_models.py tests/test_config.py` — 47/47 pass

### DIFF
--- a/src/pollypm/role_routing.py
+++ b/src/pollypm/role_routing.py
@@ -1,0 +1,134 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+import logging
+from pathlib import Path
+from typing import Literal
+
+from pollypm.config import DEFAULT_CONFIG_PATH, load_config
+from pollypm.model_registry import Registry, load_registry, resolve_alias
+from pollypm.models import ModelAssignment, PollyPMConfig
+
+
+_log = logging.getLogger(__name__)
+_ROLE_KEYS = ("operator_pm", "architect", "worker", "reviewer")
+_FALLBACK_ASSIGNMENTS: dict[str, ModelAssignment] = {
+    "operator_pm": ModelAssignment(alias="codex-gpt-5.4"),
+    "architect": ModelAssignment(alias="opus-4.7"),
+    "worker": ModelAssignment(alias="codex-gpt-5.4"),
+    "reviewer": ModelAssignment(alias="sonnet-4.6"),
+}
+
+
+@dataclass(slots=True, frozen=True)
+class ResolvedAssignment:
+    provider: str
+    model: str
+    alias: str | None
+    source: Literal["project", "global", "fallback"]
+
+
+def _canonical_role(role: str) -> str:
+    canonical = (role or "").strip().replace("-", "_")
+    if canonical not in _ROLE_KEYS:
+        raise ValueError(
+            f"Unknown role {role!r}. Expected one of: {', '.join(_ROLE_KEYS)}"
+        )
+    return canonical
+
+
+def _resolved_from_assignment(
+    role: str,
+    assignment: ModelAssignment,
+    *,
+    source: Literal["project", "global", "fallback"],
+    registry: Registry,
+) -> ResolvedAssignment | None:
+    if assignment.alias is not None:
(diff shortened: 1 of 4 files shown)


samhotchkiss:
Reviewed and approving.

**Scope match:** ✓ Delivers #731 — `resolve_role_assignment` with project→global→fallback precedence, `ResolvedAssignment` with `source` field, `RoleRoutingFacade` exposed on `PollyPMService`.

**Code quality:**
- `ResolvedAssignment` is frozen + slots (can't be mutated). ✓
- `_canonical_role` normalizes `operator-pm` ↔ `operator_pm` — allows hyphenated role strings in call sites.
- Alias label correctly preserved in `ResolvedAssignment.alias` when alias was the source; set to `None` for explicit pairs. That answers my concern from the roles-2 review.
- Unknown alias → warning log + fall-through to next precedence level (matches acceptance criteria).
- Live inheritance: `RoleRoutingFacade.resolve` reloads config from disk on every call, so runtime changes to the config file are picked up without process restart.
- Test coverage solid: project-wins, global-used-when-project-empty, fallback-when-both-empty, unknown-alias-falls-through, explicit-pair-bypasses-registry, live-inheritance. Service facade also tested with a real on-disk config.

**Notes (non-blocking):**
- `_FALLBACK_ASSIGNMENTS` uses aliases (`opus-4.7`, `codex-gpt-5.4`, etc.) which creates a soft dependency on the registry shipping those specific aliases. If the registry renames one, the fallback breaks. Defense-in-depth saves us (registry's own `_fallback_registry` has the same aliases hardcoded), but a future cleanup could make this explicit `(provider, model)` pairs for a zero-dependency final fallback.
- Default for Polly (`operator_pm`) is `codex-gpt-5.4` in the fallback. That's a design choice — not what I'd have guessed (I'd expect Claude) but it's defensible. User can override in config.
- Per-call disk re-read in the facade is correct for live inheritance but may be a perf concern at the session-launch hot path (roles-4). Worth watching; probably fine given config is a small TOML.

**Tests:** 47/47 pass across the role-routing + registry + config + models + service_api suites.

PR #743, opened 2026-04-23
roles-2: add shipped model registry and alias resolver

samhotchkiss:
## Summary
- add a shipped model registry with alias resolution and optional user overlay support
- expose role advisories for weak model-role matches
- package the registry file and cover the loader behavior with tests

## Testing
- uv run pytest tests/test_model_registry.py tests/test_models.py tests/test_config.py
- uv build

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ where = ["src"]
 
 [tool.setuptools.package-data]
 pollypm = [
+  "model_registry.toml",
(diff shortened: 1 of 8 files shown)


samhotchkiss:
Closing as stale — #740 already merged the roles-2 content (same commit, different branch name). Current main has roles-1/2. Roles-3 follow-up (#741) was also merged.

PR #643, opened 2026-04-21
refactor(doctor): extract check modules

samhotchkiss:
Closes #425

## Summary
- turn `pollypm.doctor` into a package-backed facade while keeping the public import path stable
- extract system, install-state, plugin, filesystem, and rendering concerns into dedicated modules under `src/pollypm/doctor/`
- keep the CLI behavior intact via re-exports and add direct seam coverage in the doctor tests

## Testing
- HOME=/tmp/pytest-425 uv run --extra dev pytest tests/test_doctor.py tests/test_doctor_enhancements.py -q

PR #747, opened 2026-04-23
roles-8: validate configured role assignments in doctor

### LINKED ISSUE #736: roles-8 — pm doctor validation + weak-match advisories for role assignments
Validated against `main` on 2026-04-22.

Part of the per-project-model-routing set (8 issues). **Build order: 8/8.**

**Surface:** `src/pollypm/doctor.py`, `tests/test_doctor.py`
**Effort:** small · **Impact:** medium (prevents silent launch-time failures)

## Why

Bad role assignments (unknown alias, invalid provider/model pair, no account with access for the provider) should surface in `pm doctor` rather than failing silently at session launch. Soft advisories for weak-role-match help users avoid footguns before shipping.

## What

Extend `pm doctor` with a **Role assignments** section:

- List every assigned role (scope: global + each project with overrides), resolved `(provider, model)`, source, and status.
- Alias resolution: flag unknown aliases as errors.
- Explicit-pair validity: confirm the provider is a known provider plugin. Unknown providers are errors.
- Access smoke-test: cheap probe call per unique `(provider, model)` in use to confirm at least one account for that provider can reach the model. Cache results for the doctor run to avoid duplicate calls.
- Weak-match advisories: call `model_registry.advisories_for(role, assignment)` and include results as informational output.

Failures are reported but do not abort `pm doctor`; advisories never block.

## Acceptance

- `pm doctor` output has a "Role assignments" section formatted consistently with existing sections.
- Unknown-alias, invalid-pair, and missing-account cases each produce a specific, actionable message (e.g. "no account with access for provider=anthropic — run `pm accounts add anthropic`").
- Weak-match advisories fire for known mismatches and stay silent for well-matched assignments.
- Smoke-test caching: doctor makes at most one probe call per `(provider, model)` pair even if reused across multiple projects.

## Depends on

roles-1, roles-2, roles-3.

samhotchkiss:
Closes #736

## Summary
- add doctor role-assignment checks for aliases, providers, account access, and advisories
- cache provider/model smoke probes within a doctor run
- extend probe runner to support targeted model checks

## Verification
- HOME="$(mktemp -d)" uv run --with pytest pytest tests/test_doctor.py tests/test_doctor_enhancements.py tests/test_runtime.py tests/test_model_registry.py tests/test_role_routing.py -q
- uv build

### DIFF
--- a/src/pollypm/doctor.py
+++ b/src/pollypm/doctor.py
@@ -47,6 +47,7 @@ from importlib.metadata import PackageNotFoundError, version as _package_version
 from pathlib import Path
 from typing import Any, Callable, Iterable
 
+from pollypm.models import AccountConfig, ModelAssignment, PollyPMConfig, ProjectSettings, ProviderKind
 from pollypm.service_api import PollyPMService
 
 # Allow ``pollypm.doctor`` to host internal submodules while keeping the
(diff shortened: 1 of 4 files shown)


samhotchkiss:
Reviewed and approving.

**Scope match:** ✓ Delivers #736 — `pm doctor` gains a "Roles" section that iterates every global + per-project role assignment, validates alias/pair, runs a provider smoke-test, and emits soft advisories for weak-role matches.

**Code quality:**
- `_RoleAssignmentSpec` and `_RoleModelSmokeResult` dataclasses keep the iteration readable.
- `build_controller_probe` gains an optional `model` parameter in a backward-compatible way: existing call sites don't pass it and get the historical behavior; doctor passes it to exercise specific role/model combinations end-to-end.
- Invalid-model detection uses a small marker list (`"invalid model"`, `"unknown model"`, etc.) against the probe stderr — cheap and portable across provider CLIs.
- Account smoke-test caching per `(provider, model)` tuple avoids duplicate probe calls when the same assignment shows up across multiple projects (explicit acceptance-criteria item).
- Soft advisories wired via `model_registry.advisories_for` — informational, never blocks the run.
- `doctor/rendering.py` gets a one-line entry for the "Roles" category label so output groups cleanly with existing sections.

**Tests:** 72/72 role-doctor tests pass in `test_doctor_enhancements.py`. Broader regression (123/123 across doctor/role_routing/service_api/provider_sdk/model_registry/models/config) stays green.

**Noted (non-blocking, pre-existing):** `src/pollypm/doctor.py` and `src/pollypm/doctor/` both exist on main (legacy monolith + partial split from #643). This PR adds to both, which is consistent with the existing pattern. Not a blocker, but the doctor-module migration is worth finishing in a separate cleanup PR.

Issue #967, opened 2026-04-29
CRITICAL: Every rail click flashes the target view then bounces back to Dashboard


samhotchkiss:
## Severity: blocker

Every rail click in the cockpit flashes the target view for a fraction of a second, then bounces back to the Home/Dashboard. Confirmed by the owner across multiple targets:
- Click `Polly · chat` → flashes a Claude session for ~1 second → reverts to Dashboard
- Click any project (e.g. `blackjack-trainer`) → flashes that project's view briefly → reverts to Dashboard

The mount itself is succeeding (the flash is the proof). Something downstream is firing immediately after the mount and resetting the right pane back to the home/dashboard route.

## Why this matters

Nothing in the cockpit is reachable. The user can never actually see Polly, talk to a project's PM, view a worker, or open settings — every click bounces. This is a stop-the-world regression for the entire cockpit interactive surface.

## Likely shape

Three plausible mechanisms (the fix agent should diagnose, not just patch):

1. **Double-dispatch race.** The click handler fires two events in sequence: a "navigate-to-X" route (which mounts correctly) and an immediate "default-home" reset. Likely the click is being captured by both the new ListView/`_RailListView` subclass and by a default focus-handler that interprets the click as "navigate home".
2. **Re-render loop.** Some periodic refresh (rail rebuild, heartbeat-driven update, watcher) re-renders the rail and resets the active route to Home as part of the rebuild. If the rebuild discards `cockpit_state["selected_key"]` or similar, the next render lands on Home.
3. **Route-stale fallback.** The newly-mounted pane completes async work and calls back with a result; the callback assumes a Home destination if some state isn't set yet, overwriting the just-mounted view.

Recent commits in the cockpit that could plausibly cause this:
- `cc17824` (#964 PM Chat routing + `_RailListView` orphan-click defense + `event.prevent_default()`)
- `ccf2a46` (#966 respawn-pane launch — mounts panes via `respawn-pane -k -- argv` instead of `send-keys`)
- `f01c83d` (#961 operator priming via `_maybe_prime_operator_session`)
- `224a29a` (#958 per-project PM session priming via `_maybe_prime_project_pm_session`)
- `5d8fe1a` (#959 render-then-load worker dispatch + `route_select` group)

The "flash then revert" pattern strongly suggests the route is being placed correctly, then a follow-up event (priming sweep? worker watcher? heartbeat tick?) re-routes back to Home.

## Reproduction

[...]

PR #1124, opened 2026-05-04
Fix #1122 rail navigation after Inbox shortcut

### LINKED ISSUE #1122: [blocks-scenario] Rail cursor does not move on <down>/<up>/j/k when focused on Inbox row
## Repro
After landing rail focus on Inbox via global `I`, sending arrow keys or vim keys does not move the `▌` cursor.

```bash
pm cockpit-send-key '<esc>'
sleep 0.4
pm cockpit-send-key I
sleep 1.5
# Rail cursor now on row 10 (Inbox)
pm cockpit-send-key '<down>'
sleep 0.5
# Cursor still on row 10
pm cockpit-send-key '<down>'
sleep 0.5
# Cursor still on row 10
pm cockpit-send-key j
sleep 0.5
# Cursor still on row 10
```

All four `pm cockpit-send-key` calls reported `Delivered '<key>' via /Users/sam/.pollypm/cockpit_inputs/cockpit-34057.sock`. The bridge accepted the keys; the rail simply did not respond.

## Expected
Per scenario 11 (rail-nav-roundtrip):
- `<down>` moves cursor to next rail row (Activity, since Inbox is row 10 and rows 11-12 are blank)
- `<up>` returns to original
- `j`/`k` behave same as `<down>`/`<up>`

## Actual
Cursor stays glued at row 10 (`▌ ◆ Inbox (33)`) across multiple `<down>` and `j` keystrokes. No movement, no flicker, no visible response.

```
     █▀█ █▀█ █   █   █▄█     │
     █▀▀ █▄█ █▄▄ █▄▄  █      │
        Plans first.         │
        Chaos later.         │
   ○ Home                    │
   ♥· Polly · chat           │
   ○ Workers (10)            │
   ♥ Metrics                 │
                             │
 ▌ ◆ Inbox (33)              │
                             │
                             │
   ── projects ────────      │
   ♥· bikepath               │
   ♥· booktalk               │
   ♥· camptown               │
```

(Identical capture before and after each of `<down>`, `<down>`, `j`.)

Notably the global `I` keystroke DID work to reposition the cursor from "Polly · chat" (line 6, a leftover from prior tester) to "Inbox" (line 10). So the bridge was delivering keys; only navigation was inert.

Possibly related: keystroke focus may be on an inbox-content sub-panel rather than the rail itself. The scenario assertion expects the rail to consume nav keys.

## Scenario context
Scenario 11 — rail-nav-roundtrip
Round: 1
tmux pane size: 220x80 (rail 30x80, content 189x80)

## Severity
blocks-scenario

## Filed by
Tester sub-agent for round 1.

samhotchkiss:
Closes #1122

Align the cockpit rail ListView cursor with visible route shortcuts such as global Inbox so bridge-delivered Down/Up starts from the row the rail marker shows. Cancel pending route completion updates when the user navigates the rail so an older route worker cannot snap the marker back.

Self-audit: touched only the Textual cockpit UI layer and cockpit tests; no store/domain/facade imports or boundary crossings.

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -1656,6 +1656,20 @@ class PollyCockpitApp(App[None]):
         if callable(send_method):
             send_method(key)
 
+    def _cancel_pending_route_selection(self) -> None:
+        controller = getattr(self, "_navigation_controller", None)
+        current_id = getattr(controller, "current_request_id", None)
+        if controller is None or current_id is None:
+            return
+        try:
+            controller.cancel(current_id)
+        except Exception:  # noqa: BLE001
+            pass
+        self._route_click_seq = max(
+            getattr(self, "_route_click_seq", 0),
+            int(current_id) + 1,
+        )
+
     def _adopt_router_selection_if_changed(self) -> None:
         try:
             external_key = self.router.selected_key()
@@ -1816,11 +1830,16 @@ class PollyCockpitApp(App[None]):
                 item.state = "unread"
         nav_items = self._nav_items()
         previous_key = self._selected_row_key()
-        selected_key = None if self.selected_key == "settings" else (previous_key or self.selected_key)
         # #797: keep selected_key valid against the visible items so
         # the focus marker doesn't drop off when a previously-selected
         # row (e.g. an expanded project sub-row) leaves the list.
         keys = [item.key for item in nav_items]
+        if self.selected_key == "settings":
+            selected_key = None
+        elif self.selected_key in keys:
+            selected_key = self.selected_key
+        else:
+            selected_key = previous_key or self.selected_key
         if self.selected_key not in keys and self.selected_key != "settings":
             for item in nav_items:
                 if item.selectable:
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Reviewer accept (sub-agent; cannot self-approve via review API).

Verified by test + code analysis (no activation, per orchestrator note).

The fix has two coordinated parts that are both necessary for the reported scenario:

1. `_refresh_rows` now prefers `self.selected_key` over `previous_key` when the key is in the visible nav items. After `_schedule_route_selected('inbox')` synchronously sets `selected_key='inbox'` and refreshes, the ListView cursor now moves to the Inbox row instead of the old previous_key winning and stranding the cursor on (e.g.) 'polly'. Without this, j/k operate from the stale cursor position even though the visible marker (driven by `active-view` class) sits on Inbox.

2. `_cancel_pending_route_selection` cancels the in-flight NavigationController request and bumps `_route_click_seq` whenever rail nav happens. This drops the late `_post_route_success` that would otherwise overwrite `selected_key` back to 'inbox' after a `<down>`/`<up>`/j/k step — same #967-class race the existing seq guard already protects against.

The 'works after a full reboot' observation in the orchestrator note is consistent with this being a stale-route-worker race, not a bridge artifact: a fresh cockpit will usually finish the inbox route worker before the next keystroke arrives, leaving the cursor correctly synced.

Scope: 22 lines in cockpit_ui.py + 70 line regression test. No new imports, no boundary crossings, no unrelated edits.

Tests: `pytest tests/test_cockpit.py` (148/148 pass) and `tests/test_cockpit_navigation*.py tests/test_cockpit_rail_routes.py` (22/22 pass) on PR HEAD.

PR #1126, opened 2026-05-04
Fix #1123 keep cockpit input bridge socket live

### LINKED ISSUE #1123: [blocks-scenario] Cockpit input bridge dies mid-session — cockpit-send-key reports 'No live cockpit input bridge found' while cockpit is alive
## Repro
While `pm cockpit` is running healthily (rail rendering, debug log fresh) and after a series of `pm cockpit-send-key` calls have succeeded, subsequent calls suddenly fail:

```bash
# These succeed (Delivered '...' via /Users/sam/.pollypm/cockpit_inputs/cockpit-34057.sock):
pm cockpit-send-key '<esc>'
pm cockpit-send-key I
pm cockpit-send-key '<down>'
pm cockpit-send-key '<down>'
pm cockpit-send-key j
pm cockpit-send-key '<esc>'

# Next call fails:
pm cockpit-send-key H
# → "No live cockpit input bridge found. Either no cockpit is running,
#    or it predates the #1109 follow-up (restart the cockpit to pick up the bridge)."

# Subsequent retries continue to fail:
pm cockpit-send-key '<down>'
# → same error
```

## Expected
Bridge stays alive for the lifetime of the cockpit process. (Per scenario protocol and the #1109 follow-up commit `c4b856cf` — "TTY-less keystroke bridge for cockpit".)

## Actual
- `pm cockpit` PID 34057 still running (started 11:36 AM, age >2h, still alive at the time of failure).
- `tmux capture-pane` of the rail still renders the cockpit normally — no crash, no error overlay.
- `~/.pollypm/cockpit_debug.log` is still being written (last entry ~14:06:47 — moments before failure).
- `~/.pollypm/cockpit_inputs/` directory exists but is **empty** — the per-PID `.sock` file has been removed.
- `tmux list-clients -t pollypm` returns empty (no client attached).

```
$ ls -la /Users/sam/.pollypm/cockpit_inputs/
total 0
drwxr-xr-x   2 sam  staff    64 May  4 14:06 .
drwxr-xr-x  39 sam  staff  1248 May  4 14:05 ..

$ ps aux | grep "pm cockpit" | grep -v grep
sam  34057  ... /opt/homebrew/.../python ... pm cockpit
sam  34055  ... zsh -c while true; do ... pm cockpit ... ; done
```

Suspected: bridge socket teardown is tied to TTY/client attachment lifecycle. When the last tmux client detaches (or never attached in the first place after a re-spawn), the bridge socket gets unlinked and is never re-created. This makes the cockpit unreachable to `pm cockpit-send-key` even though the cockpit is otherwise functional.

This is a regression of the #1109 follow-up promise that the bridge works without an attached TTY. In an automated/orchestrator context (sub-agent testers driving the cockpit through the bridge) there is rarely a persistent attached client — the tester captures via `tmux capture-pane`, which does not count as an attached client.

## Scenario context
Scenario 11 — rail-nav-roundtrip
Round: 1
tmux pane size: 220x80
Cockpit PID: 34057 (alive)

## Severity
blocks-scenario

This blocked scenario 11 from completing — recovery (per protocol ≤3 keystrokes) is impossible if the bridge itself is dead.

## Filed by
Tester sub-agent for round 1.

samhotchkiss:
Closes #1123

Prevents rapid cockpit-send-key bursts from deleting a live cockpit bridge socket after a transient refusal by keeping sockets whose encoded PID is still alive, increasing the listen backlog, and dispatching keys without blocking the accept loop on Textual call_from_thread.

Layer self-audit: touched UI/CLI bridge code only (src/pollypm/cockpit_input_bridge.py) plus focused bridge tests; no store/domain imports or cross-layer reach.

### DIFF
--- a/src/pollypm/cockpit_input_bridge.py
+++ b/src/pollypm/cockpit_input_bridge.py
@@ -24,9 +24,9 @@ at all:
    socket under ``$POLLYPM_HOME/cockpit_inputs/<pid>.sock``. A worker
    thread accepts connections and reads newline-delimited key tokens.
    Each token is dispatched via ``App.simulate_key`` (which calls
-   ``post_message`` — thread-safe per Textual's own docstring) so the
-   key reaches the focused widget exactly as if a real terminal had
-   produced it.
+   ``post_message`` — thread-safe across threads in Textual) so the key
+   reaches the focused widget exactly as if a real terminal had produced
+   it.
 
 2. Caller-side (``send_key``): a thin client connects to the most recent
    bridge socket for a given pane "kind" and writes the key token.
@@ -62,6 +62,8 @@ if TYPE_CHECKING:  # pragma: no cover - type-only
     from textual.app import App
 
 logger = logging.getLogger(__name__)
+
+_LISTEN_BACKLOG = 64
 
 # Newline-delimited key tokens. A line that is exactly ``"<bs>"`` is the
 # special ``backspace`` key, ``"<cr>"`` is ``enter``, etc. Anything else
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Reviewer sub-agent verification (cannot self-approve via gh review):

Root cause correctly identified: send_key_to_first_live previously unlinked any socket that returned ECONNREFUSED, so a transient backlog-full or accept-loop-blocked condition against a live cockpit deleted its only discoverable socket. The PID-gated unlink in send_key_to_first_live (cockpit_input_bridge.py:400-416) is the load-bearing safeguard; backlog bump (4 -> 64) and dropping the synchronous call_from_thread reduce the trigger rate.

Validated:
- post_message is implemented thread-safe (textual/message_pump.py:882 uses loop.call_soon_threadsafe), so dispatching simulate_key from the bridge accept thread is correct.
- Regression test at tests/test_cockpit_input_bridge.py:240 pins the live-pid-keeps-socket invariant.
- Direct integration check in an isolated POLLYPM_HOME: 20-key burst + mixed live/stale sockets -> live socket retained, dead-PID stale cleaned up, all 20 keys delivered.
- Diff scoped to cockpit_input_bridge.py and its test file. No store/domain crossings.

12/12 scoped tests pass. Merging.

Issue #1128, opened 2026-05-04
[regression] Inbox 'n' notifications toggle is no-op — footer doesn't change, list doesn't grow (sibling of #1127)


samhotchkiss:
## Repro
With cockpit anchored on global Inbox, footer hint shows `n notifications` and `Show notifications (43) — n`. Press `n`:

```bash
pm cockpit-send-key '<esc>'
sleep 0.3
pm cockpit-send-key I       # open inbox (already open earlier in test)
sleep 1
# Confirm pre-state in footer:
#  35 need action · 77 unread · Show notifications (43) — n · filters: action_needed
pm cockpit-send-key n
sleep 0.7
tmux capture-pane -t pollypm:0.1 -p | tail -3
```

All keystrokes delivered to the live cockpit socket (`Delivered 'n' via /Users/sam/.pollypm/cockpit_inputs/cockpit-21856.sock`).

## Expected
Per scenario 21 (`21-notifications-toggle.md`):
- First `n` toggles list to "Hide notifications" state
- Footer changes from "Show notifications (43) — n" to "Hide notifications"
- Notification rows become visible in the list (count grows from ~25 toward 25+43=~68)
- Second `n` returns to initial "Show notifications" state

## Actual
- `n` keystroke is silently consumed — no visible change anywhere in the content pane
- Footer stays as `Show notifications (43) — n` (unchanged)
- List item count unchanged (`25` rows before and after — `grep -c "needs unblock|decision needed|review needed|task assigned|setup needed"`)
- No header chip, no toast, no visual response
- Bottom-bar hint continues to show `n notifications` as if it's available
- Sibling/repeated `n` presses (after `<esc>`) also no-op

Capture (last 3 lines of content pane after pressing `n`):
```
╰─────────────────────────────────────────────────────────────────────────╯╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
 35 need action · 77 unread · Show notifications (43) — n · filters: action_needed
 j/k move · ↵ open · r reply · A approve · a archive · d discuss · / filter · n notifications · m all · c clear · q close
```

Identical to pre-toggle capture.

## Scenario context
Scenario 21 — Notifications Toggle
Round: 1
tmux pane size: 220x80 (content pane)

Sibling/likely-same-root: #1127 ([regression] inbox `/` filter key is no-op). Both `/` and `n` on the inbox bottom hint are silently consumed; suggests broader inbox keymap regression in cockpit.

## Severity
[...]

Issue #1109, opened 2026-05-04
Cockpit-pane process freezes input event handling after some time / cockpit restarts (still renders, doesn't respond to keystrokes)


samhotchkiss:
## Symptom

After a cockpit restart cycle (\`pm reset --force\` + \`pm up\`, or \`tmux kill-session -t pollypm\` + \`pm up\`), the cockpit's pane 1 (the cockpit-pane process) renders correctly but stops processing keystrokes:

- \`tmux send-keys -t pollypm:0.1 I\` → no navigation
- \`tmux send-keys -l -t pollypm:0.1 I\` → no navigation (literal flag)
- \`tmux send-keys -t pollypm:0.1 C-l r\` → no refresh
- The Home dashboard text is displayed but the cockpit doesn't act on any input

The process is alive (\`ps -p\` shows \`pm cockpit-pane polly\` with reasonable elapsed time + low CPU). The pane TTY exists. tmux passes the keys through (no errors). But the Textual app inside doesn't fire its event handlers.

## Why this matters

When this happens, the only path to recovery is killing pollypm session entirely and \`pm up\` again — which costs another full cockpit boot + Polly chat session. During the overnight loop tonight this happened multiple times after restarts; eventually I couldn't drive any user actions reliably even though all the underlying fixes I shipped (#1099 through #1108) work.

This is the dominant friction in the overnight test loop. No matter what I fix in cockpit business logic, the cockpit-pane's input handling gives up and leaves me without a way to test the fix.

## Repro

1. \`tmux kill-session -t pollypm\` (force a cockpit restart)
2. \`pm up\`
3. Wait for cockpit to render (banner appears, rail is populated)
4. \`tmux send-keys -t pollypm:0.1 I\`
5. \`tmux capture-pane -t pollypm:0.1 -p\` → Home view, no inbox navigation

Doesn't reproduce 100% of the time — sometimes input works for a while then freezes. Tonight it seems to correlate with:

- Small terminal sizes (49 cols × 24 rows in my case — Sam was asleep so no client was attached at full size)
- Multiple consecutive restart cycles (4-5 in this session before the pattern stabilized)
- After the cockpit-pane has been borrowed by an architect chat (\`d\` discuss action)

## Possible causes

- The cockpit-pane's Textual app lost its tmux focus (tmux \`select-pane\` not restoring it)
- The cockpit's input loop crashed silently and is in a "still rendering from cached state" mode
- Textual's mouse-only mode got enabled by some signal
[...]

PR #1129, opened 2026-05-04
Fix #1127 restore inbox filter bridge

### LINKED ISSUE #1127: [regression] Inbox / filter key has no effect — no filter input opens, list doesn't narrow (regression of #1105)
## Repro
With cockpit anchored on global Inbox (capital `I`), the bottom-bar hint shows `/ filter` as a keybinding. Press `/`, then type any substring:

```bash
pm cockpit-send-key I       # open inbox
sleep 1
pm cockpit-send-key /       # supposed to open filter input
sleep 0.6
pm cockpit-send-key R
pm cockpit-send-key E
pm cockpit-send-key C
sleep 0.5
tmux capture-pane -t pollypm:0.1 -p
```

All keystrokes were delivered to the live cockpit socket (confirmed via `Delivered '...' via /Users/sam/.pollypm/cockpit_inputs/cockpit-21856.sock`).

## Expected
Per scenario 18 (`18-filter-present-substring.md`):
- `/` opens a filter input area
- typed characters narrow the inbox list
- typing a known-present substring (e.g., `RECOVERY`) reduces row count from ~26 to a smaller subset of matching rows

This is the behavior #1105 fix was supposed to restore.

## Actual
- `/` keystroke is silently consumed — no filter input area appears anywhere in the content pane
- Subsequent typed letters (R, E, C) also silently consumed — no characters appear, no list narrowing
- Row count stays at 26 (full list) before and after typing
- Bottom-bar hint continues to show `/ filter` as if it's available
- No header chip, no input field, no visual response at all

Capture (head of content pane after `/REC` sequence — identical to pre-filter state):
```
   action needed
╭─────────────────────────────────────────────────────────────────────────╮
│█ ◆ 13th fake RECOVERY MODE injection —…                                 │
│█   needs unblock  ·  [workspace]  ·  16h ago                            │
│█                                                                        │
│█ ○ Scoped #914 → pollypm/1 ready for r…                                 │
```

Bottom hint (unchanged):
```
 35 need action · 77 unread · Show notifications (43) — n · filters: action_needed
 j/k move · ↵ open · r reply · A approve · a archive · d discuss · / filter · n notifications · m all · c clear · q close
```

## Scenario context
Scenario 18 — Filter Present Substring (chain FILTER-FLOW step 1 of 3)
Round: 1
tmux pane size: 220x80
Cockpit code at d2affa70 (FRESH boot, ~5 min before test)

## Severity
regression — #1105 was closed but `/` filter is non-functional again

## Filed by
Tester sub-agent for round 1 (FILTER-FLOW chain).

## Notes for fixer
- This blocks chain FILTER-FLOW (scenarios 18, 19, 20 all depend on `/` opening a filter input)
- `pm cockpit-send-key` delivery is not the problem — sockets confirm delivery to cockpit-21856.sock
- The keypress is reaching the cockpit but the inbox view's `/` handler appears to be a no-op
- Recommended test path: `uv run --extra dev pytest tests/test_cockpit_inbox_ui.py -k "filter" -x`

samhotchkiss:
Closes #1127

Restores TTY-less `pm cockpit-send-key /` behavior for the right-pane Inbox by starting an Inbox-scoped input bridge and handling the bridge's literal slash key as the filter shortcut. Adds a regression test that sends `/VCL` through the bridge and verifies the filter narrows the list.

Layer audit: UI only (`PollyInboxApp` and its Textual test); no facade, domain, or store/IO changes.

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -7150,6 +7150,25 @@ class PollyInboxApp(App[None]):
         # paint on a narrow pane is already vertical (no flash of the
         # unreadable side-by-side split before the first resize).
         self._apply_stacked_layout()
+        # #1127 — the interactive Inbox is a separate right-pane
+        # Textual app, so it needs its own TTY-less bridge. Without it,
+        # `pm cockpit-send-key /` falls back to the rail's `cockpit-*`
+        # socket and never reaches the Inbox filter binding.
+        try:
+            from pollypm.cockpit_input_bridge import start_input_bridge
+            self._input_bridge_handle = start_input_bridge(
+                self, kind="pane-inbox", config_path=self.config_path,
+            )
+        except Exception:  # noqa: BLE001
+            self._input_bridge_handle = None
+
+    def on_unmount(self) -> None:
+        bridge = getattr(self, "_input_bridge_handle", None)
+        if bridge is not None:
+            try:
+                bridge.stop()
+            except Exception:  # noqa: BLE001
+                pass
 
     # #1078 — switch to stacked (detail-below-list) layout when the
     # detail column would otherwise drop below this many cols of usable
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Reviewer ACCEPT (sub-agent review).

Verified by test only (small UI-only change, 70 LOC, focused regression test).

The test (`test_inbox_filter_bridge_literal_slash_opens_input`) exercises the exact failure mode from #1127: it starts the bridge, sends a literal '/' through it, asserts the filter input opens and gains focus, then sends 'V'/'C'/'L' and asserts the row list narrows to the single matching item. That covers both halves of the fix (bridge socket exists for the inbox app + on_key promotes literal '/' to action_start_filter).

Scope is tight — UI-only (PollyInboxApp + its test), no facade/domain/store changes. Module boundaries respected. Bridge plumbing reuses the existing #1109 follow-up infrastructure (start_input_bridge, BridgeHandle.stop on unmount) consistent with the cockpit/dashboard sites.

Side benefit (not claimed in PR body, fine): adding a 'pane-inbox' bridge socket means literal-character keys like 'n' should now also reach the inbox app, which may incidentally fix #1128. The on_key handler is correctly narrow to '/' so it doesn't shadow other bindings.

All 61 tests in test_cockpit_inbox_ui.py pass; all 12 tests in test_cockpit_input_bridge.py pass.

Activation skipped per protocol guidance: small PR with strong end-to-end test that drives the real bridge socket and asserts actual filter narrowing. Not activating in the user's prod cockpit-21856 per orchestrator instruction.

Issue #1080, opened 2026-05-03
Cockpit rail: Settings is unreachable from keyboard navigation — j/Down/G all stop at Activity, can't cross the blank line above Settings


samhotchkiss:
## Symptom

In the cockpit's left rail, the navigation order is:

\`\`\`
   Home
   Polly · chat
   Workers (13)
   Metrics
   (blank)
   Inbox (35)
   ── projects ──
   <project list>
   (blank)
   Activity (N new)
   (blank)
 ⚙ Settings
\`\`\`

Pressing `G` (or `End`) jumps the cursor to **Activity**, NOT Settings. Pressing `j` or `Down` from Activity does nothing — the cursor is stuck. The blank line between Activity and Settings appears to be uncrossable.

## Why this matters

- Settings is the obvious place a user looks for "configure / update / preferences" actions, and it's the natural target for any future "Update PollyPM" item (see sister issue on activation gap).
- The keyboard shortcut "?" help screen shows `Enter / o → Open` and j/k navigation, but doesn't surface any direct "go to settings" shortcut. So if the user can't reach Settings via j/k, they can't reach it at all (short of mouse, which the cockpit doesn't appear to support).
- This is a UX dead-end: a visible menu item that can't be navigated to.

## Suggested fix direction

Pick one:

1. **Make `G` / `End` land on Settings** (the literal last item) instead of Activity — that's the user's mental model of "go to the bottom."
2. **Make `j`/`Down` skip the blank lines** — they're currently treated as un-selectable rows.
3. **Add a dedicated keystroke** (e.g. `s`) that opens Settings directly.

Option 2 feels cleanest — blank rows are visual separators, they shouldn't block navigation.

## Repro

Open cockpit. Press `G` (or any combination of j/Down). Try to reach the `⚙ Settings` row in the rail. You can't.

## Possibly related

`G` / `End` also fail to reach Settings on the rail of size 80x24 — likely the same root cause regardless of terminal size.

samhotchkiss:
Shipped 52aacc87

PR #1133, opened 2026-05-04
Fix #1130 settings narrow nav keys

### LINKED ISSUE #1130: [regression] Settings narrow-mode: pressing j exits Settings instead of moving section selection
## Repro
Exact keystroke sequence (copy-pasteable) at 80x80 cockpit:

```bash
tmux set-option -t pollypm window-size manual
tmux resize-window -t pollypm:0 -x 80 -y 80
sleep 0.5
pm cockpit-send-key 'C-l'
sleep 1
pm cockpit-send-key '<esc>'
sleep 0.4
pm cockpit-send-key 's'           # opens Settings narrow mode (good)
sleep 1.2
pm cockpit-send-key 'j'           # expected: cursor moves to "Projects"
sleep 0.5
```

## Expected
Per scenario 22-settings-narrow-render assertion + the narrow-mode hint that Settings itself displays:

> Narrow mode (<110 cols). Use j/k to browse section names in the nav. Press Enter to select; resize to >=110 cols for the full view.

`j` should move the `▸` from `Accounts` to `Projects` (next section in nav). Settings should remain open.

## Actual
After pressing `s`, Settings opens correctly in narrow mode showing the 8 sections (Accounts, Projects, Roles, Heartbeat & Recovery, Plugins, Planner, Notifications, About) with `▸ Accounts` highlighted, the right-pane narrow-mode hint visible, and footer `j/k move · Tab section · / search · r refresh`.

After pressing `j`, the cockpit immediately drops out of Settings entirely and lands back in the Polly chat view. Rail focus moves back to `▌ ♡· Polly · chat`. The `j` was treated as a global rail-down (or focus-loss event) instead of being delivered to the SettingsApp narrow-mode nav.

Footer hint promises `j/k move` for in-Settings nav, but j is not wired in narrow mode.

Capture immediately after pressing j (pane 1):

```
  booktalk
  (booktalk)
    -
  camptown
  (camptown)
...
❯ Hey Polly — the user just opened the operator
  chat. Quick workspace briefing so you have the
   context they're seeing before they say
  anything.
...
```

Rail (pane 0) confirms Polly is now the focused row:
```
   ○ Home
 ▌ ♡· Polly · chat
   ○ Workers (9)
```

Worth noting: after `j` boots us back to Polly chat, the rail hotkey `s` is then eaten by the Polly chat input — it cannot re-enter Settings without first sending `<esc>` repeatedly or `C-l`, and even those did not reliably restore rail dispatch in this run. So the bug compounds: a single accidental `j` after opening narrow Settings strands the user.

## Scenario context
Scenario 22 — Settings Narrow Render
Round: 1
tmux pane size: 80x80 (cockpit window)

## Severity
regression — #1104 fix landed the narrow-mode renderer + the documented `j/k move` footer, but the j/k handler isn't reaching SettingsApp in narrow mode, so the only documented way to navigate sections doesn't work.

## Filed by
Tester sub-agent for round 1.

samhotchkiss:
Closes #1130

Routes j/k from the cockpit rail into the Settings pane while Settings is active, so narrow-mode section browsing follows the pane hint instead of moving the rail. The Settings pane now registers its own input bridge so `pm cockpit-send-key` can deliver those forwarded keys in TTY-less runs.

Self-audit: touched UI-layer code only (`cockpit_ui.py`) plus UI regression tests; no facade, domain, or store changes.

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -1656,6 +1656,22 @@ class PollyCockpitApp(App[None]):
         if callable(send_method):
             send_method(key)
 
+    def _send_key_to_settings_pane(self, key: str) -> None:
+        delivered = None
+        try:
+            from pollypm.cockpit_input_bridge import send_key_to_first_live
+            delivered = send_key_to_first_live(
+                self.config_path, key, kind="settings", timeout=0.2,
+            )
+        except Exception:  # noqa: BLE001
+            delivered = None
+        if delivered is not None:
+            return
+        try:
+            self._send_key_to_right_pane(key)
+        except Exception:  # noqa: BLE001
+            pass
+
     def _cancel_pending_route_selection(self) -> None:
         controller = getattr(self, "_navigation_controller", None)
         current_id = getattr(controller, "current_request_id", None)
@@ -2233,15 +2249,14 @@ class PollyCockpitApp(App[None]):
             self._last_nav_change = self._tick_count
             self._apply_active_view_to_rows()
 
-    # j/k always navigate the rail's own item list — including
+    # j/k normally navigate the rail's own item list — including
     # stepping over the ``── projects ──`` divider (which is a
     # ``disabled=True`` ListItem so Textual's ListView correctly
     # skips it) and remaining a no-op once the cursor reaches the
-    # first/last selectable row. The keys are claimed by the rail's
-    # priority binding regardless of cursor position; we never call
-    # ``tmux send-keys`` for j/k, which previously leaked the byte
-    # into whatever process owned the right-pane TTY (#918, reverts
-    # the right-pane forward added in #856).
+    # first/last selectable row. The one exception is Settings: its
+    # pane advertises j/k section browsing, so while Settings is active
+    # these keys are delivered to that pane instead of moving the rail
+    # cursor (#1130).
     #
     # ``⚙ Settings`` lives in a separate ``Static`` widget below the
     # nav ``ListView`` (so a long project list can scroll while the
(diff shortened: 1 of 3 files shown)


samhotchkiss:
Reviewer (sub-agent) verdict: ACCEPT — verified by tests + diff inspection (no cockpit activation; orchestrator instruction was to leave the user's running prod cockpit-14824 alone, and the change is UI-only with strong regression coverage).

- Fix wires j/k from the rail into the Settings pane via `cockpit_input_bridge` whenever `selected_key == "settings"`; the pane's existing `Binding("j,down", "nav_down")` / `Binding("k,up", "nav_up")` then move `_nav_cursor` and call `_refresh_narrow_overlay_if_active()` — so the narrow-mode hint actually does what it advertises.
- Scope is tight: one source file plus a regression test (`test_cockpit_k_on_settings_forwards_to_settings_pane`, updated `test_cockpit_j_at_last_nav_row_steps_onto_settings`) and a bridge-registration test (`test_settings_pane_registers_input_bridge`).
- Wide-mode path: the same forwarding works; `action_nav_down/up` updates the visible section cursor in the side-by-side layout, so >=110 col users keep their behavior.
- Module boundaries respected (UI-only — no facade/domain/store touch).
- Trade-off noted: the prior #1080 `k`-from-Settings round-trip back onto the last nav row is gone; users now leave Settings via the documented `q`/`escape`. Acceptable given the bug couldn't be fixed while keeping that round-trip.
- Nit (non-blocking): comment at `src/pollypm/cockpit_ui.py:2318` ("Step up off the virtual Settings row onto the last selectable nav row (#1080)") is now stale since the settings branch returns early.

Tests: 165/165 in `tests/test_cockpit.py` + `tests/test_settings_ui.py` pass.

PR #1144, opened 2026-05-04
Fix #1139 carry round start env into launches

samhotchkiss:
Closes #1139

Propagates ROUND_START_ISO_TS from the orchestrating process into local runtime launcher payloads and Docker runtime exports so spawned agent shells can bound scenario log checks to the current round.

Layer audit: touched runtime environment plumbing in src/pollypm/runtime_env.py and runtime wrappers in src/pollypm/runtimes/{local,docker}. No UI, store, or domain-layer imports were added.

PR #1147, opened 2026-05-04
Fix #1140 carry round counters into launches

### LINKED ISSUE #1140: [blocks-scenario] Scenario 37 (counter-stability) cannot run: ROUND_START_ERRNO24/HBFAIL/MISSING_WINDOW env vars not exported
## Repro
Run scenario 37-counter-stability via the smoke-loop orchestrator.

```bash
echo "ROUND_START_ERRNO24=$ROUND_START_ERRNO24"
echo "ROUND_START_HBFAIL=$ROUND_START_HBFAIL"
echo "ROUND_START_MISSING_WINDOW=$ROUND_START_MISSING_WINDOW"
```
All three are empty in the tester sub-agent's environment. Only `ROUND_START_ISO_TS=2026-05-04T20:01:00Z` was forwarded by the orchestrator.

## Expected
Per scenario fixture step, the orchestrator must export `ROUND_START_ERRNO24`, `ROUND_START_HBFAIL`, and `ROUND_START_MISSING_WINDOW` before dispatching the tester so deltas can be computed.

## Actual
Env vars empty. The fixture's own contract says: `[ -n "$round_start_errno24" ] || { file_issue "ROUND_START env vars not set"; return blocked; }` — so the scenario short-circuits to blocked.

## Round counter snapshot (for orchestrator's findings)
At handoff (timestamp 2026-05-04 15:33 local, errors.log mtime):
- total Errno 24 across whole log: 983
- total HeartbeatRail tick failed: 507
- total missing-from-tmux: 0
- last ~1h Errno 24 count: 579 (major spike — needs root-cause but cannot be attributed to "this round" without a baseline)
- last ~1h HBfail: 4
- last ~1h missing-window: 0

The Errno 24 spike (579 events in ~1h) is suspicious and worth a separate investigation, but per protocol it cannot be filed as a `regression` from this scenario because the round-start baseline is unavailable.

## Scenario context
Scenario 37 — Counter Stability
Round: 1
tmux pane size: 220x80
Companion to #1139 (scenario 36 had the same orchestrator-side env-var omission, different vars).

## Severity
blocks-scenario

## Filed by
Tester sub-agent for round 1.

samhotchkiss:
Closes #1140

Carries the explicit ROUND_START_* scenario context keys from the orchestrating process into each planned LaunchCommand env before runtime wrapping. This includes ROUND_START_ERRNO24, ROUND_START_HBFAIL, and ROUND_START_MISSING_WINDOW so counter-stability testers can compute round-local deltas.

Layer audit: touched the default launch planner in plugins_builtin/default_launch_planner and its planner test. No UI, store/IO, facade, or cross-layer imports were added.

### DIFF
--- a/src/pollypm/plugins_builtin/default_launch_planner/planner.py
+++ b/src/pollypm/plugins_builtin/default_launch_planner/planner.py
@@ -18,6 +18,8 @@ from __future__ import annotations
 
 from dataclasses import dataclass, replace
 import logging
+import os
+from collections.abc import Mapping
 from typing import TYPE_CHECKING, Callable
 
 from pollypm.models import AccountConfig, ProviderKind, SessionConfig, SessionLaunchSpec
@@ -37,6 +39,12 @@ if TYPE_CHECKING:
 from pollypm.models import CONTROL_ROLES as _CONTROL_ROLES
 
 _ROUTED_ROLES = frozenset({"operator-pm", "architect", "worker", "reviewer"})
+_ROUND_START_ENV_KEYS = (
+    "ROUND_START_ISO_TS",
+    "ROUND_START_ERRNO24",
+    "ROUND_START_HBFAIL",
+    "ROUND_START_MISSING_WINDOW",
+)
 _log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Reviewer accept (self-PR, can't formally approve).

Verified by test only (planner-internal change, well-scoped helper).

- Diff matches PR title: adds `_carry_round_start_env` helper at the planner layer that fills in any of `ROUND_START_{ISO_TS,ERRNO24,HBFAIL,MISSING_WINDOW}` from the ambient process env into `launch.env`, but never overwrites an account/session-set value.
- Test pins the exact invariant the issue describes (three new vars propagate; explicit env wins over ambient).
- No layer crossings — stays in `plugins_builtin/default_launch_planner`. Complementary to #1144 (runtime layer) and additive for the three new keys; no conflict.
- Scoped tests green: 10/10 in test_launch_planner.py, runtime/runtime_env tests still pass.
- No anti-patterns; `os.environ` read at call-time so testable via monkeypatch.

PR #1150, opened 2026-05-04
Fix #1148 handle corrupt state db

### LINKED ISSUE #1148: [regression] Corrupted state.db: pm misdiagnoses as 'pending migrations', migrate/status raise raw tracebacks
## Repro
Isolated HOME, install pm, init, register a project, then truncate state.db to garbage:
```bash
TEST_HOME=/tmp/polly-corrupt-test-$(date +%s)
mkdir -p "$TEST_HOME"
export HOME="$TEST_HOME"
unset TMUX TMUX_PANE
uv tool install --reinstall /Users/sam/dev/pollypm >/dev/null 2>&1
"$HOME/.local/bin/pm" init >/dev/null
SCRATCH=/tmp/polly-corrupt-project-$(date +%s)
mkdir -p "$SCRATCH"; cd "$SCRATCH"; git init -q; touch a; git add a; git commit -q -m init
"$HOME/.local/bin/pm" add-project "$SCRATCH"

# Corrupt the centralized store
echo "garbage" > "$HOME/.pollypm/state.db"

# Now invoke common commands a user would run
"$HOME/.local/bin/pm"                # bare
"$HOME/.local/bin/pm" status         # crashes
"$HOME/.local/bin/pm" migrate --check
"$HOME/.local/bin/pm" migrate --apply
"$HOME/.local/bin/pm" doctor         # this one IS correct
```

## Expected
A user with a corrupted `state.db` should get a clear, friendly error pointing at the corruption and a real recovery path (e.g. `pm doctor`, `pm repair`, restore-from-backup, etc.). Per scenario 52: "the cockpit either (a) shows a clear error pointing at the bad project, or (b) auto-recovers via pm doctor / similar. It does NOT silently render stale data or crash."

## Actual
Four distinct failure modes for the same underlying corruption:

1. **`pm` (bare) misdiagnoses corruption as 23 pending schema migrations** and points the user at `pm migrate --apply` as the "recommended" recovery path:
```
✗ Cannot start — 23 pending schema migrations on state.db.

Your PollyPM was upgraded since last boot and the database needs an update
before any session can connect. ...

Next: Apply the pending migrations with pm migrate --apply
...
DB: /private/tmp/polly-corrupt-test-1777932250/.pollypm/state.db
23 pending migrations:
  [state] v1: Rebuild alerts unique index
  ...
```
The DB isn't behind on migrations; it's a non-database file. The user is now being steered toward a wrong recovery action.

2. **`pm migrate --check` crashes** with a full Rich-rendered SQLAlchemy/Python traceback (50+ lines) ending in:
```
File "pollypm/cli_features/migrate.py", line 105, in migrate
    _run_check(db_path)
...
DatabaseError: (sqlite3.DatabaseError) file is not a database
```

3. **`pm migrate --apply` crashes** with a full traceback through `pollypm/store/migrations.py` -> `pollypm/storage/state.py` (50+ lines), same root cause. So the recovery path the bare-pm message recommended dumps a stack trace at the user.

4. **`pm status` crashes** with a full Rich traceback through `sqlalchemy/engine/base.py` -> `pollypm/store/engine.py:160 _set_pragmas`, ending:
```
DatabaseError: (sqlite3.DatabaseError) file is not a database
```

5. `pm doctor` IS correct — flags `storage-backend: storage backend 'sqlite' failed to initialise ((sqlite3.DatabaseError) file is not a database)`. So the diagnostic exists, it's just not used to gate the bare-pm and migrate paths.

## Scenario context
Scenario 52 — Corrupted State DB
Round: 1
tmux pane size: N/A (CLI only — see "Why no cockpit test" below)

## Why no cockpit test
Scenario fixture assumed a per-project `\$SCRATCH/.pollypm/state.db` exists; PollyPM actually centralizes state at `\$HOME/.pollypm/state.db`. After corrupting the centralized store, I did NOT run `pm up` because of #1143 (pm up leaks `HOME=/Users/sam`), which would have either polluted the canonical cockpit or simply opened the canonical (healthy) state.db, defeating the test. Filing the CLI-surface regressions instead since they reproduce the same data-integrity error-handling failure the scenario was designed to catch.

## Severity
regression — release-blocker per the scenario file ("data corruption is a real failure mode"). The misdiagnosis (item 1) is the worst: a user is actively misled to run `pm migrate --apply`, which then crashes with a traceback. There is no path through the CLI that says "your state.db is corrupt; here's how to recover" except `pm doctor`, which a user is unlikely to discover.

## Filed by
Tester sub-agent for round 1, scenario 52.

samhotchkiss:
Closes #1148

Detect existing state.db files that SQLite cannot read and render a structured corruption/recovery message instead of treating them as pending migrations or letting status/migrate crash.

Self-audit: touched the store migration facade for corruption classification and the CLI/migrate/status entrypoints that already route through the migration gate; no new cross-layer imports beyond the existing CLI-to-migration-gate path.

### DIFF
--- a/src/pollypm/cli_features/migrate.py
+++ b/src/pollypm/cli_features/migrate.py
@@ -110,7 +110,10 @@ def register_migrate_commands(app: typer.Typer) -> None:
 def _run_check(db_path: Path) -> None:
     from pollypm.store import migrations as _migrations
 
-    status = _migrations.inspect(db_path)
+    try:
+        status = _migrations.inspect(db_path)
+    except _migrations.UnusableDatabaseError as exc:
+        _migrations.exit_unusable_database(exc)
     if status.up_to_date:
         typer.echo(f"All migrations up to date ({db_path}).")
         raise typer.Exit(code=0)
@@ -227,7 +230,10 @@ def _run_apply(db_path: Path, *, force: bool = False) -> None:
         if live:
             _refuse_live_processes(live)
 
-    outcome = _migrations.apply(db_path)
+    try:
+        outcome = _migrations.apply(db_path)
+    except _migrations.UnusableDatabaseError as exc:
+        _migrations.exit_unusable_database(exc)
     if outcome.already_up_to_date:
         typer.echo("All migrations up to date.")
     else:
--- a/src/pollypm/cli_features/session_runtime.py
+++ b/src/pollypm/cli_features/session_runtime.py
@@ -247,6 +247,7 @@ def register_session_runtime_commands(app: typer.Typer, *, helpers) -> None:
     ) -> None:
         if not helpers._config_option_was_explicit():
             config_path = helpers._discover_config_path(config_path)
+        helpers._enforce_migration_gate(config_path)
         from pollypm.service_api import PollyPMService
 
         payload = PollyPMService(config_path).session_status(session_name)
--- a/src/pollypm/store/migrations.py
+++ b/src/pollypm/store/migrations.py
@@ -45,6 +45,15 @@ NAMESPACE_STATE = "state"
 NAMESPACE_WORK = "work"
 
 
+class UnusableDatabaseError(RuntimeError):
+    """Raised when ``state.db`` exists but cannot be read as SQLite."""
+
+    def __init__(self, db_path: Path, detail: str) -> None:
+        self.db_path = db_path
+        self.detail = detail
+        super().__init__(f"{db_path}: {detail}")
+
+
 @dataclass(frozen=True)
 class PendingMigration:
     """One migration that has not yet been applied to a target DB."""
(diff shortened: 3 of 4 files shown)


samhotchkiss:
Reviewer LGTM (auto-approval blocked because PR author == authenticated user).

Activated + CLI-driven verify in isolated HOME. All four originally-broken paths from #1148 now produce a structured corruption error with `pm doctor` recovery hint, exit code 2, no traceback:
- bare `pm` — no longer misdiagnoses 23 pending migrations
- `pm status` — gated through `_enforce_migration_gate`
- `pm migrate --check` — caught at `_run_check`; no SQLAlchemy traceback
- `pm migrate --apply` — caught at `_run_apply`; no longer self-recommends and crashes

Implementation: `UnusableDatabaseError` raised from a `PRAGMA quick_check` probe in `_ensure_readable_sqlite` at the front of `inspect()`, so `apply()` and the refuse-start gate both gain corruption classification for free. Detector matches the three relevant SQLite errors. `format_unusable_database_message` uses StructuredUserMessage with `pm doctor` / list-backups / `pm restore` actions.

Tests: 21/21 in tests/test_migration_gate.py pass (4 new regression tests pin the behavior). No anti-patterns; no scope creep.

PR #1154, opened 2026-05-04
Fix #1142 bootstrap fresh DB migrations

### LINKED ISSUE #1142: [blocks-scenario] Fresh install: 7 [work] migrations pending after pm init + first pm; second pm hits migration gate
## Repro
Exact sequence (copy-pasteable, isolated HOME):
\`\`\`bash
TEST_HOME=/tmp/polly-fresh-install-test-$(date +%s)
rm -rf \"$TEST_HOME\"; mkdir -p \"$TEST_HOME\"
export HOME=\"$TEST_HOME\"; export TMUX=
uv tool install --reinstall /Users/sam/dev/pollypm
\"$HOME/.local/bin/pm\" init
# first pm — runs [state] migrations silently, leaves [work] migrations unapplied
\"$HOME/.local/bin/pm\"
# second pm — blocked by migration gate
\"$HOME/.local/bin/pm\"
\`\`\`

## Expected
After \`pm init\` + first \`pm\`, the state.db is fully migrated. Subsequent \`pm\` invocations launch the cockpit normally without telling the user \"7 pending schema migrations.\"

A fresh install should not require the user to run \`pm migrate --apply\` to launch the cockpit on the second invocation.

## Actual
After \`pm init\`, \`pm migrate --check\` shows **23 pending migrations** ([state] v1-v13 and [work] v1-v7 plus a few more state). The first bare \`pm\` invocation appears to run [state] migrations implicitly, but leaves all 7 [work] migrations unapplied. The second bare \`pm\` is blocked:

\`\`\`
$ pm
✗ Cannot start — 7 pending schema migrations on state.db.

Your PollyPM was upgraded since last boot and the database needs an update
before any session can connect. Running without it risks cross-version data
corruption.

Next: Apply the pending migrations with pm migrate --apply (pm is the PollyPM CLI installed alongside pollypm).

[...]
DB: /private/tmp/polly-fresh-install-test-1777931306/.pollypm/state.db
7 pending migrations:
  [work] v1: Initial schema — baseline version
  [work] v2: Add work_sync_state table for per-adapter per-task sync tracking
  [work] v3: Add entry_type column to work_context_entries for inbox reply/read markers
  [work] v4: Add notification_staging table for pm notify priority tiering / digest rollups
  [work] v5: Add hot-path indexes for active-task and dependency-from queries
  [work] v6: Persist worker provider + transcript home so per-task archival can locate the right Claude/Codex tree at teardown (#809)
  [work] v7: Track per-execution kickoff delivery so the heartbeat sweep can force the first 'Resume work' ping past the bootstrap race (#922)
\`\`\`

The error text also says \"PollyPM was upgraded since last boot\" — but this is a brand-new install, never booted before. Misleading copy.

\`pm migrate --apply\` does work as a workaround:
\`\`\`
$ pm migrate --apply
Applied 7 migrations to /.../state.db:
  [work] v1..v7
$ pm migrate --check
All migrations up to date
\`\`\`

## First-pm output (also surfaces a separate suspicious warning)
\`\`\`
$ pm
open terminal failed: not a terminal
\`\`\`
On a fresh install with no prior session there is no rail to recover; the message implies the launch path falls into a recovery branch on the cold path. Mentioning here in case it's the same root cause as the work-migration skip. (Filing primarily about the migration gate.)

## Scenario context
Scenario 40-fresh-install — Fresh Install
Round: 1
tmux pane size: n/a (headless test, isolated HOME)
Code at fb2f5818

## Severity
blocks-scenario — a new user installing PollyPM today will hit the migration gate on their second \`pm\` invocation. This is the new-user experience the scenario file flags as release-blocker.

## Filed by
Tester sub-agent for round 1.

samhotchkiss:
Closes #1142

Bootstraps a missing workspace state.db through the migration gate so first startup applies both state and work migration families before any partial StateStore-only initialization can occur. Added a regression test that verifies the fresh missing-DB gate path leaves both migration namespaces fully current.

Self-audit: touched only the store migration gate and its focused migration-gate tests; no UI, facade, domain, or store boundary expansion beyond existing migration-runner behavior.

### DIFF
--- a/src/pollypm/store/migrations.py
+++ b/src/pollypm/store/migrations.py
@@ -446,15 +446,17 @@ def require_no_pending_or_exit(db_path: Path) -> None:
     Skipped when ``POLLYPM_SKIP_MIGRATION_GATE`` is set — ``pm migrate``
     sets this internally so the apply path can itself open the store.
 
-    A missing DB file is NOT a gate violation: it's the legitimate
-    first-boot path where ``StateStore``/``SQLiteWorkService`` will
-    create the file and apply every migration from scratch. The gate
-    protects against the "installed new code, forgot to migrate" case,
-    which by definition requires an existing older DB.
+    A missing DB file is the legitimate first-boot path. Bootstrap it
+    fully here so startup cannot create only the ``StateStore`` half of
+    the shared database and leave work-service migrations pending for
+    the next launch. The gate protects against the "installed new code,
+    forgot to migrate" case, which by definition requires an existing
+    older DB.
     """
     if bypass_env_is_set():
         return
     if not db_path.is_file():
+        _apply_all(db_path)
         return
     status = inspect(db_path)
     if status.up_to_date:
--- a/tests/test_migration_gate.py
+++ b/tests/test_migration_gate.py
@@ -278,6 +278,31 @@ def test_migration_gate_skipped_when_bypass_env_set(
     mig_mod.require_no_pending_or_exit(db_path)
 
 
+def test_refuse_start_gate_bootstraps_missing_db_fully(tmp_path: Path) -> None:
+    """Fresh startup must not leave only work migrations pending (#1142).
+
+    ``pm init`` writes config but no DB. The first bare ``pm`` reaches
+    the migration gate before Supervisor/StateStore startup. If the gate
+    simply returns, Supervisor can create a state-migrated DB without
+    opening the work service; the second ``pm`` then refuses to start on
+    pending ``[work]`` migrations.
+    """
+    db_path = tmp_path / "state.db"
+
+    mig_mod.require_no_pending_or_exit(db_path)
+
+    status = mig_mod.inspect(db_path)
+    assert status.up_to_date
+    assert (
+        status.applied[mig_mod.NAMESPACE_STATE]
+        == status.latest[mig_mod.NAMESPACE_STATE]
+    )
+    assert (
+        status.applied[mig_mod.NAMESPACE_WORK]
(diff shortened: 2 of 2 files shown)


samhotchkiss:
Reviewer verdict: ACCEPT.

Verified by activation-free path (unit test + isolated-HOME reproduction); cockpit not activated to avoid disturbing the user's prod cockpit-32440.

Findings:
- Unit test `test_refuse_start_gate_bootstraps_missing_db_fully` pins the regression: both [state] and [work] namespaces at latest after one gate call on a missing DB.
- Manual repro in /tmp/polly-fresh-1142-test (rebased PR installed into a temp venv, isolated HOME):
  - After `pm init` + one call to `require_no_pending_or_exit`: state.applied=16/16, work.applied=7/7.
  - `pm migrate --check` then prints "All migrations up to date".
- The misleading "upgraded since last boot" copy is moot for fresh installs because the gate now succeeds rather than refusing on the first launch; real upgrades with an existing older DB still see the (accurate-for-them) refusal.
- Scope is tight: 12 lines in store/migrations.py + 25 lines of test; no new boundary crossings (`_apply_all` already opened both StateStore and SQLiteWorkService pre-PR).
- Scoped tests: test_migration_gate.py (22) + test_work_service.py (61) + test_legacy_per_project_db_migration.py (7) = 90 passed.

Note: PR was branched before #1148 landed; needed a rebase locally before tests pass cleanly. `gh pr merge --squash` will produce the rebase via GitHub. Mergeability is reported as CLEAN.

PR #1173, opened 2026-05-05
Fix #1167 surface stopped worker panes

### LINKED ISSUE #1167: [regression] Stuck worker (SIGSTOP'd) not surfaced in cockpit dashboard within 60s
## Repro
```bash
# Suspend the inner codex binary for worker-itsalive
target_window=$(tmux list-windows -t pollypm-storage-closet | grep worker-itsalive | head -1 | cut -d: -f1)
node_pid=$(tmux list-panes -t "pollypm-storage-closet:$target_window" -F "#{pane_pid}")
inner=$(ps -ef | awk -v p="$node_pid" '$3==p {print $2}')
kill -STOP "$inner"

# Wait the documented 60s heartbeat sweep
sleep 60

# Observe cockpit dashboard
tmux capture-pane -t pollypm:0.1 -p | grep -E "itsalive|stuck|⚠|⏻"
```

## Expected
Per scenario 53 success criterion: stuck worker visible in Workers view (or at minimum on the dashboard "Now" panel) within 60s, with some indicator (`stuck`, `⚠`, `⏻`, or similar) and ideally an inbox alert.

## Actual
After SIGSTOP-ing the inner codex binary (verified STAT=T+) and waiting 60s, the dashboard's "Now" panel still shows:
```
   ● itsalive
     • Hi. I've skimmed the PollyPM system guide, the worker_itsalive role…  just now
```
No warning glyph, no "stuck" label, no timestamp staleness reflected ("just now" is wrong — the worker has been frozen for over a minute). No new inbox item appears either.

Verified the worker WAS stopped: `ps -p 32038 -o stat` showed `T+`. Restored with SIGCONT after observation.

I could not directly verify the dedicated Workers view because cockpit-send-key for `A`/`I`/`g` are no-ops (see #1163), but the dashboard's "Now" feed is the highest-signal user surface and contains no indicator either.

## Scenario context
Scenario 53 — worker stuck
Round: 2
tmux pane size: 189x80 (cockpit pane 0.1)

## Severity
regression — session-health monitoring is documented in the scenario as a release-blocker; the dashboard claims the worker is active "just now" while it is in fact SIGSTOP'd. A user with a wedged worker would have no signal until they manually opened the storage closet.

## Filed by
Tester sub-agent for round 2 (scenario 53).

samhotchkiss:
Closes #1167

Adds tmux process-tree detection for SIGSTOP'd pane descendants and carries that mechanical signal through heartbeat contexts. Stopped panes now raise a pane_stopped alert and mark the runtime status as stuck, so the Home dashboard shows a stuck indicator instead of refreshing as healthy.

Layer self-audit: touched store/IO-adjacent tmux probing in src/pollypm/tmux/client.py and domain heartbeat code in src/pollypm/heartbeats/*.py; no UI imports from store and no cross-layer imports added.

### DIFF
--- a/src/pollypm/heartbeats/api.py
+++ b/src/pollypm/heartbeats/api.py
@@ -308,10 +308,20 @@ class SupervisorHeartbeatAPI:
             pane_id: str | None = None
             pane_command: str | None = None
             pane_dead = False
+            pane_stopped = False
             if window is not None:
                 pane_id = window.pane_id
                 pane_command = window.pane_current_command
                 pane_dead = window.pane_dead
+                pane_pid = getattr(window, "pane_pid", None)
+                if pane_pid:
+                    try:
+                        pane_stopped = (
+                            self.supervisor.session_service.tmux
+                            .pane_has_stopped_descendant(pane_pid=int(pane_pid))
+                        )
+                    except Exception:  # noqa: BLE001
+                        pane_stopped = False
                 try:
                     raw_snapshot_path, pane_text = self.supervisor.write_snapshot(window, self.snapshot_lines)
                 except Exception:
@@ -344,6 +354,7 @@ class SupervisorHeartbeatAPI:
                     previous_log_bytes=previous.log_bytes if previous else None,
                     previous_snapshot_hash=previous.snapshot_hash if previous else None,
                     cursor=cursor,
+                    pane_stopped=pane_stopped,
                 )
             )
         return contexts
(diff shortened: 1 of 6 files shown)


samhotchkiss:
Reviewer note: Mechanical SIGSTOP detection is the right primitive for this — walking the pane process tree via ps and routing it through a new pane_stopped context flag is cleaner than time-window heuristics. Scope is tight (heartbeats + tmux/client only), layer boundaries respected (PR body self-audits), and both new regression tests pin the behavior end-to-end. 240 heartbeat/tmux tests pass. Verified by tests (small heartbeat/store internal change, no cockpit_ui touched).

PR #1190, opened 2026-05-05
Fix #1166 keep corrupt DB cockpit pane readable

### LINKED ISSUE #1166: [regression] Corrupted state.db: cockpit content pane crash-loops, error message flickers (~1Hz respawn)
## Repro

Boot pm up under an isolated HOME, then corrupt the centralized state.db while the cockpit is alive.

```bash
TEST_HOME=/tmp/polly-corrupt-test-$(date +%s)
mkdir -p "$TEST_HOME"
HOME="$TEST_HOME" uv tool install --reinstall /Users/sam/dev/pollypm
HOME="$TEST_HOME" "$TEST_HOME/.local/bin/pm" init
HOME="$TEST_HOME" "$TEST_HOME/.local/bin/pm" migrate --apply

# Use unique tmux_session in pollypm.toml to avoid colliding with canonical
# (separate finding — see comment on #1143 for related session-name leak)
sed -i '' 's/tmux_session = "pollypm"/tmux_session = "pollypm-corrupt-r2"/' \
  "$TEST_HOME/.pollypm/pollypm.toml"

nohup env -u TMUX -u TMUX_PANE HOME="$TEST_HOME" \
  PATH="$TEST_HOME/.local/bin:$PATH" \
  script -q /dev/null "$TEST_HOME/.local/bin/pm" up < /dev/null \
  > /tmp/scenario-52-up.log 2>&1 &
disown
sleep 15

# Now corrupt the centralized DB the cockpit is reading
echo "garbage" > "$TEST_HOME/.pollypm/state.db"

# Send any key to trigger a redraw / DB read
HOME="$TEST_HOME" "$TEST_HOME/.local/bin/pm" \
  --config "$TEST_HOME/.pollypm/pollypm.toml" cockpit-send-key 'C-l'

# Watch the content pane PIDs respawn
for i in $(seq 1 12); do
  sleep 1.5
  tmux list-panes -t pollypm-corrupt-r2: -F "#{pane_id}/#{pane_pid}/#{pane_current_command}"
done
```

## Expected

After corruption is detected, the cockpit content pane shows the (very nice) "Cannot use state.db — the file is not a valid SQLite database" error message and **stays on screen** so the user can read it and follow the recovery hints.

## Actual

The corruption is correctly detected (good!) and a high-quality error screen IS rendered, but the Python pane immediately exits and is respawned by the wrapper `while true` loop in zsh, then exits again, on a ~1Hz cycle. Net effect to the user: the error flickers in and out and is hard to read, and the tmux pane churns up CPU.

Pane respawn evidence (PIDs change every ~1.5s):

```
T+1s panes=%20756/50569/zsh,
T+3s panes=%20756/50569/zsh,%20973/68375/Python,
T+4s panes=%20756/50569/zsh,%20975/68452/Python,
T+6s panes=%20756/50569/zsh,%20977/68565/Python,
T+7s panes=%20756/50569/zsh,%20978/68604/Python,
T+9s panes=%20756/50569/zsh,%20980/68915/Python,
T+10s panes=%20756/50569/zsh,
T+12s panes=%20756/50569/zsh,%20983/69213/Python,
T+13s panes=%20756/50569/zsh,%20985/69352/Python,
T+15s panes=%20756/50569/zsh,%20987/69463/Python,
T+16s panes=%20756/50569/zsh,%20988/69544/Python,
T+18s panes=%20756/50569/zsh,%20990/69619/Python,
```

A race-grabbed capture of the live frame (text is correct and useful):

```
✗ Cannot use state.db — the file is not a valid SQLite database.

PollyPM could not read its central store. This is data-store corruption, not a
pending schema migration, so `pm migrate` cannot repair it.

Next: Run `pm doctor` to confirm the failure, then restore from a recent backup with `pm restore <snapshot> --confirm` or move the corrupt state.db aside before reinitializing.

Options:
  - Diagnose
      pm doctor
  - List backups
      ls ~/.pollypm/backups
  - Restore backup
      pm restore ~/.pollypm/backups/<snapshot>.db.gz --confirm

DB: /private/tmp/polly-corrupt-test-1777939166/.pollypm/state.db SQLite error:
database disk image is malformed
```

The rail (left side) keeps rendering happily off the WAL cache, so the cockpit looks "alive" but the content pane is constantly crashing.

For crash-detection only (per protocol): errors.log shows the underlying SQLite error driving the loop:

```
sqlite3.DatabaseError: database disk image is malformed
  File ".../pollypm/storage/state.py", line 566, in __init__
    self.execute("PRAGMA auto_vacuum=INCREMENTAL")
```

Plugin replay (`task_assignment_notify`) re-opens the StateStore on each `session.created`, hitting the same error every couple of seconds, which appears to drive the respawn cycle.

## Scenario context

Scenario 52 — corrupted state.db
Round: 2 (re-test after #1148 + #1143/#1155 merged)
tmux pane size: 220x80
Installed pm: pollypm 1.0.0rc2 (built from main @ ae95fedf locally)

## Severity

regression — #1148 fixed the CLI surface (good error in `pm`/`pm migrate`/`pm status`), but the cockpit-pane render path now crash-loops instead of staying on the error screen. Pre-#1148 the cockpit render was unverified (scenario was blocked). The error text quality is excellent; the only remaining bug is "render once and stay" instead of "render and exit".

## Suggested fix layer

cockpit / pane wrapper — either (a) when the cockpit detects unrecoverable DB corruption and renders the error screen, exit with a sentinel that causes the wrapper `while true` loop in zsh to NOT respawn, or (b) keep the Textual app running on the error screen instead of returning, or (c) add a backoff so the respawn cycle doesn't churn.

## Filed by

Tester sub-agent for scenario 52 round 2.

samhotchkiss:
Closes #1166

When a cockpit right-pane launch hits unrecoverable state.db corruption, PollyPM now renders the existing recovery message and keeps the pane process alive so tmux does not churn through respawns. Normal CLI migration-gate callers still exit non-zero; only cockpit-pane launch commands opt into the hold behavior.

Layer self-audit: touched the store migration error exit path and cockpit right-pane launch command generation, plus focused regression tests. No inbox action/keymap behavior changed.

### DIFF
--- a/src/pollypm/cockpit_rail.py
+++ b/src/pollypm/cockpit_rail.py
@@ -3493,7 +3493,13 @@ class CockpitRouter:
         root = shlex.quote(str(self.config_path.parent.resolve()))
         import shutil
         pm_cmd = "pm" if shutil.which("pm") else "uv run pm"
-        args = [pm_cmd, "cockpit-pane", shlex.quote(kind)]
+        args = [
+            "env",
+            "POLLYPM_HOLD_UNUSABLE_DATABASE_SCREEN=1",
+            pm_cmd,
+            "cockpit-pane",
+            shlex.quote(kind),
+        ]
         if project_key is not None:
             # #751 — inbox and activity both use --project to scope
             # their view. Other views (settings/issues/project) use
--- a/src/pollypm/store/migrations.py
+++ b/src/pollypm/store/migrations.py
@@ -439,7 +439,21 @@ def format_unusable_database_message(error: UnusableDatabaseError) -> str:
 def exit_unusable_database(error: UnusableDatabaseError, *, code: int = 2) -> None:
     """Print the corruption message and exit with a non-zero CLI code."""
     sys.stderr.write(format_unusable_database_message(error) + "\n")
+    sys.stderr.flush()
+    if hold_unusable_database_screen_enabled():
+        _hold_unusable_database_screen(code=code)
     raise SystemExit(code)
+
+
+def _hold_unusable_database_screen(*, code: int) -> None:
+    """Keep a cockpit pane alive after rendering an unrecoverable DB error."""
+    import time
+
+    try:
+        while True:
+            time.sleep(3600)
+    except KeyboardInterrupt as exc:
+        raise SystemExit(code) from exc
 
 
 # ---------------------------------------------------------------------------
@@ -451,10 +465,15 @@ def exit_unusable_database(error: UnusableDatabaseError, *, code: int = 2) -> No
 # migration CLI can actually open the store to fix the situation. Also
 # useful for tests that want to exercise the raw store.
 _BYPASS_ENV = "POLLYPM_SKIP_MIGRATION_GATE"
+_HOLD_UNUSABLE_DATABASE_ENV = "POLLYPM_HOLD_UNUSABLE_DATABASE_SCREEN"
 
 
 def bypass_env_is_set() -> bool:
     return bool(os.environ.get(_BYPASS_ENV))
+
+
+def hold_unusable_database_screen_enabled() -> bool:
+    return bool(os.environ.get(_HOLD_UNUSABLE_DATABASE_ENV))
 
 
 def set_bypass(enabled: bool) -> None:
--- a/tests/test_cockpit.py
+++ b/tests/test_cockpit.py
@@ -5702,6 +5702,10 @@ def test_right_pane_command_uses_exec_to_avoid_orphans(tmp_path: Path) -> None:
             f"{kind!r} pane command missing ``exec`` keyword (would leak "
             f"orphan on respawn-pane -k): {command!r}"
         )
+        assert "env POLLYPM_HOLD_UNUSABLE_DATABASE_SCREEN=1" in command, (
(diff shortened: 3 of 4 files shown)


samhotchkiss:
Reviewer verdict: ACCEPT.

Verified by activation in isolated HOME. After corrupting state.db and forcing a right-pane respawn, PID stayed stable across 10s of monitoring (no respawn churn) and the full error screen rendered with the recovery hints (pm doctor / pm restore / list backups / DB path + SQLite error). Process state is interruptible-sleep (Ss+) — the holding helper does its job.

CLI callers (pm / pm migrate / pm status) keep their non-zero exit because they don't set POLLYPM_HOLD_UNUSABLE_DATABASE_SCREEN. Diff is tight (57/-1 across 4 files), boundaries clean (env-var signal between cockpit launcher and store gate, no new imports across layers), and the new regression test pins both the rendering and the hold-mode path.

Scoped tests green: tests/test_migration_gate.py (23 passed) + tests/test_cockpit.py::test_right_pane_command_uses_exec_to_avoid_orphans (1 passed).

(Cannot use --approve because PR author == reviewer identity; merging via squash.)

PR #1193, opened 2026-05-05
Fix #1192 expand project subitems on enter

### LINKED ISSUE #1192: [regression] tracked-project Enter on rail opens dashboard but does not expand sub-items (Right-arrow does)
## Repro
```bash
# Anchor at top of rail (cursor on Polly chat)
pm cockpit-send-key '<esc>'
sleep 0.3
# Walk down to first tracked project (♥· / ♡·)
pm cockpit-send-key '<down>'  # Workers
pm cockpit-send-key '<down>'  # Metrics
pm cockpit-send-key '<down>'  # Inbox
pm cockpit-send-key '<down>'  # bikepath  (♡· tracked)
sleep 0.3
pm cockpit-send-key '<cr>'    # Enter — expected to expand sub-items per scenario 26
sleep 1
tmux capture-pane -t pollypm:0.0 -p
```

## Expected
Per `26-project-dashboard-subitems.md`: after Enter on a tracked project, the rail should expand to show sub-items `Dashboard`, `PM Chat`, `Tasks`, `Settings` underneath the project entry.

## Actual
Enter only opens the project Dashboard in the content pane. The rail does NOT expand — bikepath remains a single-line entry with no sub-items. To expand sub-items the user must press `<right>` instead.

Rail capture immediately after Enter:

```
   ── projects ────────      │
 ▌ ♡· bikepath               │
   ♥· booktalk               │
   ♥· camptown               │
   ○ health-coach            │
   ♥· itsalive               │
   ○ media                   │
   ♡· polly-remote           │
   ♡· pollypm                │
   ♡· russell                │
   ♡· smoketest              │
   ♡ Activity (18 new)       │
```

After pressing `<right>` (recovery), the sub-items appear:

```
   ── projects ────────      │
   ♡· bikepath               │
 ▌     Dashboard             │
       PM Chat (Archie)      │
       Tasks                 │
       Settings              │
   ♥· booktalk               │
```

## Scenario context
Scenario 26 — project dashboard sub-items
Round: 4 (chain PROJECT-DRILLDOWN)
tmux pane size: rail 30x80, content 189x80

## Severity
regression — scenario file documents Enter as the expand key (`# Enter to expand/select`) and the assertion captures rail immediately after Enter expecting sub-items. With current behavior, the assertion as written never matches — only `<right>` reveals sub-items. Scenario passes via recovery so it's not blocks-scenario.

## Filed by
Tester sub-agent for round 4.

samhotchkiss:
Closes #1192

Enter on a bare project rail row now primes the optimistic selection to the canonical project dashboard sub-item before the async route worker runs, so the project sub-items render immediately while the dashboard opens. Added a focused cockpit regression test for the first refresh before route work.

Layer self-audit: touched only cockpit UI routing behavior and its unit test; no store/domain/service boundary changes.

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -2485,6 +2485,11 @@ class PollyCockpitApp(App[None]):
     # bounce reported in #967. The seq guard short-circuits that path.
     _route_click_seq: int = 0
 
+    def _optimistic_selected_key_for_route(self, key: str) -> str:
+        if key.startswith("project:") and key.count(":") == 1:
+            return f"{key}:dashboard"
+        return key
+
     def _ensure_navigation_controller(self) -> NavigationController:
         controller = getattr(self, "_navigation_controller", None)
         if controller is not None:
@@ -2522,7 +2527,13 @@ class PollyCockpitApp(App[None]):
         # rail highlight tracks the user's intent before the route work
         # even starts. The router may correct this once it knows the
         # canonical selection (e.g. ``project:x`` → ``project:x:dashboard``).
-        self.selected_key = key
+        optimistic_key = self._optimistic_selected_key_for_route(key)
+        self.selected_key = optimistic_key
+        if optimistic_key != key:
+            try:
+                self.router.set_selected_key(optimistic_key)
+            except Exception:  # noqa: BLE001
+                pass
         request = self._ensure_navigation_controller().accept(key)
         seq = request.request_id
         self._route_click_seq = seq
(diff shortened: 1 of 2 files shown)


samhotchkiss (commented):
Verified by test only (small UI-internal change, ~13 LOC). The fix correctly threads the optimistic project:x:dashboard key through router.set_selected_key BEFORE _refresh_rows runs, so the next build_items call reads cockpit_state['selected']=project:x:dashboard and _project_rows (in plugins_builtin/core_rail_items/plugin.py) emits the Dashboard / PM Chat / Tasks / Settings sub-items in the same tick. The regression test pins the ordering invariant: set happens first, then a refresh that sees the dashboard key on both app.selected_key and router.selected_key(), and that refresh precedes the route worker. Full cockpit suite + rail/nav suites green (192 tests). Module boundaries clean (UI-only). Scope is tight. LGTM.

PR #1200, opened 2026-05-05
Fix #1199 return from architect chat to inbox

### LINKED ISSUE #1199: [blocks-scenario] Esc + I from architect chat doesn't return to Inbox — keystrokes routed to Codex, no clean back path
## Symptom

After `d` (discuss) opens an architect chat (OpenAI Codex pane embedded in cockpit content), there is no working single-key escape back to the cockpit Inbox. Specifically:

- `<esc>` produces no visible cockpit-level response — pane stays on the architect Codex chat
- Capital `I` (global open-inbox shortcut) moves the rail's `▌` cursor to Inbox but does NOT switch the content pane (still shows Codex), and the rail/content pane disagree until an additional `<cr>` is sent
- A second `<cr>` is required to commit the rail selection and finally re-render the inbox

This is similar to closed #1102 / #1151 but is reproducing live on `7248a64f`. The two-step recovery (`I` then `<cr>`) means scenario 51's escape assertion `[[ "$post" == *"action needed"* ]]` fails: after `<esc>` + `I` the post-capture is "Home" view (or Codex), not the inbox.

Additionally, while in the architect chat, individual keystrokes from `pm cockpit-send-key` get inconsistently routed:
- Some keys (lowercase letters like `e`, `f`) are interpreted as cockpit hotkeys and trigger global navigation (e.g., to Activity)
- Some keys (uppercase letters like `R`) flow into Codex's input box
- Bursts of characters get partially dropped (e.g., typing `Reflect` keystroke-by-keystroke produced `Rlc`)

This makes it impossible to inject a long/slow prompt to the architect via the bridge, which is what scenario 51 needs as a fixture.

## Repro

```bash
pm cockpit-send-key '<esc>'
sleep 0.5
pm cockpit-send-key I        # rail to Inbox
sleep 1.5
pm cockpit-send-key '<cr>'   # open inbox
sleep 1.5
# (cursor on first plan-review row, e.g. pollypm)
pm cockpit-send-key d        # opens architect chat (Codex)
sleep 5
# Now try to escape:
pm cockpit-send-key '<esc>'  # NO visible effect
sleep 1
pm cockpit-send-key I        # rail cursor moves but content pane still Codex
sleep 1.5
tmux capture-pane -t pollypm:0.1 -p
```

## Expected

Per scenario 51, after `<esc>` + `I`, the content pane shows the cockpit Inbox ("action needed" header visible). Single-key escape from architect chat should return to inbox cursor on the row that opened it.

## Actual

After `<esc>`: pane unchanged (Codex header `OpenAI Codex (v0.128.0)` still visible).
After `I`: pane shows Home dashboard ("10 projects · 13 agents · 40 inbox" / "Last 24 hours…"), NOT the inbox. Rail cursor is on Inbox but content didn't follow until a third keystroke `<cr>` was sent.

```
   10 projects  ·  13 agents  ·  40 inbox

   Last 24 hours: 82 commits across 1 project, 10 issues completed, 40 inbox items waiting for you.

 Now
   ● Polly
     managing projects and reviewing work  just now
   ● bikepath
     task next, produce the stage artifacts, advance nodes explicitly, and…  just now
   ...
```

## Why this is a scenario-blocker

Scenario 51 (`51-architect-times-out`) needs to verify the user can escape from a hung architect chat. Even when the architect responds quickly (the natural-slow path can't be tested via bridge anyway because chars get dropped), the escape flow itself is broken: `<esc>` is a no-op and capital `I` requires a second `<cr>` to actually re-render the pane.

A real user with a wedged architect would press Esc, see nothing happen, and quit cockpit — which is the exact bad-UX failure mode the scenario is meant to catch.

## Scenario context

Scenario 51 — Architect Times Out
Round: 4
tmux pane size: 220x80 (rail 30, content 189)

## Severity

blocks-scenario (escape flow is the scenario's core assertion; bridge typing breakage prevents the slow-prompt fixture)

## Filed by

Tester sub-agent for round 4.

samhotchkiss:
Closes #1199

Records the static cockpit surface before mounting a live agent, so Esc from an Inbox-started architect chat routes back to Inbox instead of only focusing the rail. Static routes clear the return key after replacing the live pane.

Layer audit: UI/router only (`cockpit_ui`, `cockpit_rail`) plus cockpit regression tests. No store/domain boundary changes.

### DIFF
--- a/src/pollypm/cockpit_rail.py
+++ b/src/pollypm/cockpit_rail.py
@@ -863,6 +863,45 @@ class CockpitRouter:
         except Exception:  # noqa: BLE001
             pass
 
+    def _return_key_for_live_mount(self, previous_key: object) -> str | None:
+        """Return the static cockpit route to restore after a live mount."""
+        if not isinstance(previous_key, str) or not previous_key:
+            return None
+        if previous_key in {
+            "dashboard",
+            "inbox",
+            "workers",
+            "metrics",
+            "settings",
+            "activity",
+        }:
+            return previous_key
+        if previous_key.startswith(("inbox:", "activity:")):
+            return previous_key
+        if previous_key.startswith("project:"):
+            parts = previous_key.split(":")
+            if len(parts) == 2:
+                return f"{previous_key}:dashboard"
+            if len(parts) >= 3 and parts[2] in {"dashboard", "settings", "issues"}:
+                return previous_key
+        return None
+
+    def _record_live_mount_return_key(
+        self,
+        plan,
+        previous_key: object,
+    ) -> None:
+        state = self._load_state()
+        if isinstance(plan, LiveAgentPane):
+            return_key = self._return_key_for_live_mount(previous_key)
+            if return_key:
+                state["mounted_return_key"] = return_key
+            else:
+                state.pop("mounted_return_key", None)
+        else:
+            state.pop("mounted_return_key", None)
+        self._write_state(state)
+
     def should_show_palette_tip(self) -> bool:
         data = self._load_state()
         return not bool(data.get("palette_tip_seen"))
@@ -2552,6 +2591,8 @@ class CockpitRouter:
         # right pane to the stale prior selection. Writing the new key
         # to disk first keeps the rail aligned with the user's intent
         # even when downstream work raises.
+        state_before_route = self._load_state()
+        previous_key = state_before_route.get("selected")
         self.set_selected_key(key)
         token = self._begin_layout_mutation()
         try:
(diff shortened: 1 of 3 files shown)


samhotchkiss:
Reviewer (Claude sub-agent): ACCEPT.

Verified by test only (UI/router-internal change, ~115 LOC, with focused regression tests).

- _record_live_mount_return_key correctly fires before _route_content_plan, capturing the previous static surface only when the new plan is a LiveAgentPane.
- _return_key_for_live_mount whitelists the static rail surfaces (dashboard / inbox / inbox:<id> / activity / activity:<id> / workers / metrics / settings) and project:<key>(:dashboard|:settings|:issues), correctly returning None for live keys (polly, project:foo:session) so Esc still falls through to focus_rail in that case.
- _show_command_view clears mounted_return_key after a static replacement so a stale return-key doesn't bleed across mounts.
- action_back_to_home reads mounted_return_key only when a live session is mounted; the else branch keeps the prior focus_rail behaviour.
- 165 cockpit tests + adjacent rail/navigation/state-store/input-bridge suites all green; both new tests cover the route_selected write side and the action_back_to_home read side. Scope is exactly what the title claims; no boundary crossings (UI + router only).
- Activation skipped per orchestrator directive (do NOT activate into prod cockpit-52323).

PR #1222, opened 2026-05-05
Fix #1110 refuse update from agent worktrees

### LINKED ISSUE #1110: pm update sometimes runs from inside an agent worktree (.claude/worktrees/agent-*) — does git reset --hard there, can clobber agent work
## Symptom

When I ran \`pm update\` during the overnight loop, the output sometimes showed:

\`\`\`
[step] repo_root=/Users/sam/dev/pollypm/.claude/worktrees/agent-a6df6ef44745e08fa
[step] fetching origin
[step] git reset --hard origin/main
[step] uv tool install --reinstall <repo>
\`\`\`

The script picked up an **agent worktree** as repo_root rather than the canonical \`/Users/sam/dev/pollypm\`. It then \`git reset --hard origin/main\` *inside that worktree* and reinstalled the user binary from there.

This happened repeatedly across the session. Each \`pm update\` invocation seemed to bind to whichever worktree was 'most recent' (likely the worktree the last Agent dispatch ran from).

## Why this matters

1. **Data loss risk**: \`git reset --hard\` inside an active worktree wipes any uncommitted work. If a parallel agent had local-only changes, they'd be gone.
2. **Surprising user-binary source**: \`~/.local/bin/pm\` ends up symlinked into a transient worktree directory; later worktree pruning could break the binary.
3. **Mostly silent**: a user running \`pm update\` doesn't expect their installed binary to be re-rooted into \`.claude/worktrees/\` and would be surprised to find it gone after worktree cleanup.

## Likely cause

\`pm update\`'s \`repo_root\` resolution probably uses the current working directory or some uv-tool-install lookup that follows the most-recent install source. When agent dispatches use \`uv tool install --reinstall /tmp/wt-*\` from inside a worktree, the install pointer drifts.

## Repro

1. Spawn a background agent with \`isolation: 'worktree'\` that runs \`pm update\` as part of its workflow (or just any \`uv tool install\` from inside the worktree).
2. After the agent finishes, run \`pm update --check-only\` from the main checkout.
3. Note that \`repo_root\` shows the agent's worktree path, not \`/Users/sam/dev/pollypm\`.

## Suggested fix direction

\`pm update\` should resolve repo_root from a deterministic canonical location:

- A config-stored absolute path written at first install (\`~/.pollypm/install_source.toml\`), OR
- The directory containing \`pyproject.toml\` whose canonical Git \`origin\` matches the expected remote, walking up from CWD only if the config path is missing
- And: refuse to operate when \`repo_root\` matches \`*/.claude/worktrees/agent-*\` — print an explicit \"this looks like an agent worktree, not the canonical install path; specify --repo-root or run from the install dir\" error

## Filed by

Overnight loop session, 2026-05-04. Observed during multiple \`pm update\` invocations.

samhotchkiss:
Closes #1110

Refuses `pm update` when the resolved repo root is at or inside `.claude/worktrees/agent-*`, before fetch/reset/reinstall can run. Adds a regression test that all side-effect seams remain untouched for an agent worktree path.

Layer audit: touched `src/pollypm/update.py` in the UI/CLI update orchestration layer and `tests/test_update.py`; no boundary crossings.

### DIFF
--- a/src/pollypm/update.py
+++ b/src/pollypm/update.py
@@ -50,6 +50,19 @@ def is_git_checkout(repo_root: Path) -> bool:
     return (repo_root / ".git").exists()
 
 
+def _is_claude_agent_worktree(repo_root: Path) -> bool:
+    """True iff ``repo_root`` points at or inside a Claude agent worktree."""
+    parts = repo_root.parts
+    for index in range(len(parts) - 2):
+        if (
+            parts[index] == ".claude"
+            and parts[index + 1] == "worktrees"
+            and parts[index + 2].startswith("agent-")
+        ):
+            return True
+    return False
+
+
 @dataclass(slots=True)
 class CommitInfo:
     sha: str
@@ -267,6 +280,23 @@ def update(
     do_reset = reset_runner or (
         lambda r, target: _do_reset_hard(r, target)
     )
+
+    if _is_claude_agent_worktree(repo):
+        msg = (
+            f"refusing to update: repo_root {repo} looks like a Claude "
+            "agent worktree, not the canonical install path. Run "
+            "`pm update` from the canonical install directory."
+        )
+        step(msg)
+        return UpdateResult(
+            ok=False,
+            check_only=check_only,
+            refused=True,
+            old_sha="",
+            new_sha="",
+            commits=[],
+            message=msg,
+        )
 
     # Refuse on in_progress work — yanking source out from under a live
     # worker corrupts logs and confuses the supervisor. The user can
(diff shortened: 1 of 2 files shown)


samhotchkiss (commented):
Activated + cockpit-driven verify. Scoped tests pass (tests/test_update.py: 14/14). Diff is tight: adds _is_claude_agent_worktree() walking parts to detect .claude/worktrees/agent-* anywhere in the path, then refuses with a clear message before fetch/resolve/reset/reinstall seams fire. Regression test asserts calls list stays empty. Module boundary clean (UI/CLI update layer only).

PR #1234, opened 2026-05-05
Fix #1233 keep live chat bridge input sticky

### LINKED ISSUE #1233: [blocks-scenario] PM Chat (discuss-with-PM) input loses focus after first character; subsequent letters drive global rail navigation
## Repro
Open Inbox, navigate to a plan-review row, press 'd' to open discuss-with-PM. The cockpit shows the architect Codex chat with a "› <placeholder>" input and the architect "Working (Ns)" status. Send characters one at a time:

```bash
pm cockpit-send-key '<esc>'
sleep 0.5
pm cockpit-send-key I            # Inbox
sleep 1.5
pm cockpit-send-key d            # discuss with PM (selected row is "Plan ready for review: pollypm")
sleep 5

# Architect chat is now visible in pane 1, with a "›" input prompt and a placeholder.
# Send a multi-character prompt one key at a time, with 0.6-1.0s between sends:
pm cockpit-send-key R            # appears in input as "R"  -> OK
sleep 0.8
pm cockpit-send-key e            # NOT appended; input still shows just "R"
sleep 0.8
pm cockpit-send-key x            # NOT appended; input still shows "R"
sleep 0.8
pm cockpit-send-key z            # NOT appended

# Following keystrokes navigate the rail / global views instead.
# Sending 'a','b','c' caused the cockpit to navigate from PM Chat (Archie)
# into the global Metrics view — i.e. the keys hit the rail's keymap.
```

## Expected
After pressing `d` to open discuss-with-PM, the chat input should retain focus across the entire typed message. All printable letters should be appended to the input text until I press Enter (send) or Esc (back to list). Scenario 51 needs this so I can submit a deliberately deep prompt to provoke a slow architect response.

## Actual
Only the first letter ("R") is captured into the chat input. Every subsequent letter is interpreted as a rail/global keybinding. For example, after a single `R` is shown in the chat input, sending `a`/`b`/`c` navigates the cockpit to the Metrics view; sending `S`-ish letters navigates to Settings; etc. The chat input never receives the 2nd+ keystroke.

This made it impossible to type the scenario 51 deep-prompt ("Reflect on the deepest meaning of this plan and provide a comprehensive 50000-word analysis."). The scenario can't progress past the prompt-typing step.

## Captures

After typing R successfully (input shows "› R"):
```
────────────────────────────────────────────────────────────────


› R

  gpt-5.5 default · ~/dev/pollypm/.pollypm/workt…
```

After then sending 'e','x','z' (still "› R" only):
```
─────────────────────────────────────────────────


› R

  gpt-5.5 default · ~/dev/pollypm/.pollypm/workt…
```

After then sending 'a','b','c' (cockpit jumped to global Metrics view, input pane is gone):
```
  Metrics

  ─────────────────────────────────────────────
  last refresh 16:40:12 UTC  ·  auto off

  ╭──────────────────────────────────────────╮
  │                                          │
  │                                          │
  ╰──────────────────────────────────────────╯

  …

  R refresh · A auto-refresh · ↑↓ select · ↵
```

Rail at that point shows the cursor moved off "PM Chat (Archie)" and onto Metrics in the global tree.

## Scenario context
Scenario 51 — Architect Times Out
Round: 5
Cockpit SHA: 305a524a
tmux pane size: rail 30x23, content 49x23 (window 80x24)

This blocks the scenario at fixture step 2 (typing the deep prompt). Without being able to type a multi-character prompt into the architect chat from the cockpit, we cannot exercise the timeout-UX path.

Notes:
- The scenario's "press <space> to focus" line did not visibly do anything — focus appears to be implicit on entering discuss view (the first keystroke does land in input, but only the first).
- This is likely the same family of bug as the cockpit-input-bridge intercept rules: characters reach the rail handler instead of the chat-input child.

## Severity
blocks-scenario

## Filed by
Tester sub-agent for round 5.

samhotchkiss:
Closes #1233

Keeps `pm cockpit-send-key` routed to a mounted live chat after the first successful right-pane delivery, even if a later CLI invocation observes stale tmux active-pane state. Escape clears the sticky bridge ownership so the cockpit rail/back path still receives the key.

Layer audit: touched CLI/UI bridge routing (`src/pollypm/cli_features/ui.py`) and input bridge tests only. Bug layer is UI/CLI. No boundary violation; no `cockpit_ui.py`, `cockpit_rail.py`, or inbox test overlap with active PRs.

### DIFF
--- a/src/pollypm/cli_features/ui.py
+++ b/src/pollypm/cli_features/ui.py
@@ -20,6 +20,7 @@ import typer
 from pollypm.config import DEFAULT_CONFIG_PATH
 
 _RIGHT_PANE_BRIDGE_BYPASS_ESCAPE_TOKENS = frozenset({"<esc>", "esc", "escape"})
+_LIVE_RIGHT_PANE_INPUT_STICKY = "live_right_pane_input_sticky"
 _HELP_KEY_TOKENS = frozenset({"?", "question_mark"})
 _HELP_CONTENT_BRIDGE_FALLBACK_KINDS = ("dashboard", "pane-inbox", "settings")
 _INBOX_BRIDGE_FIRST_TOKENS = frozenset({"/", "d"})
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Activated + cockpit-driven verify. Reproduced the bug pattern: with mounted_session set and tmux focus moved off the right pane (the failure condition), keys 'a','b','c' all routed correctly to %968 right pane, building `RRexzabc` in the architect input. Escape clears the sticky flag. Unit tests pin both stickiness and escape-clear paths. Diff tightly scoped to cli_features/ui.py bridge layer.

PR #1276, opened 2026-05-06
Fix #1270 route help modal scroll keys

### LINKED ISSUE #1270: [blocks-scenario] Help dialog (?) is unscrollable despite documenting scroll keys
## Repro
From cockpit Home (rail focused on Home):
\`\`\`bash
pm cockpit-send-key '?'      # opens help dialog
pm cockpit-send-key '<pgdn>'  # tries to scroll
pm cockpit-send-key '<down>'
pm cockpit-send-key 'f'
pm cockpit-send-key '<space>'
\`\`\`

## Expected
Help dialog scrolls to reveal the rest of the bindings. The dialog itself documents \`pagedown / Space / f → Page down\` and shows a scrollbar indicator (▄) on its right edge — both signal "scrollable content below the fold".

## Actual
NONE of the documented scroll keys (pagedown, space, f, j, down arrow) scroll the help dialog. The user sees only the first ~20 bindings ("This screen" section) and the start of "Global (anywhere in cockpit)" header, with the rest cut off. j/down keys instead leak through to the rail and move rail cursor under the dialog.

\`\`\`
   ╭──────────────────────────────────────╮
   │                                      │
   │  Keyboard shortcuts  —               │   <-- dangling em-dash, no subtitle
   │                                      │
   │  This screen                         │
   │    j / ↓                             │
   │        Down                          │
   ...
   │    pagedown / Space / f              │
   │        Page down                  ▄  │   <-- scrollbar promises more
   │                                      │
   │  Global (anywhere in cockpit)        │   <-- cut off here
\`\`\`

The help dialog title also has a trailing em-dash with no following text — looks like a missing subtitle/version.

## Scenario context
Scenario 70 — first-impression
Round: 8
tmux pane size: 80x24

## Severity
blocks-scenario — \`?\` is the universal "I need help" affordance and one of the very first things a confused user will press. If it shows truncated content with no way to scroll, the user concludes there is no help. The dialog actively misleads by listing scroll keys that do not work inside it.

## Filed by
Tester sub-agent for round 8.

samhotchkiss:
Closes #1270

When Home help opens on the visible content-pane bridge, follow-up help-modal control keys now continue to that same bridge instead of falling back to the cockpit rail bridge. The Home dashboard help title also uses the short Home label.

Layer self-audit: touched UI/CLI input routing and keyboard-help presentation/tests only; no service, domain, store, or cross-layer imports added.

### DIFF
--- a/src/pollypm/cli_features/ui.py
+++ b/src/pollypm/cli_features/ui.py
@@ -11,6 +11,7 @@ Contract:
 from __future__ import annotations
 
 import logging
+import time
 import traceback
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


samhotchkiss (commented):
Activated + cockpit-driven verify. Repro confirmed fixed: pressing ? on Home opens help modal, then <pgdn>/<space>/f/<down>/<home> all scroll the modal correctly, revealing the previously-hidden Global section. <esc> dismisses cleanly and subsequent <down> properly routes back to the rail. Scoped tests (test_cockpit_input_bridge, test_keyboard_help) all green (69 passed). Title 'Home' replaces the dangling em-dash. Bridge-tracking via CockpitRouter._load/_write_state matches the pattern already established in cli_features/ui.py.

Issue #1292, opened 2026-05-06
[cosmetic] Polly chat network-dead error overwrites input prompt; appended typed chars and no retry hint


samhotchkiss:
## Repro
With `pm dev simulate-network-dead` armed, submit a typed message in cockpit Polly chat:

```bash
pm dev simulate-network-dead
# (already inside Polly chat, marker armed)
pm cockpit-send-key p
pm cockpit-send-key i
pm cockpit-send-key n
pm cockpit-send-key g
pm cockpit-send-key '<cr>'
# error appears in <1s, replacing the empty input prompt
# now type more characters to "recover":
pm cockpit-send-key r
pm cockpit-send-key t
```

## Expected
After a failed submit, the input prompt should clear (or surface error in a separate banner/toast), and offer a recovery hint such as "press Enter to retry" or "try again". Typed characters should go into a fresh prompt, not be appended to the error string.

## Actual
Error string replaces the `❯` input prompt and stays there:

```
❯ PollyPM chat failed: network unreachable (simulated): connection refused for cockpit live chat submit
```

When the user types `r` then `t` to begin recovery, those chars are appended directly:

```
❯ PollyPM chat failed: network unreachable (simulated): connection refused for cockpit live chat submitrt
```

There is also no recovery hint in the error message — no "retry", "try again", "press X to", "check connection" etc. The user has to figure out themselves to backspace through the entire error string.

Additionally: the error briefly opens with a literal "(simulated)" tag — fine for dev fixtures but worth flagging since the same code path will fire on real failures (and "(simulated)" would be a lie there). Verifying that `(simulated)` is gated to the marker path is recommended.

## Scenario context
Scenario 91 — Network-Dead Error Is Actionable
Round: 1 (after-merge of #1232 fix)
tmux pane size: 220x80
SHA: 79d746ed (#1232 fix is included)

## Severity
cosmetic — error IS shown promptly (success criterion 1 PASSes), input IS responsive (success criterion 2 partially PASSes), but the in-prompt error placement and missing retry hint hurt recovery UX.

## Filed by
Tester sub-agent for round 1, scenario 91.

samhotchkiss:
Codex picking this up. Working in branch fix/issue-1292.

samhotchkiss:
Codex deferred — file overlap with open PR #1294. Will retry after that PR resolves.

samhotchkiss:
Reproduced in round 2, scenario 91, SHA d7cfc5b5.

[...]

PR #1316, opened 2026-05-06
Fix #1315 show network-dead chat failure

### LINKED ISSUE #1315: [regression] Polly chat network-dead failure is now silent — no error, no echo, no Polly reply
## Summary
PR #1304 (Fix #1292, SHA 20b368ad) over-corrected. Round 4 of scenario 91: with `pm dev simulate-network-dead` armed, submitting a chat prompt produces ZERO visible feedback in the cockpit Polly chat pane. No error message, no echo of the user's prompt, no Polly response. The marker file is consumed (so the dispatch path runs), but the user has no signal that anything failed — or that anything happened at all.

This is a different shape of bug than #1292 (where the error overwrote the input prompt). #1292's fix successfully kept the error out of the prompt slot, but appears to have suppressed the error UI entirely.

## Repro
```bash
# In cockpit Polly chat (rail row 2), input field idle:
pm dev simulate-network-dead   # arms; marker at ~/.pollypm/dev_network_simulation.json
ls /Users/sam/.pollypm/dev_network_simulation.json  # confirm armed

pm cockpit-send-key '<tab>'    # focus chat input (was on rail)
pm cockpit-send-key 't'
pm cockpit-send-key 'e'
pm cockpit-send-key 's'
pm cockpit-send-key 't'
pm cockpit-send-key '<cr>'

# Wait 8s, capture
sleep 8
tmux capture-pane -t pollypm:0.1 -p | tail -25
ls /Users/sam/.pollypm/dev_network_simulation.json  # gone — dispatch happened
```

## Expected
Per scenario 91 success criteria:
1. Error visible within 8s
2. Input remains usable after error
3. Error message has a recovery hint

## Actual
1. **FAIL** — no error visible within 8s. Pane content is unchanged from before the submit (last Polly bubble is the pre-submit recap).
2. PASS — input is usable; typing `h` `i` after the silent failure produced `❯ hi` correctly.
3. **FAIL by vacuity** — no error message exists, so no retry hint exists.

The marker file is consumed, confirming the dispatch fired and hit the simulated `connection refused`. But the chat pane has no visible reaction. The user's submitted "test" message is also not echoed — typical chat apps echo the user's bubble even on transport failure.

## Capture (post-submit, 8s elapsed)
```
  Active inbox: 0 item(s)
  Use the visible inbox and status context, then follow the user's lead.

⏺ Same picture — 10 projects, inbox empty. Standing by.

✻ Churned for 1s

※ recap: You're onboarding me as Polly the operator-pm; I've read SYSTEM.md and operator.md and confirmed inbox is empty across all 10 projects. Next action is whatever you want to drive —
  I'm standing by for your lead. (disable recaps in /config)

──────────────────────────────────────────────────────────────────────
❯ check the inbox
──────────────────────────────────────────────────────────────────────
  ⏵⏵ bypass permissions on (shift+tab to cycle)
```
(`check the inbox` is the dim placeholder/ghost text, verified via tmux ANSI capture showing `[0;2m` faint sequence — input is actually empty after submit.)

## Scenario context
Scenario 91 — Network-Dead Error Is Actionable
Round: 4
SHA: 20b368ad (PR #1304 merged)
tmux pane size: 220x80
Related closed: #1292 (was: error overwrites prompt). Fix went too far in the opposite direction.

## Severity
regression — PR #1304 was verified PASSing in round 3 merge note; round 4 shows the error UI is now suppressed entirely. Worse UX than the original #1292 (at least there the user knew something had failed).

## Filed by
Tester sub-agent for round 4.

samhotchkiss:
Closes #1315

Restores a visible, actionable network-dead failure in the live Polly chat pane by writing a temporary failure prompt instead of relying on an ephemeral tmux display message. The next typed key, Enter, Backspace, or Escape clears that temporary prompt before continuing so the user can recover without appending to the error text.

Layer audit: UI/CLI only. Touched the cockpit send-key bridge in src/pollypm/cli_features/ui.py and focused regression coverage in tests/test_cockpit_input_bridge.py; no facade, domain, or store changes.

### DIFF
--- a/src/pollypm/cli_features/ui.py
+++ b/src/pollypm/cli_features/ui.py
@@ -22,6 +22,7 @@ from pollypm.config import DEFAULT_CONFIG_PATH
 
 _RIGHT_PANE_BRIDGE_BYPASS_ESCAPE_TOKENS = frozenset({"<esc>", "esc", "escape"})
 _LIVE_RIGHT_PANE_INPUT_STICKY = "live_right_pane_input_sticky"
+_LIVE_CHAT_NETWORK_DEAD_PROMPT_ACTIVE = "live_chat_network_dead_prompt_active"
 _HELP_MODAL_BRIDGE_KIND = "help_modal_bridge_kind"
 _HELP_MODAL_SELECTED_KEY = "help_modal_selected_key"
 _HELP_MODAL_OPENED_AT = "help_modal_opened_at"
@@ -117,7 +118,7 @@ _RIGHT_PANE_TMUX_KEY_TOKENS: dict[str, str] = {
 }
 _LIVE_CHAT_NETWORK_DEAD_MESSAGE = (
     "PollyPM chat failed: network unreachable. "
-    "Input cleared; check connection and try again."
+    "Type again to clear this and retry; check connection."
 )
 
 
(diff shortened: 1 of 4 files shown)


samhotchkiss:
Bounce — driving scenario 91 in cockpit reproduces a regression of the original #1292 shape.

## Activated + cockpit-driven verify

Steps:
1. `pm dev simulate-network-dead`
2. Navigate to Polly chat, wait ~30s for briefing to settle (briefing consumes the marker — re-arm)
3. `pm dev simulate-network-dead` again
4. Type `t e s t` then `<cr>`

Result: within ~5s, the error text appears at the chat input prompt slot:
```
❯ PollyPM chat failed: network unreachable.
  Type again to clear this and retry; check
  connection.
```

## What's broken: error contaminates the input prompt (regression of #1292)

The fix design — write error via `send_keys` with `press_enter=False`, then send `C-u` on the next typed key — does **not** clear the error from the prompt. Reproduced just now:

- After error appears, type `h` → prompt reads `❯ ... Type again to clear this and retry; check h` — the `C-u` only killed the trailing wrap line up to the cursor; the wrapped error text on the previous two visible lines remains, with `h` appended.
- Backspace then peels off the error one char at a time — user has to backspace ~95 chars to recover an empty prompt. The PR title claims "the next typed key, Enter, Backspace, or Escape clears that temporary prompt" — observed behaviour is that none of those clear the wrapped portion.
- **Worse**: pressing `<cr>` while the error sits in the prompt **submits the error text as a chat message to Polly.** Captured: `❯ PollyPM chat failed: network unreachable. Type again to clear this and ret` was dispatched to the LLM and Polly began composing a reply (`✶ Cultivating… (2s · thinking)`). The previous bug (#1292) at least kept the user typing into a prompt that contained the error; this version turns Enter into a "submit the error to the LLM" gesture.

## Why `C-u` isn't enough

The error message wraps across multiple lines in Claude Code's input box. `tmux send-keys C-u` is interpreted by the inner TUI's readline as kill-to-beginning-of-line, which only clears one visual line. The wrapped portion above the cursor stays on screen, so the inner buffer still contains those characters when Enter fires.

## Recommended direction

The acceptance criterion the orchestrator gave was "error visible in **banner** (not input prompt) AND latency ≤ 8s AND has retry hint." Putting the error text into the live chat prompt slot via `send_keys` will keep hitting this class of contamination because the prompt is a typed buffer, not a display surface.

Suggested approaches:
1. Surface the error via the rail/cockpit toast/banner channel rather than the live right pane's input. The cockpit already has surfaces for non-input messages (status line, rail, etc.) — pick one that doesn't write into the inner TUI's stdin.
2. If reusing `display-message` (the previous approach), increase its visibility window and add a banner-line render rather than relying on tmux's overlay alone.
3. If keeping the in-prompt approach, the clear path needs to send enough `BSpace`/`C-u` sequences to wipe the *entire* wrapped buffer, not just one line; and must run **before** the user's `<cr>` reaches the inner TUI.

## Files inspected
- src/pollypm/cli_features/ui.py
- tests/test_cockpit_input_bridge.py

## Tests run
- `uv run --extra dev pytest tests/test_cockpit_input_bridge.py -x` — 44 passed

The unit tests pass because they assert call sequences against a `FakeTmux`; they don't exercise the inner TUI's response to `C-u` against a wrapped multi-line input. A regression test that catches this would need either an integration-level cockpit harness or an explicit assertion that the buffer-clear sequence sent matches the wrapped error length.

samhotchkiss:
Addressed reviewer feedback in 4c45838f

PR #1329, opened 2026-05-06
Fix #1290 speed project drilldown render

### LINKED ISSUE #1290: [cosmetic] Project drilldown render exceeds 3s budget (~4s observed)
## Repro
```bash
pm cockpit-send-key '<esc>'
sleep 0.3
pm cockpit-send-key '<esc>'
sleep 0.5
# 5 downs from Home → bikepath (first project row)
for i in 1 2 3 4 5; do pm cockpit-send-key '<down>'; sleep 0.1; done
sleep 0.3
# Time the Enter
start=$(python3 -c 'import time;print(time.time())')
pm cockpit-send-key '<cr>'
# Poll content pane for "PM:" or "needs attention" appearance
```

## Expected
Per scenario 81: pressing Enter on a project rail row opens the project Dashboard within 3s.

## Actual
Two timed samples in the same session both exceeded the 3s budget:

- **booktalk** drilldown: **3961 ms** to first content render
- **bikepath** drilldown: **4434 ms** to first content render
  - Loading placeholder ("loading project dashboard…") visible at ~906ms — good (no blank gap)
  - Final dashboard content visible at ~4434ms

Loading placeholder is present, so this is not a blank-pane stall — pure latency.

Final pane content (bikepath, after ~4.4s):
```
  bikepath   PM: Bea
  ─────────────────────────────────────────────────────────────────────
  ○ standing by
  ╭─────────────────────────────────────────────────────────────────────╮
  │ architect_bikepath (architect) is alive but standing by …           │
  ╰─────────────────────────────────────────────────────────────────────╯
  ╭─────────────────────────────────────────────────────────────────────╮
  │  Inbox                                                              │
  │  Inbox is clear for this project.                                   │
  ╰─────────────────────────────────────────────────────────────────────╯
```

## Scenario context
Scenario 81 — Project Drilldown Latency
Round: 1 (cockpit at SHA 79d746ed)
tmux pane size: 220×80
Prior observation (scenario 26+27 last loop): >2s noted; scenario 52 noted ~1.5s. This round samples are ~4s, suggesting regression vs. ≤2s historical observation.

## Severity
cosmetic (latency only — does not block the flow; loading placeholder prevents a blank pane).

## Filed by
Tester sub-agent for round 1, scenario 81.

samhotchkiss:
Closes #1290

Reuses the already-loaded project config while gathering project dashboard sections, passes prefetched task buckets into worker activity classification, opens supervisor state read-only, and caps the inline plan preview until plan focus mode. This cuts duplicated first-paint IO/render work while preserving the full plan viewer on demand.

Layer self-audit: touched UI-layer code only (src/pollypm/cockpit_ui.py) plus focused dashboard tests. No store/domain imports were added beyond existing local UI helper imports; no boundary debt.

### DIFF
--- a/src/pollypm/__main__.py
+++ b/src/pollypm/__main__.py
@@ -46,17 +46,18 @@ def _run_cockpit_pane_fast(argv: list[str]) -> bool:
 
     ``CockpitRouter`` spawns panes with ``python -m pollypm cockpit-pane ...``.
     Importing :mod:`pollypm.cli` first loads every top-level command family
-    before the pane-specific command can import its Textual app. The Inbox
-    cold path is budgeted from keystroke to first paint, so avoid that root CLI
-    import for the pane launches the cockpit itself generates. Help and any
-    unfamiliar shape fall back to the Typer command below.
+    before the pane-specific command can import its Textual app. The Inbox and
+    project-dashboard cold paths are budgeted from keystroke to first paint, so
+    avoid that root CLI import for the pane launches the cockpit itself
+    generates. Help and any unfamiliar shape fall back to the Typer command
+    below.
     """
     if len(argv) < 2 or argv[0] != "cockpit-pane":
         return False
     if any(token in _COCKPIT_PANE_HELP_FLAGS for token in argv[1:]):
         return False
     kind = argv[1]
-    if kind != "inbox":
+    if kind not in {"inbox", "project"}:
         return False
 
     parsed = _parse_cockpit_pane_options(argv)
@@ -64,6 +65,23 @@ def _run_cockpit_pane_fast(argv: list[str]) -> bool:
         return False
     config_path, project = parsed
     resolved_config_path = config_path or _default_config_path()
+    if kind == "project":
+        if not project:
+            return False
+        _paint_project_dashboard_snapshot(resolved_config_path, project)
+        from pollypm.cli_features.ui import (
+            _enforce_migration_gate,
+            _install_cockpit_debug_log_handler,
+        )
+
+        _enforce_migration_gate(resolved_config_path)
+        _install_cockpit_debug_log_handler(resolved_config_path)
+
+        from pollypm.cockpit_ui import PollyProjectDashboardApp
+
+        PollyProjectDashboardApp(resolved_config_path, project).run(mouse=True)
+        return True
+
     _paint_inbox_loading_placeholder()
     snapshot_error = _paint_inbox_snapshot(resolved_config_path, project=project)
 
@@ -80,6 +98,61 @@ def _run_cockpit_pane_fast(argv: list[str]) -> bool:
 
     PollyInboxApp(resolved_config_path, initial_project=project).run(mouse=True)
     return True
+
+
+def _paint_project_dashboard_snapshot(config_path: Path, project: str) -> None:
+    """Prepaint a cheap project-dashboard frame before Textual imports."""
+    try:
+        frame = _render_project_dashboard_snapshot(config_path, project)
+        sys.stdout.write(f"{_CLEAR_SCREEN}{frame}")
+        sys.stdout.flush()
+    except Exception:  # noqa: BLE001 - prepaint must never block launch
+        pass
+
+
+def _render_project_dashboard_snapshot(config_path: Path, project: str) -> str:
+    try:
+        from pollypm.config import load_config
+
+        config = load_config(config_path)
+        project_obj = (getattr(config, "projects", {}) or {}).get(project)
+        if project_obj is None:
+            return (
+                f"{project}   PM: Project PM\n\n"
+                "Loading project dashboard..."
+            )
+        name = (
+            project_obj.display_label()
+            if hasattr(project_obj, "display_label")
+            else (getattr(project_obj, "name", None) or project)
+        )
+        persona = getattr(project_obj, "persona_name", None)
+        pm = (
+            f"PM: {persona.strip()}"
+            if isinstance(persona, str) and persona.strip()
+            else "PM: Project PM"
+        )
+        return "\n".join(
+            [
+                f"{_plain_inbox_text(name)}   {_plain_inbox_text(pm)}",
+                "",
+                "Loading project dashboard...",
+                "",
+                "Inbox",
+                "  Loading project inbox...",
+                "",
+                "Current activity",
+                "  Loading worker state...",
+                "",
+                "Task pipeline",
+                "  Loading task pipeline...",
+                "",
+                "Recent activity",
+                "  Loading recent activity...",
+            ]
+        )
+    except Exception:  # noqa: BLE001
+        return f"{project}   PM: Project PM\n\nLoading project dashboard..."
 
 
 def _paint_inbox_loading_placeholder() -> None:
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 import gc
 import asyncio
 import json
+import logging
 import os
 import resource
 from pathlib import Path
@@ -152,6 +153,8 @@ from pollypm.cockpit_rail import CockpitItem, CockpitPresence, CockpitRouter
 
 import re as _re
 
+
+logger = logging.getLogger(__name__)
 
 _INLINE_BOLD_RE = _re.compile(r"\*\*(.+?)\*\*")
 _INLINE_ITALIC_RE = _re.compile(r"\*(.+?)\*")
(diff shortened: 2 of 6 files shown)


samhotchkiss:
## Reviewer bounce: drilldown still slow (live verify)

Activated the PR (uv tool install --reinstall + pm reset --force + phantom-client up) and timed cold drilldowns from the rail at 220x80. The PR does not deliver a meaningful speedup over the pre-PR baseline (rounds 1-6 in #1290 reported 3.6-4.4s cold; user-stated acceptance bar in this review pass is "well under 1s").

### Live timings observed on this PR (a50177b3)

| Sample | Project | Cold/Warm | Render to "PM:" |
|---|---|---|---|
| 1 | bikepath | cold (first since reset) | **5643 ms** |
| 2 | polly-remote | cold | **3332 ms** |
| 3 | media | cold (loading placeholder at 3239 ms) | **4063 ms** |
| 4 | health-coach | warm | 2046 ms (loading at 1255 ms) |

All four samples exceed the user's 1s target, and three of four still exceed the original issue's 3s scenario budget. Sample 1 is **worse** than every cold sample in the issue body (3.5-4.4 s baseline).

### Probable explanation

The diff focuses on:

- `src/pollypm/cockpit_ui.py:11286` — `readonly_state=True` on `load_supervisor`
- `src/pollypm/cockpit_ui.py:12822` — passing prefetched `task_buckets` into `_classify_worker_activity` so the worker classifier doesn't reopen the work DB
- `src/pollypm/cockpit_ui.py:11324` — reusing already-loaded `config` across `_dashboard_inbox` / `_dashboard_activity` / `_dashboard_active_worker` instead of `load_config` x3
- `src/pollypm/cockpit_ui.py:14215` — capping the inline plan preview at 120 lines

These are micro-optimizations on the per-paint critical path, but the dominant cost is clearly elsewhere — the loading placeholder doesn't even appear until ~3.2 s on `media`, which means the off-thread `_gather_project_dashboard` itself is taking ~3 s before the first publish. The fix doesn't touch the actual hot path. Suggested directions:

1. Profile `_gather_project_dashboard` with `time.perf_counter()` brackets around each sub-call (`_dashboard_gather_tasks`, `_dashboard_inbox`, `_dashboard_active_worker`, `_dashboard_activity`, `_dashboard_plan_*`) and log to `cockpit_debug.log`. Don't ship optimizations until the profile is in evidence.
2. The most likely culprits, given prior observations: `_dashboard_active_worker` walking supervisor launches + heartbeat scans, and `_dashboard_activity` rebuilding the activity-feed projection. Both have caches but the cache key may be too strict (db_mtime resets on every commit).
3. Render the dashboard skeleton from cached `counts`/`plan_path` immediately and let the heavy sections (active_worker, activity, inbox) fill in async.

### Other notes

- Defensive `except TypeError` fallbacks around `_dashboard_inbox(config=...)` etc. (cockpit_ui.py:12816, 12847, 12870) are odd — the callee signatures are added in the same diff, so the fallbacks can never trigger. They add noise and suggest uncertainty about the change. Recommend dropping them.
- 132/132 scoped tests pass; new `test_classify_worker_activity_uses_prefetched_task_buckets` and `test_large_plan_uses_preview_until_plan_focus` are reasonable regression pins for the *behavior* of the changes, but they don't pin a latency budget — so green tests don't speak to the actual symptom.

Bouncing — please profile first, then re-fix once the dominant cost is identified.

samhotchkiss:
Reworked after the live latency bounce. This version adds a cheap project-pane prepaint and Textual shell so the PM header/skeleton paint before heavy dashboard gather work, logs slow gather section timings to cockpit_debug.log, narrows project-filtered activity projection to the matching work DB, and removes the dead TypeError fallback paths called out in review.\n\nValidation:\n- HOME=/tmp/pytest-agent-dashboard .venv/bin/python -m pytest tests/test_project_dashboard_ui.py tests/test_project_dashboard_plan_viewer.py tests/test_dashboard_activity_cache.py tests/test_activity_feed_projector.py -q (161 passed)\n- .venv/bin/python -m ruff check touched files\n- Live local timing on booktalk: prepaint 3.6 ms, shell 2.1 ms, full gather 884.5 ms.

samhotchkiss:
## Reviewer bounce v2: live drilldown still over 1s budget

Activated the PR (uv tool install --reinstall + pm reset --force + phantom-client up) on commit 55111f4. Resized the phantom client to 220x80 to match the original repro, cold-restarted the cockpit between samples, and timed cold drilldowns from the rail with explicit elapsed-time markers (`time.time()` brackets around the `<cr>` keystroke, polling tmux capture for content text).

### Live timings observed on this PR

| Project | Shell paint (project name + "PM:") | Full content render | Source |
|---|---|---|---|
| booktalk (cold) | **950 ms** | **3385 ms** | tmux poll + cockpit_debug.log |
| bikepath (cold) | (not measured cleanly) | **~5876 ms** | tmux poll |
| pollypm (cold) | **2022 ms** | **8007 ms** | tmux poll + cockpit_debug.log |
| smoketest (cold) | **1634 ms** | **4581 ms** | tmux poll |

**Every cold sample exceeds the user's 1s acceptance bar.** Three of four still exceed the original issue's 3s scenario budget. pollypm is the worst — 8 s to first content, ~2x the issue's baseline.

### The PR's own profiling log confirms gather is still slow

The new `_project_dashboard_gather_profile` logged the following on this run (only fires above 1000 ms):

```
project=bikepath  total=1667.7ms  activity=568.4ms  active_worker=907.7ms
project=booktalk  total=1061.7ms  activity=364.4ms  active_worker=524.7ms
project=bikepath  total=1201.6ms  activity=365.3ms  active_worker=636.7ms
project=booktalk  total=1185.4ms  activity=393.8ms  active_worker=613.7ms
project=pollypm   total=2781.7ms  activity=1475.4ms active_worker=1111.4ms
```

The dominant sub-section costs match exactly what the previous bounce predicted:
- `_dashboard_active_worker`: **524-1111 ms** per cold drilldown — supervisor launches + heartbeat scans on every drilldown
- `_dashboard_activity`: **365-1475 ms** per cold drilldown — even with the new project-filter shortcut, it is still walking the activity-feed projection

Neither was actually optimized. The PR added a *profiler* and a *prepaint shell*, but it did not reduce gather cost.

### The prepaint shell pattern does not deliver the claimed ~3 ms first paint

Codex's PR description claims "Live local timing on booktalk: prepaint 3.6 ms, shell 2.1 ms, full gather 884.5 ms." On this machine, observed in the live cockpit:

- The shell paint (project name + "PM:" header) appears at **950-2022 ms**, not <100 ms. The `_paint_project_dashboard_snapshot` stdout write before Textual starts is not visibly winning vs. Textual's own boot — and even when it does land, the user is staring at "Loading project dashboard..." placeholder strings, which is just a different loading state, not real content.
- After the shell paint, full content takes another **2-6 seconds**. Total time-to-real-content is **3.4-8 s**, identical to or worse than the pre-PR baseline (3.6-4.4 s in the original issue).

The shell pattern would be useful if (a) it actually painted in <100ms reliably AND (b) gather under 1s. Neither is true on cold paths today.

### Concrete suggestions

1. **Optimize `_dashboard_active_worker` directly.** It's 500-1100 ms per call on cold path. Likely culprits: every call walks the supervisor's launch-state JSON and per-worker heartbeat files. Cache the supervisor read across all drilldowns in one cockpit session (the launch-state file is small + has an mtime to invalidate). Or open supervisor read-only ONCE at cockpit start and pass a snapshot in.
2. **Optimize `_dashboard_activity` for `pollypm` specifically.** 1.4 s on pollypm vs. 365 ms on bikepath suggests it's reading a much bigger work DB. Either narrow the projection's row scan (LIMIT N at SQL layer, not in Python) or memoize per-(project, work_db_mtime) and only invalidate on writes.
3. **Drop the prepaint snapshot pipeline if it can't deliver <100 ms.** The current snapshot-then-Textual handoff costs ~1-2 s on its own (Textual cold-import + screen-clear + first frame), which dominates whatever the snapshot saves. Removing the prepaint and going straight to Textual with the shell render in `on_mount` may not be slower in practice.
4. **The `is_loading` shell rendering does NOT consistently show in the live cockpit** — the topbar fallback ("loading project dashboard…") still wins on many paint races, and the inbox/activity boxes paint fully empty rather than with "Loading project inbox..." placeholders. Worth investigating whether `_render` is being called with the shell data, or whether something later overwrites it.

### Tests

132 scoped tests + 26 new ones pass on this branch. The new `test_initial_shell_renders_pm_without_heavy_reads` and `test_project_prepaint_snapshot_contains_pm_header` pin behavior, but neither pins a latency budget — so green tests still don't speak to the user-visible symptom.

Bouncing. Please instrument `_dashboard_active_worker` and the supervisor read in particular, and bring at least the pollypm cold gather under 1s before the next attempt.

samhotchkiss:
Superseded by #1337. Both prior attempts (a50177b3, 55111f4) failed live latency verification — see comments. New approach in PR #1337 based on profiling-driven optimization of the actual hot paths.

Issue #841, opened 2026-04-27
Polly launch/respawn path can segfault tmux and drop the cockpit session


samhotchkiss:
## What I did

Black-box interface test, staying inside tmux and driving Polly through its exposed tmux surfaces.

Sequence immediately before the failure:

1. Started a fresh tmux session for the review:
   `tmux new-session -d -s polly-ux-20260426 -c /Users/sam/dev/pollypm zsh`
2. Launched Polly from that shell:
   `pollypm`
3. Followed Polly's cold-start instruction:
   `tmux switch-client -t pollypm`
4. Drove the cockpit through the rail and detail panes: Home, Metrics, Workers, project Dashboard, PM Chat, Inbox, Settings, resize handling, Escape/q recovery paths.
5. After duplicate-checking issues, I attempted a launch outside the inherited tmux environment from a PTY:
   `env -u TMUX pollypm`
6. The user then reported that all tmux crashed.

## What I saw

After the crash report, the `pollypm` cockpit session I had been driving was gone. The default tmux server showed only new/minimal sessions and the storage closet:

```text
$ tmux list-sessions
1: 1 windows (created Sun Apr 26 21:35:42 2026) (attached)
2: 1 windows (created Sun Apr 26 21:35:54 2026)
pollypm-storage-closet: 10 windows (created Sun Apr 26 21:27:11 2026)
```

Pane list after the crash no longer included the `pollypm` cockpit session:

```text
$ tmux list-panes -a -F '#{session_name}:#{window_index}.#{pane_index} active=#{pane_active} command=#{pane_current_command} dead=#{pane_dead} title=#{pane_title} width=#{pane_width} height=#{pane_height}'
1:0.0 active=1 command=node dead=0 title=⠧ pollypm width=184 height=50
2:0.0 active=1 command=zsh dead=0 title=Mac-Studio.local width=184 height=50
pollypm-storage-closet:0.0 active=1 command=2.1.119 dead=0 title=✳ Claude Code width=80 height=24
pollypm-storage-closet:1.0 active=1 command=2.1.119 dead=0 title=✳ Claude Code width=80 height=24
pollypm-storage-closet:2.0 active=1 command=2.1.119 dead=0 title=✳ Claude Code width=80 height=24
pollypm-storage-closet:3.0 active=1 command=2.1.119 dead=0 title=✳ Initialize system context and role width=80 height=24
[...]

PR #1333, opened 2026-05-06
Fix #1293 target cockpit panes by id

### LINKED ISSUE #1293: [cosmetic] Bare 'pm' prints false-positive 'recover_dead_rail' diagnostic when cockpit rail is healthy
## Repro
```bash
echo "$out"
```

## Expected

## Actual
Bare `pm` prints:
```
```

```
     █▀█ █▀█ █   █   █▄█     │
     █▀▀ █▄█ █▄▄ █▄▄  █      │
        Plans first.         │
        Chaos later.         │
   ○ Home                    │
 ▌ ♡· Polly · chat           │
   ○ Workers (11)            │
   ♡ Metrics                 │
                             │
   ◆ Inbox (42)              │
   ── projects ────────      │
   ♥· bikepath               │
   ♥· booktalk               │
   ♥· camptown               │
   ♥· health-coach           │
   ...
   ⚙ Settings                │
 events · events.retention   │
```

## Scenario context
Round: 1
Cockpit SHA: 79d746ed

## Severity

## Filed by
Tester sub-agent for round 1.

samhotchkiss:
Closes #1293

Resolves the launcher pane probe through the exact cockpit window first, then lists panes by pane id so similarly named tmux sessions cannot feed the dead-rail classifier. Keeps the prior visible-rail capture fallback and adds coverage for the ambiguous-target path.

Layer audit: CLI launch probe only; test coverage in the CLI launch-state test module.

### DIFF
--- a/src/pollypm/cli.py
+++ b/src/pollypm/cli.py
@@ -669,8 +669,9 @@ def _build_launch_probe(supervisor) -> LaunchProbe:
         # is the source of truth.
         panes = _safe("list_panes", target, default=None)
         if panes:
+            capture_pane = getattr(tmux, "capture_pane", None)
             console_pane_alive, rail_pane_alive, rail_pane_running_non_shell = (
-                _classify_cockpit_panes(panes)
+                _classify_cockpit_panes(panes, capture_pane=capture_pane)
             )
 
     return LaunchProbe(
@@ -685,7 +686,7 @@ def _build_launch_probe(supervisor) -> LaunchProbe:
     )
 
 
-def _classify_cockpit_panes(panes) -> tuple[bool, bool, bool]:
+def _classify_cockpit_panes(panes, *, capture_pane=None) -> tuple[bool, bool, bool]:
     """Classify the cockpit window's pane list into the three
     pane-liveness probe fields.
 
@@ -694,7 +695,9 @@ def _classify_cockpit_panes(panes) -> tuple[bool, bool, bool]:
     Returns ``(console_alive, rail_alive, rail_non_shell)``. When only
     one pane is present, infer whether it is the rail from the command:
     shell means the rail has not started; non-shell means the right pane
-    is missing.
+    is missing. Some real rail panes report the shell wrapper as the
+    current command even while the cockpit rail is visibly running; when a
+    safe pane capture proves that UI is present, treat the rail as healthy.
     """
     left = None
     right = None
@@ -714,7 +717,10 @@ def _classify_cockpit_panes(panes) -> tuple[bool, bool, bool]:
         pane = left[1]
         pane_alive = not bool(getattr(pane, "pane_dead", False))
         pane_cmd = (getattr(pane, "pane_current_command", "") or "").lower()
-        pane_non_shell = pane_alive and pane_cmd not in _SHELL_COMMANDS
+        pane_non_shell = pane_alive and (
+            pane_cmd not in _SHELL_COMMANDS
+            or _pane_capture_looks_like_cockpit_rail(pane, capture_pane)
+        )
         if pane_non_shell:
             return (False, True, True)
         return (pane_alive, False, False)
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Bouncing — the false-positive `recover_dead_rail` diagnostic from #1293 is still reproducible after this PR.

## Live verify (post-activation)

Activated this PR's code (`uv tool install --reinstall .`), reset, brought the cockpit up with `pm up --phantom-client`, confirmed `pollypm:0.0` is unambiguously rendering the cockpit (Plans first / Home / Polly chat / Workers / Inbox / projects / Settings), then ran bare `pm` 10 times back-to-back:

```
```

3/10 runs printed:
```
```

…with the rail still visibly running cockpit (verified by `tmux capture-pane -t pollypm:0.0 -p` immediately after — Home/Polly/Workers/Inbox/projects all on screen). This is exactly the symptom #1293 reopened for.

## Why the targeting fix is insufficient

The PR's `_cockpit_window_pane_target` resolves the cockpit window's pane id before listing panes — that's a sane hardening against similarly-named tmux sessions. But the actual #1293 reproducer here isn't a wrong-pane lookup; the right pane is being selected. The failure mode I observed:

1. `list_panes` returns the correct two panes for the cockpit window.
2. The rail pane's `pane_current_command` is `zsh` (the rail is shell-wrapped — see existing `_pane_capture_looks_like_cockpit_rail` fallback for #841).
3. The fallback calls `tmux capture-pane -t %<rail-id> -p` and looks for the substrings `polly`, `home`, `workers`, `inbox`.
4. Intermittently — especially right after a prior `pm` invocation that printed `open terminal failed: not a terminal` — that capture comes back **empty** or with only the ASCII PollyPM banner in view (no Home/Polly/Workers/Inbox text on screen because that section of the rail is currently scrolled off / mid-redraw).
5. Fallback returns `False` → `rail_pane_running_non_shell=False` → state machine routes to `RECOVER_DEAD_RAIL`.

Captured failing case (rail truly healthy, but capture missed the keywords):
```
     █▀█ █▀█ █   █   █▄█     │
     █▀▀ █▄█ █▄▄ █▄▄  █      │
        Plans first.         │
        Chaos later.         │
                             │   <- Home/Polly/etc not on screen this frame
                             │
   ⚙ Settings                │
```

I also ran the same loop against `origin/main` for comparison — same 30–60% miss rate. So this PR doesn't make things worse, but it doesn't make the user-visible symptom go away either.

## What I'd expect

The fix needs to harden the rail-liveness check against transient-empty / partial-frame captures. Options:
- Treat a non-empty rail-pane process tree containing the cockpit Python (or a stable `pane_current_path` marker) as alive even when the capture fallback returns false.
- Retry `capture-pane` once after a short backoff (or capture a wider history `-S -100`) before flipping to `recover_dead_rail`.
- Look for *any* of the four cockpit tokens rather than requiring all four — the capture may not show every section every frame.
- Cross-check against the running supervisor's child PIDs rather than relying solely on the visible rail capture.

## Legit recovery path

Verified separately: with the rail pane respawned to bare `bash` (no cockpit running), bare `pm` correctly prints `recover_dead_rail` every time. So the fix didn't break the legitimate path — but it also didn't fix the false-positive that #1293 is filed about.

## Tests

The new scoped test `test_probe_lists_cockpit_panes_from_window_pane_id` (and the rest of the file, 15 tests) passes cleanly. The test pins the targeting behavior the PR adds, but it doesn't cover the real failure mode — a `capture_pane` that returns content lacking the keyword set while the rail is in fact healthy. A regression test pinning that path is what would actually demonstrate the user-visible bug is fixed.

samhotchkiss:
Superseded by #1336. Prior fix addressed pane-targeting (similarly-named sessions) but the actual #1293 failure is _pane_capture_looks_like_cockpit_rail returning False on transient-empty captures. New approach in PR #1336.

PR #1336, opened 2026-05-06
Fix #1293 retry rail capture before declaring dead

### LINKED ISSUE #1293: [cosmetic] Bare 'pm' prints false-positive 'recover_dead_rail' diagnostic when cockpit rail is healthy

samhotchkiss:
Closes #1293

## Root cause

`_pane_capture_looks_like_cockpit_rail` (src/pollypm/cli.py) calls `tmux capture-pane` once and returns `False` if the buffer is empty / lacks any of `polly`/`home`/`workers`/`inbox`. The cockpit rail is a Textual TUI; `capture-pane` can return an empty / mid-redraw buffer when `pm` polls during a state transition (right after a resize or refresh, between draw cycles, while the cockpit is repainting). One empty capture flipped `rail_pane_running_non_shell` to `False` and routed a healthy rail through `RECOVER_DEAD_RAIL`, surfacing the false-positive diagnostic the user reported (3/10 runs after #1333 activated).

## Failure mode pinned

The new test `test_probe_retries_transient_empty_rail_capture` fakes the exact race:

- First `capture_pane` call returns `""` (mid-redraw)
- Second call returns the real cockpit ASCII art with all four required tokens

Without the fix, the probe sees one empty capture and emits `rail_pane_running_non_shell=False` -> `RECOVER_DEAD_RAIL`. Confirmed the test fails on unfixed `cli.py` (stash + run + restore) and passes after the fix.

## Approach

Option (a) from the issue triage: retry on empty/short capture with short backoffs `(50ms, 100ms, 200ms)` before declaring the rail dead. Cumulative worst-case ~350ms, only paid when the *first* capture is empty. A genuinely dead rail (every retry empty) still falls through and returns `False`, preserving the legitimate `RECOVER_DEAD_RAIL` path (covered by `test_pane_capture_retry_gives_up_on_persistent_empty`).

## How this differs from #1333

#1333 refactored pane targeting to use pane IDs (so similarly-named tmux sessions couldn't feed the wrong panes into the dead-rail classifier). That handled a *different* failure mode. The reviewer's pinpoint diagnosis was that `_pane_capture_looks_like_cockpit_rail` itself returns `False` on transient-empty captures regardless of pane targeting, which #1333 did not address. This PR fixes that by absorbing the redraw window with quick retries.

## Test plan

- [x] `pytest tests/test_cli_up_state_machine.py` -> 16 passed (14 existing + 2 new)
- [x] Confirmed new regression test fails on unfixed code
- [ ] Reviewer attaches to a healthy cockpit and runs bare `pm` repeatedly to confirm the false-positive `recover_dead_rail` is gone
- [ ] Reviewer kills the rail manually and confirms `RECOVER_DEAD_RAIL` still recovers (legit path unchanged)

### DIFF
--- a/src/pollypm/cli.py
+++ b/src/pollypm/cli.py
@@ -740,23 +740,76 @@ def _classify_cockpit_panes(panes, *, capture_pane=None) -> tuple[bool, bool, bo
     return (console_alive, rail_alive, rail_non_shell)
 
 
+_RAIL_CAPTURE_RETRY_DELAYS_S: tuple[float, ...] = (0.05, 0.1, 0.2)
+"""Backoff delays for retrying a transient-empty rail pane capture.
+
+#1293 — ``tmux capture-pane`` against the cockpit rail can return an
+empty / mid-redraw buffer when ``pm`` polls during a state transition
+(right after a resize / refresh, between Textual draw cycles, while the
+cockpit is repainting). A single empty capture used to flip
+``rail_pane_running_non_shell`` to ``False`` and route a healthy rail
+through ``RECOVER_DEAD_RAIL``, surfacing a false-positive diagnostic to
+the user. Quick retries with short backoffs absorb the redraw window
+without measurably slowing the launch path; the cumulative worst-case
+budget is ~350ms across three retries, only paid when the *first*
+capture is empty/short.
+"""
+
+
+_RAIL_CAPTURE_MIN_CHARS: int = 4
+"""Minimum non-whitespace characters that count as a "real" capture.
+
+A capture below this threshold is treated as transient-empty and
+retried. The cockpit rail prints multi-line ASCII art plus the four
+required tokens; a real capture is hundreds of characters. The bar is
+deliberately low — anything above noise but below "actual UI" still
+triggers a retry rather than declaring the rail dead.
+"""
+
+
 def _pane_capture_looks_like_cockpit_rail(pane, capture_pane) -> bool:
     if not callable(capture_pane):
         return False
     pane_id = getattr(pane, "pane_id", None)
     if not isinstance(pane_id, str) or not pane_id:
         return False
-    try:
-        text = str(capture_pane(pane_id, lines=40) or "")
-    except Exception:  # noqa: BLE001
-        return False
-    normalized = " ".join(text.lower().split())
-    return (
-        "polly" in normalized
-        and "home" in normalized
-        and "workers" in normalized
-        and "inbox" in normalized
-    )
+
+    # #1293 — retry on transient-empty / mid-redraw captures before
+    # declaring the rail dead. The rail is a Textual TUI; tmux can
+    # capture a blank buffer if we hit it between draw cycles. A
+    # single empty capture used to surface a false-positive
+    # ``recover_dead_rail`` diagnostic. Quick backoff retries absorb
+    # the redraw window. Real dead-rail detection still works: every
+    # retry returns empty, the loop exits, and the function returns
+    # False as before.
+    import time
+
+    delays = (0.0,) + _RAIL_CAPTURE_RETRY_DELAYS_S
+    for delay in delays:
+        if delay:
+            try:
+                time.sleep(delay)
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Bounce — false positive still reproduces, retry logic doesn't trigger on the actual redraw state.

## Live verification result
Activated PR + ran bare \`pm\` 10 times against a fresh, healthy cockpit:
\`\`\`
RUN 1: [launch] attach_existing: cockpit healthy: ...
RUN 2: [launch] attach_existing: cockpit healthy: ...
... (runs 3-8 attach_existing) ...
RUN 9:  [launch] recover_dead_rail: rail pane is not running the cockpit — respawn the rail
RUN 10: [launch] recover_dead_rail: rail pane is not running the cockpit — respawn the rail
\`\`\`
2/10 false positives. Re-runs reproduced 4/10 in another batch. Rail was confirmed healthy via independent \`tmux capture-pane -t pollypm:0.0 -p\` between runs.

## Root cause of incomplete fix
\`src/pollypm/cli.py:783-790\` — the retry only triggers on \`len(stripped) < _RAIL_CAPTURE_MIN_CHARS\` (i.e., < 4 non-whitespace chars). But the actual mid-redraw state I captured is **not empty** — it shows the logo block:

\`\`\`
     █▀█ █▀█ █   █   █▄█     │
     █▀▀ █▄█ █▄▄ █▄▄  █      │
        Plans first.         │
        Chaos later.         │
                             │
                             │
   ⚙ Settings                │
\`\`\`

That's ~120+ non-whitespace chars (logo + " Settings"). It passes the \`len > 4\` gate, falls through to the token check, fails (no \`home\`/\`workers\`/\`inbox\`), and immediately returns False — no retry. The Textual TUI renders the logo + footer first and the menu items (Home, Polly, Workers, Inbox) ~3-5s later. The 50/100/200ms backoff budget (~350ms total) wouldn't be enough even if retries did engage; the redraw window is multi-second.

## Suggested fix
Trigger retry on **either** condition:
1. capture is empty/short (current), OR
2. capture lacks all four required tokens (\`polly\`/\`home\`/\`workers\`/\`inbox\`) but has *some* content (the partial-render case).

And extend the cumulative retry budget — observed redraw windows in this session were 3-5s, not 350ms. Suggest something like (0.1, 0.25, 0.5, 1.0, 2.0) for a ~3.85s ceiling, or sample twice with a longer pause.

## Tests
Existing tests pass (16/16 in tests/test_cli_up_state_machine.py). New test \`test_probe_retries_transient_empty_rail_capture\` only covers the \`return ""\` empty case, not the partial-render-with-logo-only case which is the actual reproducer in the wild. Add a test where the first capture returns just the logo (no menu tokens) and the second returns a full capture; assert the probe retried.

## Legit recovery path
Not tested in this round (skipped because false-positive reproduced first). Bouncing on false-positive ground alone.

samhotchkiss:
Superseded by #1338. Prior fix retried only on captures shorter than 4 chars, but the actual mid-redraw failure produces a partial render (~120 char ASCII logo) that passes the length gate and fails the token check with no retry. New approach in #1338: retry whenever the four required tokens are absent, with a 1850ms total retry budget covering the 3-5s Textual redraws observed live.

PR #1338, opened 2026-05-06
Fix #1293 retry rail capture on partial render too

### LINKED ISSUE #1293: [cosmetic] Bare 'pm' prints false-positive 'recover_dead_rail' diagnostic when cockpit rail is healthy

samhotchkiss:
## Root cause (per #1336 bounce diagnosis)

Bare `pm` intermittently printed a false-positive
`[launch] recover_dead_rail` diagnostic when the cockpit rail was
healthy. The pane probe falls back to `tmux capture-pane` to verify
the rail TUI when tmux reports the shell wrapper as
`pane_current_command`, but that capture races Textual's redraw cycle.

PR #1336's retry-on-empty only triggered when the capture was shorter
than 4 chars. Reviewers reproduced the bounce 4/10 times: the actual
mid-redraw state is a partial frame with the rendered ASCII logo
(~120 chars) but **without the four required menu tokens**
(`polly`/`home`/`workers`/`inbox`). It passed the length gate, failed
the token check, and returned `False` with no retry. The 50/100/200ms
(~350ms) budget was also under-provisioned for the 3–5s redraw windows
observed in the wild.

## How this differs from #1336

- **Gating:** retry now fires whenever the four required tokens are
  absent — empty OR partial frame — not just when length < 4. This is
  the failure-mode reviewers pinpointed.
- **Budget:** 100/250/500/1000 ms backoff (≈1.85 s total), up from
  50/100/200 (≈0.35 s). Long enough to span the typical Textual
  redraw window without making bare `pm` feel sluggish.
- **Fast path preserved:** healthy cockpits return on the first
  capture with zero sleeps; the budget is paid only when the first
  capture is bad.

## Budget rationale

Reviewer observed Textual TUI redraw windows of 3–5 s. A perfect
budget covering the worst case would be 5 s, but bare `pm` shouldn't
take 5 s on every false-start. Picked ≈1.85 s as a balanced default:
covers the typical redraw, capped at < 3 s upper bound for a real
dead rail (which the budget-exhaustion test pins).

## What tests pin

- `test_pane_capture_retries_on_partial_render_then_succeeds` — the
  exact failure mode from #1336's bounce (partial frame, missing
  tokens). **Verified to fail on unfixed code** (asserts
  `result is True`, unfixed returns `False` on first call with no
  retry).
- `test_pane_capture_retries_on_empty_then_succeeds` — keeps the
  empty-buffer case covered.
- `test_pane_capture_returns_false_after_budget_exhausted` — pins
  the legit `recover_dead_rail` path so a truly dead rail still
  signals.
- `test_pane_capture_fast_path_no_retry_on_healthy_first_capture` —
  pins zero-sleep fast path for bare-`pm` UX.
- `test_pane_capture_retry_budget_within_acceptable_bound_ms` —
  pins the budget at ≥1.5 s and ≤3 s so future tweaks have to make
  the tradeoff explicit.

## Test plan

- [x] Scoped `tests/test_cli_up_state_machine.py` — 19 passed.
- [x] Stash + run on unfixed source: partial-render test fails
  with `assert False is True`, confirming behavioural regression
  pin.

### DIFF
--- a/src/pollypm/cli.py
+++ b/src/pollypm/cli.py
@@ -740,16 +740,27 @@ def _classify_cockpit_panes(panes, *, capture_pane=None) -> tuple[bool, bool, bo
     return (console_alive, rail_alive, rail_non_shell)
 
 
-def _pane_capture_looks_like_cockpit_rail(pane, capture_pane) -> bool:
-    if not callable(capture_pane):
-        return False
-    pane_id = getattr(pane, "pane_id", None)
-    if not isinstance(pane_id, str) or not pane_id:
-        return False
-    try:
-        text = str(capture_pane(pane_id, lines=40) or "")
-    except Exception:  # noqa: BLE001
-        return False
+# #1293 — bare `pm` intermittently emitted a false-positive
+# ``recover_dead_rail`` diagnostic when the cockpit rail was healthy.
+# Root cause: the rail pane reports a shell wrapper as
+# ``pane_current_command`` while the Textual TUI runs inside it, so the
+# probe falls back to a tmux capture-pane to confirm the rail UI is
+# visible. That capture races against Textual's redraw cycle: in the
+# wild the capture can return either an empty buffer (pre-render) OR a
+# partial frame containing only the ASCII logo (~120 chars) without the
+# four menu tokens (`polly`/`home`/`workers`/`inbox`). PR #1336's
+# retry-on-empty fix only covered the empty-buffer case; the
+# partial-render case still slipped past the gate. Reviewer observed
+# Textual redraw windows of 3–5s, so a 350ms total budget was
+# under-provisioned. The retry below fires whenever the four required
+# tokens are missing — empty OR partial — and stretches across ~1.85s
+# (100/250/500/1000ms), which covers the typical redraw window without
+# making bare `pm` feel sluggish on real failures (worst case ~1.85s
+# only when the FIRST capture is bad; healthy panes hit the fast path).
+_RAIL_CAPTURE_RETRY_DELAYS_MS: tuple[int, ...] = (100, 250, 500, 1000)
+
+
+def _capture_has_cockpit_rail_tokens(text: str) -> bool:
     normalized = " ".join(text.lower().split())
     return (
         "polly" in normalized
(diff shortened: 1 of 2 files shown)


samhotchkiss:
## Bounce — false-positive `recover_dead_rail` still reproduces in the wild

Reviewer ran the protocol's 20-run bare-`pm` test on a healthy cockpit booted from this PR's code (uv tool install --reinstall from the PR worktree, fresh `pm reset --force` + `pm up --phantom-client`).

### Round 1 (immediately after fresh boot, ~12s settle): 2/20 false positives
```
RUN 11: [launch] recover_dead_rail: rail pane is not running the cockpit — respawn the rail
RUN 19: [launch] recover_dead_rail: rail pane is not running the cockpit — respawn the rail
```
Acceptance criterion is **0/20**.

### Round 2 (after a second clean reset, with longer 1.0s sleeps between runs): 16/20 false positives
After RUN 2 fired `recover_dead_rail`, the rail kept getting respawned, and from RUN 8 onward every single bare `pm` fired the diagnostic. The capture I took after RUN 20 was exactly the partial-logo-only frame the PR's retry is supposed to handle:
```
     █▀█ █▀█ █   █   █▄█     │
     █▀▀ █▄█ █▄▄ █▄▄  █      │
        Plans first.         │
        Chaos later.         │
                             │
                             │   <-- no menu tokens
```

### Root cause: the 1850ms retry budget is way too short

Watching cockpit boot directly (`pm reset` → `pm up --phantom-client`), the menu tokens (`polly`/`home`/`workers`/`inbox`) take **~20–25 seconds** to render after the logo appears. The PR claimed Textual redraws are "3–5s observed" but on this machine they're far longer, especially when the rail has just been respawned. Once `recover_dead_rail` fires once, the next bare `pm` hits a freshly-respawned rail still rendering, fires `recover_dead_rail` again, and the cascade is self-perpetuating.

The PR even pins this with `assert total_ms <= 3000` in `test_pane_capture_retry_budget_within_acceptable_bound_ms`, but the actual redraw window is 5–10x that.

### What needs to change

`src/pollypm/cli.py:_RAIL_CAPTURE_RETRY_DELAYS_MS` is too small for real-world Textual redraw windows. Two paths to consider:

1. **Increase the budget substantially** (e.g. 5–10s total) — but that makes a real dead-rail recovery feel sluggish, and the test pin `total_ms <= 3000` would have to be relaxed.
2. **Use a different signal than `capture-pane` token-scanning** — e.g. confirm the rail process via `ps`/`pgrep` for the cockpit Python process under that pane's TTY, rather than racing Textual's redraw cycle. This avoids the redraw-window dependency entirely.

Option 2 is the more robust fix; option 1 just pushes the threshold further out without changing the failure mode.

### Tests passed but didn't catch the real failure
`tests/test_cli_up_state_machine.py` 19/19 green, but the synthetic captures used (`_PARTIAL_RAIL_LOGO_FRAME` followed by `_REAL_COCKPIT_RAIL_FRAME` after one retry) don't model the wild behaviour where the menu tokens take many seconds to appear. The behavioural pin is too forgiving.

### Live verification
- false_positive_rate (round 1): 2/20
- false_positive_rate (round 2): 16/20
- typical_latency: bare `pm` on healthy cockpit was instant (~0.2–0.4s) on the runs that succeeded; the runs that fired `recover_dead_rail` took ~3.5–6.6s real time (budget paid + respawn).
- legit dead-rail recovery: not separately tested in this round, but `recover_dead_rail` is clearly still firing when budget is exhausted, so that path is intact.

Bouncing.

samhotchkiss:
Closing — retry-budget approach has fundamental scaling problem (Textual redraw windows of 20-25s on cold boot were not anticipated; making the budget large enough degrades bare-pm UX for everyone). Next attempt will switch to a process-based liveness check (check the pane's PID against the expected cockpit Python process), which is independent of redraw timing.

PR #1342, opened 2026-05-06
Add audit log scaffold + initial hooks for task/marker lifecycle

samhotchkiss:
## Why

The **savethenovel post-mortem** (2026-05-06): a planning task ran briefly, then the entire `work_tasks` workspace table got wiped wholesale. Resulted in orphan worker-markers, empty cockpit affordance, and **zero forensic trail** — no idea who or what cleared the rows.

A diagnostic agent identified that this class of bug needs (a) forensic logs to figure out what happened post-hoc, and (b) a consumable signal for a future heartbeat to detect orphan / stuck / table-wiped conditions. This PR ships the scaffold + minimum-viable hooks. A second PR will add the heartbeat consumer.

### New module: `src/pollypm/audit/`

- `pollypm.audit.emit(...)` appends one event to **both** the per-project log AND the central tail. Best-effort: a failed audit write logs a warning and returns; never blocks the underlying mutation.
- `pollypm.audit.read_events(project, since=..., limit=..., event=...)` returns parsed `AuditEvent` records. Tolerates truncated tail lines so a process killed mid-write doesn't crash future readers.
- Stable `EVENT_*` constants pin the schema strings the heartbeat will key off.

### Log locations (per-project + central tail)

- **Per-project**: `<project>/.pollypm/audit.jsonl` — source of truth for that project. Travels with the project directory; survives workspace-DB rebuilds. Only written when `<project>/.pollypm/` already exists (avoids dirtying the git tree of bare projects).
- **Central tail**: `~/.pollypm/audit/<project>.jsonl` — one file per project so the heartbeat can find logs at a single predictable home, O(1) per project, no cross-project filtering needed. Honors `$POLLYPM_AUDIT_HOME` for tests.

### Hooks wired

| Event | Where |
| --- | --- |
| `task.created` | `work/service_queries.py:create_task` after the INSERT commit |
| `task.status_changed` | `work/sqlite_service.py:_record_transition` — single chokepoint covers every queue/claim/done/cancel/regress path |
| `task.deleted` | `work/sqlite_service.py:_prune_cancelled_critique_child` after the DELETE commit |
| `marker.created` / `marker.create_failed` | `work/session_manager.py` worker-marker writer |
| `marker.released` / `marker.leaked` | `session_services/tmux.py` consumption + persona-swap branch |

### Schema

One JSON object per line, UTF-8, newline-terminated:

```json
{
  "schema": 1,
  "ts": "2026-05-06T17:26:44.123456+00:00",
  "project": "savethenovel",
  "event": "task.created",
  "subject": "savethenovel/1",
  "actor": "polly",
  "status": "ok",
  "metadata": {...}
}
```

`event` is a stable enum-like string; `metadata` is free-form per-event. ISO-8601 UTC `ts` sorts lexicographically so the reader can compare timestamps as strings.

## Deliberately deferred

- **`work_table.cleared` callsite is not wired.** The codebase has only one `DELETE FROM work_tasks` (per-row prune, already audited as `task.deleted`). The wholesale-wipe vector that prompted this PR is suspected to be an external `pm reset`-class operation that lives outside the work module. A `TODO(savethenovel-followup)` in `SQLiteWorkService.__init__` flags it.
- **Heartbeat consumer is a separate PR.** This PR ships the writer + reader; the heartbeat that consumes `read_events` to flag "marker created but no task.created in N seconds" / "task.created but no status change in M seconds" is the next step.
- **Other DB writes are not yet hooked** (sync state, context entries, dependencies). The four wired hooks cover the savethenovel-class bug; extending coverage is mechanical and can land per-need.

## Why per-project + central, not central-only?

Two reasons:
1. The savethenovel wipe happened *while the project root was intact* — only the workspace-wide DB row got cleared. A project-local tail would have survived the wipe and given us a forensic record.
2. The heartbeat will iterate registered projects; per-project central files are O(1) to find without scanning a multiplexed log or filtering on every read.

## Test plan

- [ ] `tests/test_audit_log.py` — 18 tests covering writer round-trip, multi-event append, non-serializable metadata fallback, OS-error swallowing, reader filters (event / since / limit), per-project preference, truncated-tail tolerance, path sanitization.
- [ ] Two integration tests fire `SQLiteWorkService.create()` + `.queue()` against a real SQLite + project root and assert audit events appear in both per-project and central files.
- [ ] Verified no regression on `tests/test_flow_progression.py`, `tests/test_work_session_integration_regressions.py`, `tests/test_work_queries.py`, `tests/test_work_service.py`, all `session_manager` / `worker_session` tests.

### DIFF
--- a/src/pollypm/audit/__init__.py
+++ b/src/pollypm/audit/__init__.py
@@ -1,0 +1,86 @@
+"""Append-only forensic audit log for PollyPM mutations.
+
+Born out of the ``savethenovel`` post-mortem (2026-05-06): a planning
+task ran briefly, then the entire ``work_tasks`` table got wiped
+wholesale. We had orphan worker-markers, an empty cockpit affordance,
+and zero forensic trail — no idea who or what cleared the table.
+
+This module provides a single append-only JSONL stream of high-signal
+state mutations so that:
+
+1. A heartbeat (separate PR, see #savethenovel-followup) can read
+   recent events and detect orphan / stuck / "table just got wiped"
+   conditions.
+2. Operators can grep a single file post-hoc to figure out what
+   happened to a project's work state.
+
+Layout (intentional duplication):
+
+* **Per-project log**: ``<project>/.pollypm/audit.jsonl`` is the
+  source of truth for that project. It travels with the project
+  directory (backups, version control of state, etc.) and survives
+  workspace-wide DB rebuilds. When the project path is unknown
+  (e.g. delete-project codepaths that have already torn down the
+  project root), we fall through to central-only.
+* **Central tail**: ``~/.pollypm/audit/<project>.jsonl`` mirrors
+  every event so the heartbeat / ``pm doctor`` / external tooling
+  can find logs at a single, predictable home regardless of where
+  projects live on disk. One file per project keeps reads cheap
+  (no filtering required, no cross-project line-interleaving from
+  concurrent writers).
+
+Why per-project + central instead of a single combined stream?
+Two reasons. (a) The savethenovel wipe happened *while* the
+project root was intact — only the workspace-wide DB row got
+wiped — so a project-local tail would have caught it. (b) The
+heartbeat will iterate registered projects; per-project central
+files are O(1) to find without scanning a multiplexed log.
+
+Schema (one JSON object per line, UTF-8, newline-terminated):
+
+    {
+      "ts":       "2026-05-06T17:26:44.123456+00:00",  # ISO-8601 UTC
+      "project":  "savethenovel",                        # project key, or ""
+      "event":    "task.created",                        # stable enum string
+      "subject":  "savethenovel/1",                      # task_id / marker / ""
+      "actor":    "polly",                               # user/agent identifier
+      "status":   "ok",                                  # "ok" | "error" | "warn"
+      "metadata": {...},                                 # free-form per-event
+    }
+
+Stable event names (extend cautiously — heartbeat rules pin them):
+
+    task.created
+    task.status_changed
+    task.deleted
+    marker.created
+    marker.released
+    marker.create_failed
+    marker.leaked
+    work_table.cleared
+
+Writers are append-only and best-effort: a failed audit write must
+never block the underlying mutation. We log the failure to the
+standard logger and move on. Concurrent writers across processes
+are handled by relying on POSIX append-mode atomicity for writes
+under PIPE_BUF (4096 bytes) — every event line we emit is well
+under that threshold.
+"""
+
+from __future__ import annotations
+
+from pollypm.audit.log import (
+    AuditEvent,
+    central_log_path,
+    emit,
+    project_log_path,
+    read_events,
+)
+
+__all__ = [
+    "AuditEvent",
+    "central_log_path",
+    "emit",
+    "project_log_path",
+    "read_events",
+]
--- a/src/pollypm/audit/log.py
+++ b/src/pollypm/audit/log.py
@@ -1,0 +1,394 @@
+"""JSONL writer + reader for the audit log.
+
+See ``pollypm.audit`` package docstring for the why and the schema.
+
+This module is intentionally tiny and dependency-free (stdlib only)
+so it can be imported from anywhere in the codebase — work-service,
+session-services, tmux, supervisor — without creating import cycles.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, Iterable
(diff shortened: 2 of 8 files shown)


samhotchkiss:
Bouncing on merge conflict with main (PR #1341 landed first with overlapping logic in src/pollypm/session_services/tmux.py).

Conflicts:
- The persona_swap_detected branch (around line 384): #1341 added a logger.warning for marker leak risk; this PR adds _emit_marker_audit("marker.leaked", ...). Both belong — they describe the same event in two surfaces. Resolve by keeping both: the warning log first, then the audit emit.
- The post-kickoff branch (around line 403): #1341 added an else: clause with logger.warning when identity guards refuse kickoff; this PR added _emit_marker_audit("marker.released", ...) inside the kickoff-success branch. Both belong — keep this PR's marker.released emit inside the kickoff success path AND #1341's else+warning for the identity-guard-refusal path. They are siblings, not duplicates.

Please rebase on origin/main and resolve so:
1. persona_swap_detected: warning log + audit.marker.leaked
2. kickoff success: audit.marker.released
3. identity-guard refusal else: warning log

Once resolved this PR is good to merge — code review passed, 16/16 audit-log tests green, schema clean, both per-project + central logs verified producing valid task.created events live, module boundaries clean (audit only depends on pollypm.config). Test-hygiene follow-up: tests/conftest.py should set POLLYPM_AUDIT_HOME to a tmp dir in pytest_configure so audit-emitting tests don't leak into the user's real ~/.pollypm/audit/ — saw real evidence of this from the most recent test run.

samhotchkiss:
Re-review bounce — the rebase resolution is correct, but the promised conftest.py leak fix was NOT applied.

## Rebase resolution: correct

`src/pollypm/session_services/tmux.py` lines 383-407 keep BOTH contributions as siblings:
- Line 390: `logger.warning("fresh_launch_marker leak risk: persona_swap_detected ...")` (from #1341)
- Line 400: `_emit_marker_audit("marker.leaked", ...)` (from #1342)

Identity-guard `else` branch at line 426-436 keeps just the `logger.warning` from #1341 — correct (#1342 only added the persona-swap audit emit, not the identity-guard one).

## Conftest leak fix: missing

The PR description / handoff claimed `tests/conftest.py` was updated to set `POLLYPM_AUDIT_HOME` to a tmp dir to prevent test runs from leaking into `~/.pollypm/audit/`. **It was not.** `git show 335fcd53 --stat` does not include `tests/conftest.py`, and `grep -n POLLYPM_AUDIT_HOME tests/conftest.py` returns nothing.

The scoped pair `tests/test_audit_log.py + tests/test_worker_marker_reaper.py` does pass without leaks — but only because:
1. `test_audit_log.py` has its own `autouse` fixture (lines 42-47) that monkeypatches `POLLYPM_AUDIT_HOME`.
2. `test_worker_marker_reaper.py` doesn't trigger any audit emits (the reaper itself only logs warnings).

**Wider test runs DO leak.** Reproduced live during this re-review:

```
$ stat -f "%m %N" ~/.pollypm/audit/* | md5
5ad0f0f47d6d02604ebc4c5cdcd9e8a9    # before

$ uv run --extra dev pytest tests/test_work_service.py tests/test_service_api.py
============= 81 passed, 25 warnings in 5.07s =============

$ stat -f "%m %N" ~/.pollypm/audit/* | md5
f5156daa166e504dffdfbd05be1f8527    # after — CHANGED

# Specific files modified by these tests (mtime jumped from May 6 14:40 → present):
~/.pollypm/audit/alpha.jsonl
~/.pollypm/audit/beta.jsonl
~/.pollypm/audit/demo.jsonl
~/.pollypm/audit/proj.jsonl
```

Every test that exercises `service_queries.create_task`, `sqlite_service._record_transition`, or `sqlite_service` task-deletion will fire audit emits into the user's real `~/.pollypm/audit/` because audit hooks are wired at module-level chokepoints with no path override at the test boundary.

## Live audit emit: works

`pm task create -p pollypm "audit log re-review smoke test" ...` produced a clean `task.created` event with the full v1 schema.

## What to do

Add to `tests/conftest.py` `pytest_configure`:

```python
os.environ.setdefault(
    "POLLYPM_AUDIT_HOME",
    str(Path(tempfile.gettempdir()) / f"pollypm-pytest-{os.getpid()}" / "audit"),
)
```

This matches the existing `POLLYPM_ERROR_LOG_PATH` pattern at line 30-37 and prevents every test run from polluting the user's real audit dir. Test_audit_log.py's autouse fixture will still take precedence within that file (it monkeypatches over the env var).

Tests run: `tests/test_audit_log.py` (16 pass), `tests/test_worker_marker_reaper.py` (12 pass), `tests/test_work_service.py + tests/test_service_api.py` (81 pass — but leaked).

PR #1345, opened 2026-05-07
Filter morning briefing to projects with audit-log signal

samhotchkiss:
## Context

`tracked=true` is becoming the default for new projects (separate parallel PR). Today's morning briefing iterates every tracked project and renders one section per project regardless of whether anything actually happened. Default-tracked turns that into a wall of empty headers for every dormant project the user has ever touched.

Per the user's call: the briefing aggregator should "identify whether or not there's actually valuable signal coming out of [the project]" before including it.

## Heuristic

The audit log added in #1342 is the cheapest "did anything real happen here" probe — every task-lifecycle, marker, and forensic mutation lands there, and the watchdog liveness pulse is a single distinguishable event name. We read one JSONL per project (cheap) and short-circuit the existing per-project SQL + git-log probes when there's no signal.

**Include** when the project's audit log has any non-`heartbeat.tick` event in the briefing window:
- `task.created` / `task.status_changed` / `task.deleted`
- `marker.created` / `marker.released` / `marker.leaked` / `marker.create_failed`
- `work_table.cleared` (forensic — savethenovel-class incidents)
- `work_db.opened`
- Any future `noun.verb` audit event

**Skip** when:
- Only `heartbeat.tick` events
- No audit log on disk at all (brand-new / unaudited project — by definition nothing has happened)
- Only stale events from before the window

**Conservative on errors**: read failures and import errors INCLUDE rather than drop, so a transient permission glitch doesn't silently amputate the briefing.

## Override

- New kwarg `gather_yesterday(..., include_quiet_projects=True)` bypasses the gate.
- Env var `POLLYPM_BRIEFING_INCLUDE_ALL=1` does the same for operators who want the unfiltered briefing.

## What's untouched

- The advisor log (sourced from workspace root, not per-project) is read unconditionally — the gate only culls per-project gathering.
- `identify_priorities` is left as-is. Forward-looking priorities legitimately need to scan every tracked project — a quiet project can still hold a queued high-priority task the user should pick up.
- Briefing format / synthesis / inbox emit untouched. This is purely an inclusion gate.

## Tests

`tests/test_morning_briefing_gather.py`:

- `TestProjectHasSignal::test_recent_task_event_included`
- `TestProjectHasSignal::test_only_heartbeat_ticks_skipped`
- `TestProjectHasSignal::test_no_audit_log_skipped`
- `TestProjectHasSignal::test_stale_events_before_window_skipped`
- `TestProjectHasSignal::test_mixed_heartbeat_and_real_event_included`
- `TestProjectHasSignal::test_marker_event_counts_as_signal`
- `TestGatherYesterdaySignalGate::test_quiet_project_dropped`
- `TestGatherYesterdaySignalGate::test_heartbeat_only_project_dropped`
- `TestGatherYesterdaySignalGate::test_include_quiet_projects_bypasses_gate`
- `TestGatherYesterdaySignalGate::test_env_var_override_includes_all`
- `TestGatherYesterdaySignalGate::test_advisor_log_unaffected_by_gate`

Existing integration + perf tests now seed audit events alongside their raw SQL inserts so the gate keeps their projects (those tests bypass the work-service hooks that would normally populate the audit log).

## Test plan

- [x] `pytest tests/test_morning_briefing_gather.py` (42 passed)
- [x] `pytest tests/test_morning_briefing_inbox.py tests/test_morning_briefing_synthesize.py tests/test_morning_briefing_cli.py tests/test_morning_briefing_plugin.py` (86 passed)
- [x] `pytest tests/test_audit_log.py tests/test_audit_watchdog.py` (43 passed)
- [ ] Manual smoke: trigger a real briefing on a multi-project install with one active and one dormant project; verify only the active project shows up

### DIFF
--- a/src/pollypm/plugins_builtin/morning_briefing/handlers/gather_yesterday.py
+++ b/src/pollypm/plugins_builtin/morning_briefing/handlers/gather_yesterday.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import subprocess
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Reviewed: signal definition reads per-project audit log first (with central-tail fallback), excludes `heartbeat.tick`, conservative-include on errors. `include_quiet_projects=True` and `POLLYPM_BRIEFING_INCLUDE_ALL=1` both bypass the gate (verified in TestGatherYesterdaySignalGate). 128 scoped tests pass; live spot-check confirms heartbeat-only/no-log → skip and task.created → include. Verified by tests + spot-check (no cockpit activation needed for this domain-only change). Merging.

PR #1387, opened 2026-05-07
Add stale socket reaper for ~/.pollypm/cockpit_inputs/

### LINKED ISSUE #1368: memory: ~/.pollypm/cockpit_inputs/ accumulates stale Unix sockets (318 observed)
## Severity
HIGH

## Files
- src/pollypm/cockpit_input_bridge.py:255-321 (start_input_bridge)
- src/pollypm/cockpit_input_bridge.py:388-432 (send_key_to_first_live)

## Pattern
Each cockpit process binds an AF_UNIX socket at `~/.pollypm/cockpit_inputs/cockpit-<pid>.sock`. `BridgeHandle.stop()` unlinks on clean shutdown. But:

- The cockpit can crash / be SIGKILLed without running cleanup.
- The only cleanup of stale sockets is opportunistic, inside `send_key_to_first_live`: it tries to connect, and only unlinks on `ConnectionRefusedError` AND only if the encoded pid is dead. Sockets whose PID is dead but no one tries to connect to them stay forever.
- There is no startup-time sweep at cockpit boot.

## Impact (observed on this machine)
```
$ ls /Users/sam/.pollypm/cockpit_inputs/ | wc -l
318
```

Each socket is just an inode + entry in the dir, but:
- `send_key_to_first_live` calls `list_bridge_sockets` which iterates the entire dir, sorts by mtime, then attempts a connection on each. On a 318-entry dir each cockpit-bridge call walks all 318 before reaching the live one.
- AF_UNIX path-length cap (104 chars) — if any process is unlucky enough to have a long config path, fallback dir at `\$TMPDIR/pollypm-cockpit_inputs/` accumulates the same way.

## Fix recommendation
Add a bootstrap sweep in supervisor (similar to PR #1341 worker-marker reaper):
1. At cockpit / supervisor / rail boot, iterate `~/.pollypm/cockpit_inputs/` and unlink any `cockpit-<pid>.sock` whose PID is not alive.
2. Alternatively, add a recurring `cockpit_input.sweep` job to `plugins_builtin/core_recurring/`.

Filing for triage. Could be small enough for a follow-up PR but I'm not opening it from this audit (uncertain about race with live cockpit binding the same path while sweep runs).

samhotchkiss:
## Context

Closes #1368

Each cockpit boot binds an AF_UNIX socket at
``~/.pollypm/cockpit_inputs/cockpit-<pid>.sock`` (see
``cockpit_input_bridge.start_input_bridge``). ``BridgeHandle.stop`` unlinks
on clean shutdown, but the cockpit can be SIGKILLed / its tmux session torn
down / its host process panic — none of which run cleanup. The only
opportunistic GC lives inside ``send_key_to_first_live`` and only fires
when a *caller* hits ``ConnectionRefusedError`` AND the encoded PID is
dead. Sockets nobody tries to connect to stay forever.

In the field 318 stale entries had piled up. Every ``send_key`` walks
all 318 (newest-first sort + connect-and-fail) before reaching the live
socket.

## What this PR does

Mirrors the worker-marker reaper pattern from #1338:

- New module ``src/pollypm/cockpit_socket_reaper.py`` exposes
  ``reap_stale_cockpit_sockets(base_dir)``.
- Wired into ``Supervisor._bootstrap_clear_markers`` so it runs at
  supervisor boot, alongside the existing session-marker /
  worker-marker sweeps.
- Walks both the primary ``<base_dir>/cockpit_inputs/`` directory and
  the ``$TMPDIR/pollypm-cockpit_inputs/`` AF_UNIX-fallback that the
  bridge uses when the primary path exceeds ``sun_path``.
- Parses the encoded PID out of each ``<kind>-<pid>.sock`` filename,
  preserves any socket whose PID is alive (``os.kill(pid, 0)``), and
  unlinks the rest.
- Emits a ``socket.reaped`` audit event (new
  ``EVENT_SOCKET_REAPED`` constant in ``pollypm.audit.log``) under the
  ``_workspace`` project key per reap, so leak rates can be
  forensically counted from ``~/.pollypm/audit/_workspace.jsonl``.

## Why bootstrap-only

At supervisor boot no cockpit is alive yet, so every reapable socket
is by definition dead — no race with the live cockpit binding the
same path. A runtime sweeper would race with the bridge's pre-bind
unlink. The issue explicitly flagged this concern; we sidestep it by
restricting the sweep to bootstrap.

## Staleness criteria

A ``.sock`` entry is reaped when ALL of:
- Filename parses as ``<kind>-<pid>.sock`` with an integer PID.
- ``os.kill(pid, 0)`` returns ``ProcessLookupError`` (PID is gone).
- The on-disk entry is actually a socket inode (not a regular file).

We chose PID-from-filename over ``lsof`` per-socket because at 318+
files the ``lsof`` approach is N kernel round-trips, whereas
``os.kill`` is constant-time per file. The bridge already encodes the
PID into the filename for exactly this reason
(``_socket_filename`` / ``_socket_owner_pid``).

## Test plan

- [x] Unit: ``test_reaps_socket_whose_pid_is_dead`` — real Unix
      socket with a dead PID gets unlinked.
- [x] Unit: ``test_preserves_socket_owned_by_live_process`` — current
      test process's PID is preserved.
- [x] Unit: ``test_reaps_only_stale_when_mixed`` — mixed live/stale
      directory: only stale ones go.
- [x] Unit: ``test_skips_files_with_unparseable_names`` — random
      ``.sock`` files in the dir are left alone.
- [x] Unit: ``test_skips_non_socket_file_with_dead_pid_name`` —
      regular files (not socket inodes) are preserved even if they
      happen to be named ``cockpit-<dead-pid>.sock``.
- [x] Unit: ``test_handles_missing_directory`` — fresh install where
      ``cockpit_inputs/`` doesn't exist is a no-op.
- [x] Unit: ``test_sweeps_tmpdir_fallback_directory`` — the
      ``$TMPDIR/pollypm-cockpit_inputs`` fallback dir is also swept.
- [x] Integration: ``test_emits_socket_reaped_audit_event`` —
      reap path writes a ``socket.reaped`` line to the central
      audit tail.
- [x] Integration: ``test_bootstrap_path_creates_then_reaps_only_stale``
      — full create-then-reap cycle.
- [x] Existing ``tests/test_supervisor.py`` (73 tests) still pass.
- [x] Existing ``tests/test_audit_log.py`` + ``test_audit_watchdog.py`` (43 tests) still pass.
- [x] Existing ``tests/test_cockpit_input_bridge.py`` + ``test_worker_marker_reaper.py`` (66 tests) still pass.

All sockets in tests are real AF_UNIX inodes — no kernel-state mocks.

### DIFF
--- a/src/pollypm/audit/log.py
+++ b/src/pollypm/audit/log.py
@@ -59,6 +59,10 @@ EVENT_WORK_DB_OPENED = "work_db.opened"
 # the fleet quantify whether the fix is holding without grepping
 # ``errors.log``.
 EVENT_WORKER_THREAD_LEAKED = "worker.thread_leaked"
+# #1368 — emitted by ``cockpit_socket_reaper`` when it unlinks a stale
+# ``cockpit-<pid>.sock`` whose owning PID is no longer alive. Lets
+# operators forensically count leak rates without scraping logs.
+EVENT_SOCKET_REAPED = "socket.reaped"
 
 
 @dataclass(slots=True, frozen=True)
@@ -408,6 +412,7 @@ __all__ = [
     "EVENT_WORK_TABLE_CLEARED",
     "EVENT_WORK_DB_OPENED",
     "EVENT_WORKER_THREAD_LEAKED",
+    "EVENT_SOCKET_REAPED",
     "AuditEvent",
     "central_log_path",
     "emit",
--- a/src/pollypm/cockpit_socket_reaper.py
+++ b/src/pollypm/cockpit_socket_reaper.py
@@ -1,0 +1,276 @@
+"""Reaper for stale ``~/.pollypm/cockpit_inputs/*.sock`` files (#1368).
+
+Background
+----------
+:mod:`pollypm.cockpit_input_bridge` binds an AF_UNIX socket per cockpit
+process at ``<base_dir>/cockpit_inputs/<kind>-<pid>.sock``. The
+``BridgeHandle.stop`` happy-path unlinks on clean shutdown, but:
+
+* ``SIGKILL`` / crash / tmux session teardown skip cleanup entirely.
+* The only opportunistic GC is inside
+  :func:`cockpit_input_bridge.send_key_to_first_live`, which only
+  unlinks a socket if a caller actually attempts a connection AND
+  receives ``ConnectionRefusedError`` AND the encoded PID is dead.
+
+In practice the directory accumulates one entry per crashed cockpit
+boot. On the reporting machine 318 stale entries had piled up; every
+``send_key`` walks all 318 (newest-first sorted) before reaching the
+live socket.
+
+This module mirrors :mod:`pollypm.work.worker_marker_reaper` for the
+session-marker / worker-marker case: a bootstrap-time sweep that runs
+once at ``Supervisor.bootstrap_tmux`` and unlinks any
+``cockpit-<pid>.sock`` whose PID is no longer alive.
+
+Staleness signal
+----------------
+The socket filename already encodes the cockpit PID (see
+``cockpit_input_bridge._socket_filename``). We re-use that PID for the
+liveness check rather than running ``lsof`` per socket — at 318+ files
+the ``lsof`` approach is N round-trips through the kernel, whereas
+``os.kill(pid, 0)`` is constant-time per file.
+
+A socket is reaped when ALL of the following hold:
+
+* Filename parses as ``<kind>-<pid>.sock`` with an integer PID.
+* ``os.kill(pid, 0)`` returns ``ProcessLookupError`` (or any
+  non-permission-error variant signalling the PID is gone).
+* The on-disk entry is actually a socket file (not a regular file
+  someone dropped here by mistake).
+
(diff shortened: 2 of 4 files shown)


samhotchkiss:
Bouncing for rebase: confirmed merge conflict in `src/pollypm/audit/log.py` against `origin/main` (`mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`).

PR #1385 (commit 27674426) merged to main and added `EVENT_WORKER_THREAD_LEAKED` in the same constants block + `__all__` list where this PR adds `EVENT_SOCKET_REAPED`. `git merge-tree --write-tree --messages origin/main pr-1387` reports:

```
Auto-merging src/pollypm/audit/log.py
CONFLICT (content): Merge conflict in src/pollypm/audit/log.py
```

Per review protocol, conflicts get bounced for rebase rather than fixed in-place. Please rebase onto current `main` and re-push; the resolution is mechanical (keep both new event constants + both new `__all__` entries). The reaper logic itself looks fine on inspection (bootstrap-only, PID-from-filename + `os.kill(pid, 0)`, sweeps both primary and `$TMPDIR` fallback dirs).

PR #1389, opened 2026-05-07
Hoist SQLiteWorkService construction to work module factory

samhotchkiss:
## Summary

Closes #1369.

- Adds `pollypm.work.create_work_service` as the canonical public construction surface, composing `resolve_work_db_path` with the `SQLiteWorkService` constructor.
- Migrates **every** direct `SQLiteWorkService(...)` callsite outside `pollypm.work/` — 38 sites across 27 files — onto the factory. After this PR a grep for `SQLiteWorkService(` outside `pollypm.work/` returns zero hits (excluding the existing class-as-callable injection in `cockpit_inbox_items.py`).
- Documents the canonical pattern in `pollypm.work.__init__` and the new `pollypm.work.factory` module.
- Bonus fix: the `notification_staging` prune handler in `plugins_builtin/core_recurring/maintenance.py` used to fall back to `_config.project.state_db` — the *messages-side* DB — when the store path was unavailable. `notification_staging` is a work-DB table, so that fallback would silently stamp the work schema onto the messages DB on fresh installs. The handler now routes through the factory so the canonical work DB is selected regardless.

## Factory API

```python
from pollypm.work import create_work_service

# Canonical: resolver picks the workspace-root DB.
with create_work_service(project_path=project.path) as svc: ...

# Escape valve (legacy migration, per-project scans, explicit overrides).
with create_work_service(db_path=db_path, project_path=project.path) as svc: ...
```

Signature:
```
create_work_service(
    *,
    db_path: str | Path | None = None,
    project_path: str | Path | None = None,
    config: PollyPMConfig | None = None,
    project_key: str | None = None,
    sync_manager: SyncManager | None = None,
    session_manager: object | None = None,
) -> SQLiteWorkService
```

## Test plan

- [x] Added `tests/test_work_service_factory.py` with 10 tests covering the public contract: factory is exported, resolver is called only when `db_path` is omitted, `config`/`project_key` are forwarded, the constructed service round-trips, both `str` and `Path` are accepted, the context-manager pattern works.
- [x] Added a regression guard test that scans `src/pollypm/` for any future direct `SQLiteWorkService(...)` construction outside the work package and fails CI if found.
- [x] All work-service / supervisor / heartbeat tests pass (394 tests in `pytest -k "factory or work_service or supervisor or heartbeat"`).
[...]

Issue #1004, opened 2026-04-30
Multi-DB sync gap: pm task list, cockpit, and workspace work_tasks see different task data (follow-up to #1003)


samhotchkiss:
## Severity: minor (state-source layering — cockpit and CLI see different data than the workspace DB)

Follow-up from #1003. The Bug A surface (`<project>/7` 'Plan ready for review' stub) was fixed in commit `4481c7f` by hiding pm-notify-inbox tasks from the Tasks pane and `pm task list`. This issue tracks the underlying multi-DB sync defect that produced the symptom.

## Summary

`pm task list --project smoketest` returned 7 tasks at one moment and 'No tasks found' a few minutes later. `SELECT * FROM work_tasks WHERE project = 'smoketest'` against `/Users/sam/.pollypm/state.db` consistently returned 0 the whole time. The task data is living in a transient or worktree-scoped storage that the workspace-root DB doesn't have.

The #1003 subagent diagnosed the root cause:

> `db_resolver` short-circuits on per-project DB existence regardless of content; architect notify writes are landing in the workspace-root DB while empty per-project DBs exist beside them.

So we have at least two task storage locations that aren't kept in sync — `pm task list` consults one, the cockpit consults another, the architect writes to a third. They drift; whatever the user sees depends on which path won the most recent read.

## Repro

1. `pm project new --yes <fresh_project>` — wait for planning to complete.
2. From a fresh terminal: `pm task list --project <project>` repeatedly. Observe results flicker between full task list and 'No tasks found' as background sync runs.
3. Compare to `sqlite3 /Users/sam/.pollypm/state.db 'SELECT project, task_number, work_status FROM work_tasks WHERE project = "<project>"'` — typically 0 rows.
4. Compare to `sqlite3 /Users/sam/dev/<project>/.pollypm/state.db 'SELECT project, task_number FROM work_tasks WHERE project = "<project>"'` if the file exists — may have rows or may not.

## Adjacent

The same multi-DB sync gap was hit yesterday: `pm task next` returned `pollypm/1` as available while `pm task get pollypm/1` returned 'not found'. Same flavor — different reads see different DBs.

## Repro environment

- HEAD `4481c7f` on `origin/main` (the #1003 Bug A fix is in)
- bikepath at `/Users/sam/dev/bikepath` and smoketest at `/Users/sam/dev/smoketest` are still in this state right now

## Instructions for the fix agent

[...]

PR #1423, opened 2026-05-07
Watchdog _gather_open_tasks: route through create_work_service factory

### LINKED ISSUE #1419: watchdog: _gather_open_tasks ignores post-#1004 workspace DB layout
## Context

Followup from #1414 / PR #1415 review.

`_gather_open_tasks` in `src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py` hardcodes:

```python
db_path = project_path / ".pollypm" / "state.db"
if not db_path.exists():
    return []
```

In the post-#1004 collapsed layout, `work_tasks` lives at `<workspace_root>/.pollypm/state.db`, not at `<project_path>/.pollypm/state.db`. On the user's machine the per-project DB exists but is empty (legacy file), and the real data lives at `/Users/sam/dev/.pollypm/state.db`. Net effect: `role_session_missing` rule silently no-ops in production — it found 0 open tasks for `savethenovel` even though `savethenovel/10` was at `status=review` with no reviewer window.

## Fix

Use `pollypm.work.db_resolver.resolve_work_db_path(project=project_key)` instead of the hardcoded path.

## Test gap

The integration tests in `tests/test_audit_watchdog.py` mock `_gather_open_tasks`, which is why this slipped through. The real-DB path needs a fixture that exercises the resolver.

## Severity

Medium — `task_review_stale` and `worker_session_dead_loop` still work (they read audit-log events), but `role_session_missing` is the rule the savethenovel/10 case explicitly targeted (no `reviewer-savethenovel` window). The PR's demo accidentally relied on `task_review_stale` firing instead.

samhotchkiss:
Closes #1419. Was using legacy per-project DB path; #1004 collapsed those into the workspace DB.

<details>
<summary>Why the rule was silently no-oping in prod</summary>

`_gather_open_tasks` in `src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py` hardcoded:

```python
db_path = project_path / ".pollypm" / "state.db"
if not db_path.exists():
    return []
```

In post-#1004 layouts that file is either missing or an empty leftover — the canonical `work_tasks` table now lives at `<workspace_root>/.pollypm/state.db`. Net effect: the open-task list always came back empty and `role_session_missing` could not fire. PR #1415's savethenovel demo only worked because `task_review_stale` (which doesn't depend on this query) covered the same case.

</details>

## Fix

Route through `pollypm.work.create_work_service` (the factory from #1389) with no `db_path` override, so the resolver picks the canonical workspace DB. `project_key` and `project_path` are forwarded for resolver warnings + project-aware audit metadata. Rule logic untouched — only the data source.

## Regression tests

Two new tests in `tests/test_audit_watchdog.py`:

- `test_gather_open_tasks_routes_through_factory_to_workspace_db` — seeds an `in_progress` task into a workspace-layout DB, leaves the legacy `<project>/.pollypm/state.db` deliberately absent, points the resolver at the canonical DB, and asserts both that `_gather_open_tasks` returns the seeded task and that `role_session_missing` fires through `_scan_one_project`. Verified: fails on the unfixed code (returns 0 tasks), passes with the fix.
- `test_gather_open_tasks_returns_empty_on_factory_failure` — pins the broad-except contract so a future "raise on missing config" refactor of the factory can't silently break the heartbeat.

## Test plan

- [x] `pytest tests/test_audit_watchdog.py` — 44 passed (was 42 before, +2 new)
- [x] Sanity check: revert the production fix and confirm the new regression test fails as expected, then restore the fix

### DIFF
--- a/src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py
+++ b/src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py
@@ -134,20 +134,25 @@ def _config_from_payload(payload: dict[str, Any]) -> WatchdogConfig:
 def _gather_open_tasks(project_key: str, project_path: Path | None) -> list[Any]:
     """Best-effort load of in-flight tasks for ``role_session_missing``.
 
+    Routes through :func:`pollypm.work.create_work_service` (the factory
+    landed in #1389), which delegates path resolution to
+    :func:`pollypm.work.db_resolver.resolve_work_db_path`. Post-#1004
+    the canonical DB lives at ``<workspace_root>/.pollypm/state.db`` —
+    a per-project ``<project>/.pollypm/state.db`` is at best stale and
+    at worst empty (#1419), so we deliberately do NOT pass a hand-built
+    ``db_path`` here. ``project_key`` is forwarded for resolver warnings
+    and ``project_path`` for project-aware audit metadata only.
+
     Returns an empty list on any failure — the watchdog keeps working
     even when the work-service is unreachable; the new rule simply
     no-ops for that project.
     """
-    if project_path is None:
-        return []
     try:
         from pollypm.work import create_work_service
 
-        db_path = project_path / ".pollypm" / "state.db"
-        if not db_path.exists():
-            return []
         with create_work_service(
-            db_path=db_path, project_path=project_path,
+            project_path=project_path,
+            project_key=project_key,
         ) as svc:
             list_fn = getattr(svc, "list_nonterminal_tasks", None)
             if not callable(list_fn):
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Bouncing for rebase: PR #1422 merged moments ago and touches the same files (`src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py` and `tests/test_audit_watchdog.py`). GH reports `mergeable: CONFLICTING`.

Please rebase onto current `main`, resolve the audit_watchdog.py / test_audit_watchdog.py conflicts, re-run `pytest tests/test_audit_watchdog.py tests/test_work_service_factory.py`, and push. The diff itself looks right — `_gather_open_tasks` correctly drops the hand-built `db_path` and routes through `create_work_service(project_path=..., project_key=...)` so the resolver picks the canonical workspace DB — this is purely a merge-order issue.

Activation Protocol v2 / pr-review bot.

PR #1451, opened 2026-05-08
Fix #1447 dedupe PM chat context re-entry

### LINKED ISSUE #1447: PM Chat re-injects context line every time the user re-enters
## Summary

Pressing `c` (chat with PM) from the project dashboard always sends a fresh context line into Polly's chat, even when the user is re-entering the same chat for the same project. The chat history fills with duplicate `re: project/<key> "dashboard discussion"` (or refinement primer / plan-now) lines.

## Repro

1. Open the cockpit on a project dashboard.
2. Press `c` — chat opens with a context line like `re: project/X "dashboard discussion"`.
3. Navigate back (e.g. `Ctrl-b ←` then `I` to inbox, or rail navigation).
4. Press `c` again on the same project dashboard.
5. Same context line is re-sent into Polly's chat as a fresh message.

## Expected

The context line should be sent at most once per (project, context-kind, recent-window). Re-entering an already-active chat should route the cockpit pane without re-injecting the same primer. Optional toast: "Re-attached to existing PM Chat for <project>."

## Actual

`_dispatch_to_pm_sync` always calls `_perform_pm_dispatch(cockpit_key, context_line)` with no recency check, so every keypress sends a new line.

## Where it lives

- `src/pollypm/cockpit_ui.py:16416` — `action_chat_pm` builds the context line (refinement primer / plan-now / dashboard discussion / blocker context).
- `src/pollypm/cockpit_ui.py:10097` — `_dispatch_to_pm_sync` always dispatches; no dedupe.
- `_perform_pm_dispatch` is the actual sender.

## Proposed direction

Cheapest: track the last `(cockpit_key, context_line)` pair + timestamp on the cockpit app instance. In `_dispatch_to_pm_sync`, if the same pair was sent within the last N seconds (e.g. 60–300s), still route the cockpit pane but skip the `_perform_pm_dispatch` send and surface a "re-attached" toast instead. The "blocker context already on screen" path at `cockpit_ui.py:16466` already implements a similar "don't re-investigate" guard for one variant — generalize that pattern for all branches of `action_chat_pm`.

Edge case: refinement primer with a different version bump should still send (the primer text changes when the plan version changes, so the dedupe key naturally differs).

## Severity

Mid. Doesn't break the chat, but burns tokens, fills Polly's context, and risks confusing her about which "discussion" the user actually wants.

## Surfaced by

User report on 2026-05-08 during a multi-stream debugging session.

samhotchkiss:
Closes #1447

Adds a recent-context guard for project dashboard PM Chat dispatches: identical cockpit/context pairs within the reattach window route back to the PM pane without sending another primer line. The actual route/send behavior is split so re-entry can reuse routing without injecting duplicate chat text.

Layer audit: project dashboard UI only; no service/store/domain boundary changes.

Tests:
- `PYTHONPATH=src /Users/sam/dev/pollypm/.venv/bin/python -m pytest tests/test_project_dashboard_ui.py::test_pm_dispatch_reattaches_recent_duplicate_context tests/test_project_dashboard_ui.py::test_c_keybinding_dispatches_to_pm -q`

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -14335,6 +14335,7 @@ class PollyProjectDashboardApp(App[None]):
         "j/k scroll \u00b7 g/G top/bottom \u00b7 v explainer "
         "\u00b7 o editor \u00b7 p back \u00b7 q exit"
     )
+    _PM_CONTEXT_REATTACH_WINDOW_SECONDS = 300.0
 
     def __init__(self, config_path: Path, project_key: str) -> None:
         super().__init__()
@@ -14488,6 +14489,7 @@ class PollyProjectDashboardApp(App[None]):
         # timer is the deferred-commit handle so ``u`` can cancel it.
         self._pending_plan_approval: dict | None = None
         self._pending_plan_approval_timer = None
+        self._last_pm_context_dispatch: tuple[str, str, float] | None = None
 
     # ------------------------------------------------------------------
     # Lifecycle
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Reviewer-bot pass: implementation matches issue #1447's proposed direction (track (cockpit_key, context_line, timestamp) on app instance, 300s sliding window, route pane without re-injecting context on duplicate re-entry). Clean factoring of _route_pm_target out of _perform_pm_dispatch. Both new tests pass on the PR branch in isolation. Module boundaries respected (UI-only). No overlap with parent-agent's uncommitted dashboard DB-resolution work (different functions, ~4500 lines apart). Merging.

PR #1452, opened 2026-05-08
Fix #1441 quiet notify fallback note

### LINKED ISSUE #1441: pm notify: confusing fallback warning when --user-prompt-json omitted
## Symptom

When the architect (or other caller) invokes \`pm notify\` at immediate priority without \`--user-prompt-json\`, the CLI emits this warning:

> Warning: posting an immediate-priority user-facing notify without --user-prompt-json. The dashboard's Action Needed card will fall back to [...] question for a first-class action surface.

Observed during savethenovel watch loop, tick 22, in architect pane after architect handled /12 review autonomously. The warning is correct, but it's printed to a context where it looks like an error.

## Why this matters

The architect surfaces this warning to a user-facing log. Users who don't understand the internals see a "Warning:" prefix and assume something is broken. It also doesn't actually offer a remediation — just informs.

## Suggested directions

Pick one:

1. Suppress the warning unless a specific debug flag is set; instead, just degrade silently to the fallback rendering.
2. Reword to a non-warning prefix ("note:" or just informational) since this is an expected fallback path.
3. If \`--user-prompt-json\` is genuinely the right call, make architect/agents pass it always (then the warning doesn't fire and the dashboard gets the better surface).

(3) is the most useful — fix the call sites, then the warning becomes truly exceptional.

## Discovered via

Round 2 watch loop on savethenovel; surfaced in observations at tick 22 (architect pane).

samhotchkiss:
Closes #1441

Suppresses the missing --user-prompt-json fallback diagnostic by default for immediate user-facing notify messages, so role panes no longer stream a Warning-looking line into user-facing logs. Developers can opt back in with POLLYPM_NOTIFY_USER_PROMPT_FALLBACK_NOTE=1, which prints a lowercase note instead of Warning:.

Layer audit: touched CLI command handling and CLI notify tests only; no service/domain/store imports or boundary changes.

### DIFF
--- a/src/pollypm/cli_features/session_runtime.py
+++ b/src/pollypm/cli_features/session_runtime.py
@@ -58,6 +58,13 @@ def _fake_recovery_inbox_override_enabled() -> bool:
     raw = os.environ.get("POLLYPM_DEV_FAKE_RECOVERY_INBOX", "").strip().lower()
     return raw in {"1", "true", "yes", "on"}
 
+
+def _notify_user_prompt_fallback_note_enabled() -> bool:
+    """Return True when missing-user-prompt fallback notes should be printed."""
+    raw = os.environ.get("POLLYPM_NOTIFY_USER_PROMPT_FALLBACK_NOTE", "")
+    return raw.strip().lower() in {"1", "true", "yes", "on"}
+
+
 # ``<project>/<N>`` matches the canonical task id form. When ``pm send``
 # receives an argument matching this shape we translate it to the
 # per-task worker window name (#924) so the user does not have to know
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Quiet-by-default fallback note matches the issue's options (1)+(2). Scope is tight (CLI session_runtime + scoped CLI test only), no boundary crossings, regression test pins both the silent default and the POLLYPM_NOTIFY_USER_PROMPT_FALLBACK_NOTE=1 opt-in. Verified by test only (worktree pytest tests/test_cli_notify.py: 32 passed); no activation due to active workspace work.

PR #1454, opened 2026-05-08
Fix #1442 audit raw task deletes

### LINKED ISSUE #1442: audit gap: tasks deleted from work_tasks without task.cancelled/deleted event
## Symptom

`advisor_review` tasks `savethenovel/27`, `/28`, `/29`, `/30` were auto-queued by `advisor.tick` at 2026-05-07T19:17:21-19:17:47Z (visible as `task.created` + `task.queued` events in `~/.pollypm/audit/savethenovel.jsonl`).

By tick 31 (~19:34:31Z), the rows were absent from both `work_tasks` and `work_transitions` — but **no `task.cancelled` or `task.deleted` audit event was emitted**.

DB-truth and audit-truth diverged silently.

## Why this matters

- The audit log is supposed to be the source of truth for task lifecycle. If tasks can vanish from the DB without an audit trail, the log can no longer be used to reconstruct state or debug why a task disappeared.
- For the auto-unstick mechanism (#1414), this is dangerous: a watchdog rule that reads the audit log will see stale `task.created`/`queued` events and treat the task as still pending — potentially escalating ghosts.
- Distinct from #1440 (`stuck_draft` spam, where the rule observes too much): this is the inverse — a sweeper acts without observability.

## Repro

1. Let `advisor.tick` auto-queue tasks at the cadence (no manual setup needed for projects with active advisors).
2. Wait one or two cadence cycles.
3. Query `work_tasks` for the auto-queued task numbers — gone.
4. Grep `~/.pollypm/audit/<project>.jsonl` for `task.cancelled` or `task.deleted` — absent.

## Suggested directions

1. **Find the sweeper** — likely a cleanup pass that deletes rows directly via SQL rather than going through the WorkService transition machinery. Route it through `cancel_task()` (or equivalent) so the standard `task.cancelled` event fires.
2. **Audit invariant** — add an integration test: every row deletion from `work_tasks` must be preceded by either `task.cancelled` or `task.deleted` in the audit log.
3. **Symmetric guard** — the audit module could refuse to emit `task.created` for a task that already has a previous `task.cancelled` (helps catch resurrection bugs too).

(1) is the right surgical fix; (2) prevents regression.

## Discovered via

Round 2 watch loop on savethenovel; tick 28 captured the auto-queue; tick 31 captured the disappearance.

samhotchkiss:
Closes #1442

Adds a SQLite delete-trigger outbox for work_tasks and flushes it through SQLiteWorkService so raw row deletes become task.deleted audit events. Covers the advisor_review-shaped repro where a queued task is removed by direct SQL outside the work-service transition machinery.

Layer audit: touched store/IO-owned work schema and work-service persistence only, plus focused schema/audit regression tests; no UI or service-facade imports added.

### DIFF
--- a/src/pollypm/work/schema.py
+++ b/src/pollypm/work/schema.py
@@ -126,6 +126,43 @@ CREATE INDEX IF NOT EXISTS idx_work_tasks_priority
 -- column the (pre-migration) legacy work_tasks table doesn't have
 -- yet, so it must live in the ensure helper rather than WORK_SCHEMA.
 
+CREATE TABLE IF NOT EXISTS work_task_delete_audit_outbox (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    project TEXT NOT NULL,
+    task_number INTEGER NOT NULL,
+    title TEXT NOT NULL DEFAULT '',
+    flow_template_id TEXT NOT NULL DEFAULT '',
+    previous_status TEXT NOT NULL DEFAULT '',
+    created_by TEXT NOT NULL DEFAULT '',
+    deleted_at TEXT NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS idx_work_task_delete_audit_outbox_project
+    ON work_task_delete_audit_outbox(project, task_number);
+
+CREATE TRIGGER IF NOT EXISTS trg_work_tasks_delete_audit_outbox
+AFTER DELETE ON work_tasks
+BEGIN
+    INSERT INTO work_task_delete_audit_outbox (
+        project,
+        task_number,
+        title,
+        flow_template_id,
+        previous_status,
+        created_by,
+        deleted_at
+    )
+    VALUES (
+        OLD.project,
+        OLD.task_number,
+        OLD.title,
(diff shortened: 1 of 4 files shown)


samhotchkiss:
LGTM. Trigger-based outbox is a clean way to catch raw-SQL deletes — the audit emit is decoupled from the deleter, so any future sweeper (sqlalchemy, raw SQL, ad-hoc cleanup) is covered automatically. Flush on both open and close gives reasonable durability.

Verified by tests only (store-layer change, regression test mirrors the #1442 advisor_review repro: queue task -> raw DELETE -> reopen -> assert task.deleted with previous_status=queued).

- tests/test_audit_log.py + tests/test_work_schema.py: 43 passed
- broader sweep (-k "work_service or sqlite_service or work_schema or audit"): 305 passed
- migration v8 -> v9 bump propagated to both legacy-walk tests
- module boundaries respected (work/* only, no UI imports)
- scope tight; the removed per-call emit in `_prune_cancelled_critique_child` is correctly replaced by the unified flush

Self-PR — GH blocks self-approve, leaving as comment per reviewer protocol.

PR #1455, opened 2026-05-08
Fix #1443 watchdog cadence dedupe

### LINKED ISSUE #1443: watchdog: cadence pulses occasionally drop a beat (single 5-min skip)
## Symptom

The audit watchdog runs on a 5-minute cadence (`@every 5m` recurring handler in `audit_watchdog.py`). Across the round-2 savethenovel watch loop, the cadence skipped a beat **twice** in the same session:

- **First skip**: 19:32Z pulse missing. Last seen 19:27:55Z, next at 19:37:59Z (gap = ~10m).
- **Second skip**: 19:57Z pulse missing. Last seen 19:52:41Z, next at 20:02:45Z (gap = ~10m).

Both were single-beat skips followed by clean recovery on the next 5m boundary. Spacing between the two skips: ~25 minutes.

## Why this matters

Auto-unstick (`task_review_stale` rule) relies on the cadence firing within `30 min - 5 min = 25 min` of the threshold crossing to escalate before the next status check. A single dropped beat is *currently* absorbed by the safety margin (the 5-min cadence is much shorter than the 30-min stale threshold). But:

1. If two beats drop in a row, escalation latency exceeds 10 min; user-perceived staleness gets visibly worse.
2. Rules with shorter horizons (e.g. if anyone tunes a watchdog rule below 10 min) would miss fires entirely.
3. Operationally, missed pulses make the watchdog log untrustworthy — anyone debugging "why didn't the rule fire" has to reason about whether a missed pulse is a rule bug or a scheduler issue.

This was witnessed live in the loop: /14's auto-unstick #4 still succeeded, but *only* because the second cadence pulse came back at 20:02:45Z, just 1m56s after the threshold. With one more dropped beat it would have stretched to 6m+.

## Repro

Watch `~/.pollypm/audit/_workspace.jsonl` or the project audit logs over a multi-hour window. Look for `audit.finding` clusters — they should appear at consistent 5-minute intervals (timestamps like XX:02:41, XX:07:41, XX:12:41 with small drift). When a beat is dropped, you'll see ~10-min gap between two clusters where there should be one.

In the savethenovel/round-2 case:
- ~/.pollypm/audit/savethenovel.jsonl
- ~/.pollypm/audit/_workspace.jsonl

Filter for `audit.finding` events and look for gaps >7 minutes between consecutive clusters.

## Hypotheses

1. **Scheduler clock drift / jitter**: the `@every 5m` registration may use wall-clock arithmetic that drifts and misses an aligned slot.
2. **Long-running pulse blocks next pulse**: if a single watchdog tick takes longer than 5 min for some reason (rare but possible with state-based detection paths added in #1435), the next tick is suppressed by the scheduler's "skip if previous still running" guard.
3. **Daemon hiccup**: rail_daemon momentarily unhealthy (CPU pressure, GC, etc.) and missed a fire.

(2) is most likely — state-based detection scans every project's `work_tasks` on each pulse, which is more expensive than the prior event-window scan. Worth profiling pulse duration and confirming it never approaches 5 min.

## Suggested directions

1. Log every pulse start/end with duration; alert if any pulse takes >2.5 min (half the cadence).
2. If pulse-too-long is the cause, run pulses on overlap-allowed scheduler (let two ticks queue/run sequentially rather than dropping one).
3. Add a self-healing rule: at the start of each pulse, check `last_pulse_at` from state; if gap >7 min, log a `watchdog.cadence_skip` finding so the gap is visible without external observers.

## Discovered via

Round 2 watch loop on savethenovel; ticks 31-32 captured first skip (19:32), ticks 36-37 captured second skip (19:57). Both single-beat, spaced ~25 min apart.

## Related

- #1435 (state-based detection paths added — possible perf cause)
- #1414 (auto-unstick architecture; this is a watchdog reliability gap inside it)

samhotchkiss:
Closes #1443

Adds a durable queue-side recency check for deduped recurring jobs and has the heartbeat skip same-window @every pulses without advancing local cadence. This prevents an active watchdog job from consuming the next beat and prevents a second HeartbeatRail from double-firing the same dedupe key seconds later.

Self-audit: touched heartbeat scheduling/domain code and the durable job queue API only; no UI or service facade changes.

### DIFF
--- a/src/pollypm/heartbeat/tick.py
+++ b/src/pollypm/heartbeat/tick.py
@@ -139,6 +139,9 @@ class Heartbeat:
                 entry.first_seen_at = now
 
             if self._is_due(entry, now, prev):
+                if self._dedupe_recent_or_active(entry, now):
+                    result.skipped_not_due += 1
+                    continue
                 payload_snapshot = _snapshot_payload(entry.payload)
                 job_id = self.queue.enqueue(
                     entry.handler_name,
@@ -186,6 +189,30 @@ class Heartbeat:
 
         # Unknown schedule type — be strict, don't fire.
         return False
+
+    def _dedupe_recent_or_active(
+        self,
+        entry: RosterEntry,
+        now: datetime,
+    ) -> bool:
+        """Coalesce same-window ``@every`` dedupe keys across live rails.
+
+        ``JobQueue.enqueue(dedupe_key=...)`` suppresses only queued/claimed
+        rows. If a cadence job finishes quickly, a second HeartbeatRail with a
+        slightly different in-memory anchor can otherwise enqueue the same
+        handler again seconds later. Conversely, if the previous row is still
+        active, treating that deduped enqueue as a fire consumes the next beat.
+        """
+        if not entry.dedupe_key or not isinstance(entry.schedule, EverySchedule):
+            return False
+        checker = getattr(self.queue, "has_recent_or_active_dedupe", None)
+        if not callable(checker):
+            return False
+        since = now - entry.schedule.interval
+        try:
+            return bool(checker(entry.dedupe_key, since=since))
+        except Exception:  # noqa: BLE001
+            return False
 
     def _every_is_due(
         self,
--- a/src/pollypm/jobs/queue.py
+++ b/src/pollypm/jobs/queue.py
@@ -363,6 +363,45 @@ class JobQueue:
             label="JobQueue.enqueue",
         )
 
+    def has_recent_or_active_dedupe(
+        self,
+        dedupe_key: str,
+        *,
+        since: datetime,
+    ) -> bool:
+        """Return True if ``dedupe_key`` is active or already fired since ``since``.
+
+        ``enqueue(dedupe_key=...)`` only dedupes queued/claimed rows. A
+        second HeartbeatRail can therefore enqueue the same cadence handler
+        seconds after the first one completes. Recurring ticks use this
+        read-side guard to coalesce same-window pulses across rails without
+        permanently reserving the dedupe key.
+        """
+        if not dedupe_key:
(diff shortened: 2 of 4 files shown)


samhotchkiss:
Reviewer LGTM (self-PR, comment in lieu of approval).

Verified by test only — additive, sealed-domain change (no UI/cli/rail_daemon touches that mandate activation; activation would conflict with 4 uncommitted parallel fixes in the working tree).

Diff is tight and faithful to the issue:
- src/pollypm/jobs/queue.py adds JobQueue.has_recent_or_active_dedupe(key, since=) — durable read-side guard returning True for queued/claimed rows OR any enqueued_at > since.
- src/pollypm/heartbeat/tick.py: only @every entries with a dedupe_key consult the new check; on hit, the entry is treated as not-fired (last_fired_at left intact), so the next beat is NOT consumed — addresses the +26s/+27s double-pulse pattern in #1443 without re-introducing the original 5-min skip.

Tests pin both failure modes:
- test_every_dedupe_active_job_does_not_consume_next_beat (covers 19:32/19:57 skip)
- test_every_dedupe_recent_job_blocks_second_rail_double_pulse (covers 20:07:46+20:08:12 / 20:27:52+20:28:19 doubles)
- test_has_recent_or_active_dedupe_tracks_active_and_recent_rows (queue contract)

Scoped tests green: tests/test_heartbeat_tick.py + tests/test_job_queue.py = 59 pass. Broader sweep (heartbeat_plugin_api + audit_watchdog + core_recurring_migration) = 119 pass.

No scope creep, no anti-patterns, no boundary violations. The bare `except Exception` in _dedupe_recent_or_active is justified (read-side guard must never crash the tick) and is properly noqa'd.

PR #1458, opened 2026-05-08
Fix #1445 add cockpit account recovery actions

### LINKED ISSUE #1445: cockpit: Accounts pane needs session-reassignment + controller-assignment actions
## Symptom

The cockpit's Accounts pane (`src/pollypm/cockpit_ui.py` lines 4004-4390) is read-only beyond add/remove/refresh/relogin. There is no UI to:

1. **Reassign which account a session uses** (e.g., move `heartbeat` or `operator` from one account to another).
2. **Set the controller account** (`pollypm.controller_account`).
3. **Block account removal when sessions still reference it** (right now removal relies on downstream errors).

Today, doing any of these requires either dropping into the legacy `pm accounts-ui` TUI (which only supports `operator` and `controller`, not arbitrary sessions like `heartbeat`) or hand-editing `~/.pollypm/pollypm.toml`.

## Why this matters

When a Claude or Codex account loses access (subscription disabled, org-policy change, expired login), the user has to migrate all sessions pinned to that account before they can remove it. With no UI, this becomes:

- Remember which sessions pin which account
- Edit the TOML directly
- Restart sessions manually

A common, foreseeable failure mode (auth dropping mid-cascade) becomes a config-spelunking exercise. Encountered live during the 2026-05-07 savethenovel watch loop — the user's `claude@swh.me` account dropped Claude Code access, blocking the `/15` worker and silently impacting `operator` + `heartbeat` (both pinned to that account). Surfacing this as a tracked gap.

## What exists today (do not re-implement)

The underlying API is already in place — only the UI is missing:

- `PollyPMService.switch_session_account(session_name, account_key)` — works for any session
- `PollyPMService.set_controller_account(account_key)` — works
- Legacy TUI wires:
  - `control_tui.action_switch_operator` (line 1477) — operator-only
  - `control_tui.action_make_controller` (line 1466)

## What to add (cockpit Accounts pane)

1. **Per-session reassignment**: in the account-detail view, show sessions referencing that account; for each, a button/picker to pick a different account; calls `switch_session_account()`. Should handle restart of the session post-switch.
2. **Make-controller button**: in the account-detail view; calls `set_controller_account()`.
3. **Safe-remove flow**: before calling remove, walk sessions for any pin to this account and either (a) require user to reassign first or (b) offer a one-click "reassign all to <picker>" before removing.

Optionally, the legacy `pm accounts-ui` should also gain the same per-session reassignment (it currently hard-codes `operator`), but the cockpit pane is the primary surface.

## Suggested layout

In `cockpit_ui.py`:
- Account-detail panel (around line 6214) gets two new buttons: "Make Controller", "Reassign Sessions…".
- "Reassign Sessions…" opens a modal listing each session with its current account and a dropdown of available accounts. Submit triggers `switch_session_account()` for changed rows and shows session-restart status.
- Remove button (existing) gets the safe-remove guard.

## Acceptance

- A user can migrate `heartbeat`, `operator`, and the controller off a failing account from inside the cockpit without ever touching the TOML.
- Removal is blocked (or auto-cascades a reassignment flow) when any session still pins the account.

## Discovered via

2026-05-07 savethenovel watch loop — user needed to migrate `heartbeat` + `operator` + `controller_account` off a Claude account whose org disabled Claude Code access. Mid-loop diagnostic confirmed `accounts-ui` only handles operator/controller and the cockpit handles neither.

## Related

- #1444 (no detector for worker auth-block) — sister issue: when an account fails, this UI is what the user reaches for to recover.

samhotchkiss:
Closes #1445

Adds cockpit Accounts-pane recovery actions for making a selected account the controller, reassigning sessions pinned to an account, and blocking removal until controller/session references are moved. The reassign path uses the existing PollyPMService controller/session switch facade and persists the selected session account/provider after successful switches so removal no longer hits stale config references.

Layer audit: UI-only cockpit/settings changes plus focused UI tests; no store or domain imports added.

### DIFF
--- a/src/pollypm/cockpit_settings_accounts.py
+++ b/src/pollypm/cockpit_settings_accounts.py
@@ -40,6 +40,14 @@ SETTINGS_ACCOUNT_ACTIONS: tuple[SettingsAccountAction, ...] = (
         label="Refresh Usage",
     ),
     SettingsAccountAction(
+        button_id="settings-account-make-controller",
+        label="Make Controller",
+    ),
+    SettingsAccountAction(
+        button_id="settings-account-reassign-sessions",
+        label="Reassign Sessions",
+    ),
+    SettingsAccountAction(
         button_id="settings-account-remove",
         label="Remove",
         variant="error",
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -4097,6 +4097,29 @@ def _collect_recent_tasks_by_account(
     return recent
 
 
+def _settings_session_refs_by_account(config) -> dict[str, list[dict[str, object]]]:
+    """Return configured session references grouped by pinned account."""
+    refs: dict[str, list[dict[str, object]]] = {}
+    sessions = getattr(config, "sessions", {}) or {}
+    for session_name, session in sorted(sessions.items()):
+        account_key = str(getattr(session, "account", "") or "")
+        if not account_key:
+            continue
+        provider = getattr(session, "provider", "")
+        provider_value = getattr(provider, "value", str(provider or ""))
+        refs.setdefault(account_key, []).append(
+            {
+                "name": str(getattr(session, "name", session_name) or session_name),
+                "role": str(getattr(session, "role", "") or ""),
+                "project": str(getattr(session, "project", "") or ""),
+                "provider": provider_value,
+                "enabled": bool(getattr(session, "enabled", True)),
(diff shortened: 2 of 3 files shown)


samhotchkiss:
Reviewed: cockpit_ui.py edits are in lines 4097-6589 (Settings Accounts pane); parent's uncommitted _dashboard_task_db_paths fix is at line 11948 — no overlap. Scoped tests/test_settings_ui.py: 21 passed. CI clean, mergeable. Merging.

Issue #1367, opened 2026-05-07
Module boundary: circular import pairs require lazy-import workarounds (25+ pairs)


samhotchkiss:
## Violation

A scan of all `pollypm/*.py` for top-level vs in-function imports of `pollypm.*` found 25+ pairs where module A lazy-imports B *and* B top-imports A. Lazy imports inside function bodies are a strong indicator of an unresolved circular dependency.

#### Cockpit cluster (worst offenders)
- `pollypm.cockpit_rail` <-lazy- `pollypm.cockpit`  AND  `pollypm.cockpit` -top-> `pollypm.cockpit_rail`
- `pollypm.cockpit_palette` <-lazy- `pollypm.cockpit_ui`  AND  `pollypm.cockpit_ui` -top-> `pollypm.cockpit_palette`
- `pollypm.cockpit_inbox` <-lazy- `pollypm.cockpit_inbox_items`  AND  `pollypm.cockpit_inbox_items` -top-> `pollypm.cockpit_inbox`
- `pollypm.cockpit_activity` <-lazy- `pollypm.cockpit_ui`  AND  `pollypm.cockpit_ui` -top-> `pollypm.cockpit_activity`
- `pollypm.cockpit_sections.dashboard` <-lazy- `pollypm.cockpit`  AND  `pollypm.cockpit` -top-> `pollypm.cockpit_sections.dashboard`

#### Work / sqlite_service (intra-package — service split is leaky)
- `pollypm.work.service_sync` <-lazy- `pollypm.work.sqlite_service` (and reverse top)
- `pollypm.work.service_worker_session_manager` <-lazy- `pollypm.work.sqlite_service`
- `pollypm.work.service_queries` <-lazy- `pollypm.work.sqlite_service`
- `pollypm.work.service_dependency_manager` <-lazy- `pollypm.work.sqlite_service`
- `pollypm.work.service_transition_manager` <-lazy- `pollypm.work.sqlite_service`
- `pollypm.work.service_notifications` <-lazy- `pollypm.work.sqlite_service`
- `pollypm.work.service_transitions` <-lazy- `pollypm.work.sqlite_service`

#### Heartbeats / supervisor
- `pollypm.heartbeats.local` <-lazy- `pollypm.supervisor`  AND reverse top
- `pollypm.heartbeats.api` <-lazy- `pollypm.supervisor`  AND reverse top
- `pollypm.schedulers.base` <-lazy- `pollypm.supervisor`  AND reverse top

[...]

PR #1462, opened 2026-05-08
Refs #1367 Break plugin validation import cycle

samhotchkiss:
Refs #1367

This fixes one low-risk circular-import slice: `plugin_validate` now imports `ExtensionHost` only under `TYPE_CHECKING`, so `plugin_host` can import `validate_plugin` normally instead of inside `_install`.

Self-audit: touched only the plugin service/facade validation path (`src/pollypm/plugin_host.py`, `src/pollypm/plugin_validate.py`). No UI, domain, store, or cross-layer imports added.

Remaining scope for #1367: current import scan still reports 13 pairs, including CLI feature pairs, cockpit pairs, onboarding, activity_feed, advisor inbox, and work service query/notification pairs. This PR intentionally leaves those for follow-up slices.

### DIFF
--- a/src/pollypm/plugin_host.py
+++ b/src/pollypm/plugin_host.py
@@ -46,6 +46,7 @@ from pollypm.plugin_api.v1 import (
     check_requires_api,
 )
 from pollypm.plugin_trust import warn_third_party_extension_trust_once
+from pollypm.plugin_validate import validate_plugin
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Reviewer-bot accept (self-PR, cannot approve via review): scoped tests green (92 passed: test_plugin_validate, test_plugins, test_plugin_interop, test_plugin_boundary_conformance). Tight diff matches title — top-level import of validate_plugin in plugin_host, TYPE_CHECKING-guarded ExtensionHost in plugin_validate (only used in annotations). Both modules import cleanly in either order. Merging.

PR #1467, opened 2026-05-08
Fix #1444 detect stale in-progress tasks

### LINKED ISSUE #1444: watchdog: no detector for worker alive-but-auth-blocked (silent stall)
## Symptom

A task can sit in `in_progress` indefinitely if the worker pane is alive but its underlying client (Claude Code, in this case) is blocked from making API calls. Observed live during round-2 savethenovel watch:

- `savethenovel/15` claimed at 20:03:09Z, marker created 20:03:12Z
- Worker pane (`pollypm-storage-closet:17`) is alive and responding to nudges, but every prompt returns:
  > Your organization has disabled Claude subscription access for Claude Code
- Five+ minutes elapsed with zero file changes and zero progress events.
- No watchdog rule fires.

## Why this matters

`task_review_stale` (the rule that drives auto-unstick #1414) only fires for tasks in `review` state. Tasks stuck in `in_progress` because the worker can't actually work hit no detection. If a user's auth quietly degrades (subscription expires, org-policy change, rate-limit ban), every cascade wedges silently — the auto-unstick mechanism does nothing because none of its rules apply.

## What's missing

A watchdog rule for tasks in `in_progress` that haven't moved in N minutes. Something like:

- `task_progress_stale` — if a task has been in `in_progress` for >M minutes (perhaps M = 20m for `implement`, longer for `test`/`review`) without a `task.transition` or `worker.heartbeat`, fire a finding and dispatch architect with brief: "worker for X is alive but unproductive — investigate auth, sandbox, or worker logic."

The rule needs to distinguish *new* in_progress tasks (just claimed, not yet stuck) from genuinely stalled ones. A heartbeat cadence on the worker side would help — but even without it, time-since-claimed + zero git activity is a reasonable heuristic.

## Repro

Disable Claude Code subscription/auth in your org settings, then claim a task from a savethenovel-style cascade. Worker pane will print the auth message; task will sit in_progress; no watchdog rule fires.

(In this session it happened organically — user's org-level Claude Code access dropped mid-cascade.)

## Related

- #1414 — auto-unstick architecture; this is the mirror gap (in_progress vs review).
- #1437 — Codex sandbox-write block (also a worker-side failure mode).
- #1439 — Russell asleep at the wheel (different pane wedge mode).
- #1440 — `stuck_draft` no escalation (sister gap on the draft side).

This trio (#1437, #1439, #1440, this one) plus #1442 paint a consistent picture: each task lifecycle state needs a stuck-detector, and the rules that exist are too rigid about which state they cover.

## Discovered via

Round 2 watch loop on savethenovel; tick 38 captured the wedge.

samhotchkiss:
Closes #1444

Adds a watchdog `task_progress_stale` detector for tasks currently stuck in `in_progress` without recent state, task-context, or worker heartbeat activity, and dispatches the architect brief with auth/sandbox recovery guidance. The recurring watchdog now hydrates `in_progress` tasks so the state-based rule has transition history.

Self-audit: touched watchdog domain code in `src/pollypm/audit/watchdog.py`, the recurring watchdog adapter in `src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py`, and focused watchdog tests only. No UI/store boundary changes.

### DIFF
--- a/src/pollypm/audit/watchdog.py
+++ b/src/pollypm/audit/watchdog.py
@@ -39,6 +39,11 @@ Detection rules (each returns a list of :class:`Finding`):
    following ``cancel_grace_seconds`` window. Indicates a planning
    stall after a self-cancel (savethenovel/1's class — a worker
    self-cancels but Polly never queues a follow-up).
+5. ``task_progress_stale`` — a live task currently at ``in_progress``
+   whose worker has produced no transition / heartbeat / task-context
+   activity for longer than ``progress_stale_seconds``. Catches the
+   alive-but-unproductive worker class, including auth-blocked panes
+   that keep accepting nudges but cannot make API calls.
 
 Severity is ``"warn"`` for every rule today. The recovery hint is
 attached to each finding so the alert message can include a
@@ -81,6 +86,7 @@ __all__ = [
     "RULE_STUCK_DRAFT",
     "RULE_CANCEL_NO_PROMOTION",
     "RULE_TASK_REVIEW_STALE",
+    "RULE_TASK_PROGRESS_STALE",
     "RULE_ROLE_SESSION_MISSING",
     "RULE_WORKER_SESSION_DEAD_LOOP",
     "RULE_TASK_ON_HOLD_STALE",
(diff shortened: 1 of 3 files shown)


samhotchkiss:
Reviewer LGTM (self-PR, can't formally approve). Verified by scoped tests (77 passed in tests/test_audit_watchdog.py). Diff is tightly scoped to audit/watchdog.py + the recurring adapter + tests — no UI/store boundary touches, no scope creep. State-based detector (preferred) + audit-event fallback with dedupe is well-structured; new `task_progress_stale` rule wires into `_DISPATCHABLE_RULES` so the architect gets the brief with auth/sandbox/quota guidance. Note: parent agent's #1474 (Heartbeat auth_broken → recovery + filter unhealthy accounts) addresses the same #1444 from a complementary angle — that PR catches the auth-failure pattern at the heartbeat layer and filters unhealthy accounts at spawn. This PR catches stuck in_progress tasks regardless of cause via watchdog detection. They're orthogonal mechanisms — both ship together for layered detection.

PR #1469, opened 2026-05-08
Fix #1446 suppress pytest work-db open audit

### LINKED ISSUE #1446: pytest sweeps spam canonical audit log with work_db.opened events
## Summary

`pytest` test runs write `work_db.opened` audit events into the production canonical audit log (the workspace `audit.jsonl` resolved by `resolve_work_db_path`). During a long sweep, the canonical audit log accumulates ~10 `work_db.opened` rows/second, masking real signal during incident diagnosis.

## Repro

1. Real PollyPM workspace at `/Users/sam/dev` with the canonical audit log being written.
2. Run a wide pytest sweep, e.g. `uv run pytest tests/ -x -q`.
3. Tail the canonical audit log while the sweep is running — fills with `work_db.opened` events from the test process.

## Expected

Test fixtures isolate `work_db.opened` (and similar lifecycle audit events) from the real canonical log. Two options:
- `tmp_path`-based config fixtures fully redirect the audit-log writer, OR
- `work_db.opened` is gated on a test-mode check and no-ops when set.

## Actual

The events land in the production `audit.jsonl`. Verified observationally on 2026-05-08 ~01:00–01:11Z during two concurrent fix-subagent test sweeps (advisor.tick throttle/resolver + auth_broken failover). Caused the watch-loop observer subagents to flag it explicitly: the audit log was \"drowning in pytest `work_db.opened` rows.\"

## Severity

Diagnostic noise → mid. Doesn't break product behavior, but masks signal during multi-stream debugging. Hit twice during a single session today; worse with parallel pytest runs.

## Proposed direction

Find the producer of `work_db.opened` (likely `pollypm.work.create_work_service` or the SQLite-service constructor's audit emit). Either:
- Wrap the emit in a guard that no-ops under a `POLLYPM_DISABLE_AUDIT_EMIT` (or similar) env var that conftest sets for the test session.
- Have the test conftest override the audit-log path to a tmp file alongside the tmp DB.

Whichever matches existing test-isolation patterns elsewhere in the repo.

## Surfaced by

Watch-loop observer subagents during ticks 50–52 on 2026-05-08.

samhotchkiss:
Closes #1446

Suppress the high-frequency work_db.opened audit breadcrumb when POLLYPM_DISABLE_WORK_DB_OPENED_AUDIT is set, and set that guard for pytest by default. Audit-log tests explicitly clear the guard so production emit behavior remains covered, including the per-project log path added by #1454.

Layer self-audit: touched domain work-service constructor plus test configuration/coverage only; no UI, facade, or store boundary reach.

### DIFF
--- a/src/pollypm/work/sqlite_service.py
+++ b/src/pollypm/work/sqlite_service.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 import hashlib
 import json
 import logging
+import os
 import sqlite3
 import subprocess
 from datetime import UTC, datetime
@@ -29,6 +30,13 @@ from pathlib import Path
 from typing import Any, Protocol
 
 logger = logging.getLogger(__name__)
+
+_DISABLE_WORK_DB_OPENED_AUDIT_ENV = "POLLYPM_DISABLE_WORK_DB_OPENED_AUDIT"
+
+
+def _work_db_opened_audit_disabled() -> bool:
+    value = os.environ.get(_DISABLE_WORK_DB_OPENED_AUDIT_ENV, "")
+    return value.strip().lower() not in {"", "0", "false", "no", "off"}
 
 
 # #894 — register the work_service module as an emitter that routes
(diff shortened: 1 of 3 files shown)


samhotchkiss:
Reviewer pass (self-PR, comment-only). Verified by test only (no cockpit/UI touch). Fix matches Path 1 from the issue: env-var gate POLLYPM_DISABLE_WORK_DB_OPENED_AUDIT, set by conftest, explicitly cleared in the audit-log test fixture so production emit path stays covered. Regression test pins the disabled-guard behavior. Truthiness logic correct (0/false/no/off/empty disable; everything else enables). tests/test_audit_log.py 22/22 pass; tests/test_work_service*.py + read_write_consistency 85/85 pass. Layer-clean: domain only. No conflict with parent's open PRs (#1470/#1471/#1474/#1475).

PR #1470, opened 2026-05-08
Cockpit: prefer workspace canonical DB over per-project legacy DB

samhotchkiss:
## Summary

Reorder `_candidate_dbs` (cockpit_tasks.py) and `_dashboard_task_db_paths` (cockpit_ui.py) so the workspace-root canonical `state.db` is checked before the project-local one. Per #1004 the per-project DB is deprecated; both helpers were still preferring it, which caused the task-drilldown Live tab to show "No active worker session" even when canonical had the in_progress row + work_sessions row.

## Context

Reproduced live during today's debugging session: clicked savethenovel → task /15 → Live and saw "No active worker session attached to this task" while the worker pane (`%11933`) was actively coding Module 6. Diagnosed as `_get_svc()` picking the per-project DB first because it had 68 stale advisor.tick rows; that DB had zero `work_sessions`, so `get_worker_session(active_only=True)` returned None.

## What changed

- `src/pollypm/cockpit_tasks.py:_candidate_dbs` — reorder so workspace root is added first, project-local second, config-state-db override last.
- `src/pollypm/cockpit_ui.py:_dashboard_task_db_paths` — same reorder for the dashboard mirror.
- `tests/test_cockpit_tasks_candidate_dbs.py` — new file with two regression tests:
  - `test_candidate_dbs_prefers_workspace_when_both_exist` — workspace wins when both DBs exist with rows.
  - `test_candidate_dbs_falls_back_to_per_project_when_workspace_missing` — legacy back-compat when only per-project exists.

## Test plan

- [x] `tests/test_cockpit_tasks_candidate_dbs.py` (2 new tests, pass)
- [x] Targeted suite: `test_audit_watchdog.py`, `test_legacy_per_project_db_migration.py`, `test_workspace_scope_resolution.py` (88 passed)
- [x] Broader cockpit_tasks-touching tests (118 passed)
- [x] Reinstall + manual verification: `pm cockpit-pane` task drilldown Live tab now resolves correctly through the workspace DB

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1476, opened 2026-05-08
Fix #1436 orphan marker age threshold

### LINKED ISSUE #1436: [regression] orphan_marker rule's age threshold is off by ~7x; fires false positives on fresh markers
## Repro

savethenovel watch tick 13 (2026-05-07T18:00 UTC area):
- savethenovel/12 marker created ~4 min ago.
- `orphan_marker` rule fired TWICE claiming the marker has been alive for >30 min.
- Off by approximately 7x (4 min reported as >30 min).

## Hypothesis

Time arithmetic bug. Likely candidates:
- Timezone confusion: comparing UTC ISO timestamp against local-time-naive datetime, treating local time as UTC and getting a 7-hour offset (Mountain Daylight = UTC-6/-7).
- Float vs int ms/s mismatch.
- Wrong field used (e.g.  of task instead of marker file mtime).

## Impact

- Alarm fatigue: every fresh task gets an orphan_marker finding within minutes of starting.
- Audit log noise.
- Doesn't dispatch the architect (orphan_marker is finding-only, not in dispatch set), so no false escalations — but pollutes the signal.

## Cross-references

- #1414, #1415, #1428, #1435 — sibling watchdog work; this rule is from PR #1344 (the original watchdog scaffold).

samhotchkiss:
Closes #1436

Updates orphan_marker so the 30 minute threshold is evaluated against marker.created age instead of treating fresh marker.created events inside the scan window as already orphaned. The regression test covers an old task with a fresh marker to prevent false positives on active work.

Self-audit: touched only the audit watchdog detector and its focused tests; no layer-boundary changes.

### DIFF
--- a/src/pollypm/audit/watchdog.py
+++ b/src/pollypm/audit/watchdog.py
@@ -20,11 +20,11 @@ no new alert channel is introduced).
 Detection rules (each returns a list of :class:`Finding`):
 
 1. ``orphan_marker`` — ``marker.created`` for ``task-<project>-<N>``
-   with no matching ``marker.released`` within the window AND no
-   ``task.status_changed`` to a terminal state (``done`` / ``cancelled``
-   / ``abandoned``) in that window. Catches the savethenovel pattern
-   where a worker is launched and the task is then cancelled but the
-   on-disk marker lingers.
+   older than the configured threshold with no matching
+   ``marker.released`` AND no ``task.status_changed`` to a terminal
+   state (``done`` / ``cancelled`` / ``abandoned``). Catches the
+   savethenovel pattern where a worker is launched and the task is
+   then cancelled but the on-disk marker lingers.
 2. ``marker_leaked`` — any ``marker.leaked`` event in the window.
    Each occurrence is a finding because the persona-swap-detected
    branch firing at all means a worker was redirected — we want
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Reviewed against /Users/sam/Desktop/pollytests-2026-05-04/05-reviewer-protocol.md.

Conflict check vs parent's open PRs (#1470/#1471/#1474/#1475): clean — touches only `src/pollypm/audit/watchdog.py` + `tests/test_audit_watchdog.py`, different territory from cockpit, advisor handlers, heartbeat/sessions, and test_task_assignment_notify.

Fix correctness: the actual bug per #1436 issue body is that orphan_marker had no age check at all — any `marker.created` event inside the 30-min read window fired immediately, so a 4-min-old marker tripped the >30-min message. Diff adds `if ts > cutoff: continue` so the event must be older than `window_seconds` before it counts. Threshold semantics now match the docstring + log message.

Note on the stuck-dedupe hypothesis: I traced the firing-then-silent pattern. Scanner reads via `read_events(since=now-window_seconds*2)` (60-min lookback by event timestamp), so `marker.created` naturally scrolls out at T+60min and the rule goes silent regardless of dedupe. There is no per-finding dedupe tracker for orphan_marker (the alert dedupe via `upsert_alert` session name is not a permanent-suppression cache; `was_recently_dispatched` is escalation-only and orphan_marker is finding-only). The 12-fires-then-silent observation is consistent with cadence ticks while the event was in window, then the event aging out — that's a separate window-exit behavior, not in scope for #1436's stated bug.

Tests: ran `tests/test_audit_watchdog.py` on worktree of d79a1bf1 — 78 passed. New regression test `test_orphan_marker_recent_marker_ignored_even_when_task_is_old` correctly pins the marker-age vs task-age distinction.

LGTM — merging.

Issue #1365, opened 2026-05-07
Module boundary: cross-plugin coupling — advisor/task_assignment_notify both reach into project_planning.plan_presence; itsalive/magic/morning_briefing reach into core_agent_profiles


samhotchkiss:
## Violation

Plugins under `pollypm.plugins_builtin/X` import from `pollypm.plugins_builtin/Y` — a sign the imported piece should be hoisted into the plugin API or a shared module.

\`\`\`
3  core_recurring -> task_assignment_notify
1  advisor -> project_planning           (plan_presence)
1  task_assignment_notify -> project_planning  (plan_presence)
1  itsalive -> core_agent_profiles
1  magic -> core_agent_profiles
1  morning_briefing -> core_agent_profiles
\`\`\`

Specific imports:

- `src/pollypm/plugins_builtin/advisor/handlers/assess.py:61`
  \`from pollypm.plugins_builtin.project_planning.plan_presence import ...\`
- `src/pollypm/plugins_builtin/task_assignment_notify/handlers/sweep.py:94`
  \`from pollypm.plugins_builtin.project_planning.plan_presence import ...\`
- `src/pollypm/plugins_builtin/itsalive/plugin.py:9`
  \`from pollypm.plugins_builtin.core_agent_profiles.profiles import ...\`
- `src/pollypm/plugins_builtin/magic/plugin.py:5`
  \`from pollypm.plugins_builtin.core_agent_profiles.profiles import StaticPromptProfile\`
- `src/pollypm/plugins_builtin/morning_briefing/...` (cf. core_agent_profiles import)
- `src/pollypm/plugins_builtin/core_recurring/*` (3 imports of task_assignment_notify; check if these are sweep helpers that belong in a shared module)

## Why it's wrong

- Plugins should be independently loadable / disablable. Cross-plugin imports mean disabling one plugin breaks others.
- Both `plan_presence` (a pure predicate) and `core_agent_profiles.profiles` (prompt strings + a lightweight protocol) are used by core *and* sibling plugins — clear signs they belong as foundational modules.

## Proposed fix

Bundle this with #1361 and #1363:

1. Hoist `core_agent_profiles` to `pollypm.agent_profiles` (foundational), keep the plugin shell as a registration-only module.
2. Hoist `project_planning.plan_presence` to a foundational `pollypm.plans` (or `pollypm.plan_presence`) module, leaving the plugin's CLI/handlers in place.
[...]

PR #1471, opened 2026-05-08
Advisor.tick: restore throttle, route writes through canonical resolver

samhotchkiss:
## Summary

Two compounding bugs in advisor.tick that together caused 68 stale "Advisor review" rows to silently pile up on the savethenovel project, invisible to anything reading the canonical workspace DB.

## Bug 1 — Throttle permanently disabled

`has_in_progress_advisor_task` short-circuited to `False` whenever `work_service is None`. In production it's always None (the tick handler is invoked with `payload={}`), so the throttle never triggered. Every 30-min tick created a fresh advisor task even when 67 prior reviews were already queued.

Fix: lazy-build a work_service when called without one — same pattern already used by `enqueue_advisor_review` and `has_project_stagnation_candidate`.

## Bug 2 — Bypassing the canonical DB resolver

`_resolve_state_db` in `advisor/db_paths.py` checked `<project>/.pollypm/state.db` first and only walked up. Per #1004, `pollypm.work.db_resolver.resolve_work_db_path` always returns the workspace canonical DB and per-project DBs are deprecated. Advisor was silently routing all writes to per-project — invisible to engines reading workspace canonical (cockpit, `pm task list`, etc).

Fix: delete `advisor/db_paths.py`, route advisor through `resolve_work_db_path` like the rest of the system.

## Closes / refs

Partial fix for #1437 (the write-path side). Paired with PR #1450's per-task-worker `--add-dir` fix on the read side.

Note: PR #1463 ("route plugin transition reads through storage") was held earlier today specifically because it imports from the now-deleted `db_paths.py`. After this PR merges, #1463 will need rebase OR can be closed as redundant.

## Test plan

- [x] `tests/test_advisor_plugin.py` extended (existing tests pass; new throttle + resolver tests added)
- [x] Reinstall + observed live: advisor.tick now writes new rows to `/Users/sam/dev/.pollypm/state.db` (canonical) instead of `/Users/sam/dev/savethenovel/.pollypm/state.db` (per-project). Canonical max(task_number) advanced 26 → 91 over the test window.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1477, opened 2026-05-08
Refs #1365 hoist plan presence gate

samhotchkiss:
Refs #1365

Hoists the plan-presence gate into `pollypm.plan_presence` and leaves `project_planning.plan_presence` as a compatibility alias. Updates advisor and task_assignment_notify to consume the shared module so those sibling plugins no longer import through project_planning.

Layer audit: domain/shared module plus plugin handler import rewires and focused tests only; no UI/store changes in the committed slice.

Remaining scope: `core_recurring` still imports `task_assignment_notify.api`; `core_agent_profiles` plugin imports in itsalive/magic/morning_briefing are already absent on current main.


samhotchkiss:
Reviewed via worktree. Verdict: clean self-PR — merging.

**Hoist quality:**
- New `src/pollypm/plan_presence.py` is a verbatim move; no `self` smuggling (it was already a module-level function set, no class refactor).
- Old `project_planning/plan_presence.py` becomes a 15-line shim using `sys.modules[__name__] = _shared` so that legacy callers (`cockpit_ui`, `cockpit_tasks`, `cockpit_rail`, `recovery/state_reconciliation`) and `mock.patch("...project_planning.plan_presence._find_approved_plan_task")` continue to resolve to the same object.
- Verified `p1 is p2` and that all caller modules import cleanly under the shim.
- `assess.py` and `sweep.py` updated to top-level imports of the new module (also tidies their previous mid-file import block — nice).

**Tests run on PR worktree:**
- `test_plan_presence_gate.py` + `test_dashboard_plan_staleness_cache.py`: 33 passed (latter still uses old monkeypatch path; shim aliasing carries it).
- `test_task_assignment_notify*`, `test_advisor_plugin`, `test_advisor_assess`, `test_state_drift_reconciliation`: 186 passed.

**Conflict scan vs open parent PRs (#1470/#1471/#1474/#1475):** no file overlap. #1471 touches `advisor/advisor_tick.py` + `detect_changes.py`; this PR touches `advisor/assess.py` — sibling files, no merge conflict.

LGTM.

PR #1498, opened 2026-05-08
cockpit: auto-redraw rail on focus-regain + periodic stale-paint tick

### LINKED ISSUE #1497: cockpit: rail UI can stale on focus-loss / SIGWINCH; needs auto-redraw + focus-event hook
## Symptom (live tonight)

User reported the cockpit's left rail UI "died" — pane showed Polly's Claude session but no rail UI rendered. Workaround was tmux detach + reattach (`Ctrl-b d` then `tmux attach`), which forces a full re-render. rail-daemon process (PID 1570) was alive throughout the dead-rail period.

## Root cause

`PollyCockpitApp` (in `cockpit_ui.py`) has:
- A 0.8s `_tick` that polls state and calls `_refresh_rows()` when state changed, but never calls `self.refresh()` directly. The implicit Textual repaint relies on widget state changing.
- An `on_resize` handler that DOES force a redraw (good).
- **No `on_app_focus` handler**, so when the tmux pane regains focus there's no recovery hook.
- No periodic stale-paint refresh — if Textual believes nothing changed but tmux dropped a redraw signal, the rail stays blank until the next genuine state change.

When tmux loses focus, the client disconnects and reattaches, or SIGWINCH gets dropped, Textual can fail to repaint. Detach+reattach works because reattaching forces Textual to redraw the entire tree.

## Proposed fix shape

1. **Periodic stale-paint refresh** in `_tick`: at a `_STALE_PAINT_INTERVAL` cadence (~5 ticks ≈ 4s at 0.8s/tick), call `self.refresh(layout=False)` defensively. Cheap, recovers from dropped redraw signals within ~4s.
2. **`on_app_focus` handler** on `PollyCockpitApp` that runs the same recovery path as `on_resize` (`call_after_refresh(self._recover_after_resize)`). Restores the rail the moment the pane is observed again.
3. Tests via source inspection (the full Textual harness is unreasonable here): assert `_STALE_PAINT_INTERVAL` exists, `_tick` references it + calls `self.refresh()`, `on_app_focus` exists and routes through `_recover_after_resize`.

## Severity

Medium-high — silently breaks the user's primary visibility into project state. They have no way to know the rail is stale rather than empty without manual intervention.

## Verify

After the fix activates: trigger the dead-rail scenario by tmux detach + don't reattach for ~10s; reattach; rail should be current. Or kill+restart the tmux client; rail should redraw on focus. With the periodic refresh, even sustained tmux disruption recovers within ~4s.

samhotchkiss:
Closes #1497

## Summary

Tmux focus-loss / client reattach / SIGWINCH drops can leave the cockpit's left rail visibly blank even though widget state is current and `rail-daemon` is alive. The user's manual workaround was tmux detach + reattach (forces a full re-render). This wires two non-invasive auto-recovery paths so the workaround stops being needed.

## Changes

- **`src/pollypm/cockpit_ui.py`**:
  - New `_STALE_PAINT_INTERVAL = 5` class constant on `PollyCockpitApp` (~4s at the 0.8s tick).
  - `_tick` now calls `self.refresh(layout=False)` on that cadence — defensive non-layout repaint that catches any dropped Textual redraw signal within a few seconds. Cheap; doesn't reflow the list.
  - New `on_app_focus(self, _event: events.AppFocus)` handler. Forwards to `call_after_refresh(self._recover_after_resize)` — the same recovery path `on_resize` already uses (`_recover_cockpit_render(force_render=True)`). Restores the rail the moment the pane is observed again.

- **`tests/test_cockpit_rail_redraw.py`** (new): regression sentinels via source inspection — assert the constant exists, `_tick` references it + calls `self.refresh()`, `on_app_focus` exists and funnels through `_recover_after_resize`. Source inspection rather than a full Textual harness because constructing a real `PollyCockpitApp` pulls in a config + supervisor graph that's unreasonable for unit scope.

## Test plan

```
uv run --extra dev pytest tests/test_cockpit_rail_redraw.py tests/test_import_boundary.py tests/test_cockpit.py -x
```

Local: 5/5 new + 8/8 boundary + 186/186 cockpit smoke (= 199/199) pass.

## Activation-required for full verify

This touches `cockpit_ui.py`. Reviewer should activate (`uv tool install --reinstall .` + `pm reset --force` + `pm up --phantom-client`) and confirm:

- After ~4s of any rail oddity (e.g., shrunk pane, switched away and back, killed+reattached client), the rail is redrawn — no manual detach+reattach needed.
- Focus-out doesn't cause flashes during normal use (the refresh is `layout=False` and gated on the existing 0.8s tick — should be invisible during steady state).

## Boundary discipline

Read `docs/conventions.md` Import boundaries + `docs/architecture.md` Service API v1 + `tests/test_import_boundary.py` header before coding. No Supervisor direct imports added, no `_method` reach-through, no `_conn` SQL, no cross-plugin private imports, no CLI-as-runtime-dep. No allow-list additions. `tests/test_import_boundary.py` is green.

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -1854,6 +1854,12 @@ class PollyCockpitApp(App[None]):
     _LAYOUT_CHECK_INTERVAL = 38  # ~30s at 0.8s/tick
     # Force GC every ~2 minutes
     _GC_INTERVAL = 150
+    # Stale-paint defensive refresh: tmux focus loss / SIGWINCH drops can
+    # leave Textual without a redraw signal even though widget state is
+    # current. Force a non-layout refresh on this cadence so the rail
+    # recovers within a few seconds without operator intervention. See
+    # the rail-died report — detach + reattach was the manual workaround.
+    _STALE_PAINT_INTERVAL = 5  # ~4s at 0.8s/tick
 
     def _tick(self) -> None:
         self._tick_count += 1
@@ -1926,6 +1932,15 @@ class PollyCockpitApp(App[None]):
         if self._tick_count % self._LAYOUT_CHECK_INTERVAL == 0:
             try:
                 self._enforce_rail_width()
+            except Exception:  # noqa: BLE001
+                pass
+        # Stale-paint defense: force a Textual repaint every few ticks
+        # so the rail recovers from focus-loss / disconnect / SIGWINCH
+        # drops where Textual's own redraw signal got lost. layout=False
+        # because we only need a paint, not a reflow.
+        if self._tick_count % self._STALE_PAINT_INTERVAL == 0:
+            try:
+                self.refresh(layout=False)
             except Exception:  # noqa: BLE001
                 pass
 
@@ -3109,6 +3124,19 @@ class PollyCockpitApp(App[None]):
         self._recover_cockpit_render(force_render=False)
 
     def on_resize(self, _event: events.Resize) -> None:
+        self.call_after_refresh(self._recover_after_resize)
+
+    def on_app_focus(self, _event: events.AppFocus) -> None:
+        """Force a full rail redraw when the tmux pane regains focus.
+
+        When the tmux client disconnects/reattaches, the user switches
+        away to another pane and back, or any other event drops the
+        Textual redraw signal, the rail can be left visibly blank
+        even though widget state is current and rail-daemon is alive.
+        Detach + reattach has been the manual workaround. Hooking
+        AppFocus restores the rail automatically the moment the pane
+        is observed again.
+        """
         self.call_after_refresh(self._recover_after_resize)
 
     def _recover_after_resize(self) -> None:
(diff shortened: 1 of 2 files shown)


samhotchkiss (commented):
Boundary check: clean. 8/8 import_boundary tests green on the PR branch. Light-activation verify done in /tmp/wt-review-1498:
- import pollypm.cockpit_ui — clean
- from pollypm.cockpit_ui import * — clean (no decorator/syntax errors)
- tests/test_cockpit_rail_redraw.py — 5/5 pass

Code review:
- _STALE_PAINT_INTERVAL = 5 with periodic self.refresh(layout=False) is the right minimal repaint cadence — light enough to not jank, frequent enough to recover from stale state within seconds.
- on_app_focus handler hooks Textual's focus-regain — exactly what tmux focus-events deliver after detach+reattach.
- Regression sentinels are the correct gate for Textual redraw bugs (audit anticipated this — direct redraw tests are not feasible without an interactive Textual harness).

Tests: 5/5 + 8/8 + 186/186 green per PR body.

Approving and merging — completes the rail-death recovery fix you saw earlier.

PR #1499, opened 2026-05-08
pm task approve: recover from pre-merge 'would be overwritten' on safe files

samhotchkiss:
Closes #1496

## Summary

`pm task approve <task>` was silently blocking on every project where the worker's `task/<slug>-N` branch and `main` had each added different lines to `.gitignore`. The python dirty-tree gate (`_status_is_only_pollypm_scaffold`) admits the case, but git's merge engine has its own pre-merge guard that doesn't share that allowance — so `git merge` errors out with `error: Your local changes to the following files would be overwritten by merge: .gitignore` *before* any `MERGE_HEAD` or conflict state exists. The existing union-safe resolver (`_resolve_union_safe_conflict`) was never reached.

Live repro tonight: coffeeboardnm/1 shipped a deliverable (live URL verified, `pm itsalive verify ok=True`) but the task state machine was stuck in `review` because three approve attempts each created an auto-stash (`preserve-main-gitignore-before-coffeeboardnm-1-approve-{,2}`, `preserve-reviewer-gitignore-before-coffeeboardnm-1-approve`) and never advanced.

## Fix

When `git merge --no-ff` fails before producing conflict state, parse the refused-files list from stderr. If every refused file is in `_UNION_SAFE_MERGE_FILES`, reset those files to HEAD (`git checkout HEAD -- <file>`) and retry the merge. The union-safe resolver then handles any actual add/add conflict that surfaces during the proper merge.

Targeted recovery — does **not** restructure the existing flow, doesn't add a new worktree, doesn't change behavior for non-safe-file conflicts, and is a no-op for the already-working union-safe path.

## Repro from #1496

\`\`\`
✗ Could not auto-merge \`task/coffeeboardnm-1\` into \`main\`. Resolve the repo state and retry approve. Git said: error: Your local changes to the following files would be overwritten by merge:
  .gitignore
  Please commit your changes or stash them before you merge.
  Aborting
\`\`\`

The repeat-attempt history was visible as auto-stashes:

\`\`\`
stash@{0}: On main: preserve-main-gitignore-before-coffeeboardnm-1-approve-2
stash@{1}: On pollypm/coffeeboardnm/reviewer/reviewer_coffeeboardnm: preserve-reviewer-gitignore-before-coffeeboardnm-1-approve
stash@{2}: On main: preserve-main-gitignore-before-coffeeboardnm-1-approve
\`\`\`

## Test plan

`tests/test_task_approve_merge_gitignore.py`:

[...]

PR #1484, opened 2026-05-08
Refs #1356 hoist session runtime command callbacks

samhotchkiss:
Refs #1356

Hoists the session/runtime Typer command callbacks out of register_session_runtime_commands and registers bound module-level callbacks instead. The registration function drops from 798 LOC to 34 LOC while preserving the existing command signatures/help surface and closure-style helper capture.

Layer self-audit: CLI layer only (src/pollypm/cli_features/session_runtime.py). No store/domain imports were added; existing lazy imports remain in the command bodies.

Remaining scope: other >200 LOC offenders still include cockpit_ui._dashboard_inbox, session_runtime.notify, heartbeats.local._process_session, cockpit_inbox._gather_worker_roster, and the remaining functions listed on #1356.

Tests:
- uv run --extra dev ruff check src/pollypm/cli_features/session_runtime.py
- uv run --extra dev pytest tests/test_cli_notify.py tests/test_notification_tiering.py tests/test_inbox_dedup.py
- uv run --extra dev pytest tests/test_cli_help_examples.py tests/test_cli_send_force.py tests/test_inbox_fake_recovery_injection_gate.py
- uv run --extra dev pytest tests/test_import_boundary.py
- uv run pm --help
- uv run pm notify --help

Known existing failure observed outside touched files:
- uv run --extra dev pytest tests/test_import_boundary.py tests/test_cockpit_contract_boundaries.py fails test_tmux_pane_window_methods_stay_inside_allowlist on pre-existing offenders in src/pollypm/cli_features/ui.py, src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py, and src/pollypm/work/worker_marker_reaper.py.

PR #1502, opened 2026-05-08
pm CLI: replace functools.partial command callbacks with signature-preserving wrappers (fixes #1500)

### LINKED ISSUE #1500: pm CLI broken: typer crashes on partial(callback, helpers) registration after #1484
## Symptom

After #1484 merged, building \`pm\` from current \`main\` produces a binary that crashes on every invocation:

\`\`\`
TypeError: functools.partial(<function launch at 0x...>, <module 'pollypm.cli' from '...'>) is not a module, class, method, or function.
\`\`\`

Trace shows typer's \`get_params_from_function\` calling \`get_type_hints\` on the partial, which raises because partial isn't a function/method/class.

## Repro (live tonight)

1. \`cd /tmp/wt-pm-install\` (worktree on origin/main HEAD = 7e478e50, post-#1484, post-#1499)
2. \`uv tool install --reinstall .\`
3. \`pm task status coffeeboardnm/1\` → TypeError above

Reverting to \`ec677590\` (pre-#1484) and reinstalling produces a working pm. So the regression is squarely in #1484.

## Why #1484's review missed this

The review noted "pm --help / notify --help / rail-daemon --help / reset --help / send --help all preserve signatures" — but those were probably run inside the test environment / before a fresh install. The bug only manifests on actual \`uv tool install --reinstall\` of the PR's tree, then a fresh pm invocation. \`tests/test_session_runtime.py\` (158/158) doesn't go through typer's registration path.

## Why it matters

Anyone who reinstalls pm from current \`main\` gets a completely non-functional CLI. That includes the operator, codex's tooling, every reviewer's activation step, and the whole heartbeat-self-heal verification path.

Currently masked because running daemons (heartbeat, rail-daemon, cockpit) already loaded the old code into memory at session start. New invocations are what break.

## Proposed fix shape

Two options:

1. **Stop using \`partial\` for typer command registration.** Find a way to attach the helpers without losing the inspectable signature — module-level wrapper functions, a closure inside a registration helper, or \`functools.wraps\` to copy the wrapped function's signature onto the partial.

2. **Update typer/main.py to unwrap partials.** Probably out of scope (third-party).

Option 1 is the right path. Each command should be registered with its real function or a thin wrapper that preserves the signature.

## Tests

Add a smoke test that imports \`pollypm.cli\` and instantiates the Typer app — the registration crash will fire there. \`tests/test_pm_cli_registration.py\`. This catches the regression at unit-test speed without needing a full reinstall.

## Severity

High — silent regression that breaks every fresh install of the PM CLI from \`main\`.

## Related

Filed as part of overnight magic-hardening pass. Discovered while activating the merged \`task-approve-gitignore-union\` (#1499) fix. Augments the audit's items.

samhotchkiss:
Closes #1500

## Summary

Typer's command-registration path calls \`typing.get_type_hints\` on each registered callable. Python 3.13's \`typing.py\` rejects \`functools.partial\` objects ("not a module, class, method, or function") — so the \`partial(callback, helpers)\` pattern introduced in #1484 produced a \`TypeError\` at every fresh \`uv tool install --reinstall .\` of pm.

Live repro tonight (post-#1484, post-#1499):

\`\`\`
TypeError: functools.partial(<function launch at 0x...>, <module 'pollypm.cli' from '...'>) is not a module, class, method, or function.
\`\`\`

The fix preserves the spirit of #1484 (closures hoisted out of \`session_runtime\`) without using \`partial\` for the typer-registered callable.

## Fix

\`_bind_session_command\` now returns a real wrapper function instead of a \`partial\`:

- \`__signature__\` drops the bound \`helpers\` parameter so typer sees only the user-facing options/arguments.
- \`__annotations__\` drops \`helpers\` to match.
- \`__name__\`/\`__doc__\`/\`__module__\` preserved via \`functools.wraps\`.

The wrapper closes over \`helpers\` and forwards \`(helpers, *args, **kwargs)\` to the original callback. From typer's introspection point of view it's a normal function — \`get_type_hints\` succeeds.

## Why #1484's review missed this

The merged review claimed "pm --help / notify --help / rail-daemon --help / reset --help / send --help all preserve signatures" — those checks must have run inside the test environment / on already-imported modules, not after a real \`uv tool install --reinstall\` and shell invocation. \`tests/test_session_runtime.py\` tests the bound callable's behavior but not typer's registration path.

This PR adds \`tests/test_pm_cli_registration.py\` which does walk that path (\`typing.get_type_hints\`, \`inspect.signature\`, a Typer-runner-driven \`--help\`) so the regression cannot recur silently.

## Test plan

\`\`\`
uv run --extra dev pytest tests/test_pm_cli_registration.py tests/test_cli.py tests/test_cli_help_examples.py tests/test_session_manager.py tests/test_import_boundary.py
\`\`\`

111/111 pass locally (5 new + 35 cli + 35 cli-help + 27 session_manager + 8 import_boundary).

**Install-time check (the one that proves the regression is gone):**

\`\`\`
uv tool install --reinstall .
pm --help
\`\`\`

Confirmed locally: pm --help renders the usage text instead of crashing on \`get_type_hints\`. Verified before pushing.

## Boundary discipline

Read \`docs/conventions.md\` "Import boundaries (enforced in CI)" + \`docs/architecture.md\` Service API v1 + \`tests/test_import_boundary.py\` header before coding. Changes are entirely internal to \`pollypm.cli_features.session_runtime\` (no Supervisor direct imports added or removed, no \`_conn\` SQL, no cross-plugin imports, no allow-list changes).

8/8 \`tests/test_import_boundary.py\` green.

### DIFF
--- a/src/pollypm/cli_features/session_runtime.py
+++ b/src/pollypm/cli_features/session_runtime.py
@@ -13,7 +13,8 @@ Contract:
 
 from __future__ import annotations
 
-from functools import partial
+import functools
+import inspect
 import json
 import os
 import re
@@ -205,11 +206,37 @@ def _hold_review_tasks_for_notify(
 
 
 def _bind_session_command(callback, helpers):
-    bound = partial(callback, helpers)
-    bound.__name__ = callback.__name__
-    bound.__doc__ = callback.__doc__
-    bound.__module__ = callback.__module__
-    return bound
+    """Bind ``helpers`` as the first arg of ``callback`` while preserving the
+    inspectable signature Typer needs.
+
+    Typer's command-registration path calls :func:`typing.get_type_hints` on
+    the registered callable. ``functools.partial`` objects don't satisfy
+    ``get_type_hints``'s "module/class/method/function" check (Python 3.13's
+    typing.py raises ``TypeError`` on them), so we build a real wrapper
+    function whose signature drops the bound first parameter — Typer then
+    sees a normal function with the user-facing options/arguments only.
+    """
+    sig = inspect.signature(callback)
+    parameters = list(sig.parameters.values())
+    if not parameters:
+        bound_param_name = None
+        new_sig = sig
+    else:
+        bound_param_name = parameters[0].name
+        new_sig = sig.replace(parameters=parameters[1:])
+
+    @functools.wraps(callback)
+    def wrapper(*args, **kwargs):
+        return callback(helpers, *args, **kwargs)
+
+    wrapper.__signature__ = new_sig
+    if bound_param_name is not None:
+        wrapper.__annotations__ = {
+            name: annotation
+            for name, annotation in callback.__annotations__.items()
+            if name != bound_param_name
+        }
+    return wrapper
 
 
 def launch(
--- a/tests/test_pm_cli_registration.py
+++ b/tests/test_pm_cli_registration.py
@@ -1,0 +1,103 @@
+"""Smoke test that catches typer-introspection regressions on command callbacks.
+
+Issue #1500: ``functools.partial`` wrappers on Typer command callbacks tripped
+``typing.get_type_hints`` (Python 3.13's typing.py raises ``TypeError`` because
+partials don't satisfy "module/class/method/function"). The whole pm CLI
+became non-functional on every fresh install. This test re-registers the CLI
+in a clean Typer app and exercises the same introspection path Typer takes
+internally — any future regression that breaks ``inspect.signature`` /
+``get_type_hints`` on a registered command surfaces here at unit-test speed.
+"""
+
+from __future__ import annotations
+
+import inspect
+import typing
+
+import typer
+from typer.testing import CliRunner
+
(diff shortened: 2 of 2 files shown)


samhotchkiss (commented):
Boundary check: clean. session_runtime.py is the right surgical seam for the fix.

Code review:
- Replacing functools.partial with a signature-preserving wrapper in _bind_session_command is exactly the right fix. functools.wraps preserves the inspectable callable for typer.
- The new test (tests/test_pm_cli_registration.py) catches the regression at unit speed: typing.get_type_hints, inspect.signature, and a Typer-runner --help invocation. Three angles, each catches a different failure mode.
- Fork verified `uv tool install --reinstall . && pm --help` works on the branch — the install-time regression IS gone.

Tests: 111/111 (5 new registration + 35 cli + 35 cli-help + 27 session_manager + 8 import_boundary). Test-only-safe per protocol since the fork already did install-and-invoke verification.

This unblocks every activation-required review tonight. Approving and merging.

Issue #1506, opened 2026-05-08
worker: pm send-up CLI for posting mid-task milestone updates to operator inbox


samhotchkiss:
## Symptom

Workers iterate silently between \`pm task claim\` and \`pm task done\`. The operator (Polly) sees the start and the finish but nothing in between — "render started", "research done", "deploy started" all happen inside the worker pane and never surface unless the operator asks. That gap is exactly why coffeeboardnm/1 looked stalled to the user during the visual-explainer render this past session: real progress was happening (\`Rendering poc-plan.html… 6m 57s · ↓ 23.6k tokens\`); nobody upstream could see it.

## Repro (live tonight)

1. Worker on coffeeboardnm/1 spent 14m 30s rendering HTML between research-done and deploy-start.
2. Polly's chat (\`pollypm:0\`) showed only the original deploy-verify message; no progress narration.
3. The user pinged me to ask if "#1 was actually moving forward" because there was no operator-visible signal.

The work was happening — there was just no path for the worker to surface mid-task milestones without going through Polly's heartbeat (which is alert-shaped, not progress-shaped).

## Proposed fix

A new \`pm send-up "<message>"\` CLI:

- Emits a single inbox card with category \`worker_milestone\` + the worker's message text.
- Auto-detects task id from \`POLLYPM_TASK_ID\` env var or \`--task <project/N>\` flag.
- Dedupes the same message text from the same task within a 5 minute window (so a retry-loop worker doesn't spam Polly).
- Card lands at \`priority="normal"\` — informational, not alert. Polly sees it the next time she scans her inbox; no heartbeat wakeup, no operator interrupt.

## Tests

- Helper-level: title build, body build, dedupe window logic.
- Module-level: emit_worker_milestone happy/empty/missing/dedupe/create-failure paths.
- CLI: happy path, missing task id (clean error), env var fallback, dedupe (silent success), JSON output.

## Severity

Medium — magic-feel improvement. Without it, every long-running task looks stalled to operator until the worker reports done.

samhotchkiss:
Merged via #1507, SHA 9dd9cc2b8d37432bed0fbc293fd72584e7073b8b

PR #1505, opened 2026-05-08
heartbeat: distinguish progress_signal from liveness; mark stalled-but-running roles flow=quiet

samhotchkiss:
Closes #1504

## Summary

Generalises the per-question detection added in #1493 (merged as #1495) to the broader "alive but silent" pattern. Workers and other non-event-driven roles that produce no fresh transcript output for ≥3 ticks now have their \`pm status\` reason prefixed with \`flow=quiet:\` so operators can distinguish "actively working" from "silently stalled."

Repro tonight: \`pm status\` showed \`status=healthy running=yes\` on every architect/reviewer/worker line, all with \`reason="No new transcript output since last heartbeat"\` — looked all-green while coffeeboardnm/1 sat for 25+ minutes.

## Changes

- **\`src/pollypm/heartbeats/base.py\`** — \`HeartbeatCursor.quiet_tick_count: int = 0\`. Default keeps legacy cursors.json payloads loadable.
- **\`src/pollypm/heartbeats/api.py\`** — \`update_cursor()\` accepts optional \`quiet_tick_count: int = 0\`; persisted via the existing \`asdict\` round-trip.
- **\`src/pollypm/heartbeats/local.py\`** — three module constants (\`_QUIET_TICKS_FOR_FLOW_MARKER=3\`, \`FLOW_QUIET_REASON_PREFIX="flow=quiet: "\`, \`_PROGRESS_SIGNAL_EXEMPT_ROLES\` covering heartbeat-supervisor / operator-pm / reviewer / architect). \`_process_session()\` reads the prior counter from \`context.cursor\`, increments on \`(verdict="unclear", reason=_NO_TRANSCRIPT_REASON)\` for non-exempt roles, resets otherwise. After threshold, calls \`_set_session_status(..., reason=flow=quiet: <original>)\`.
- **\`tests/test_heartbeat_progress_signal.py\`** — new file, 14 tests.
- **\`tests/test_heartbeat_plugin_api.py\`** — \`FakeHeartbeatAPI.update_cursor\` updated to accept the new kwarg (single line + record it).

## Pairs with

#1495 (heartbeat-question-detect, merged) covers the specific "agent asked operator a question" stall. **This** PR covers every other "alive but silent" case. Together they're defense in depth on the broader pattern.

## Test plan

\`\`\`
[...]

PR #1495, opened 2026-05-08
heartbeat: detect worker awaiting operator question, surface to inbox

samhotchkiss:
Closes #1493

## Summary

Workers that end a turn with a question to the operator and sit at the empty `❯` prompt now classify as `awaiting_operator` instead of `legitimate_idle`. The heartbeat emits a `worker_question` alert with the question text so polly / the cockpit operator inbox sees it instead of `pm status` silently reporting `healthy`.

## What changed

- **`pollypm.idle_placeholders`** — new `pane_ends_with_unanswered_question(pane_text) -> str | None`. Verifies the pane tail shows the empty `❯` prompt, walks back through the assistant's last block (skipping TUI footer chrome, `✻`/`✶` status spinners, box-drawing rules), returns the joined text iff it ends in a question pattern. Returns `None` when the operator already responded (a `❯ <text>` line sits between the question and the prompt) or when the pane is mid-turn.
- **`pollypm.heartbeats.stall_classifier`** — `StallClass` extended with `awaiting_operator`. `StallContext` gets `awaiting_operator_question: str | None`. `classify_stall` checks the question field *before* the `pane_is_idle_placeholder` short-circuit so the question wins over the silent-idle promotion.
- **`pollypm.heartbeats.local`** + **`pollypm.supervisor_alerts`** — both stall-detection callsites now compute `pane_ends_with_unanswered_question` alongside `pane_is_idle_placeholder` and route an `worker_question` alert via the existing `_emit_routed_alert` / `_route_signal` paths when the new class fires. No new emit infrastructure; existing inbox surface picks it up.

## Test plan

- `tests/test_stall_classifier.py` — 7 new cases (all passing):
  - placeholder + question pattern → `awaiting_operator`
  - architect / `worker_pollypm` + question → `awaiting_operator` (priority over their legitimate_idle paths)
  - placeholder + no question → `legitimate_idle` (regression guard)
  - real coffeeboardnm pattern detected
  - normal idle (\"task complete\") returns None
  - already-answered (`❯ go ahead` between question and prompt) returns None
  - mid-turn pane returns None
- `tests/test_import_boundary.py` — green. No allow-list additions.
- 242-test broader sweep across `heartbeat / stall / idle_placeholder / supervisor_alerts` — green.

## Manual reproduction

The exact failure this addresses:

[...]

PR #1509, opened 2026-05-08
heartbeat: detect actual pane crashes and auto-respawn active roles within 1 tick

### LINKED ISSUE #1508: heartbeat: auto-respawn architect/reviewer panes on actual crash (vs nudge-stall)
## Symptom

When a worker/architect/reviewer pane crashes (Claude error, OOM, manual close that leaves the tmux pane alive but the Claude/codex process gone), the heartbeat marks it unhealthy via the \`shell_returned\` alert but does NOT auto-restart. The session sits dead until either:

- The operator notices and runs \`pm reset\` manually, OR
- The \`_MAX_NUDGES_BEFORE_RECOVERY\` (=6) circuit-breaker fires — which it never does because there's no Claude PID to nudge in the first place.

For active projects, this means a dead role can sit dead for many minutes before recovery.

## Repro pattern (live tonight)

Heartbeat reports many sessions as \`status=healthy running=yes\` with reason "No new transcript output since last heartbeat" — some of those might actually be crashed (Claude exited, tmux pane back at \`zsh\`) without anyone noticing. Generalises the same liveness-vs-progress conflation that #1493 (heartbeat-question-detect, merged via #1495) and #1501 (heartbeat-progress-signal, merged via #1505) addressed for the silent-but-alive case.

## Proposed fix shape

In \`_process_session\` (\`src/pollypm/heartbeats/local.py\`), right after the \`shell_returned\` alert block:

1. **Detect role-crash:** \`pane_command in {bash,zsh,sh,fish}\` AND \`role in {worker, architect, reviewer}\` AND \`has_pending_work=True\`.
2. **Auto-respawn within this tick** by calling \`_recover_session(failure_type="role_crashed", ...)\`. Reuses the existing recovery path — no new respawn mechanism.
3. **Don't auto-respawn idle role panes** — \`has_pending_work=False\` paths are intentionally fine to leave at the shell prompt.
4. **Crash-loop guard:** after \`_CRASH_LOOP_ATTEMPT_THRESHOLD\` (=2) cumulative recoveries on the session, escalate to a \`crash_loop\` operator alert instead of respawning again.

## Tests

\`tests/test_heartbeat_role_respawn.py\` (new) covers:

- Crashed architect / reviewer / worker with pending work → \`recover_session\` called with \`failure_type="role_crashed"\` in this tick.
- Crashed role with NO pending work → no respawn.
- Crash-loop threshold exceeded → \`crash_loop\` alert, no respawn.
- Stalled-but-alive role still uses the existing nudge ladder (don't break #1495 / #1505).
- Sentinel: role list locked to \`worker/architect/reviewer\`; threshold locked to 2.

## Severity

Medium — silent dead-role can sit dead for many minutes; degrades the magic-feel of self-healing PollyPM. Fix is targeted and unit-testable.

## Related

Filed as part of the overnight magic-hardening pass (audit at \`/tmp/magic-audit-issues.md\`, P2 role-respawn-on-crash). Pairs with #1495 + #1505 as defense in depth.

samhotchkiss:
Closes #1508

## Summary

When a worker/architect/reviewer pane returns to a shell prompt (Claude/codex process exited but tmux pane itself is still alive) AND the role has active project work, the heartbeat now immediately calls \`recover_session(failure_type=\"role_crashed\", ...)\` instead of waiting for the nudge ladder's \`_MAX_NUDGES_BEFORE_RECOVERY\` ticks. The nudge ladder can't fire when there's no Claude PID to nudge — so without this fix, dead roles sit dead until the operator notices.

(The commit message references #1506 — that was the in-progress slug; the actual issue filed is #1508.)

## Distinct from existing recovery paths

- \`pane_dead\` (whole tmux pane gone) — already auto-recovered via the existing block above \`shell_returned\` in \`_process_session\`.
- Stalled-but-running (Claude PID alive, just silent) — uses the existing nudge ladder via #1495 (heartbeat-question-detect) / #1505 (heartbeat-progress-signal).
- Idle role panes (no pending work) — intentionally NOT auto-respawned; the existing \`shell_returned\` alert is the right operator surface for that case.

## Crash-loop guard

After \`_CRASH_LOOP_ATTEMPT_THRESHOLD\` (=2) cumulative \`recovery_attempts\` on the session, an additional crash escalates to a \`crash_loop\` operator alert (severity=error) instead of respawning yet again. The threshold uses the existing \`session_runtime.recovery_attempts\` counter, so any combination of recoveries (auth, missing window, pane_dead, role_crashed) counts toward the cap. Conservative — biases toward escalation rather than infinite respawn.

## Files

- \`src/pollypm/heartbeats/local.py\` — two new class constants (\`_CRASH_RECOVERY_ROLES\`, \`_CRASH_LOOP_ATTEMPT_THRESHOLD\`) + the new role-crash detection block immediately after the \`shell_returned\` alert.
- \`tests/test_heartbeat_role_respawn.py\` (new) — 9 regression tests.

## Test plan

50/50 green locally including 8/8 \`tests/test_import_boundary.py\`. Plus broader sweep \`tests/test_heartbeats_local_classify.py tests/test_heartbeat_tick.py tests/test_heartbeat_plugin_api.py\` 76/76 green.

## Boundary discipline

Read \`docs/conventions.md\` Import boundaries + \`docs/architecture.md\` Service API v1 + \`tests/test_import_boundary.py\` header before coding. No Supervisor direct imports added, no \`_method\` reach-through, no \`_conn\` SQL, no cross-plugin private imports, no CLI-as-runtime-dep. No allow-list additions.

## Recovery path reused

\`_recover_session\` → \`api.recover_session\` → \`Supervisor.maybe_recover_session\` (the same path \`pane_dead\` already uses, plus auth_broken / missing_window). No new respawn mechanism. The \`failure_type=\"role_crashed\"\` is a new label so logs and event streams can distinguish role-crash recoveries.

### DIFF
--- a/src/pollypm/heartbeats/local.py
+++ b/src/pollypm/heartbeats/local.py
@@ -532,6 +532,20 @@ class LocalHeartbeatBackend(HeartbeatBackend):
     )
     _WORKER_MUTATING_SESSION_ROLES = frozenset({"worker"})
 
+    # #1506 — role-respawn-on-crash. Roles that can crash (Claude/codex
+    # process exits leaving the tmux pane at a shell prompt) and need
+    # auto-respawn when there is active work to do. Operator/heartbeat
+    # roles are deliberately excluded — those crashes are rare and the
+    # operator-attention surface is elsewhere.
+    _CRASH_RECOVERY_ROLES = frozenset({"worker", "architect", "reviewer"})
+
+    # After this many cumulative recovery attempts on a session, an
+    # additional crash escalates to the ``crash_loop`` alert instead of
+    # respawning yet again. Conservative — uses the existing
+    # ``recovery_attempts`` counter from session_runtime so any
+    # recovery (auth, missing window, pane_dead, role_crashed) counts.
+    _CRASH_LOOP_ATTEMPT_THRESHOLD = 2
+
     _AUTH_FAILURE_PATTERNS = (
         "authentication failure",
         "not authenticated",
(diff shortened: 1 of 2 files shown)


samhotchkiss (commented):
Boundary check: clean. Touches only heartbeats/local.py + new test file. No Supervisor direct imports, no _conn SQL, no allow-list additions.

Code review:
- Reuses existing recover_session → Supervisor.maybe_recover_session path with new failure_type='role_crashed'. Doesn't fork a new respawn mechanism.
- Crash-loop guard at threshold=2 cumulative recovery_attempts — bounded escalation.
- Idle role panes (no pending work) intentionally left at shell prompt — existing shell_returned alert path handles those. Correct scoping.
- Pairs with #1495 (question detect), #1505 (progress signal), and now #1509 (crash detect) as a complete heartbeat self-heal triad.

Tests: 9 new + 19 stall_classifier + 14 progress_signal + 8 boundary = 50/50 scoped, plus 76/76 broader heartbeat sweep.

Note on commit message: fork's commit message references #1506 (a slug confusion); PR body's 'Closes #1508' is canonical and correct. Cosmetic, doesn't affect linkage.

Approving and merging — this completes the full P0/P1/P2 audit.

Issue #1546, opened 2026-05-09
Heartbeat cascade: tiered self-heal + generic safety-net probes + tier-3 operator handoff + tier-4 escalated mode + product-broken terminal state


samhotchkiss:
## The model

```
heartbeat tick →
  TIER 1: run specific mechanical rules
          if any matched & healed → done

  GENERIC SAFETY-NET PROBES: is this project making forward progress
                              OR cleanly idle?
          yes → done
          no  → handoff to TIER 2

  TIER 2: project PM (Cole, Archie, etc.) — full project context
          resolved? → done
          stuck?    → handoff to TIER 3

  TIER 3: Polly (operator), normal mode
          resolved? → done
          stuck?    → escalate to TIER 4

  TIER 4: Polly with the gloves off — \"do everything possible to fix\",
          broader authority, fewer guard-rails, wider scope
          resolved? → done
          exhausted? → TERMINAL

  TERMINAL: set product state = broken (first-class flag, visible
            everywhere, blocks new work) + urgent structured inbox
            message to Sam (what was tried, what failed, hypothesis,
            forensics location)
```

## Why this shape

Every tier handles only what it can confidently handle. Mixing tiers — heartbeat reasoning about ambiguous causes, PM trying to fix things outside its scope, no escalation path beyond Polly — leads to silent wedges or destructive over-correction.

**The generic safety-net is the crucial piece.** Specific tier-1 rules are deny-list — they catch known bad shapes. Without a default-deny safety net that asks \"is this project making progress, OR cleanly idle?\", anything we didn't think to write a rule for slips through silently. Exactly what happened on 2026-05-08/09 with coffeeboardnm: 52 advisor-task cancellations, three structural bugs, none caught by any of the 13 existing rules.

Specific rules become **fast-paths for known shapes**, not the safety mechanism. The safety mechanism is the generic probe + escalation chain.

## What the safety-net probes look like

Each is project-scoped, state-shape-agnostic:

- **Queue without motion** — project has queued tasks but zero claim / execution-advance / status-change events in N minutes
- **Churn without progress** — K tasks transitioned create→cancel without intermediate execution in N minutes
- **Alert past referent** — any open alert whose underlying entity (task, session, plan) is gone or terminal
- **Lane referenced by work but missing** — any queued task naming a role with no `<role>-<project>` lane in storage closet
- **Human-gate aging** — task at `user_approval` (or other human-owned node) parked > N days

[...]

PR #1552, opened 2026-05-09
Heartbeat cascade foundation: tier registry + safety-net probes + tier-3 dispatch + rejection-loop detector + product-broken plumbing (#1546)

samhotchkiss:
## Summary

Implements the foundation for the heartbeat cascade described in #1546 — tier registry, safety-net probes, tier-3 operator dispatch, rejection-loop detector, product-broken plumbing — and a minimal proof-of-life of each piece end-to-end. Sam will land specific rules in follow-up PRs against this stable foundation.

`Closes part of #1546`

## In-scope pieces

1. **`Finding.tier` field** + `TIER_*` constants (`1`, `2`, `3`, `4`, `terminal`). Every existing detector now sets a tier. Default is `TIER_TERMINAL` so a forgotten rule stays observe-only (the safe default).
2. **`_TIER1_HEALERS: dict[rule_name, Callable]` registry** in `audit_watchdog.py`. The pre-#1546 if/elif self-heal branches were replaced by registry dispatch. Existing 3 self-heal rules (`duplicate_advisor_tasks`, `plan_review_missing`, `legacy_db_shadow`) are byte-identical (registry adapters wrap the original helpers and translate counter shapes). New entries: `role_session_missing`, `state_db_missing`.
3. **Generic safety-net probe framework** — `ProbeContext` dataclass + `register_safety_net_probe(...)` decorator. Ships **one** probe end-to-end: `queue_without_motion` (queued tasks but no claim / execution / status-change activity for 30 min). Threshold tunable via `WatchdogConfig`.
4. **Rejection-loop detector** (NEW tier-1 detector → tier-2 dispatch). K=3 consecutive rejections at the same review node within W=2h with reject reasons clustered on shared error tokens. **Does NOT auto-cancel** — hands structured evidence to the project PM via the existing `_maybe_dispatch_to_architect` path.
5. **Tier-3 operator dispatch** — `_maybe_dispatch_to_operator(finding, ...)` symmetric to `_maybe_dispatch_to_architect`. Creates inbox task via the same plumbing `pm notify` uses (`store.enqueue_message` + `svc.create`), with `roles.operator='user'`, label `notify`, dedup key by `(rule, project, subject)`, throttled at 90 min via the audit log itself. Emits `EVENT_WATCHDOG_OPERATOR_DISPATCHED`.
6. **Product-broken first-class flag** — `workspace.product_state` row in `state.db` (new `workspace_state` key-value table). Helpers: `set_product_state_broken`, `get_product_state`, `clear_product_state`, `is_product_broken`. When set, new task queueing refuses with `ProductBrokenError` (gate hooked into `create_task`, raw-sqlite read on the hot path with mtime-keyed cache). `pm doctor` integration: new `check_product_state` reports healthy / broken in diagnostics. Auto-tripping is deferred (tier-4).
7. **Urgent-inbox structured-message helper** — `build_urgent_human_handoff(tried, failed, hypothesis, forensics_path) -> InboxItemBuilder`. Body opens with hypothesis (one paragraph), lists what was tried + what failed, ends with forensics path. Labels: `urgent`+`notify`, priority `immediate`. Not auto-triggered yet.
8. **One new tier-1 rule**: `state_db_missing` — registered project has no `.pollypm/state.db` (only `state.db.legacy-*` archives or nothing). Heals via `enable_tracked_project`. Audit event: `audit.project_tracked_mode_repaired`.
9. **Filter widening**: `_detect_role_session_missing` now includes `"queued"` (the coffeeboardnm cancel-loop class). Moved from `_DISPATCHABLE_RULES` to `_TIER1_HEALERS`. Heal action spawns the missing lane via `pm worker-start` plumbing (idempotent — re-uses existing session lookup). Audit event: `audit.worker_lane_spawned`. Skips `--role worker` (per-task workers are provisioned by `pm task claim`).

## Evidence-not-hypotheses principle

Audited `format_unstick_brief` for the `Options: (a) X, (b) Y, (c) Z` solution-menu pattern the issue calls out and refactored every offending branch. Briefs now hand structured evidence + a "generate hypotheses fresh from the evidence" prompt rather than a menu of pre-loaded options. CLI levers are listed as plain hints, not enumerated as a menu.

New `tier_handoff_prompt(evidence, question) -> str` helper for tier-2/3/4 dispatches. Structurally refuses solution-menu shapes:
- Function signature accepts only `evidence` + `question` — no `solutions` parameter.
- Evidence dict is rejected at runtime if it carries solution-menu keys (`options`, `solutions`, `recommended_actions`, ...).
- Lint test asserts the rendered prompt body has no `(a) ... (b) ...` pattern.

## Public API added

- `Finding.tier`, `Finding.evidence` fields on the dataclass.
- `TIER_1`, `TIER_2`, `TIER_3`, `TIER_4`, `TIER_TERMINAL` constants in `pollypm.audit.watchdog`.
- `tier_handoff_prompt(evidence, question)` and `SolutionMenuError` in `pollypm.audit.watchdog`.
- `build_urgent_human_handoff(tried, failed, hypothesis, forensics_path, ...)` and `InboxItemBuilder` in `pollypm.audit.watchdog`.
- `register_safety_net_probe(probe)` and `ProbeContext` in `pollypm.audit.watchdog`.
- `RULE_REJECTION_LOOP`, `RULE_STATE_DB_MISSING`, `RULE_QUEUE_WITHOUT_MOTION` rule constants.
- `OPERATOR_DISPATCH_THROTTLE_SECONDS`, `REJECTION_LOOP_THRESHOLD`, `REJECTION_LOOP_WINDOW_SECONDS`, `QUEUE_MOTION_THRESHOLD_SECONDS` constants.
- `emit_operator_dispatched(...)`, `was_recently_operator_dispatched(...)` in `pollypm.audit.watchdog`.
- `_TIER1_HEALERS` registry, `_maybe_dispatch_to_operator(...)`, `_StateDbProbe` dataclass in `pollypm.plugins_builtin.core_recurring.audit_watchdog`.
- `pollypm.storage.product_state` module: `ProductState`, `ProductBrokenError`, `PRODUCT_STATE_BROKEN`, `set_product_state_broken`, `get_product_state`, `clear_product_state`, `is_product_broken`.
- `StateStore.set_workspace_state`, `StateStore.get_workspace_state`, `StateStore.clear_workspace_state` (new key-value table backing the product-state plumbing).
- `EVENT_WATCHDOG_OPERATOR_DISPATCHED`, `EVENT_WATCHDOG_WORKER_LANE_SPAWNED`, `EVENT_WATCHDOG_PROJECT_TRACKED_MODE_REPAIRED` audit events.

## Deferred (out of scope per the issue)

- Tier-4 "do everything possible" mode for Polly (separate spec needed).
- Auto-tripping product-broken state.
- UI surfaces for product-broken (rail banner, dashboard takeover).
- Modifying detection logic of the 13 existing rules — only their classification + registry plumbing changed.
- Anything coffeeboardnm-specific.

## Spec calls Sam should sanity-check before merging

1. **Rejection-loop cluster predicate**: I picked `min_overlap=1` (a single shared symbol-bearing token like `better-sqlite3` or `node25` is enough to call a cluster). The issue says "exact-match not required; stem-overlap or shared error-token signal is enough". The stop-list filters generic words so single-token overlap is meaningful, but if false positives bite we can crank it to `min_overlap=2`.
2. **`Finding.tier` semantics**: I marked `rejection_loop` findings as `tier=TIER_2` (the dispatch tier) rather than `TIER_1` (the issue's "tier-1 detector" framing). The dispatcher branches on `tier`, so consistency with the dispatch leg felt right. Documented in the detector body.
3. **Operator throttle**: 90 min (1.5x the architect throttle). Tier-3 surfaces in the user's cockpit so I went wider than tier-2's 30 min. Tunable via `OPERATOR_DISPATCH_THROTTLE_SECONDS`.
4. **Product-broken bypass labels**: tasks tagged `watchdog`, `urgent`, or `notify` skip the create-task gate so the cascade itself can keep dispatching to the operator inbox even when the workspace is broken. If you want a stricter list, it's a one-line edit in `_PRODUCT_BROKEN_BYPASS_LABELS`.
5. **`role_session_missing` filter widening**: now fires for `queued` too. The coffeeboardnm `52-cancellation` cycle was the motivating case. Re-runs are idempotent because the existing-session lookup is preserved.
6. **`state_db_missing` heal calls `enable_tracked_project`**: the helper re-runs the scaffold step and re-detects project kind. Idempotent on already-healthy projects. If the missing-DB shape is a corruption (file deleted) rather than a never-scaffolded one, the heal should still work (the resolver creates the DB on first work-service open).
7. **Product-broken gate hot path**: `_enforce_product_state_gate` runs on every `create_task` call. To avoid spending tens of ms per call on a multi-GB workspace state.db, the gate uses raw `sqlite3` (read-only URI, immutable=1) plus an `mtime_ns`-keyed cache. Benchmarked at ~29 ms for the cold call and ~0.4 ms for cached calls.

## Test plan

- [x] Targeted scoped runs green:
  - `tests/test_audit_watchdog.py` — all 95 existing tests pass unchanged.
  - `tests/test_heartbeat_cascade_foundation.py` — all 38 new tests pass.
  - `tests/test_work_service.py` + `tests/test_doctor.py` + `tests/test_advisor_*.py` + `tests/test_chat_flow.py` etc. — pass.
  - 753-test scoped run across audit / watchdog / work_service / doctor / heartbeat / cascade / product all pass.
- [x] Pre-existing failures (verified failing on `origin/main` BEFORE this PR; deselected for the broader run):
  - `tests/test_cockpit_contract_boundaries.py::test_tmux_pane_window_methods_stay_inside_allowlist` — `list_panes` / `list_windows` allowlist drift in unrelated files (`cli_features/ui.py`, `task_assignment_notify.py`, `worker_marker_reaper.py`, the existing `_gather_storage_windows` in `audit_watchdog.py`).
  - `tests/test_cockpit_project_dashboard.py::test_render_project_dashboard_integration` — dashboard rendering string mismatch.
  - `tests/test_docs_command_references.py::test_live_doc_command_references_match_cli` — `web-api-spec.md` references `pm serve` / `pm api regen-token` (from `feat/web-api-phase1` parent branch).
- [x] Integration-style tests:
  - Rejection-loop with mocked tmux send-keys → architect dispatch.
  - State-db-missing → tier-1 healer invokes `enable_tracked_project` (mocked).
  - Operator dispatch with mocked inbox sink → audit event lands.
  - Lint test asserts every rendered tier-handoff prompt has no solution-menu language.
- [x] Idempotence: `_self_heal_role_session_missing` safe to call repeatedly.
- [x] Product-broken roundtrip: set/get/clear, refusal in `create_task`, bypass for watchdog labels.

### DIFF
--- a/src/pollypm/audit/log.py
+++ b/src/pollypm/audit/log.py
@@ -89,6 +89,29 @@ EVENT_PLAN_SUCCESSOR_CREATED = "plan.successor_created"
 # for the dedup to work.
 EVENT_WORKER_SESSION_REAPED = "worker.session_reaped"
 EVENT_WATCHDOG_ESCALATION_DISPATCHED = "watchdog.escalation_dispatched"
+# #1546 — operator-tier dispatch event. Fires when a tier-3 finding is
+# routed to the user's inbox via the ``pm notify``-shaped path. The
+# throttle window queries this event for dedup so a heartbeat-process
+# restart doesn't reset the per-rule cooldown. Symmetric to
+# ``EVENT_WATCHDOG_ESCALATION_DISPATCHED`` but for the operator leg of
+# the cascade — the leg that exists when a tier-2 architect dispatch
+# has already been tried (or is structurally unavailable) and the next
+# rung up is a human in the loop. The event metadata carries
+# ``finding_type``, ``subject``, and ``inbox_task_id`` (when the inbox
+# task creation succeeded) so forensic reads can correlate with the
+# inbox row directly.
+EVENT_WATCHDOG_OPERATOR_DISPATCHED = "watchdog.operator_dispatched"
+# #1546 — fires when the watchdog spawns a missing role lane (the
+# ``role_session_missing`` self-heal action). One event per spawned
+# lane; metadata carries ``role``, ``project``, and ``task_subject``
+# so forensic reads can confirm which queued / in-flight task drove
+# the spawn.
+EVENT_WATCHDOG_WORKER_LANE_SPAWNED = "audit.worker_lane_spawned"
+# #1546 — fires when the watchdog repairs a tracked project whose
+# canonical ``.pollypm/state.db`` is missing (only legacy archives
+# remain). Metadata carries ``project_key`` and ``project_path`` so
+# forensic reads can confirm which project was repaired.
+EVENT_WATCHDOG_PROJECT_TRACKED_MODE_REPAIRED = "audit.project_tracked_mode_repaired"
 # #1413 — emitted whenever the supervisor / on-demand path provisions a
 # project-scoped role session (e.g. reviewer-<project>) that did not
 # previously exist. The watchdog (#1414) reads this stream to surface
@@ -451,6 +474,9 @@ __all__ = [
     "EVENT_PLAN_SUCCESSOR_CREATED",
     "EVENT_WORKER_SESSION_REAPED",
     "EVENT_WATCHDOG_ESCALATION_DISPATCHED",
+    "EVENT_WATCHDOG_OPERATOR_DISPATCHED",
+    "EVENT_WATCHDOG_WORKER_LANE_SPAWNED",
+    "EVENT_WATCHDOG_PROJECT_TRACKED_MODE_REPAIRED",
     "AuditEvent",
     "central_log_path",
     "emit",
--- a/src/pollypm/audit/log.py
+++ b/src/pollypm/audit/log.py
@@ -101,6 +101,14 @@ EVENT_WATCHDOG_ESCALATION_DISPATCHED = "watchdog.escalation_dispatched"
 # task creation succeeded) so forensic reads can correlate with the
 # inbox row directly.
 EVENT_WATCHDOG_OPERATOR_DISPATCHED = "watchdog.operator_dispatched"
+# #1546 — fires when the operator-dispatch leg attempted an inbox
+# write but the write raised. Carries the same ``finding_type`` /
+# ``subject`` metadata as ``EVENT_WATCHDOG_OPERATOR_DISPATCHED`` so
+# the next tick can detect the failure case and not throttle on
+# behalf of a row that never landed. Distinct event so the throttle
+# query (which only counts ``operator_dispatched`` rows) doesn't
+# accidentally suppress retries.
+EVENT_WATCHDOG_TIER3_DISPATCH_FAILED = "watchdog.tier3_dispatch_failed"
 # #1546 — fires when the watchdog spawns a missing role lane (the
 # ``role_session_missing`` self-heal action). One event per spawned
 # lane; metadata carries ``role``, ``project``, and ``task_subject``
@@ -475,6 +483,7 @@ __all__ = [
     "EVENT_WORKER_SESSION_REAPED",
     "EVENT_WATCHDOG_ESCALATION_DISPATCHED",
     "EVENT_WATCHDOG_OPERATOR_DISPATCHED",
+    "EVENT_WATCHDOG_TIER3_DISPATCH_FAILED",
     "EVENT_WATCHDOG_WORKER_LANE_SPAWNED",
     "EVENT_WATCHDOG_PROJECT_TRACKED_MODE_REPAIRED",
     "AuditEvent",
--- a/src/pollypm/audit/watchdog.py
+++ b/src/pollypm/audit/watchdog.py
@@ -71,6 +71,7 @@ from pollypm.audit.log import (
     EVENT_TASK_CREATED,
     EVENT_TASK_STATUS_CHANGED,
     EVENT_WATCHDOG_ESCALATION_DISPATCHED,
+    EVENT_WATCHDOG_OPERATOR_DISPATCHED,
     EVENT_WORKER_SESSION_REAPED,
     AuditEvent,
 )
@@ -94,6 +95,14 @@ __all__ = [
     "RULE_PLAN_REVIEW_MISSING",
     "RULE_LEGACY_DB_SHADOW",
     "RULE_PLAN_MISSING_ALERT_CHURN",
+    "RULE_REJECTION_LOOP",
+    "RULE_STATE_DB_MISSING",
+    "RULE_QUEUE_WITHOUT_MOTION",
+    "TIER_1",
+    "TIER_2",
+    "TIER_3",
+    "TIER_4",
+    "TIER_TERMINAL",
     "ON_HOLD_HUMAN_NEEDED_TAG",
     "ON_HOLD_ARCHITECT_TAG",
     "scan_events",
@@ -101,11 +110,18 @@ __all__ = [
     "emit_heartbeat_tick",
     "emit_finding",
     "emit_escalation_dispatched",
+    "emit_operator_dispatched",
     "format_unstick_brief",
+    "tier_handoff_prompt",
+    "build_urgent_human_handoff",
     "watchdog_alert_session_name",
     "WATCHDOG_ALERT_TYPE",
     "format_finding_message",
     "ESCALATION_THROTTLE_SECONDS",
+    "OPERATOR_DISPATCH_THROTTLE_SECONDS",
+    "REJECTION_LOOP_THRESHOLD",
+    "REJECTION_LOOP_WINDOW_SECONDS",
+    "QUEUE_MOTION_THRESHOLD_SECONDS",
 ]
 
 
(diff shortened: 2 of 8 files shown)


samhotchkiss:
Codex review findings for PR #1552:

- High: `state_db_missing` checks `<project>/.pollypm/state.db` as canonical at `src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py:1604`, but the canonical DB after the workspace-root migration is `<workspace_root>/.pollypm/state.db`. Healthy projects without per-project DBs can alert every tick, and the healer at `:1304-1305` calls `enable_tracked_project()` rather than creating the per-project DB it is checking for. Please align detection and repair with the workspace-root DB resolver.

- High: The product-broken gate reads `DEFAULT_CONFIG_PATH` / `config.project.state_db` at `src/pollypm/work/service_queries.py:83-88`, while `create_task()` only calls `_enforce_product_state_gate(labels=labels)` at `:202`. This ignores the actual `SQLiteWorkService` DB path. A broken flag in the workspace DB may not block task creation, and a broken default/user DB can block an unrelated workspace. Please enforce against the service's actual DB/config.

- High: The product-broken gate caches only on the main DB mtime and opens SQLite with `immutable=1` at `src/pollypm/work/service_queries.py:99-117`. With WAL enabled, `workspace_state` updates can live in `state.db-wal`, so the gate can keep using stale healthy or stale broken state until checkpoint/main-file mtime changes. Please avoid immutable/WAL-blind reads or include WAL-aware invalidation.

- High: `role_session_missing` still only considers `worker`, `architect`, and `reviewer` at `src/pollypm/audit/watchdog.py:1401-1405`. Actual advisor-review tasks are created with `roles={"advisor": "advisor"}`, so the widened detector misses the production advisor lane. The new test uses a worker-shaped advisor role and does not match production. Please include the real advisor role shape in detection and tests.

- Medium: Tier-3 operator inbox dispatch ignores the loaded runtime config/workspace. `src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py:1498` calls `_resolve_db_path()` without passing the handler's config path or services context, so alternate-config heartbeat runs can write inbox tasks to the wrong DB while the audit log says dispatch succeeded. Please thread the active config/DB path into `_create_operator_inbox_task()`.

- Medium: The tier field is not actually driving dispatch. `queue_without_motion` is emitted as tier 2 (`src/pollypm/audit/watchdog.py:2331-2332`) but routed to operator/tier 3 by a rule-name set (`src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py:1768`), and tier 3 has no emitters. This will mislead UI/audit consumers that trust `Finding.tier`. Please make tier and dispatch policy agree.

- Medium: `rejection_loop` does not enforce consecutive rejections. `src/pollypm/audit/watchdog.py:2024-2055` filters rejected executions by node and takes the newest K, ignoring intervening approvals or progress. This can escalate tasks that are not actually stuck in a same-node rejection loop. Please require consecutive same-node rejection events or reset on progress/approval.

- Medium: `workspace_state` was added to the base schema but not as a numbered migration. Existing DBs opened read-only by doctor can hit `no such table: workspace_state` and report probe-read failures instead of healthy/no flag. Please add a migration or make readonly probes handle missing schema as no state.

- Medium: Operator dispatch emits the throttle audit event before inbox creation (`src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py:1431-1456`). If the inbox write fails, future ticks can throttle on an audit row with no inbox task and no evidence body. Please emit attempt/failure/success separately or throttle only after a successful inbox write.

Verification: `git diff --check origin/main...origin/feat/heartbeat-cascade-1546` passed in review.

samhotchkiss:
## Codex review resolutions (commit 3110a5c6)

**HIGH-1 — `state_db_missing` checks wrong path**: Probe now resolves the canonical DB through `pollypm.work.db_resolver.resolve_work_db_path` (the same source of truth `pm task list` uses) and reports presence of `<workspace_root>/.pollypm/state.db`, not the deprecated per-project location. The healer opens the canonical DB through `create_work_service` so the schema replay materialises the missing file. New test `test_state_db_missing_silent_for_project_without_per_project_db` guards the regression.

**HIGH-2 — Product-broken gate uses wrong DB path**: `_enforce_product_state_gate` now takes a `db_path=` argument; `create_task` threads `service._db_path` so the gate consults the workspace the service is actually bound to. New test `test_product_broken_gate_uses_service_db_path_not_global_default` shows workspace A (broken) blocks while workspace B (healthy) succeeds, even when global config points at B.

**HIGH-3 — Gate cache is WAL-blind**: Dropped `immutable=1` (the URI flag that tells SQLite to ignore the WAL file) and folded `*-wal` mtime into the cache key. New test `test_product_broken_gate_reflects_wal_writes_without_checkpoint` flips `workspace_state` and immediately re-runs the gate to assert the new state is visible without a checkpoint.

**HIGH-4 — `role_session_missing` still misses advisor**: Candidate-role list extended to include `advisor` for both `review` and the default branch. Verified production advisor task creation at `plugins_builtin/advisor/handlers/advisor_tick.py:262` uses `roles={"advisor": "advisor"}`. New tests `test_role_session_missing_fires_for_queued_advisor_task` and `test_role_session_missing_silent_when_advisor_lane_present` use the production role shape.

**MED-1 — Tier-3 dispatch ignores active config**: `_maybe_dispatch_to_operator` accepts `config_path` and forwards it to `_create_operator_inbox_task`, which now resolves the work DB via `resolve_work_db_path(config=loaded_cfg)` instead of `_resolve_db_path(".pollypm/state.db", project=...)`. Mirrors the architect dispatch's config plumbing.

**MED-2 — Tier field doesn't drive dispatch**: Bumped `queue_without_motion` to `TIER_3` (it's an operator-level concern by design and was already routed via `_OPERATOR_DISPATCHABLE_RULES`). The tier-mapping test and the integration tests were updated to expect `TIER_3`.

**MED-3 — `rejection_loop` doesn't enforce consecutive**: Detector now walks each node's rows newest→oldest and counts only the consecutive run at the head — any non-rejection (approval, status change at that node) breaks the run. New test `test_rejection_loop_silent_when_approval_breaks_run` covers reject → approve → reject ordering.

**MED-4 — `workspace_state` not in numbered migration**: Added migration 17 (`CREATE TABLE IF NOT EXISTS workspace_state ...`) and made `StateStore.get_workspace_state` catch `sqlite3.OperationalError` so read-only doctor opens against pre-#1546 DBs return `None` instead of raising. Two new tests: `test_workspace_state_migration_idempotent_on_old_db` and `test_get_workspace_state_tolerates_missing_table`.

**MED-5 — Operator dispatch emits throttle audit before inbox creation**: Reordered: inbox write happens FIRST, throttle-engaging `operator_dispatched` event only emits on success. Inbox failures emit a distinct `watchdog.tier3_dispatch_failed` event for forensics so the next tick re-attempts cleanly. New test `test_maybe_dispatch_to_operator_inbox_failure_does_not_throttle_next_tick` simulates a failure then a success and asserts the second call dispatches.

### Judgment calls
- HIGH-1 healer keeps a fallback to `enable_tracked_project` for the degenerate case where the resolver can't find a workspace DB. The primary path opens the workspace DB through `create_work_service`, which is idempotent.
- MED-2 chose to bump `queue_without_motion` to TIER_3 (per recommendation) rather than dispatching by `tier` field, because the field-driven contract still needs a separate codepath for the per-rule routing decisions (rejection_loop is "tier-2 architect", queue stalls are "tier-3 operator", etc.). The set-membership map remains the routing table; the tier field now matches it.
- HIGH-3 used WAL-mtime in the cache key instead of dropping the cache or forcing a checkpoint. Cheapest fix that respects the hot-path constraint.

### Test counts
- Heartbeat cascade tests: 38 → 47 passing (9 new).
- Audit watchdog tests: 95 unchanged passing.
- Targeted runs over `tests/test_work_service.py`, `tests/test_work_queries.py`, `tests/test_doctor_*` clean.

PR #1551, opened 2026-05-09
feat(web-api): Phase 1 — read endpoints + bearer auth + SSE (#1547)

samhotchkiss:
Closes #1547. Implements Phase 1 of the Web API spec (#1438; design in `docs/web-api-spec.md`, contract in `docs/api/openapi.yaml`).

## Summary

- New CLI command `pm serve --port N [--host H] [--allow-remote]` runs the FastAPI app as a peer to the cockpit.
- New CLI command `pm api regen-token` rotates the bearer token at `~/.pollypm/api-token`.
- All Phase 1 read endpoints from §8 of the spec, plus the SSE activity stream from §4.
- The API is cockpit-independent: every read goes through `pollypm.service_api.v1.PollyPMService` and `pollypm.work.factory.create_work_service` (#1389), so `state.db` and `audit.jsonl` stay the only sources of truth.

## In scope (Phase 1, per spec §11)

- `GET /api/v1/health` — liveness + version (no auth per §3).
- `GET /api/v1/projects` and `GET /api/v1/projects/{key}` — list + drilldown.
- `GET /api/v1/projects/{key}/tasks` — paginated task list (`?status=`, `?limit=`, `?cursor=`).
- `GET /api/v1/projects/{key}/plan` — structured plan body (summary, judgment_calls, body, critic_synthesis).
- `GET /api/v1/tasks/{project}/{n}` — task detail (transitions, executions, relationships, plan when applicable).
- `GET /api/v1/inbox` and `GET /api/v1/inbox/{id}` — list + thread detail.
- `GET /api/v1/events` — SSE stream with `?since=`, `?project=`, `?event=` glob, `Last-Event-ID` reconnect, and 15-second keep-alive comments.
- Bearer token at `~/.pollypm/api-token` (mode 0600, 256-bit url-safe). Auto-generated on first launch; `pm api regen-token` rotates and prints once.
- Errors follow §6 `{ "error": { code, message, hint? } }`; validation errors include `details[]` per §6.
- `/api/v1/openapi.json` exposes the runtime contract with `bearerAuth` declared.
- New runtime deps: `fastapi`, `uvicorn`. New dev deps: `httpx`, `openapi-spec-validator`.

## Deferred to later phases

- Phase 2 (#1548): write endpoints — approve, reject, queue, register, replan, chat, inbox reply / archive.
- Phase 3 (#1549): `GET /api/v1/doctor`, idempotency-key persistence, rate limiting, TLS guidance.

## Spec sections this implementation maps to

[...]

Issue #1547, opened 2026-05-09
[web-api] Phase 1 — Read endpoints + bearer auth + SSE


samhotchkiss:
Implementation Phase 1 of the Web API spec landed in #1438 (`docs/web-api-spec.md`, `docs/api/openapi.yaml`).

## Scope (quoted from `docs/web-api-spec.md` §11)

> ### Phase 1 — Read + auth + SSE
>
> - `pm serve` skeleton with FastAPI, bearer-token auth, OpenAPI doc at `/api/v1/openapi.json`, `pm api regen-token` CLI.
> - `GET /api/v1/health`
> - `GET /api/v1/projects`, `GET /api/v1/projects/{key}`
> - `GET /api/v1/projects/{key}/tasks`, `GET /api/v1/tasks/{project}/{n}`
> - `GET /api/v1/projects/{key}/plan`
> - `GET /api/v1/inbox`, `GET /api/v1/inbox/{id}`
> - `GET /api/v1/events` (SSE)
>
> Outcome: frontend can render the dashboard and stream live updates.

## Endpoint operations from the §8 endpoint table that fall in this phase

| Method | Path | Purpose |
|--------|------|---------|
| GET | `/api/v1/health` | Liveness + version (no auth) |
| GET | `/api/v1/projects` | List registered projects |
| GET | `/api/v1/projects/{key}` | Project drilldown |
| GET | `/api/v1/projects/{key}/tasks` | Task list (cursor-paginated) |
| GET | `/api/v1/projects/{key}/plan` | Structured plan body |
| GET | `/api/v1/tasks/{project}/{n}` | Task detail |
| GET | `/api/v1/inbox` | List inbox items |
| GET | `/api/v1/inbox/{id}` | Inbox item detail |
| GET | `/api/v1/events` | SSE stream (`?since=&project=&event=`) |

## Architectural musts

- New CLI command `pm serve --port N` runs a FastAPI app as a peer to the cockpit (separate process).
- Bearer token auth: token at `~/.pollypm/api-token` (mode `0600`); `pm api regen-token` rotates.
- All reads go through the canonical `pollypm.service_api.v1.PollyPMService` facade and `pollypm.work.factory.create_work_service` (#1389). No second source of truth.
- `pm serve` must work with the cockpit down — no tmux / supervisor / rail-daemon dependency.
- Errors follow §6 format: `{ "error": { "code", "message", "hint?" } }`.
- All paths under `/api/v1/...`.
[...]

Issue #1432, opened 2026-05-07
[regression] Stale rail_daemon processes accumulate across cockpit cycles


samhotchkiss:
## Repro

savethenovel watch session 2026-05-07: tick 5 found multiple rail_daemon processes alive concurrently:
- PID 85456 (current, 96.8% CPU)
- PID 39551 (stale, from prior boot)
- PID 96105 (stale, from prior boot)

These should be at most one. The cockpit's `pm reset --force && pm up --phantom-client` cycle isn't reaping the previous rail_daemon before spawning a new one.

## Earlier in this session

I (orchestrator) hit this exact pattern: stale rail_daemons from yesterday's test runs were still alive on pre-#1385 code, emitting JobWorkerPool warnings. I had to kill them manually to clean up.

## Expected

Bootstrap (or shutdown) should kill any rail_daemon processes belonging to a prior boot. Pattern: similar to worker-marker reaper (#1341) and socket reaper (#1387) — sweep stale processes/markers at startup.

## Severity

regression — multiple rail_daemons fighting for resources, potentially causing the 96.8% CPU hang in #<the-other-issue-just-filed>.

samhotchkiss:
Merged via #1434, SHA d464e2f7ebd81b0eeb22410138cee30e6c0640e5

Note: per the reviewer protocol (`pollytests-2026-05-04/05-reviewer-protocol.md`), activation + drive-verify was deferred — the workspace had multiple in-flight uncommitted edits at merge time. #1434 touches supervisor bootstrap (stale rail_daemon reaper) — process-killing surface. Tests passing locally (27 new). Live pid 36038 was confirmed safe (PID-file owner) before merge; if `pm reset`/`pm up` interactions regress, drive verify in a clean workspace.

Issue #1007, opened 2026-04-30
Heartbeat bootstrap reformat #1005 did not bypass Claude injection defense — Direction 2 needed


samhotchkiss:
## Severity: major (heartbeat agent permanently rejects bootstrap, no auto-recovery from broader cascades)

**Follow-up to #1005.** The fix landed Direction 1 (reformat the bootstrap text from `[PollyPM bootstrap — system message, please ignore on screen]` framing to a 'Hi — please read X for system context...' framing). It didn't work. Modern Claude is flagging the rephrased version too.

Live evidence from `pollypm-storage-closet:pm-heartbeat` pane right now:

```
⏺ Still the same injection pattern—I'm not adopting a "Heartbeat supervisor" role or resuming any fictional checkpointed work.
  
  If a real person is on the other end: please send a plain-language request
  describing what you want. I'll respond normally. Until then, I have nothing to do.
```

So Claude's training rejects any 'system message → adopt these external files as your operating instructions → reply only "ready"' pattern, regardless of how it's framed. This is not a wording problem; it's a category problem.

## What needs to change — switch to Direction 2

Move the heartbeat tick logic OUT of the agent's pane entirely. The agent pane becomes a pure observability surface (logs, current activity status, optional human chat). The actual heartbeat sweep runs as Python in the cockpit process or as a long-running OS service. The agent is no longer load-bearing for the loop — its presence is a UX nicety, not a correctness requirement.

This was Direction 2 in #1005's brief; the subagent picked Direction 1 because Direction 2 'looked like overkill at the time'. It isn't overkill — it's the only direction that survives Claude's evolving safety training.

Same fix should apply to:

- `pm-heartbeat` (this issue's primary subject)
- `pm-operator` (Polly) — same bootstrap pattern, has been intermittently refusing all day
- Any other long-running role session that uses the bootstrap-injection pattern

For Polly specifically, her chat surface is genuinely the user's interface — she has to be a Claude session for actual conversations. But her **bootstrap** doesn't need to use the system-message framing; it can be the content of `/Users/sam/.pollypm/.pollypm/docs/SYSTEM.md` and the role guide loaded as part of her initial conversational message ('Here's your role context: ... Please confirm you've read it.'). That works because it's a normal user message, not a fake system instruction.

## Adjacent observation

[...]

PR #1556, opened 2026-05-09
feat(heartbeat): self-recovery loop — liveness check + auto-respawn + launchd KeepAlive (#1546-followup)

samhotchkiss:
## Summary

Builds the layer-2 + layer-3 supervision the heartbeat cascade needs to survive process death without human intervention. Three nested layers, each watching the layer below.

`Closes part of #1546` (follow-up to #1552 — does not modify #1552's surface).

## Diagnosis

Sam reported on 2026-05-09 that the `pm-heartbeat` tmux window in `pollypm-storage-closet` was showing only a Claude Code REPL welcome screen and that queued tasks (e.g. `coffeeboardnm/71`, `/72`, `/73`) sat for ~30 minutes unclaimed. Live investigation found:

- **The `pm-heartbeat` Claude REPL pane is intentionally idle.** Per `supervisor.py:333-340` (#1007), the heartbeat tick loop runs as Python in `HeartbeatRail` (a daemon thread inside the cockpit/`rail_daemon` process). Bootstrapping the `pm-heartbeat` pane as a "Heartbeat supervisor" agent tripped Claude's prompt-injection defense (#1005, #1007), so the pane is now observability-only — a dormant REPL the user can use ad-hoc. Direction 2 from #1007: stop trying to bootstrap the pane.
- **The actual heartbeat (`pollypm.rail_daemon` pid 56615) was alive and ticking.** Diagnosed via the new `pm heartbeat status` command:
  ```
  rail_daemon: alive (alive)
    pid: 56615
    last tick: 3s ago
  ```
  `rail_daemon.log` showed continuous role-routing entries every ~30s; `last_heartbeat_at()` from `state.db` returned a timestamp seconds old. No crash signatures, no OOMs, no sandbox failures in `~/.pollypm/errors.log` for the last 24h.
- **What likely caused the queued-task stall** is a separate issue downstream of the heartbeat — maybe a stuck claim-row, a worker-marker race, or a dedupe-key wedge — but is **out of scope** for this PR. The user's principle ("when the recovery loop fails, fix the loop") still applies: this PR builds the recovery-loop-for-the-heartbeat that was missing, so the **next** time the rail daemon truly does die, it self-recovers without needing the user to notice.

The structural gap was real even though the symptom was misdiagnosed: nothing watched the rail daemon. If it had genuinely died (OOM, segfault, sandbox kill, `pm reset --force` racing the daemon's `atexit` handler), the cockpit would never have noticed and queued tasks would have piled up indefinitely. This PR closes that gap.

## Design — what watches the watchdog

Three nested layers, each watching the layer below:

| Layer | What it watches | What it can recover from | Where it lives |
|---|---|---|---|
| 1. **Tick-loop retry** (already existed) | individual `tick()` calls | one bad tick (SQLite WAL contention, plugin handler exception) | `HeartbeatRail._tick_loop` — try/except around each tick |
| 2. **External process supervisor** (NEW) | the rail daemon process | process-level death (OOM, segfault, kill -9) and stuck-but-alive (no tick for 180s) | cockpit periodic timer (60s) + `pm heartbeat` cron path |
| 3. **OS-level supervisor** (NEW, opt-in) | layers 1 + 2 | the cockpit + cron path being themselves dead; machine reboot | macOS LaunchAgent with `KeepAlive=true` |

The "what watches the watchdog" answer at each level:
- Layer 1's watcher is layer 2 — when the tick loop wedges (no tick for 180s), layer 2 kills + respawns.
- Layer 2's watcher is layer 3 — when the cockpit dies AND the cron job isn't installed, launchd respawns the daemon directly.
- Layer 3's watcher is the operating system — `KeepAlive` is OS-enforced.

Layer 2 is the load-bearing change. Layer 3 is opt-in because writing to `~/Library/LaunchAgents/` is a user-level action; the install command makes it explicit.

### Layer 2: `rail_daemon_supervisor`

`src/pollypm/rail_daemon_supervisor.py` (new). Two staleness signals checked from outside the daemon's process:

1. **Process-level**: `rail_daemon.pid` is missing or names a dead PID.
2. **Tick-level**: most recent `messages` row scoped to `heartbeat` is older than `stale_tick_seconds` (default 180s — the cadence sweep runs every 10s, so 180s = ~18 missed sweeps).

Action when dead/stuck:
1. SIGTERM the existing PID (only when `stuck_no_tick`; SIGKILL after 3s grace if needed).
2. Unlink stale `rail_daemon.pid`.
3. Spawn fresh daemon via the existing `cli._spawn_rail_daemon` (the daemon's own `_claim_pid_file` lock guards against doubled instances).
4. Emit `daemon.revived` audit event.

Throttled to one revival per 60s — without throttling, a daemon that dies-on-boot (e.g. broken plugin) gets respawned in a tight loop.

Wired into:
- **Cockpit periodic timer** (60s cadence). Captures the boot-time decision (`_rail_daemon_was_external`) so it only respawns when the cockpit deferred to an external daemon — never in cockpit-hosted-rail mode where respawning would race the cockpit's own ticker thread.
- **`pm heartbeat` cron path** (defense-in-depth — fires every minute even when the cockpit is dead).

### Layer 3: `rail_daemon_launchd`

`src/pollypm/rail_daemon_launchd.py` (new). Builds + installs `~/Library/LaunchAgents/com.pollypm.rail-daemon.plist` with:
- `KeepAlive: true` — restart on any exit.
- `RunAtLoad: true` — boot on login.
- `ThrottleInterval: 30` — backoff for crash loops.
- StandardOut/Err → `~/.pollypm/rail_daemon.log` (same file `cli._spawn_rail_daemon` writes to).

Activated via `pm heartbeat install-launchd` / `pm heartbeat uninstall-launchd`.

### Diagnostic surface

`pm heartbeat status [--json]` (new) — mirrors `diagnose_rail_daemon` so operators can see what the supervisor sees without taking action.

### Audit event

`daemon.revived` (new, symmetric to `daemon.reaped` from #1432). Carries `state`, `previous_pid`, `last_tick_age_s`, `revived`, `spawn_error`, `kill_signal`. Lets operators quantify chronic revivals — frequent revivals indicate a real problem (OOM, leak, crash loop) that needs investigation, not just trust in the supervisor.

## Public API added

- `pollypm.rail_daemon_supervisor`: `RevivalDecision`, `RevivalResult`, `diagnose_rail_daemon`, `check_and_revive_rail_daemon`, `revive_if_needed`, `query_last_heartbeat_iso`, `should_revive`.
- `pollypm.rail_daemon_launchd`: `build_plist_dict`, `install_launchd_keepalive`, `uninstall_launchd_keepalive`, `plist_path_for_label`.
- `pollypm.audit.log.EVENT_DAEMON_REVIVED` constant.
- CLI: `pm heartbeat status`, `pm heartbeat install-launchd`, `pm heartbeat uninstall-launchd`.
- `PollyCockpitApp._tick_rail_liveness`, `_resolve_rail_liveness_interval`, `_cockpit_uses_external_rail_daemon`. Tunable cadence via `POLLYPM_RAIL_LIVENESS_INTERVAL_SECONDS`.

## Test plan

- [x] **40 new tests** — all green:
  - `tests/test_rail_daemon_supervisor.py` (18) — diagnose purity, throttle behaviour, signal failure, spawn-error reporting, idempotence under throttle.
  - `tests/test_rail_daemon_launchd.py` (14) — plist shape, install/uninstall flow, missing `launchctl`, `--no-load` opt-out, returncode handling.
  - `tests/test_cockpit_rail_liveness_wiring.py` (8) — env-var override, garbage fallback, cockpit-hosted-rail skip, external-mode invocation.
- [x] **Existing rail-daemon family tests** still pass: 68 across `test_rail_daemon.py`, `test_rail_daemon_reaper.py`, plus the 32 new tests above.
- [x] **Cockpit / heartbeat tests** still pass: 323/324 in `test_cockpit.py` + `test_supervisor.py` + `test_supervisor_alerts.py` + `test_heartbeat_tick.py` (one pre-existing failure — see below).
- [x] **Live smoke**: `pm heartbeat status` against the running workspace correctly reports `pid=56615 alive, last heartbeat tick 3s ago` — confirms the diagnose path queries `messages` correctly.
- [x] **Manual revival simulation**: in `test_revive_calls_spawner_when_pid_missing`, a missing pid file triggers exactly one spawn-callback invocation; in `test_revive_idempotent_under_throttle`, a back-to-back call returns `throttled` without spawning.

### Pre-existing failures (verified failing on `origin/main` BEFORE this PR; deselected for the broader run)

- `tests/test_cockpit_contract_boundaries.py::test_tmux_pane_window_methods_stay_inside_allowlist` — `list_panes` / `list_windows` allowlist drift in `cli_features/ui.py`, `audit_watchdog.py`, `task_assignment_notify.py`, `worker_marker_reaper.py`. Same offenders as PR #1552 documented.
- `tests/test_packaged_worker_guide.py::test_packaged_worker_guide_matches_canonical` — `docs/worker-guide.md` and the packaged copy under `src/pollypm/defaults/docs/` have diverged on main.
- `tests/test_docs_command_references.py::test_live_doc_command_references_match_cli` — `docs/web-api-spec.md` references `pm serve` / `pm api regen-token` from PR #1551 / #1547.
- `tests/test_supervisor.py::test_bootstrap_returns_before_provider_stabilization_finishes` — also failing on `origin/main` in identical fashion.
- `tests/test_supervisor_alerts.py::test_supervisor_alert_helper_updates_and_nudges` — also failing on `origin/main` (worker nudge string drift; assertion compares stale "you appear stalled" string against the now-canonical "task X is queued" message).

## Out of scope

- **Why the queued tasks actually stalled.** The reported symptom (queued `coffeeboardnm/71`-`/73` sitting unclaimed for 30 min while the rail daemon was healthy) is a downstream issue — likely in the claim-loop, dedupe-key handling, or worker-marker provisioning, but **not** in heartbeat liveness. Worth a separate investigation issue.
- **The misleading dormant `pm-heartbeat` pane.** The Claude REPL screen confused the diagnosis — a future surface improvement could put a static "this pane is observability-only — heartbeat runs in `pollypm.rail_daemon` pid X" message in the pane to reduce confusion. Separate spec.
- **Auto-trigger of layer-3 install on first launch.** Currently opt-in via `pm heartbeat install-launchd`. Could be promoted to a launch-plan action with user consent, but that's a separate UX decision.
- **Detection of degraded-but-alive daemons.** A daemon whose plugin registry shrank silently (handler registration failures during init) still ticks but doesn't fire all roster handlers. That's a tier-2 detection problem (`audit_watchdog`), not a layer-2 supervision problem. The `daemon.revived` event will not catch it.
- **Linux/systemd equivalent of the launchd plist.** Layer 3 is macOS-only today. Sam's machine is macOS so this is fine for v1; a `systemd --user` unit can ship later.

## Spec calls Sam should sanity-check

1. **`stale_tick_seconds=180`**. The cadence sweep is `@every 10s`, so 180s = ~18 missed sweeps. If a busy plugin handler ever blocks the worker pool for >180s legitimately, layer 2 will kill the daemon mid-handler. I picked 180s as a balance between "respond fast to real wedges" and "tolerate slow handlers"; feel free to bump if a chronic false-positive bites.
2. **Throttle of 60s**. Pairs with the cockpit timer cadence (60s) — exactly one revival attempt per cockpit tick. The cron path's separate `_LAST_CRON_RAIL_REVIVAL_AT` module-level state is process-scoped (cron spawns fresh procs every minute, so it never accumulates a stale value).
3. **`KeepAlive: true` (always restart).** I went with the strongest "respawn on any exit" guarantee. The rail daemon's PID-file claim lock prevents doubled instances, so this should be safe — but if you want to gate restarts on a non-zero exit code only, swap to `KeepAlive: { SuccessfulExit: false }`.
4. **`_cockpit_uses_external_rail_daemon` is sticky.** Captured at boot time inside `_start_core_rail` and never re-evaluated. If the user runs `pm up` mid-cockpit-session to start a daemon manually, the cockpit's watcher won't pick it up until next restart. That's intentional (avoids double-ticker race) but worth confirming.
5. **`daemon.revived` carries the previous PID** even though that PID is now dead. Useful for forensics but means the `subject` field of the audit row references a defunct process. If you want a stable per-workspace subject, use `_workspace` instead.
6. **launchd `ThrottleInterval=30`.** macOS's documented minimum is 10s. I picked 30s so layer 3's restart cadence is faster than layer 2's 60s throttle for legitimate restarts, but slow enough that a crash-loop doesn't pin a CPU.

## Reference

- Issue #1546 — heartbeat cascade architecture.
- PR #1552 — tier registry, safety-net probes, tier-3 dispatch foundation. **Does not overlap** — that PR builds the cascade ladder (tier-1 / tier-2 / tier-3 / tier-4); this PR builds the layer-1 / layer-2 / layer-3 supervision that ensures the cascade has a process to **run in**.

### DIFF
--- a/src/pollypm/audit/log.py
+++ b/src/pollypm/audit/log.py
@@ -70,6 +70,16 @@ EVENT_SOCKET_REAPED = "socket.reaped"
 # can quantify how often the field machine is accumulating sibling
 # daemons without scraping ``rail_daemon.log``.
 EVENT_DAEMON_REAPED = "daemon.reaped"
+# #1546-followup — emitted by ``rail_daemon_supervisor`` when the
+# layer-2 watchdog (cockpit periodic timer or ``pm heartbeat`` cron
+# path) detects a dead/stuck rail daemon and respawns it. Carries
+# ``role`` (``"rail"``), ``state`` (the diagnose decision label),
+# ``previous_pid``, ``last_tick_age_s``, ``revived``, ``spawn_error``,
+# and ``kill_signal``. Lets operators quantify how often the layer-2
+# supervisor catches a dead daemon — chronic revivals indicate a real
+# problem (OOM, leak, crash-loop) that needs investigation, not just
+# trust in the supervisor.
+EVENT_DAEMON_REVIVED = "daemon.revived"
 # #1398 — plan task evolution. ``plan.version_incremented`` fires when
 # a plan task is refined in place (same task_id, version bump);
 # ``plan.successor_created`` fires when a replan creates a new task
@@ -486,6 +496,8 @@ __all__ = [
     "EVENT_WATCHDOG_TIER3_DISPATCH_FAILED",
     "EVENT_WATCHDOG_WORKER_LANE_SPAWNED",
     "EVENT_WATCHDOG_PROJECT_TRACKED_MODE_REPAIRED",
+    "EVENT_DAEMON_REAPED",
+    "EVENT_DAEMON_REVIVED",
     "AuditEvent",
     "central_log_path",
     "emit",
--- a/src/pollypm/cli.py
+++ b/src/pollypm/cli.py
@@ -1088,6 +1088,44 @@ def _spawn_rail_daemon(config_path: Path) -> None:
         )
 
 
+#: Module-level state for the cron-path rail-daemon liveness check.
+#: Persists across the (short-lived) ``pm heartbeat`` invocation only —
+#: cron spawns a fresh process every minute, so this never accumulates
+#: a stale value. Cockpit's analogous state lives on the App instance.
+_LAST_CRON_RAIL_REVIVAL_AT = None
+
+
+def _revive_rail_daemon_if_dead(config_path: Path) -> None:
+    """Spawn a fresh rail daemon when the existing one is dead/stuck.
+
+    Called from the ``pm heartbeat`` cron path as defense-in-depth.
+    The cockpit's periodic timer is the primary supervisor; this is
+    the second layer that fires even when the cockpit is itself dead.
+
+    Best-effort. Swallows all exceptions so a crashed liveness check
+    never breaks the heartbeat sweep that follows.
+    """
+    global _LAST_CRON_RAIL_REVIVAL_AT
+    try:
+        from pollypm.rail_daemon_supervisor import revive_if_needed
+        from pollypm.config import load_config
+
+        cfg = load_config(config_path)
+        pid_path = _rail_daemon_pid_path()
+        result = revive_if_needed(
+            config_path=config_path,
+            state_db_path=cfg.project.state_db,
+            pid_path=pid_path,
+            last_revival_at=_LAST_CRON_RAIL_REVIVAL_AT,
+        )
+    except Exception:  # noqa: BLE001
+        return
+    if result.revived:
+        from datetime import UTC, datetime as _dt
+
+        _LAST_CRON_RAIL_REVIVAL_AT = _dt.now(UTC)
+
+
 def _spawn_phantom_client(session_name: str) -> bool:
     """Spawn a detached pty-backed ``tmux attach`` client to keep the
     cockpit's input loop alive when no real terminal is attached (#1109).
--- a/src/pollypm/cli.py
+++ b/src/pollypm/cli.py
@@ -1117,6 +1117,12 @@ def _revive_rail_daemon_if_dead(config_path: Path) -> None:
             state_db_path=cfg.project.state_db,
             pid_path=pid_path,
             last_revival_at=_LAST_CRON_RAIL_REVIVAL_AT,
+            # #1556 review: cron must NOT spawn a daemon on missing_pid
+            # — the cockpit may be hosting the rail in-process. Cron
+            # still acts on dead_process / stuck_no_tick (those imply
+            # a daemon WAS recorded and lost, which doesn't happen in
+            # cockpit-hosted mode).
+            cron=True,
         )
     except Exception:  # noqa: BLE001
         return
--- a/src/pollypm/cli_features/alerts.py
+++ b/src/pollypm/cli_features/alerts.py
@@ -103,6 +103,15 @@ def heartbeat(
     from pollypm import cli as cli_mod
 
     supervisor = cli_mod._load_supervisor(config_path)
+    # #1546-followup — defense-in-depth rail-daemon liveness check.
+    # The cockpit's periodic timer is the primary watcher when it's
+    # open; this cron-driven path covers the case where the cockpit
+    # itself is dead so nothing else would notice a stale daemon.
+    # Best-effort: never blocks the heartbeat sweep.
+    try:
+        cli_mod._revive_rail_daemon_if_dead(config_path)
+    except Exception:  # noqa: BLE001
+        pass
     alerts = supervisor.run_heartbeat(snapshot_lines=snapshot_lines)
     tick_result = cli_mod._tick_core_rail_if_available(supervisor)
     cli_mod._drain_and_stop_core_rail_if_available(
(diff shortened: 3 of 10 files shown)


samhotchkiss:
Codex review findings for PR #1556:

- Critical: The supervisor can kill an unrelated process if a stale daemon PID has been reused. `_pid_is_alive()` at `src/pollypm/rail_daemon_supervisor.py:191` only checks `os.kill(pid, 0)`, `_classify_state()` treats any live PID with a stale tick as `stuck_no_tick` at `:300-311`, and `revive_if_needed()` sends SIGTERM/SIGKILL to that PID at `:516-517`. If the original daemon died and the OS reused the PID for another user process, this PR can terminate that unrelated process. Please verify PID identity before signaling, e.g. command/executable plus the active `--config` path, and treat unknown live PIDs as stale/dead without signaling. Add a reused-PID/non-daemon test.

- Critical: Concurrent revival can unlink a fresh daemon's PID file and start two daemons. `revive_if_needed()` diagnoses first, then unconditionally unlinks `pid_path` at `src/pollypm/rail_daemon_supervisor.py:550-552` before spawning at `:561-564`. If cockpit and cron both diagnose `missing_pid` or `dead_process`, one can spawn a daemon, then the other can unlink that fresh PID file and spawn another, bypassing the daemon's own PID-file claim guard. Please use a supervisor lock or compare-and-unlink only when the file still contains the diagnosed PID, and re-check immediately before spawning.

- High: Cron heartbeat can spawn a daemon while cockpit is intentionally hosting the rail in-process. `pm heartbeat` calls `_revive_rail_daemon_if_dead()` every run (`src/pollypm/cli_features/alerts.py:111-112`, `src/pollypm/cli.py:1113-1118`), and `missing_pid` can spawn a daemon. Cockpit has a guard for in-process rail hosting at `src/pollypm/cockpit_ui.py:1781-1784`, but the cron path does not. A user running cockpit-hosted mode plus the heartbeat cron can get duplicate rail tickers. Please make cron revival opt-in for expected headless-daemon mode, or have it skip `missing_pid` and only repair stale/dead PID files.

Verification: `git diff --check origin/main...origin/feat/heartbeat-self-recovery` passed during review.

samhotchkiss:
Codex review fixes pushed in 824fb7fb.

**CRITICAL-1 — PID identity verification (`rail_daemon_supervisor._pid_matches_daemon`)**
Before signaling, the supervisor now reads the recorded process's argv via `/proc/<pid>/cmdline` on Linux and falls back to `ps -p <pid> -o args=` on macOS. The match requires both the `pollypm.rail_daemon` module marker AND, when the daemon's argv carries `--config <path>`, that the path matches the supervisor's `config_path`. Mismatches classify as `dead_process` so the kill step is skipped entirely. `_terminate_with_grace` re-verifies identity before BOTH `SIGTERM` and `SIGKILL` (the grace window is another reuse opportunity). Daemons launched with no explicit `--config` still match — preserves compatibility with legacy spawns.

**CRITICAL-2 — concurrent revival race**
Wrapped the diagnose-recheck-kill-unlink-spawn critical section in a cross-process `fcntl.flock` on `<pid_path>.supervisor.lock` (with O_EXCL fallback for non-POSIX). Three layers of defense:
1. **Lock**: per-fd flock serializes cockpit + cron + manual `pm up`.
2. **Post-lock recheck**: re-runs `diagnose_rail_daemon` inside the lock — a sibling that held the lock while we waited may have already revived.
3. **Compare-and-unlink**: only delete the PID file if it still names the diagnosed PID; never clobber a fresh one.

Lock-busy callers return `RevivalResult(spawn_error="lock_busy")` and stand down.

**HIGH-1 — cron-vs-cockpit-hosted race**
Added `cron: bool = False` to `check_and_revive_rail_daemon`. When `cron=True`, `missing_pid` is downgraded to no-op (cockpit-hosted mode leaves no PID file by design). `dead_process` and `stuck_no_tick` still revive — those imply a daemon was recorded and lost, which never happens in cockpit-hosted mode. `cli._revive_rail_daemon_if_dead` now passes `cron=True`; the cockpit periodic timer continues with the full path (`cron=False`).

**Test coverage (all new)**
- Identity-mismatch live PID classifies as `dead_process` (`test_unrelated_live_pid_classified_as_dead_not_stuck`).
- `_terminate_with_grace` refuses signaling when identity mismatches (`test_revive_skips_signaling_when_pid_identity_mismatches`).
- Cmdline matcher: positive / negative / legacy-no-config / no-cmdline (`test_pid_matches_daemon_uses_cmdline`).
- Concurrent revival via `threading.Barrier` — deterministic, no sleeps — asserts exactly one spawn (`test_concurrent_revival_spawns_exactly_once`).
- Post-lock recheck stands down when sibling revived concurrently (`test_post_lock_recheck_stands_down_when_daemon_revived_concurrently`).
- Cron flag: skips `missing_pid`, still revives `dead_process`, doesn't regress non-cron callers (3 tests).

**Run**: `uv run --extra dev pytest tests/test_rail_daemon_supervisor.py tests/test_cockpit_rail_liveness_wiring.py tests/test_cli.py tests/test_cli_up_state_machine.py tests/test_rail_daemon.py tests/test_rail_daemon_launchd.py tests/test_rail_daemon_reaper.py -q` → **120 passed**.

samhotchkiss:
Codex re-review findings after commit `824fb7f`:

- Critical: concurrent revival is still racy. `check_and_revive_rail_daemon` releases the supervisor lock immediately after `spawn_fn` returns, but the real spawner is just `Popen` and returns before the child writes `rail_daemon.pid`. A second supervisor can acquire the lock, re-check `missing_pid`, and spawn again. The daemon's own PID claim is also check-then-write rather than atomic, so two children can both pass the missing-file check. Refs: `src/pollypm/rail_daemon_supervisor.py:812`, `:938`, `src/pollypm/cli.py:1074`, `src/pollypm/rail_daemon.py:47`.

Suggested fix: keep the supervisor lock until the spawned child has written a live matching PID, or make the daemon PID claim atomic with a shared flock / `O_EXCL` lock. Add a test where the spawner returns before writing the PID file.

- High: PID identity is safer for unrelated processes, but config identity is still too loose. `_pid_matches_daemon` accepts any `pollypm.rail_daemon` process with no `--config`, even when supervising a non-default config, and uses substring matching for explicit config paths. That can misidentify another workspace's daemon as this one and allow termination on stale ticks. Ref: `src/pollypm/rail_daemon_supervisor.py:291`.

Suggested fix: parse argv boundaries and compare resolved config paths exactly; only accept missing `--config` when the supervised config is the default config.

Verification: isolated PR archive passed `uv run --frozen --extra dev pytest tests/test_rail_daemon_supervisor.py tests/test_cockpit_rail_liveness_wiring.py tests/test_rail_daemon_launchd.py` (48 tests). A targeted timing probe with a Popen-like spawner that returns before writing the PID file produced two spawn calls.

samhotchkiss:
Round-2 review addressed in eaba2ac0.

**CRITICAL — concurrent revival race (lock released before child writes PID):**
Did **both** approaches recommended by the review.
1. Lock-held-during-startup: new `_wait_for_child_pid(pid_path, expected_config, timeout=5.0s)` polls the PID file and verifies via `_pid_matches_daemon` *inside* the supervisor lock right after `spawn_fn` returns. On timeout we log and release anyway (never deadlock). Default `child_wait_seconds=5.0` (`DEFAULT_CHILD_PID_WAIT_SECONDS`) — covers cold-import latency on the slowest field machines (~1-2s) without making cron callers exceed their tick budget.
2. O_EXCL atomicity in the daemon: `rail_daemon._claim_pid_file` now uses `os.open(..., O_WRONLY | O_CREAT | O_EXCL, 0o644)` so two children that both pass the existence check still see exactly one winner. Loser hits `FileExistsError` and exits cleanly with the existing "another daemon already running" log.

New regression test (`test_concurrent_revive_when_spawner_returns_without_writing_pid`): two threads, A's spawner does *not* write the PID file, A holds `child_wait_seconds=1.0`, B has `lock_timeout_seconds=0.3`. Asserts B sees `lock_busy` and never invokes its spawner — the central guarantee that even a buggy spawner can't double-spawn.

New atomicity test (`test_daemon_pid_claim_is_atomic`): 8 threads racing into `_claim_pid_file` → exactly 1 winner.

**HIGH — config identity matching too loose:**
Rewrote `_pid_matches_daemon` to:
1. `shlex.split` argv (boundary-correct), match the entry point on argv tokens (not joined-cmdline substring).
2. New `_extract_config_arg` helper handles both `--config <path>` and `--config=<path>` forms.
3. Compare resolved `Path.resolve(strict=False)` values element-by-element — substring collision (e.g. `/x/pollypm.toml` substring inside `/other/contains-x-pollypm.toml-decoy.yaml`) is now rejected.
4. Missing `--config` only matches when our supervisor's own `config_path` resolves to `DEFAULT_CONFIG_PATH`.

New tests: `test_non_default_supervisor_does_not_match_default_config_daemon`, `test_config_match_uses_argv_boundary_not_substring`, `test_config_match_handles_equals_form`. Old "no-config daemon accepted unconditionally" assertion in `test_pid_matches_daemon_uses_cmdline` updated to the stricter behavior.

**Test results:** `pytest tests/test_rail_daemon_supervisor.py tests/test_cockpit_rail_liveness_wiring.py tests/test_rail_daemon_launchd.py -q` → **53 passed** (48 baseline + 5 new). Wider rail-related sweep (incl. `test_rail_daemon.py`, `test_rail_daemon_reaper.py`) → 89 passed.

Issue #1544, opened 2026-05-09
Project dashboard rejects rail-listed project: "health-coach is not a tracked project — open the project picker to choose"


samhotchkiss:
## Symptom

The rail lists `health-coach` with the same active heart glyph used for every other healthy project:

```
   ── projects ────────
   ♥⚠ CoffeeBoardNM
   ♡· bikepath
   ♥· booktalk
   ♥· camptown
   ♥· health-coach          ← here
   ♡· itsalive
   ○ media
   ♥· polly-remote
   ♥· pollypm
   ♥· russell
   ◆ savethenovel
   ♡· smoketest
```

Render its dashboard with `pm cockpit-pane project health-coach` and the surface refuses with this header line:

```
  health-coach   is not a tracked project — open the project picker to choose

  ────────────────────────────────────────────────────────────────────────────

  ╭──────────────────────────────────────────────────────────────────────────╮
  │                                                                          │
  ╰──────────────────────────────────────────────────────────────────────────╯

  ╭─────────────────────────────────────────────────────────────────────────╮
  │                                                                         │
  │  Inbox                                                                  │
  │                                                                         │
  ╰─────────────────────────────────────────────────────────────────────────╯
```

The banner is empty. The Inbox is empty. The dashboard renders the scaffolding but bails on content with an internal-vocabulary error — "not a tracked project" — and tells the user to "open the project picker to choose" without explaining what to choose, why, or how.

## Why this is fix-tier (not polish)

Two issues stack:

### 1. Rail/dashboard contradict on the same project key

The rail (`pollypm:0.0`) lists `health-coach` as an active heart-with-dot project — same glyph as 6 other healthy projects in the list. A user who navigates to it and is told "this is not a tracked project" cannot reconcile those two statements. One of the surfaces is lying.

Per the existing tracked-filter asymmetry pattern (rail badge + cockpit inbox include paused; recovery / briefing filter to tracked), the project dashboard's filter behavior isn't documented, but currently it's diverging from rail in a way that **breaks navigation**. A user clicking a rail entry should not be bounced to a picker.

### 2. The error copy uses internal vocabulary and offers no real action

`is not a tracked project — open the project picker to choose`

- **"tracked project"** is internal-data-model terminology. A user has no way to know what makes a project tracked vs not.
- **"open the project picker to choose"** — choose what? From where? No keystroke hint, no link. Sounds like a workflow that doesn't exist.

[...]

Issue #1542, opened 2026-05-09
media project: rail glyph lies, PM name shows placeholder, and paused-task list has broken text wrap


samhotchkiss:
## Symptom

`media` is a project with a paused root task (Library-wide music dedupe). Open it from the rail and three separate things look wrong on one surface.

### 1. Rail glyph contradicts dashboard state

Rail entry: `○ media`

Dashboard for the same project:

```
  media   PM: Project PM

  ────────────────────────────────────────────────────────────────────────────
  ◆ needs attention
  ╭──────────────────────────────────────────────────────────────────────────╮
  │ Paused: 1 task is on hold                                                │
  ╰──────────────────────────────────────────────────────────────────────────╯
```

The rail uses `○` for `media` (same glyph it uses for `Home` and `Workers (14)` — i.e., "neutral / no chat" or similar). But the dashboard for this same project says `◆ needs attention` with a paused root hold blocking downstream work. **Rail says quiet, dashboard says please-look-at-me.**

A user scanning the rail looking for "what needs me right now" would skip past `media` because the glyph signals quiet. They'd miss a paused task that the dashboard itself is flagging as needs-attention.

(For comparison: `savethenovel` has a "no plan yet" state and shows `◆` in the rail; `coffeeboardnm` has alerts and shows `◆` / `♥⚠` / `♡⚠` across ticks. So the rail glyph IS dynamic — it's just not picking up paused tasks.)

### 2. PM name shows the placeholder string `Project PM`

Header: `media   PM: Project PM`

Compare:
- `coffeeboardnm` → `PM: Cole`
- `bikepath` → `PM: Archie`
- `savethenovel` → `PM: Archie`
- `media` → `PM: Project PM`

`Project PM` reads as a placeholder string that leaked into the UI rather than a real PM name. Either:
- The project never got a named PM assigned (and the fallback string is hideous), or
- The user actually named the PM "Project PM" (in which case, fine — but ask whether that's likely)

For the entire product to feel alive, every project needs a named PM with personality. `media` showing `Project PM` undoes that for this project specifically.

### 3. Paused-task panel has broken indentation on wrap

```
  ╭─────────────────────────────────────────────────────────────────────────╮ 
  │                                                                         │ 
  │  Inbox                                                                  │▅
  │                                                                         │ 
  │  On hold                                                                │ 
  │    These are the root holds keeping downstream work waiting.            │ 
[...]

Issue #1519, opened 2026-05-08
Project Dashboard task counts double-count by unioning canonical + legacy per-project DBs


samhotchkiss:
## Symptom

For `coffeeboardnm`, the dashboard "Task pipeline" line reads:

> ✎ 0 draft · ○ 6 queued · ◆ 0 in progress · ◉ 0 review · ▣ 0 blocked · ⏸ 0 on hold · ✓ 0 done

But the canonical workspace DB (`/Users/sam/dev/.pollypm/state.db`) actually has:

```
cancelled    26
done          1
draft         2
queued        1
```

Total non-cancelled: 4. Pipeline shows 6. The extra 5 come from the legacy per-project DB (`/Users/sam/dev/coffeeboardnm/.pollypm/state.db`), which still holds 5 stale `queued` advisor_review rows.

## Root cause

`src/pollypm/cockpit_ui.py:11857` `_dashboard_task_db_paths` returns BOTH the workspace canonical DB and the per-project legacy DB. The docstring acknowledges the dual return:

> Order (#1004): workspace-root canonical DB first, per-project `<project>/.pollypm/state.db` as a legacy fallback. Must mirror `pollypm.cockpit_tasks.PollyTasksApp._candidate_dbs` — a divergence here surfaces as split-brain rendering between the dashboard (which unions across DBs) and the Tasks pane (which picks the first DB with matching rows).

`_dashboard_gather_tasks` (cockpit_ui.py:11981) then iterates every returned path and unions the results. So canonical + legacy rows are summed, producing inflated counts. #1470 / #1004 noted this and made other call sites prefer-canonical, but the dashboard's gather still unions.

## Expected

The dashboard should mirror the Tasks pane's "first DB with matching rows" semantics — or, better, the watchdog should auto-migrate stale per-project legacy DB rows to canonical and the dashboard should only ever read canonical. (The existing `pollypm.storage.legacy_per_project_db.migrate_legacy_per_project_dbs` helper exists but isn't fired automatically.)

Two-part fix:

1. **Dashboard read path**: prefer canonical and stop at the first DB with matching rows for the project, matching `_candidate_dbs`.
2. **Self-healing**: a watchdog rule that detects legacy per-project DB rows for projects also present in canonical, and either migrates them or alerts. (Mirrors the #1510 / #1511 self-healing pattern.)

## Related

- #1004 / #1470 / #1471 (canonical-vs-legacy resolver work)
[...]

PR #1557, opened 2026-05-09
fix(rail+dashboard): rail glyph reflects paused-with-work; dashboard renders non-tracked projects

samhotchkiss:
## Summary

Two contradictions on the same project key, fixed together because they live on the rail/dashboard surface the user navigates between.

**#1542 — rail glyph mismatch (paused-with-work).** The rail rollup only inspected the legacy per-project DB at `<project>/.pollypm/state.db`. The dashboard's task gather walks the canonical workspace-root DB first (post-#1519). A project with a paused `on_hold` task in the canonical DB rolled up to `ProjectRailState.NONE` and rendered as `○ media`, while the dashboard for the same project correctly painted `◆ needs attention`. Rail says quiet, dashboard says please-look-at-me.

Fix: mirror the dashboard's DB-resolution order in the rollup. Walk workspace-root first, fall back to per-project legacy, union the rows. Existing `◆ project-yellow` glyph for `_is_waiting_on_user` tasks now fires for canonical-DB-only projects too.

**#1544 — dashboard refuses rail-listed project.** `pm cockpit-pane project health-coach` bounced to `health-coach   is not a tracked project — open the project picker to choose a tracked project`. Two stacked failures:

1. The lookup did a single `config.projects.get(project_key)` with no slug-normalization, so the rail's hyphenated label form silently missed the underscore-keyed config entry (`health_coach`).
2. The error copy then used internal-data-model jargon (`tracked`) and pointed at a non-existent affordance (`project picker`).

Fix:
- New helper `_resolve_project_key` resolves the lookup through three escalating attempts: direct match, slug-normalized match (via `slugify_project_key`), and case-folded scan. Lands on the right row whether the user types `health-coach` / `Health-Coach` / `health_coach`.
- Bail-state topbar copy now reads "no project registered with that key — run **pm project new <path>** to register one, or press **q** to go back" — plain language, real CLI command, real keystroke.

## Design choice — Option A (matches rail)

Sam's #1544 issue note suggested two options:
- **Option A:** dashboard renders for any project the rail shows.
- **Option B:** rail visually mutes non-tracked projects.

I chose **A**. The rail intentionally includes paused / non-tracked projects per the existing "Tracked-Filter Asymmetry" memory pattern (rail badge + cockpit inbox panel include paused; recovery / briefing helpers filter to tracked). The dashboard's bail was a separate key-resolution bug, not a tracked-filter mismatch — once the slug normalization lands, the dashboard renders for any project the rail shows, including `tracked=False` rows (the gather has never filtered on the tracked flag). Option B would lose the rail's information density.

## Watchdog self-heal — deferred

Sam's #1544 note suggests a "rail-listed project marked non-tracked → re-mark or remove from rail" rule. After investigation: that rule contradicts the existing "Tracked-Filter Asymmetry" memory pattern (rail INTENTIONALLY includes paused/non-tracked). The slug-normalization fix lands on the right config row regardless of the tracked flag, so the rule has no concrete trigger today. Documented in the commit message; will revisit if a real demote-by-mistake sweep shows up.

## Test plan

- [x] `tests/test_cockpit_rail_rollup_db_resolution.py` — 4 tests covering workspace-root-only / legacy-only / both-empty cases + the YELLOW rollup end-to-end.
- [x] `tests/test_project_dashboard_key_resolution.py` — 8 tests covering direct / slug-normalized / case-insensitive / combined lookup + the bail-copy invariants.
- [x] Existing rail / dashboard test sweeps green: `test_cockpit_project_state.py`, `test_cockpit_rail_*.py`, `test_cockpit.py`, `test_core_rail*.py` (277 tests).
- [x] Pre-existing flakes in `test_project_dashboard_ui.py` (7 failures) reproduce on a clean checkout of `main`; not introduced by this PR.

Refs #1542 #1544

### DIFF
--- a/src/pollypm/cockpit_rail.py
+++ b/src/pollypm/cockpit_rail.py
@@ -1793,56 +1793,123 @@ class CockpitRouter:
         project_path = getattr(project, "path", None)
         if not isinstance(project_path, Path):
             return [], False
-        db_path = project_path / ".pollypm" / "state.db"
-        if not db_path.exists():
-            return [], False
         from pollypm.plan_presence import has_acceptable_plan
         from pollypm.work import create_work_service
         from pollypm.work.project_aliases import project_storage_aliases
 
-        work = create_work_service(db_path=db_path, project_path=project_path)
-        try:
-            # #1092 — match the dashboard's alias-aware lookup. The work
-            # DB stores tasks under the display name (``booktalk``) while
-            # the rollup receives the slugified config key, so a single
-            # ``list_tasks(project=project_key)`` silently returned zero
-            # tasks for projects whose key and display name diverge.
-            # The dashboard correctly counts ``on_hold`` via the alias
-            # union — without it here, the rail rollup falls through to
-            # ``ProjectRailState.NONE`` and the held-task project shows
-            # up in the rail with the idle ``♥·`` glyph (no marker)
-            # instead of the ``◆`` "needs attention" indicator.
-            aliases = project_storage_aliases(config, project_key)
-            seen_ids: set[str] = set()
-            tasks: list = []
-            for alias in aliases:
-                for task in work.list_tasks(project=alias):
-                    tid = getattr(task, "task_id", None)
-                    if tid and tid in seen_ids:
-                        continue
-                    if tid:
-                        seen_ids.add(tid)
-                    tasks.append(task)
-            planner = getattr(config, "planner", None)
-            plan_dir = str(getattr(planner, "plan_dir", "docs/plan") or "docs/plan")
-            global_enforce = bool(getattr(planner, "enforce_plan", True))
-            project_enforce = getattr(project, "enforce_plan", None)
-            enforce_plan = (
-                project_enforce if project_enforce is not None else global_enforce
-            )
-            if not enforce_plan:
-                return tasks, False
-            plan_blocked = bool(tasks) and not has_acceptable_plan(
-                project_key,
-                project_path,
-                work,
-                plan_dir=plan_dir,
-            )
-            return tasks, plan_blocked
-        except Exception:  # noqa: BLE001
+        # #1542 — mirror the dashboard's DB-resolution order. Pre-#1542
+        # the rail rollup only looked at the legacy per-project DB
+        # (``<project>/.pollypm/state.db``) — the dashboard's task
+        # gather (``_dashboard_task_db_paths``) walks the workspace-root
+        # canonical DB FIRST and falls back to the per-project legacy
+        # DB only if the canonical query came up empty. A project whose
+        # tasks live in the canonical workspace DB (the post-#1519 norm)
+        # rolled up to ``ProjectRailState.NONE`` here, even though the
+        # dashboard correctly painted ``◆ needs attention`` from the
+        # same data. The result was the rail glyph (``○``) and dashboard
+        # banner contradicting on the same project key — cf. #1542's
+        # ``media`` repro: rail says quiet, dashboard says please-look-
+        # at-me. We now walk both DBs in the canonical-first order and
+        # union the rows so the rail picks up paused-with-work projects.
+        candidate_db_paths: list[Path] = []
+        seen_paths: set[Path] = set()
+
+        def _add_candidate(path: object) -> None:
+            try:
+                candidate = Path(path)
+            except TypeError:
+                return
+            if candidate in seen_paths:
+                return
+            try:
+                if not candidate.exists():
+                    return
+            except OSError:
+                return
+            seen_paths.add(candidate)
+            candidate_db_paths.append(candidate)
+
+        project_settings = getattr(config, "project", None)
+        workspace_root = getattr(project_settings, "workspace_root", None)
+        if workspace_root is not None:
+            _add_candidate(Path(workspace_root) / ".pollypm" / "state.db")
+        _add_candidate(project_path / ".pollypm" / "state.db")
+        state_db = getattr(project_settings, "state_db", None)
+        if state_db is not None:
+            _add_candidate(state_db)
+
+        if not candidate_db_paths:
             return [], False
-        finally:
-            work.close()
+
+        aliases = project_storage_aliases(config, project_key)
+        planner = getattr(config, "planner", None)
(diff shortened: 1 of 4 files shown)


samhotchkiss:
Codex review findings for PR #1557:

- High: `_project_tasks_for_rollup()` now unions rows from every candidate DB, which reintroduces stale legacy DB split-brain in the rail glyph. The loop walks every DB at `src/pollypm/cockpit_rail.py:1857`, appends rows at `:1883-1885`, and never stops after the canonical workspace DB yields matching rows. The dashboard/task pane contract is first DB with matching rows wins, so a clean canonical workspace DB can still get padded by stale legacy per-project `on_hold` rows and show a needs-attention glyph. Please mirror `_dashboard_gather_tasks()`: stop after the first DB that yields matching task rows, and only fall back when it yields none. Add a regression where canonical has rows and legacy has stale `on_hold` rows.

- High: Project-key resolution updates the gathered data but not the app's effective routing key. `_gather_project_dashboard()` resolves a hyphenated key to the canonical key at `src/pollypm/cockpit_ui.py:14117-14120`, but `PollyProjectDashboardApp.self.project_key` remains the original input. Follow-on actions still route through `self.project_key`, for example PM chat at `:16978-16980` and inbox jump at `:17244-17247`. `pm cockpit-pane project health-coach` can therefore render data for `health_coach` but route `c`, `i`, and task jumps back through `health-coach`. Please route with `data.project_key` after load or update the app's effective project key, and add a hyphen-input navigation/action test.

Verification: PR tests from a temp archive passed (`12 passed`), and a custom stale-DB check confirmed the current rollup returns both canonical and legacy rows.

samhotchkiss:
Addressed both Codex HIGH findings in 06864c67:

**HIGH-1 (`_project_tasks_for_rollup` legacy split-brain)** — Replaced the union-every-DB loop with a "first DB with matching rows wins" walk that mirrors `_dashboard_gather_tasks`. The break is placed after the work-service `finally` so the legacy candidate never reopens once the canonical DB has yielded rows. Regression test (`test_rollup_first_db_with_rows_wins_no_legacy_padding`): canonical DB has one queued task, legacy DB has a stale `on_hold` row, rollup surfaces only the canonical row.

**HIGH-2 (project-key not adopted after resolver)** — Both `_refresh` and `_first_refresh_completed` now copy `data.project_key` back onto `self.project_key` once the gather succeeds, so PM chat, inbox jump, and task jumps all route through the canonical key after a hyphenated/case-folded input resolves. Bail path (`data is None`) leaves the original input intact so the error copy can still name the missing key. Regression tests cover key adoption + actual `jump_to_inbox` / `jump_to_project` routing call args.

Tests: `tests/test_cockpit_rail_rollup_db_resolution.py` + `tests/test_project_dashboard_key_resolution.py` (38 tests, all green). Full `test_project_dashboard_ui.py` was run; the 33 failures it surfaces are pre-existing Textual-harness `data is None` flakes on `bef7751c` (verified by re-running with the fixes stashed) and unrelated to this PR.

samhotchkiss:
Codex re-review finding after commit `06864c6`:

- High: the rollup's first-DB-with-rows gate still counts notification-shaped task rows as real work. The dashboard filters notify/FYI rows with `is_notify_inbox_task` before deciding whether a DB has pipeline rows, but `_project_tasks_for_rollup()` stops on canonical DB rows even when they are notify-only. Refs: `src/pollypm/cockpit_rail.py:1881`, `:1919`; dashboard comparison at `src/pollypm/cockpit_ui.py:12283`, `:12367`.

Impact: canonical DB with only notify/FYI rows plus a legacy DB with real paused work makes the dashboard fall back and show `on_hold`, while the rail stops at canonical and rolls up `WORKING`.

Suggested fix: import/use `is_notify_inbox_task` in `_project_tasks_for_rollup`, skip notify rows before adding to `seen_ids` / `db_tasks`, and keep the `if db_tasks: break` gate based on filtered rows.

Verification: targeted PR tests passed in an isolated archive (`17 passed`). A focused reproduction confirmed rail saw only a notify row and returned `ProjectRailState.WORKING` while dashboard counts showed `{'on_hold': 1}`.

samhotchkiss:
## Round 2 — HIGH: notify-only canonical DB latched the gate

Pushed `29e3fd7d` addressing Codex's re-review finding.

**Root cause:** `_project_tasks_for_rollup`'s "first DB with rows wins" gate counted notification-shaped rows (rejection feedback, plan_review handoff, supervisor alerts, …) as real work, while the dashboard's `_dashboard_gather_tasks` filters them with `is_notify_inbox_task` BEFORE deciding whether a DB has pipeline rows. Canonical DB with only notify rows + legacy DB with real paused work → dashboard falls back to legacy (`on_hold`), rail stops at canonical and rolls up `WORKING`. Mismatched glyph and dashboard.

**Fix:** mirror the dashboard exactly. Skip notify rows BEFORE the gate and `seen_ids`. Notify tids are intentionally NOT added to the dedup set so a same-tid real-work row in a later DB can still surface (defensive against legacy/canonical drift).

```python
for task in work.list_tasks(project=alias):
    tid = getattr(task, "task_id", None)
    if tid and tid in seen_ids:
        continue
    if is_notify_inbox_task(task):   # NEW — mirrors cockpit_ui:12283
        continue
    if tid:
        seen_ids.add(tid)
    db_tasks.append(task)
```

The trailing `if db_tasks: break` gate is now structurally equivalent to "filtered rows only" since notify rows never reach `db_tasks`.

**Test added:** `test_rollup_falls_through_canonical_with_only_notify_rows` — seeds canonical with ONLY a notify-shaped row (`roles.operator = "user"` + `notify` label, `chat` flow), legacy with a real `on_hold` task, asserts rollup falls through to legacy (`on_hold` in statuses, notify title excluded) AND end-to-end the rail rolls up to `YELLOW` not `WORKING`.

**Tests run:**
- `tests/test_cockpit_rail_rollup_db_resolution.py` (6 passed, including new regression)
- `tests/test_notify_task.py` (17 passed)
- `tests/ -k 'rail and not dashboard'` (268 passed in 287s)
- The `dashboard and not rail` scope hangs at the same point on the unmodified base commit, so the failure is pre-existing and unrelated.

Issue #1539, opened 2026-05-09
Project dashboard loading state shows empty boxes for ~10s — feels broken


samhotchkiss:
## Symptom

When entering a project dashboard (observed: `coffeeboardnm`), the surface shows the header `coffeeboardnm   loading project dashboard…` followed by three completely empty rounded rectangles (banner, Inbox, Current activity) for roughly 8–12 seconds before content actually populates.

```
  coffeeboardnm   loading project dashboard…

  ────────────────────────────────────────────────────────────────────────────

  ╭──────────────────────────────────────────────────────────────────────────╮
  │                                                                          │
  ╰──────────────────────────────────────────────────────────────────────────╯

  ╭─────────────────────────────────────────────────────────────────────────╮
  │                                                                         │
  │  Inbox                                                                  │
  │                                                                         │
  │                                                                         │
  │                                                                         │
  ╰─────────────────────────────────────────────────────────────────────────╯

  ╭─────────────────────────────────────────────────────────────────────────╮
  │                                                                         │
  │  Current activity                                                       │
  │                                                                         │
```

The Inbox and Current activity panels even keep their headings during the loading state, which makes them look like they've **finished loading and have nothing in them** — until content suddenly appears 10s later. Mixed signal.

## Why it matters

A fresh user opening Polly cold will spend their first 10 seconds in a project staring at empty boxes that look populated-but-empty, with a small grey "loading…" hint at the top they may not notice. That's the worst kind of latency feel: not "obviously loading" and not "obviously empty" — just neutral and a little broken.

This violates Sam's friction→latency→magic priority. Cold open should feel **alive**, not blank.

## What would feel better

A few options, in order of effort:

1. **Skeleton lines inside each panel** (cheap): replace the empty interiors with 2–3 dim placeholder bars per panel during the loading window. Standard pattern, immediately reads as "loading" rather than "empty".
2. **Centered loading state with spinner** during the cold-fetch window — hide the panel scaffolding entirely until content is ready, show a single centered `⠋ Pulling latest activity for coffeeboardnm…` with a spinner frame.
3. **Optimistic last-render** — keep showing whatever was on screen the previous time this dashboard was opened (rail-style memo), then quietly diff in updates. Most "loading" is just confirming we already know — show what we know first.

Option 1 is probably the right balance: keeps the layout stable so content doesn't pop into place, but signals "loading" clearly. Option 3 is the most magical but riskier.

## Suggested copy if (2) is picked

[...]

PR #1558, opened 2026-05-09
polish(dashboard): skeleton loading state replaces empty rectangles (#1539)

samhotchkiss:
## Summary

- Project-dashboard cold-fetch window (8-12s on populated installs) painted empty rounded rectangles with bare ``Inbox`` / ``Current activity`` headings — read as "loaded but empty" until content populated.
- Seeds each section body Static with two or three dim block-character bars so the panels read as intentionally placeholdered during the load window. ``_render`` already calls ``.update(...)`` on every body when data lands so the skeleton is replaced in the same beat.
- Layout stays stable across the swap so content does not pop into place.

## Design choice

Picked **option 1** from the issue (skeleton lines) over option 2 (centered spinner) or option 3 (optimistic last-render). Rationale:

- Skeleton lines keep the layout stable so content does not jump when real data lands. A centered spinner would have to swap out the whole panel scaffolding on data arrival, which causes the very "pop into place" jump the issue was asking us to avoid.
- Optimistic last-render is the most magical but requires a render cache + diffing layer that is bigger than a polish PR should carry.

The bar color (``#1e2730``) matches the panel border so the skeleton sits dimmer than the existing ``proj-empty`` placeholder text — a clear "this is loading" signal without yelling for attention.

## Test plan

- [ ] Open a cold project dashboard via ``pm cockpit-pane project <key>`` — bodies show two/three dim block-character bars during the cold-fetch window
- [ ] Bars are replaced by real content the moment data arrives (no flash, no jump)
- [ ] Loaded panels still read identically to before (skeleton is purely a load-window artifact)

Refs #1539


samhotchkiss:
Codex review finding for PR #1558:

- Medium: Skeleton placeholders remain visible forever on unknown-project or failed-gather dashboards. This PR seeds body sections with skeleton markup at `src/pollypm/cockpit_ui.py:14784-14809` and `:14848-14852`, but the `data is None` render path only updates the topbar and returns at `:15245-15256`. If the project is missing or the first gather fails, the topbar says the project is not tracked while the body still shows loading bars. Please clear or hide all skeleton-backed body sections in the `data is None` branch, and add coverage for the missing-project render path.

Verification: `git diff --check origin/main...origin/polish/dashboard-skeleton-1539` passed in review.

samhotchkiss:
## Codex review — MED-1 addressed (cdc3e7de)

**MED-1: Skeleton bars stay visible forever on bail paths**

Fix: added `_clear_skeleton_bodies()` and call it from both short-circuit sites:
- `_render` — `data is None` branch (unknown / untracked project key).
- `_first_refresh_failed` — first-gather exception path. Codex caught the `data is None` site; the failed-gather path has the same shape and is fixed in the same commit.

The helper blanks `now_body` / `pipeline_body` / `plan_body` / `activity_body` / `inbox_body`. `__init__` seeds these with `_SKELETON_TWO_LINE` / `_SKELETON_THREE_LINE`; without an explicit clear the bars rendered indefinitely on both bail paths.

**Coverage**
- `test_render_clears_skeleton_bodies_when_data_is_none` — drives the unknown-project path on a real `PollyProjectDashboardApp(config_path, "untracked-key")`, asserts the block glyph (`█`) count is zero on every body after `_render()`, and asserts the topbar carries the "not a tracked project" copy.
- `test_first_refresh_failed_clears_skeleton_bodies` — drives the failed-gather path, asserts topbar shows "Error loading project" and `_first_refresh_running` flips to False, asserts no skeleton glyphs remain.

Both pass. The full `tests/test_project_dashboard_ui.py` suite is 158 passed / 1 failed; the single failure (`test_inbox_section_omits_need_action_count_when_cards_show_full_set`) reproduces on the unmodified pre-fix tip — preexisting, unrelated.

**Bonus cleanup**
Per the #1559 review note, this branch had unrelated issue-state churn (0022 / 0024 / 0025 markdown moves). Reverted those — net `git diff origin/main...HEAD` is now exactly the cockpit_ui skeleton changes + tests:
```
src/pollypm/cockpit_ui.py          |  75 ++++++++++++++++++++++--
tests/test_project_dashboard_ui.py | 115 ++++++++++++++++++++++++++++++++++++-
```

Issue #1545, opened 2026-05-09
coffeeboardnm dashboard: "Current activity" panel is empty while a worker is actively running, and plan-ready alert doesn't acknowledge in-flight work


samhotchkiss:
## Symptom

`coffeeboardnm` has had `architect_coffeeboardnm` actively working for ~8 minutes (visible in the live tmux pane: code edits, test runs, deploy attempts, git status checks). Yet the dashboard for the same project shows:

```
  CoffeeBoardNM   PM: Cole

  ────────────────────────────────────────────────────────────────────────────
  ◆ alert
  ╭──────────────────────────────────────────────────────────────────────────╮
  │ Plan's ready — POC plan: ingest every NM event May–Jul 2026 —            │
  │ visual-explainer deliverable, deploy to coffeeboardnm.itsalive.co. Press │
  │ → to review together · 1 alert                                           │
  ╰──────────────────────────────────────────────────────────────────────────╯
  ╭──────────────────────────────────────────────────────────────────────────╮
  │  [→] Review the plan together — POC plan: ingest every NM event May–Jul  │
  │ 2026 — visual-explainer deliverable, deploy to coffeeboardnm.itsalive.co │
  ╰──────────────────────────────────────────────────────────────────────────╯

  ╭─────────────────────────────────────────────────────────────────────────╮
  │  Inbox                                                                  │
  │  Inbox is clear for this project.                                       │
  ╰─────────────────────────────────────────────────────────────────────────╯

  ╭─────────────────────────────────────────────────────────────────────────╮
  │  Current activity                                                       │   ← empty
```

Two issues stack:

### 1. "Current activity" panel is empty while a worker IS active

The whole purpose of the Current activity panel — judging by its name and its placement at the bottom of the dashboard — is to show what's happening right now. It's empty.

What's actually happening (per the live tmux scrollback for that worker):
- 8m 22s of continuous work on coffeeboardnm
- Last activity: "I added a Phase-1 Week-1 ingestion scaffold in src/coffeeboardnm, with fixture-backed parsers for Ticketmaster-style JSON, venue-page JSON-LD, and ICS feeds, plus a generator at scripts/build_event_preview.py."
- Two things blocked: git commit (sandbox metadata path), itsalive deploy (DNS).
- Worker is now on a new prompt iteration.

A user opening the dashboard right now would see "Plan's ready, press → to review" and have no idea that there's a worker spinning that hit a deploy block 30 seconds ago.

[...]

Issue #1526, opened 2026-05-09
supervisor._sweep_stale_alerts orphan-clears plan_missing every heartbeat tick (root cause of #1524 flash)


samhotchkiss:
## Root cause of #1524 flash, identified via stack-trace instrumentation

After #1525 landed (#1524 fix Part A — deterministic precompute in the task_assignment sweep), the banner flash continued. Live evidence: alerts in `~/.pollypm/state.db` showed a new `plan_gate-coffeeboardnm` row created and closed every ~30s, even though my manual `task_assignment_sweep_handler({})` call leaves the alert stable.

I instrumented both `StateStore.clear_alert` and `SQLAlchemyStore.clear_alert` to dump a stack trace on every `plan_missing` close. **Two distinct closers fire:**

1. **Owned closer** — `task_assignment_sweep_handler → _task_assignment_sweep_body → _clear_plan_missing_alert` (`sweep.py:2198`). Behaves correctly post-#1525: only closes when the project genuinely isn't plan-missing.
2. **Generic stale-alert sweep** — `pm heartbeat → supervisor.run_heartbeat → supervisor._sweep_stale_alerts (supervisor.py:2604) → SQLAlchemyStore.clear_alert`. **This is the bug.**

The generic stale-alert sweep at `src/pollypm/supervisor.py:2570-2606` walks every open alert on every heartbeat tick and clears any whose `session_name` isn't in the launch plan's tracked sessions. The `plan_gate-<project>` session name is **synthetic** — it is never in the launch plan by design (mirrors the documented same-shape lifecycle of `no_session` / `no_session_for_assignment:` alerts) — so this sweep classifies it as orphaned and clears it on every tick.

The same code already has the right escape hatch for the same shape:

```python
# supervisor.py:2587-2591
if (
    alert_type == "no_session"
    or alert_type.startswith("no_session_for_assignment:")
):
    continue
```

The comment at lines 2576-2586 articulates the exact same bug pattern for `no_session`:

> `no_session` alerts are emitted by the task_assignment sweep on synthetic session names that intentionally aren't in the launch plan (the whole point of the alert is "no session is running for this role"). The owning sweep clears them when a real session comes online; this generic "not in plan → orphan" predicate is too lenient and produced a false-clean: open on the first sweep, cleared on the second even though no worker process exists. Leave them alone.

[...]

Issue #1524, opened 2026-05-08
Project banner + rail glyph flash between 'Waiting on you' and 'Queued/idle' on every sweep tick


samhotchkiss:
## Symptom

For projects with an open `plan_review` (e.g. `coffeeboardnm`, `savethenovel` after #1511's watchdog backfilled them), the Project Dashboard banner and the new rail glyph (#1520) flash between two states roughly every sweep tick:

- "Waiting on you: this project has no plan yet — press a to review · 1 alert" + rail `◆ CoffeeBoardNM`
- "Queued: 1 task waiting for a worker" + rail `♥· CoffeeBoardNM` (idle / heartbeat fall-through)

This makes the user-facing "you owe me a decision" signal unreliable — half the time it disappears.

## Verified facts

- Alerts table at `~/.pollypm/state.db` (the right DB, supervisor reads from here): `plan_gate-coffeeboardnm` row exists with `sender='plan_missing'`, `state='open'`. `supervisor.status()` correctly returns it with `alert_type='plan_missing'`.
- Run `task_assignment_sweep_handler({})` manually → `plan_missing_alerts: 2, projects_skipped: 1, auto_claim_skipped_plan_missing: 1`. The alert IS re-upserted after manual emit.
- Between sweep ticks, the rail and banner alternate. The alert row in the DB also alternates: I queried it once and got `id 174450, 174451`; after a later sweep, `id 174520, 174521` — different IDs. **Something is clearing the alerts and a later sweep is re-emitting them with new ids.**

## Probable root cause

`src/pollypm/plugins_builtin/task_assignment_notify/handlers/sweep.py`:

- **Pass 1** (line 2036-2050) sweeps the workspace canonical DB with `project_path=None`, plan-presence gate intentionally skipped → does NOT add anything to `plan_missing_projects`.
- **Workspace-project auto-claim loop** (line 2061-2079) calls `_auto_claim_next` per project, which adds projects to `plan_missing_projects` when an auto-claim is blocked on plan. This is conditional on `_auto_claim_enabled_for_project`.
- **Pass 2** (line 2081-2129) walks per-project DBs:
  ```python
  if project_key and project_key not in plan_missing_projects:
      _clear_plan_missing_alert(services, project=project_key)
  ```
  When the per-project DB is missing (true for `coffeeboardnm` post #1519 watchdog drain), this branch fires. If the project did NOT get added to `plan_missing_projects` by the auto-claim path on this tick, **the alert gets cleared**.

[...]

PR #1560, opened 2026-05-09
fix(supervisor): stale plan-ready alert clears when underlying task moves to done

samhotchkiss:
## Summary

#1545 — coffeeboardnm dashboard kept showing a stale "Plan's ready" alert for `coffeeboardnm/1` long after that task moved to `done`. The plan_missing alert family is owned by the task_assignment sweep (which Part A of #1524 made deterministic via `_precompute_plan_missing_projects`), but the per-project clear path can miss when the alert's referenced task itself terminates without the project's queue otherwise advancing — leaving the alert open across consecutive ticks even though its message body is now misleading.

## What changed

Extend `_sweep_stale_alerts` with a terminal-task path that runs **before** the existing skip-list:

1. New helper `pollypm.supervision.alert_task_lookup.extract_task_ids` returns candidate task IDs an alert references. Understands all four reference shapes alerts use today:
   - `stuck_on_task:<project>/<num>` — task id in alert_type suffix
   - `no_session_for_assignment:<project>/<num>` — same shape
   - `review-<project>-<num>` — task number in session_name scope
   - `plan_missing` — task id parsed out of message body (`queued task <id> is waiting`)
2. New `Supervisor._lookup_task_status` walks the canonical workspace DB first, with per-project legacy DB fallback (mirroring the dashboard's resolution order from #1004 / #1519).
3. When any candidate's `work_status` is in `{done, cancelled, abandoned}`, the alert clears and an `audit.alert_cleared_task_terminal` event is stamped to the forensic event stream.

## Watchdog rule shape

The terminal-task path is the self-heal — it lives directly inside the heartbeat sweep, runs every tick, and is idempotent (`clear_alert` is keyed on `(session_name, alert_type, status='open')`). No separate detector / dispatcher needed.

## Coexistence with #919 / #1526 skip-list

The existing skip-list guards (`no_session`, `no_session_for_assignment:*`, `plan_missing`) **remain intact**. The terminal-task path runs first and only fires when it can identify the underlying task AND that task is terminal. When the task is *still live*, the task_assignment sweep stays the sole owner of those alert families. A source-level test (`test_existing_skip_list_guards_remain_intact`) pins this contract so a future refactor can't silently regress the guard.

## What this PR does NOT address

`#1545`'s "Current activity panel empty while a worker is actively running" symptom is left for a separate fix. This PR addresses the stale-alert half only — Sam's brief explicitly tagged the activity-panel piece as informational context that could be deferred.

## Test plan

- [x] `tests/test_alert_task_lookup.py` — 15 unit tests pinning task-id extraction across every alert reference shape, plus negative cases (`no_session` has no task ref, message body is only parsed for known body-bearing types).
- [x] `tests/test_sweep_stale_alerts_terminal_task.py` — 6 integration tests covering: open-task → alert stays, done-task → alert clears, audit-event payload, idempotence across repeat sweeps, plan_missing-via-message-body (the witness shape), source-level skip-list invariants.
- [x] `tests/test_cancel_clears_no_session_alerts.py::TestStaleAlertGuardIntact` — existing guard tests still pass.
- [x] `tests/test_supervisor_alerts.py`, `tests/test_supervisor.py`, `tests/test_ghost_project_alerts.py`, `tests/test_review_pending_alerts.py` — 150 tests pass; the one failure (`test_supervisor_alert_helper_updates_and_nudges`) reproduces on a clean checkout of `main` and is unrelated.
- [x] Full test suite — only pre-existing failures repro (`test_cockpit_contract_boundaries.py`, `test_project_dashboard_ui.py` flakes); no new failures introduced.

Refs #1545


samhotchkiss:
Codex review findings for PR #1560:

- High: Terminal-task lookup misses alias-named projects in legacy per-project DBs. `_lookup_task_status()` looks up `self.config.projects.get(project_key)` using the literal task-id project segment at `src/pollypm/supervisor.py:2755`, then only adds the per-project legacy DB when that exact config key is found at `:2789-2792`. Projects can store task IDs under aliases/display names such as `health-coach` while the config key is `health_coach`; if the terminal task exists only in `<project>/.pollypm/state.db`, the lookup returns `None` and the stale alert remains open. Please resolve the task-id project segment through the existing project alias/slug logic before building candidate DB paths, and add a regression test where config key and stored task-id segment differ.

- Medium: The live-task regression test does not assert the alert stays open. `tests/test_sweep_stale_alerts_terminal_task.py:180-191` builds `open_pairs` but never asserts on it; it only checks that no terminal-task audit event was emitted. Because the fixture intentionally uses `worker_demo`, which is not the configured tracked session name, the existing orphan-clear path can clear the live-task alert and the test still passes. Please assert the alert remains open for a live task, using a tracked session name or a skip-list family such as `plan_missing`, and keep the separate no-terminal-audit assertion.

Verification: review ran `git diff --check` and syntax compilation of the PR versions of `supervisor.py`, `alert_task_lookup.py`, and the two new test files; all passed.

samhotchkiss:
Addressed codex-review findings.

**HIGH-1 (alias-named project / legacy DB miss)** — `_lookup_task_status` now resolves the task-id project segment through `project_storage_aliases` when `self.config.projects.get(project_key)` misses. The new helper `_resolve_project_by_alias` walks every configured project and returns the one whose alias set (slug, hyphen/underscore variants, display name, on-disk dir, casing variants) contains the task-id project segment. With the resolver wired in, the per-project legacy DB at `<project>/.pollypm/state.db` is added to the candidate list and the lookup returns `done` instead of `None`, so the stale alert finally clears.

Regression test `test_terminal_clear_resolves_alias_named_project_in_legacy_db` pins the exact topology: config key `health_coach`, project name / on-disk dir `health-coach`, terminal task seeded only in the per-project legacy DB, alert references `health-coach/N`. Includes a unit-level pre-flight on `_lookup_task_status` and an integration assertion on the full sweep + audit-event path.

**MED-1 (live-task test was a false-pass)** — The original test built `open_pairs` and never asserted on it; using a non-tracked `worker_demo` session with `stuck_on_task:` alert_type meant the existing orphan-clear path could clear the alert and the test would still pass on the (separate) no-terminal-audit assertion. Switched the alert to `plan_missing` (skip-listed by #1526 so orphan-clear cannot fire) and added a positive `("plan_gate-demo", "plan_missing") in open_pairs` assertion alongside the no-terminal-audit check.

**Bonus** — Reverted the two unrelated `chore(issues)` commits (`58d6fd08`, `fe5360ff`) so the PR diff is now scoped to the supervisor + tests only.

Test run: `tests/test_sweep_stale_alerts_terminal_task.py` + `tests/test_alert_task_lookup.py` — 22 passed. (One pre-existing failure in `tests/test_supervisor_alerts.py::test_supervisor_alert_helper_updates_and_nudges` reproduces against `origin/main` and is out of scope for this PR.)

samhotchkiss:
Codex re-review finding after commit `d1ee41a`:

- Low: `src/pollypm/supervision/alert_task_lookup.py:34` imports `Iterable` but never uses it. `ruff` flags this as `F401`, so this can fail a lint gate even though the behavior tests pass. Please remove `from typing import Iterable`.

Validation run: `uv run pytest tests/test_alert_task_lookup.py tests/test_sweep_stale_alerts_terminal_task.py -q` passed (`22 passed, 22 warnings`). `uv run ruff check ...` also reported existing `E402` issues in `src/pollypm/supervisor.py`; those appear pre-existing for that file, but the `F401` is newly introduced by #1560.

samhotchkiss:
Addressed Codex re-review F401 in `3e1847cb`: dropped the unused `from typing import Iterable` from `src/pollypm/supervision/alert_task_lookup.py`. `uv run --extra dev ruff check` on the file is now clean; `uv run --extra dev pytest tests/test_alert_task_lookup.py tests/test_sweep_stale_alerts_terminal_task.py -q` → 22 passed.

Pre-existing E402 issues in `src/pollypm/supervisor.py` confirmed unrelated to this PR (reproduce on main).

Issue #1737, opened 2026-05-19
[design] Postgres + pgvector migration — eliminate per-project SQLite


samhotchkiss:
> **Mode:** Design / scoping. **Not a PR.** Output is a plan; implementation fans out into slices A-I below.

## TL;DR

Replace PollyPM's SQLite storage layer with a single PostgreSQL instance (+ `pgvector` for semantic memory recall). The two perf reviews on #1634 and the Move A design (#1664) both bottomed out at the same structural cost: **per-project SQLite fanout — ~60–100 file opens per cockpit/rail refresh on a 12-project workspace, plus a single-writer `flock` dance and a dual-source connection bug** (`cockpit_inbox.pm_inbox_awaits_user_list` opens BOTH `create_work_service(...)` and `SQLAlchemyStore("sqlite:///...")` against the same file). The wedges (#1605/#1608/#1630) moved work off the main thread but did not change the cost.

This issue scopes the **storage replacement** instead of yet another rail wedge. PG eliminates the fanout entirely (one pool, one DB), gives Polly real concurrent writers, and unblocks **pgvector** for semantic memory recall (Sam wants this anyway, and the current `FileMemoryBackend.recall` is FTS5+bm25+recency — no embeddings).

The migration is much cheaper now than a week ago because the storage seam already exists:

- Every non-storage caller of `sqlite3` is routed through `pollypm.storage/` facades (#1376/#1596/#1617/#1625/#1681/#1682).
- `work.service_*` is decoupled from `SQLiteWorkService` via structural Protocols (#1717/#1722/#1725/#1729/#1736).
- `readonly_uri()` already centralises read-only URI construction (#1674/#1690).
- `[storage] backend = "..."` + `pollypm.store_backend` entry-point group is **already wired** (`src/pollypm/store/registry.py`, `src/pollypm/config.py:549`). A stub `PostgresStore` exists at `src/pollypm/store/backends/postgres_stub.py` to keep the protocol dialect-neutral.

The work is a **new `PgWorkService`** (replacing `SQLiteWorkService`) + a **real `PostgresStore`** (replacing `SQLAlchemyStore` in the singleton registry) + one-shot migrate tool + test-fixture replacement + pgvector embedding column. Cutover is hard — no dual-mode shim — because the protocols are clean and the alternative (run both in parallel) is the worst of both worlds.

---

### 1.1 On-disk files

| Path | Owner | Notes |
|------|-------|-------|
[...]

PR #1748, opened 2026-05-19
fix(pg-migrate): stamp project_key on migrated rows from source DB project

samhotchkiss:
## Summary

After `pm storage migrate-to-pg --commit` against a real workspace, every migrated row landed with `project_key = ''`, breaking every cockpit query that filters by project. The migration tool only injected `project_key` when a source row's table omitted the column entirely, and only on three tables.

- `checkpoints` / `worktrees` / `token_samples` / `token_usage_hourly` / `architect_resume_tokens` carry `project_key` natively but still landed empty if the source value was blank — they now get the descriptor key as a backfill.
- `work_tasks` ignored its own per-row `project` column, which already binds each row to a project even in the workspace-DB case where the source descriptor's `project_key` is empty — `work_tasks.project` is now wired as the per-table fallback.
- `_convert_row` now resolves `project_key` with a documented precedence: source row's non-empty value → fallback column (e.g. `work_tasks.project`) → source DB descriptor. Empty source values are rewritten in place so the per-row column order stays stable across the `executemany` batch (no batch flush / regeneration overhead).

Refs #1737 (pg-cutover umbrella).

## Test plan

- [x] `uv run pytest tests/test_pg_migration_tool.py -x -q` — 19 passed (13 previous + 6 new project_key stamping unit tests, covering: spec coverage, `work_tasks` project fallback, descriptor injection for `messages`, source-row precedence on `checkpoints`, in-place override on `worktrees`, workspace-DB regression for `work_tasks`).
- [x] `uv run pytest tests/test_pg_migration_tool.py tests/test_pg_schema.py tests/test_pg_storage_facades.py -q` — 49 passed.
- [ ] Re-run `pm storage migrate-to-pg --commit` against a real workspace and confirm `SELECT count(*) FROM work_tasks WHERE project_key = ''` returns 0.


samhotchkiss:
⚠️ Heads-up: this PR includes 22 unrelated files under `issues/01-ready/` and `issues/05-completed/` (around 302 of the 526 added lines) that look like watchdog/architect artifacts from the working tree, not part of the migration-tool fix.

The actual fix is just 2 files (~224 lines):
- `src/pollypm/storage/pg_migration_tool.py` (+64/-14)
- `tests/test_pg_migration_tool.py` (+160)

Recommend dropping the `issues/**` paths from the branch before merge (or merging them separately if those issue notes are wanted on main).

samhotchkiss:
Noise removed — diff now shows only the real changes (2 files).

Issue #1044, opened 2026-05-02
plan_gate alerts firing duplicate IDs simultaneously — write-time dedupe broken across multiple alert types


samhotchkiss:
## Symptom

\`pm alerts\` output (live just now):

\`\`\`
- #71470 warn plan_gate-pollypm/plan_missing: Project 'pollypm' has no approved plan yet — queued task pollypm/1 is waiting on it. Open the project and press c to ask the PM to plan it.
- #71469 warn plan_gate-pollypm/plan_missing: Project 'pollypm' has no approved plan yet — queued task pollypm/1 is waiting on it. Open the project and press c to ask the PM to plan it.
- #71468 warn plan_gate-russell/plan_missing: Project 'russell' has no approved plan yet — queued task russell/1 is waiting on it. Open the project and press c to ask the PM to plan it.
- #71467 warn plan_gate-polly_remote/plan_missing: Project 'polly_remote' has no approved plan yet — queued task polly_remote/5 is waiting on it. Open the project and press c to ask the PM to plan it.
- #71466 warn plan_gate-polly_remote/plan_missing: Project 'polly_remote' has no approved plan yet — queued task polly_remote/5 is waiting on it. Open the project and press c to ask the PM to plan it.
\`\`\`

**Two open \`plan_gate-pollypm/plan_missing\` alerts** (#71470 + #71469) and **two open \`plan_gate-polly_remote/plan_missing\` alerts** (#71467 + #71466) — same alert type, same project, same body text, same severity, both \`open\` simultaneously.

I first noticed this for \`plan_gate-russell\` two ticks ago (also doubled up at the time). It's now affecting at least three alert types in the same family.

## Why this matters (priority lens: friction + alert trust)

\`pm alerts\` is the operator's primary diagnostic surface. Two identical lines with different IDs:

1. **Doubles the perceived alert count** — what the operator reads as \"5 plan_gate alerts\" is actually \"3 plan_gate states, double-emitted.\"
2. **Forces the operator to mentally dedupe** before reasoning about scope. Magic-prompt fit: \"is the most-likely next action the easiest action?\" — no, it's \"first, figure out which of these are the same.\"
3. **Bloats the audit trail** — \`pm activity --kind alert\` will record N copies of the same plan_missing for the same project, complicating any future history-of-this-alert query (related to #1033, where alert-cleared events are missing — the create side has the inverse problem).

## Likely root cause

[...]

PR #1753, opened 2026-05-19
feat(pg): register and implement PostgresStore (Slice I, #1737)

samhotchkiss:
## Summary

Closes the cutover gap that left `[storage] backend = "postgres"` raising

```
StoreBackendNotFound: Storage backend 'postgres' is not installed.
Available backends: sqlite.
```

The in-tree stub at `pollypm.store.backends.postgres_stub:PostgresStore` was deliberately kept unregistered so the Store protocol stays dialect-neutral. Slice I ships a real `PgStore` alongside it and wires the entry point.

## What landed

- **New `PgStore`** at `src/pollypm/store/backends/pg_store.py` — full Store protocol port over psycopg, reusing `pollypm.storage.pg_pool.get_rw_pool()` / `get_ro_pool()`. Lazy `apply_migrations()` on first use mirrors the sqlite "open store, schema is current" contract.
- **Entry point** registered in `pyproject.toml`:
  ```toml
  [project.entry-points."pollypm.store_backend"]
  sqlite = "pollypm.store.sqlalchemy_store:SQLAlchemyStore"
  postgres = "pollypm.store.backends.pg_store:PgStore"
  ```
- **Forward migration 0002_alert_dedupe_tuple** — pg's initial unique alert index was on `(scope, subject)` which diverged from sqlite's `(scope, sender, recipient, type)`. Aligned so `upsert_alert` dedupes on session+alert_type identically on both backends.
- **21 new tests** in `tests/test_pg_store.py` against the schema-isolated `pg_schema_pool` fixture.

## Store-protocol method status

| Method | Status |
| --- | --- |
| `append_event` | implemented (synchronous; pg pool is deep enough that we skip the background EventBuffer the sqlite path uses) |
| `record_event` | implemented |
| `enqueue_message` | implemented |
| `upsert_message` | implemented (SELECT-then-(UPDATE\|INSERT) with IntegrityError retry, mirroring sqlite #1044) |
| `update_message` | implemented |
| `close_message` | implemented |
| `query_messages` | implemented (IN-list, since, limit, unknown-filter rejection) |
| `prune_messages` | implemented |
| `upsert_alert` / `clear_alert` | implemented (occurrences bump + alert.cleared event emit) |
| `transaction()` | implemented (yields a **psycopg** Connection, not a SQLAlchemy one — see divergence below) |
| `_delete_all_messages_for_tests` | implemented (DELETE within the test schema's search_path) |
| `url` / `dispose` / `close` | implemented |
| `execute(stmt)` | **deferred — raises `NotImplementedError`** with a three-question message. Honouring SQLAlchemy Executables on pg would require dragging SQLAlchemy onto the psycopg stack, defeating the migration's purpose. No v1 caller hits `execute()` against pg. |

## Divergences (non-papered-over)

1. `transaction()` yields a **psycopg `Connection`**, not `sqlalchemy.engine.Connection`. Every in-tree caller of `store.transaction()` is already pg-aware (Slice C facades) or reaches the DB through typed methods on the store.
2. `execute(stmt)` raises (above).
3. `append_event` is synchronous on pg (no EventBuffer drain thread). Protocol only promises fire-and-forget at the call site; semantics preserved.

## Test plan

```
uv run pytest tests/test_pg_store.py -x -q
```
output: `21 passed in 3.10s`.

Broader pg suite (`tests/test_pg_pool.py test_pg_schema.py test_pg_storage_facades.py test_pg_store.py test_pg_migration_tool.py`): `78 passed in 9.92s`.

Pre-existing failures in `test_pg_work_service_parity.py::TestSliceBStubsRaiseLoudly` are unrelated to Slice I (verified by re-running against `git stash`).

- [x] Tests green (`uv run pytest tests/test_pg_store.py -x -q` → 21 passed in 3.10s)
- [x] `pm reset --force` succeeds under `[storage] backend = "postgres"` after `uv tool install --reinstall --force --with 'psycopg[binary]' --with psycopg_pool --with pgvector .`
- [x] Operator config reverted to sqlite

Refs #1737.

Generated with Claude Code.


samhotchkiss:
Noise removed — diff now shows only the real changes (7 files).

PR #1798, opened 2026-05-19
test(pg): recover stacked K-tests parts 5+6 (#1737)

samhotchkiss:
Recovers ~54k LOC of sqlite-test deletions that were lost when the stacked PRs #1792 and #1795 merged in the wrong order — #1795 squash-merged against #1792's branch as base instead of main, so only #1795's own ~42-line diff landed. #1792 auto-closed with conflict status.

This PR re-targets the head of `pg-tests-migrate-finish-2` (which carries parts 4+5+6 commits) against main. Part 4 is already merged via #1786, so the diff here is just parts 5 + 6's deletions.

Squash-merges into one commit.

Refs #1737


samhotchkiss:
Conflicts with main due to stacked-squash merge order issue. Reopening fresh in a follow-up.

PR #2002, opened 2026-05-20
chore(path-resolution): migrate 8 hottest raw-join sites (closes #1966, refs #1972)

samhotchkiss:
## Summary

Slice 2 of the #1972 design fix. Routes the 8 highest-traffic raw ``/ ".pollypm" /`` joins through the typed helpers from PR #1996. Each is a hot-path constructor (work-service factory / DB resolver / runtime services / heartbeat) that was a doubled-pollypm-path leak vector whenever the caller happened to be operating on ``GLOBAL_CONFIG_DIR`` itself as the project root.

**Migrated sites:**
- ``audit/log.py:project_log_path`` — #1966 leak site #1. → ``project_audit_log_path``
- ``work/session_manager.py:1790`` — #1966 leak site #2 (transcript archive). → ``project_transcripts_dir`` + ``"tasks" / task_id``
- ``runtime_services.py:105`` — work-service init in the runtime factory (every process boot). → ``project_state_db_path``
- ``work/service_factory.py:25`` — inbox/cockpit ``open_project_work_service``. → ``project_state_db_path``
- ``work/db_resolver.py:106,126`` — both the legacy per-project path and the workspace-default path. → ``project_state_db_path``
- ``work/first_shipped.py:125`` — first-shipment DB lookup. → ``project_state_db_path``
- ``work/inbox_actions.py:28`` — inbox-action DB lookup. → ``project_state_db_path``
- ``heartbeats/local.py:186,1437`` — heartbeat work-DB lookups (both project-cfg and task-claim paths). → ``project_state_db_path``

**Lint-gate baseline shrink:** 155 → 145 raw joins. Entries that hit zero are removed; non-zero entries get their count reduced.

## Stack

Based on ``feat/typed-pollypm-path-helpers`` (PR #1996). Merge order: #1996 then this.

## Test plan

- [x] ``pytest tests/test_no_raw_pollypm_path_joins.py`` — 4/4 pass (ratchet test still green with reduced baseline)
- [x] ``pytest tests/test_audit_log.py tests/test_doubled_pollypm_path_regression.py tests/test_doubled_path_preconditions.py`` — 32/32 pass
- [ ] **Sam to verify:** spot-check the import-time-safety of the lazy imports added inside hot paths (each migrated site adds a function-scoped ``from pollypm.projects import ...``).

## Refs

Closes #1966. Refs #1972, #1996.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2003, opened 2026-05-20
chore(path-resolution): migrate remaining raw-join sites (refs #1972)

### LINKED ISSUE #1972: [regression] doubled-pollypm-path returns despite #1954 — root cause analysis
## Symptom (this session)

After PR #1898 and #1954 merged, the doctor's ``doubled-pollypm-path`` alert fired again. The user observed ~138 files under ``~/.pollypm/.pollypm/`` post cockpit restart: ``state.db``, ``audit.jsonl``, ``transcripts/.ingestion-state.json``, ``artifacts/checkpoints/{architect,reviewer,operator,heartbeat}_*/``, ``project-guides/``, ``dossier/``, ``INSTRUCT.md``, etc.

## Immediate cause (this session, not the architectural issue)

The user is running an old installed binary. ``/Users/sam/.local/share/uv/tools/pollypm/`` mtime is ``May 19 16:40``, but PR #1954 merged at ``May 19 21:32``. The installed ``pm`` predates the fix.

Verified by spot-check: the installed ``pollypm/knowledge_extract.py:271`` and ``session_intelligence.py:193`` still build raw ``project_root / ".pollypm" / "transcripts"`` joins — the exact code paths #1954 rewrote in source. The currently-running ``pm rail_daemon`` (PID 55557, started 14:55) and ``pm cockpit`` (PID 75333, started 23:31) both reference ``/Users/sam/.local/bin/pm`` and so were dispatching the pre-fix code. A ``uv tool install --force pollypm`` + full process restart would have stopped the new accumulation.

This is consistent with the user observation only if cockpit was restarted but the underlying ``rail_daemon`` / heartbeats were not, OR the user reinstalled but did not reinstall ``pollypm`` from a post-#1954 wheel.

## The actual root cause (architectural)

#1898 introduced ``_resolve_pollypm_root()`` (collapses ``~/.pollypm + .pollypm = ~/.pollypm``). #1954 routed 8 leak sites through it. But the codebase has **52 raw ``project_path / ".pollypm" / "..."`` constructions** that bypass the resolver entirely:

```
$ grep -rn 'project_path / "\.pollypm"\|project.path / "\.pollypm"\|workspace_root / "\.pollypm"' src --include='*.py' | wc -l
52
```

Each of these silently produces ``~/.pollypm/.pollypm/...`` if its caller happens to be operating on ``config.project.root_dir`` (which equals ``GLOBAL_CONFIG_DIR`` for user-global installs — see ``config.py:_parse_project_settings`` where ``base.name != "config"`` falls through to ``root_dir = base = ~/.pollypm``). The resolver doesn't fire because nothing CALLS it; the joins are inline ``Path /`` operations.

#1966 already identifies two more concrete sites (``audit/log.py:329 project_log_path``, ``work/session_manager.py:1790`` transcript archive). Both still write the doubled path on source HEAD. There are more (``runtime_services.py:105`` builds ``project_root / ".pollypm" / "state.db"`` and ``mkdir(parents=True)`` — guarded today only because ``workspace_root`` is usually populated, not because the resolver intervenes).

PR #1954's commit message even says: ``Out of scope: read-only path constructions that wouldn't accumulate files when doubled``. The author KNEW more sites existed and punted. That's the architectural pattern that's failing.

## Why #1898 + #1954 keep regressing

The model is wrong. ``_resolve_pollypm_root()`` is OPT-IN — a caller has to CHOOSE to use it. So:

1. Every new feature that does ``some_path / ".pollypm" / "..."`` is a potential new leak site.
2. Every refactor that changes which subsystem calls which is a potential new leak site.
3. We will keep playing whack-a-mole forever unless the abstraction is enforced.

## Recommended fix (design-shaped, deferred per RC stability rules)

A. Introduce ``project_state_db_path(project_path)``, ``project_audit_log_path(project_path)``, ``project_worktrees_dir(project_path)`` etc. as the ONLY API for these paths. Each calls ``_resolve_pollypm_root()`` so the doubled-path guard always fires.

B. Add a lint check (or a ``ruff`` custom rule, or a CI grep gate) that fails on any new ``/ ".pollypm" /`` literal join in ``src/``. The pattern is mechanically detectable.

C. Migrate the 52 existing call sites in batches over multiple PRs.

D. Optional belt-and-suspenders: add an ``__init__`` precondition to ``SQLiteWorkService`` that raises if ``db_path`` contains ``.pollypm/.pollypm``. That converts the silent doubled-path leak into a loud crash — much easier to catch in CI / smoke tests. The doctor's ``doubled-pollypm-path`` check is post-hoc.

## Why this is being filed as STOP-and-report, not fix-now

Per the v1 RC stability override and the operator's explicit instruction in the investigation prompt: ``If the root cause is design-shaped (not a bug-shaped fix) and would need a broader refactor, STOP and report — don't bandage it again.``

The piecemeal patching of additional leak sites (which #1966 is poised to do) WILL close the symptom for the currently-known callers, but will NOT prevent the next refactor or feature PR from re-introducing the bug. The fix shape (A–D above) is broader than v1 RC tolerates without explicit operator approval.

## Recommended immediate actions

1. Reinstall ``pm`` from current HEAD (``uv tool install --force --reinstall pollypm`` from this checkout, or wait for the next release wheel) — this stops the active leak now that #1954 is in source.
2. Restart ``pm rail_daemon`` + ``pm cockpit`` + ``pm heartbeat`` so they load the new code.
3. Close #1966 by either (a) routing the two specific leak sites it identifies (bandage), or (b) doing A–D above (design fix). Operator's call.
4. Defer the design fix unless operator green-lights the broader refactor.

Refs #1810, #1898, #1950, #1954, #1966.

samhotchkiss:
## Summary

Slice 3 of the #1972 design fix. Mechanical port of every remaining raw ``/ ".pollypm" /`` join in ``src/pollypm/`` through the typed helpers from #1996. Reduces the lint-gate baseline from **145 → 0** across **63 files**.

After this PR merges, the lint gate (#1996) is an absolute "no raw joins" rule.

## Migration patterns

- **Project-rooted state.db** (``X / ".pollypm" / "state.db"``) → ``project_state_db_path(X)``. The single largest pattern.
- **Project-rooted subdirs** (worktrees / plugins / gates / flows / rules / magic / config / docs / content / inbox / worker-markers / advisor-log / project-guides / control-prompts) → corresponding ``project_<thing>_dir`` helper.
- **Per-account session-markers** (``account.home / ".pollypm" / "session-markers"``) → ``project_session_markers_dir(account.home)``.
- **Global ``Path.home() / ".pollypm" / X``** module-level constants → ``from pollypm.config import GLOBAL_CONFIG_DIR`` + ``GLOBAL_CONFIG_DIR / X``.
- **Helper-module ``Path.home()`` use inside functions** → ``global_pollypm_dir()``.

## Files migrated (63)

Heaviest: ``cockpit_ui.py`` (13), ``doctor.py`` (10), ``config_patches.py`` (10), ``work/session_manager.py`` (7), ``plugin_host.py`` (6). Then ``work/gates.py``, ``work/flow_engine.py``, ``supervisor.py``, ``rules.py``, ``plugins_builtin/project_planning/cli/project.py``, ``cockpit_project_settings.py``, ``agent_profiles/defaults.py`` (4 each). Then 16 files @ 2 sites, ~30 files @ 1 site.

## Stack

Based on ``chore/migrate-raw-joins-batch-1`` (PR #2002), which is in turn based on ``feat/typed-pollypm-path-helpers`` (PR #1996). Merge order: #1996, #2002, then this.

## Test plan

- [x] ``pytest tests/test_no_raw_pollypm_path_joins.py`` — 4/4 pass (ratchet test passes with empty baseline)
- [x] AST parse + import-time smoke across every migrated module — clean.
- [ ] Smoke run of audit + doubled-path regression tests (in flight — pg testcontainer setup is slow)
- [ ] **Sam to verify:** the migrations preserve behaviour for non-doubled-path inputs (the helpers are pure pass-through except for the ``GLOBAL_CONFIG_DIR`` case).
- [ ] **Sam to verify:** circular-import risk on hot paths — most migrations use function-scoped lazy imports for safety; the few module-top imports (``GLOBAL_CONFIG_DIR``) are from ``pollypm.config`` which only depends on ``pollypm.models``.

## Refs

Refs #1972, #1996, #2002. PR 4 of the slice closes #1972 with the verification artifact.


samhotchkiss:
Blocking merge until the raw-join migration is actually complete. On the head of the stack I still see production raw joins that the lint gate misses because it excludes whole files instead of only helper/constant lines:\n\n`rg -n '/\s*"\.pollypm"' src/pollypm`\n\nRemaining code hits include:\n- `src/pollypm/config.py:1223` — `base_dir = root / ".pollypm"` outside the `GLOBAL_CONFIG_DIR` constant\n- `src/pollypm/projects.py:796` — `work_db = project.path / ".pollypm" / "state.db"`\n- `src/pollypm/projects.py:803` — `worktree_root = project.path / ".pollypm" / "worktrees"`\n\nThe test doc says `config.py` is allowed only for the constant and `projects.py` only for helper construction, but the implementation excludes those files wholesale. Please migrate those operational joins through typed helpers or make the gate line-scoped so future joins in those files are not invisible.

samhotchkiss:
@samhotchkiss addressed in `10357ebcc` (force-pushed to `chore/migrate-raw-joins-batch-2`).

Migrated the 3 operational raw joins through the typed helpers from PR #1996:

- `src/pollypm/config.py:1223` — `_build_example_config()` was `base_dir = root / ".pollypm"`; now `project_pollypm_dir(root)` (lazy import to dodge the config <-> projects circular). The resolver also collapses the doubled-path case, so `pm example-config` rooted at `~/.pollypm` no longer emits `~/.pollypm/.pollypm/state.db`.
- `src/pollypm/projects.py:796` — `rename_project_slug()` was `project.path / ".pollypm" / "state.db"`; now `project_state_db_path(project.path)`.
- `src/pollypm/projects.py:803` — same function, was `project.path / ".pollypm" / "worktrees"`; now `project_worktrees_dir(project.path)`.

Verified:
- `_current_counts()` still `{}` (lint-gate stays at zero baseline).
- `_build_example_config(GLOBAL_CONFIG_DIR)` returns `base_dir=~/.pollypm` (not the doubled path).
- All 4 tests in `tests/test_no_raw_pollypm_path_joins.py` pass locally.

The line-scoping you asked for is in PR #2004 (stacked on this branch — rebased after this push). That PR replaces the whole-file exclusion with an AST-driven line-scoped allowlist so future ops joins inside `config.py` / `projects.py` can't hide behind the previous wholesale skip.

samhotchkiss:
Re-review blocker: this still leaves a raw `.pollypm` join in `src/pollypm/state_cache/refresher.py:93`:

```py
return Path.home() / ".pollypm" / "audit"
```

Because this PR empties `_BASELINE` in `tests/test_no_raw_pollypm_path_joins.py`, that remaining source join violates the #1972 contract. Route this through the typed/global helper surface, e.g. `global_pollypm_dir() / "audit"`, or add a narrowly justified `# noqa: pollypm-path-join` if there is a real reason the helper API cannot apply.

I did not merge this PR. Also note: the local pytest run for `tests/test_no_raw_pollypm_path_joins.py` hung without output after several minutes in this environment, but the source offender above is visible directly from the tree scan.

samhotchkiss:
@samhotchkiss follow-up after the rebase of `chore/migrate-raw-joins-batch-1` on main pulled in #2000.

**New head:** `56f9067c9`.

The rebase brought `state_cache/refresher.py` (from #2000) into the diff. That file has an import-cycle fallback at line 93 — `Path.home() / ".pollypm" / "audit"` — that the line-scoped lint gate (PR #2004) correctly catches as a new offender. Without an action the zero-baseline ratchet fails on this stack head.

Fixed in `56f9067c9` by tagging that single line with `# noqa: pollypm-path-join` + an inline comment explaining the rationale: the fallback only fires when the lazy `from pollypm.config import DEFAULT_CONFIG_PATH` above has already failed (partially-initialised package, circular import). Routing through `pollypm.projects.global_pollypm_dir` would re-introduce the exact cycle the lazy import guards against. The literal is intentional.

This is the per-line escape hatch the lint gate's docstring documents for "constructing a fixture path in the package install tree, where no project root exists" — same shape: a leaf-module fallback where the typed helper isn't available.

Verification:
- `_current_counts()` returns `{}` after the noqa.
- `test_lint_gate_ratchet` passes (4/4 lint-gate tests green).
- No behavioural change.

GitHub now reports `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

samhotchkiss:
@samhotchkiss — addressed in commit `56f9067c9` (after your re-review at 16:55Z).

The `refresher.py:93` raw join is intentional and now carries an inline-justified `# noqa: pollypm-path-join` per your offered alternative. The justification comment reads:

> import-cycle fallback. This leaf module mirrors `audit.log._central_root` without importing it (cycle risk via state_cache → audit → state_cache). When the lazy `pollypm.config` import above fails, we cannot route through the typed helper (`pollypm.projects.global_pollypm_dir`) without re-introducing the cycle. The literal is intentional.

The line-scoped gate (now on #2004) correctly flagged this as a new offender — that's actually live proof the gate works. Hence the explicit noqa with justification rather than routing through the helper.

samhotchkiss:
Re-review status: the previous raw-join blocker is addressed. I verified `tests.test_no_raw_pollypm_path_joins._current_counts()` returns `{}` on this head, and the `state_cache/refresher.py` fallback now has a line-scoped `# noqa: pollypm-path-join` with a concrete import-cycle justification.

I retargeted this PR from the already-merged `chore/migrate-raw-joins-batch-1` branch to `main` so it can actually land into the default branch. After retargeting, GitHub reports `mergeStateStatus: DIRTY`. The stack now needs a rebase onto current `main` (which includes #2006) before it can merge.

I did not merge this PR. `compileall` passed locally; the focused pytest run for `tests/test_no_raw_pollypm_path_joins.py` hung again in this environment, so I stopped it.

samhotchkiss:
Blocking issue found in the latest pass:

- `src/pollypm/project_paths.py:34` imports `project_control_prompts_dir` from `pollypm.projects`, then `src/pollypm/project_paths.py:208-216` defines a local function with the same name and calls `project_control_prompts_dir(Path(project_path))`. At runtime that resolves to the local function, not the imported helper, so `pollypm.project_paths.project_control_prompts_dir(...)` now recurses until `RecursionError`.

Repro from the PR worktree:

```py
from pathlib import Path
from pollypm.project_paths import project_control_prompts_dir
project_control_prompts_dir(Path('/tmp/example'))
# RecursionError: maximum recursion depth exceeded
```

Please alias the imported helper, e.g. `from pollypm.projects import project_control_prompts_dir as _project_control_prompts_dir`, and call the alias from the wrapper. I verified the broad raw-join migration otherwise reports `_current_counts() == {}` and `compileall` passes, but this runtime path helper regression needs fixing before merge.

samhotchkiss:
@samhotchkiss — fixed in `526efa4082`. Aliased the import as `_project_control_prompts_dir` so the wrapper at :208-216 calls the helper, not itself.

Verified with the exact repro you gave:
```
$ python -c "from pathlib import Path; from pollypm.project_paths import project_control_prompts_dir; print(project_control_prompts_dir(Path('/tmp/example')))"
/private/tmp/example/.pollypm/control-prompts
```
(no RecursionError)

Issue #1966, opened 2026-05-20
doubled-pollypm-path regression: audit.jsonl still leaks after #1954


samhotchkiss:
## Symptom

Doctor's ``doubled-pollypm-path`` alert is firing again. ``~/.pollypm/.pollypm/`` was recreated after a cockpit restart on the user's machine. PR #1898 introduced the ``_resolve_pollypm_root()`` guard; PR #1954 (closes #1950) audited the scaffold/transcripts/curator/memory/maintenance sites and routed them through the guard. Despite those merges, additional leak sites remain.

## Root cause

``src/pollypm/audit/log.py:project_log_path`` (line 329) builds the per-project audit log path with a literal ``".pollypm"`` join:

```python
def project_log_path(project_path: Path | str | None) -> Path | None:
    ...
    p = Path(project_path)
    return p / ".pollypm" / "audit.jsonl"
```

When the work-service emits ``work_db.opened`` (or any other event) with ``project_path = ~/.pollypm`` — i.e. the ``pollypm`` project itself — the audit row lands at ``~/.pollypm/.pollypm/audit.jsonl``. Confirmed in the live leak file:

```
$ head -1 ~/.pollypm/.pollypm/audit.jsonl
{"schema":1,"ts":"2026-05-20T03:18:22...","project":"_workspace","event":"work_db.opened","metadata":{"project_path":"/Users/sam/.pollypm"}}
```

The PR #1954 audit explicitly lists ``audit/log.py`` as out-of-scope ("read-only path constructions"), but ``project_log_path`` is the WRITE side — called from ``emit()`` on every audit event and immediately followed by ``_append_line`` which ``mkdir(parents=True, exist_ok=True)`` and writes.

A secondary leak site: ``src/pollypm/work/session_manager.py:1790`` builds ``self._project_path / ".pollypm" / "transcripts" / "tasks" / task_id`` for transcript archive — same doubled-path failure mode when ``_project_path == ~/.pollypm``.

## Fix

Route ``project_log_path`` (audit/log.py) and the transcript archive dir (work/session_manager.py) through ``project_pollypm_dir`` / ``_resolve_pollypm_root`` so the doubled-path guard collapses ``~/.pollypm/.pollypm`` back to ``~/.pollypm``.

Refs #1810, #1950, #1954.

Issue #1950, opened 2026-05-20
[regression] ~/.pollypm/.pollypm/ artifact STILL accumulating after PR #1898 fix — cockpit startup writes scaffold via doubled path


samhotchkiss:
## Symptom

PR #1898 (closed #1810) added `_resolve_pollypm_root()` to deduplicate the case where `project_path == GLOBAL_CONFIG_DIR`. After tonight's session cleanup, I manually moved the existing 1.9 GB / 281k-file `~/.pollypm/.pollypm/` tree to a backup. **Then I restarted the cockpit.**

Within 2 minutes, the fresh cockpit had written 12 new files (43.8 KiB) to `~/.pollypm/.pollypm/` again:

```
/Users/sam/.pollypm/.pollypm/transcripts/.ingestion-state.lock
/Users/sam/.pollypm/.pollypm/transcripts/.ingestion-state.json
/Users/sam/.pollypm/.pollypm/docs/rules-manifest.md
/Users/sam/.pollypm/.pollypm/docs/SYSTEM.md
/Users/sam/.pollypm/.pollypm/docs/reference/itsalive.md
/Users/sam/.pollypm/.pollypm/docs/reference/rules-and-magic.md
/Users/sam/.pollypm/.pollypm/docs/reference/commands.md
/Users/sam/.pollypm/.pollypm/docs/reference/tasks.md
/Users/sam/.pollypm/.pollypm/docs/reference/accounts.md
/Users/sam/.pollypm/.pollypm/docs/reference/plugins.md
/Users/sam/.pollypm/.pollypm/docs/reference/sessions.md
/Users/sam/.pollypm/.pollypm/docs/reference/operator-runbook.md
```

`pm doctor` confirms: `! doubled-pollypm-path: ~/.pollypm/.pollypm/ holds 12 files (43.8 KiB, newest 0.0d old)`.

## Likely root cause

The PR #1898 fix only patched `project_*_dir()` helpers in `src/pollypm/projects.py`. But:

- **`ensure_project_scaffold(project_path)` in `src/pollypm/supervisor.py`** (and similar helpers) writes the rules-manifest / SYSTEM.md / docs/reference/ tree directly via `project_path / ".pollypm" / ...` — bypassing `_resolve_pollypm_root()`.
- **transcript ingestion** writes `transcripts/.ingestion-state.{lock,json}` via its own path construction.

[...]

Issue #1810, opened 2026-05-19
pm doctor: 180s runtime + 281k-file doubled-path artifact at ~/.pollypm/.pollypm/ + 5 sqlite-flavored failures during pg cutover


samhotchkiss:
Ran `pm doctor` at tick #20 of the overnight smoketest loop (HEAD 00733c75, pg cutover gate RED, Checks 2+3 still failing on K-source-ripout + K-config-strip). Several distinct V1-RC-relevant findings surfaced; bundling here per the 1-issue-per-tick discipline.

## Bug A — pm doctor takes 180 seconds wall-clock (3 min)

```
pm doctor 2>&1  89.72s user 30.09s system 53% cpu 3:43.83 total
55 checks · 43 passed · 9 warnings · 3 errors  (180.58s)
```

Anyone running `pm doctor` to diagnose a live problem will wait 3 minutes before getting feedback. The diagnose loop becomes "run, wait 3 min, read 200 lines, re-run". For RC, doctor should either be <30s or expose per-section timing so slow checks can be isolated/excluded.

**Likely slow checks** (per the recent tick #14 doctor.py module review enumerating ~750 LOC of sqlite-flavored checks):
- `check_state_migrations` + `check_work_migrations` — open every sqlite DB to read version
- `check_state_db_size` — file-stat the 4.3 GB state.db
- `check_doubled_pollypm_path` — `find` over 281,984 files (see Bug B below)
- `check_task_assignment_sweeper_dbs` — opens all 12 tracked-project state DBs
- `check_db_layout_canonical` — directory walk across all project paths

Half the runtime is probably scanning sqlite artifacts that are about to be deleted by K-source-ripout. A `--fast` mode that skips sqlite-flavored checks, or auto-skip when `storage_backend=postgres` is active, would cut this to <30s and be the right shape post-cutover anyway.

## Bug B — ~/.pollypm/.pollypm/ holds 281,984 files / 1.8 GB and is being actively written to

```
! doubled-pollypm-path: ~/.pollypm/.pollypm/ holds 281962 files (1452392.8 KiB, newest 0.0d old)
```

`du -sh ~/.pollypm/.pollypm` confirms 1.8 GB. `find ~/.pollypm/.pollypm -type f | wc -l` returns 281,984. The artifact is **actively being written to today** (newest = 0.0d) — this is not a stale legacy artifact, it's a live emitter.

Top subdirs by file count:
```
[...]

PR #1980, opened 2026-05-20
chore(sqlite-ripout): PR B — remove SQLiteWorkService + SQLAlchemyStore + EventBuffer (refs #1971)

samhotchkiss:
## Summary

PR B of the 3-PR sqlite-ripout sequence ([#1971](https://github.com/samhotchkiss/pollypm/issues/1971)). The big one: deletes the sqlite work-service, its 7 helper modules, the `SQLAlchemyStore` class, and the `EventBuffer` class. Re-routes the 11 `SQLAlchemyStore(f"sqlite:///...")` constructions to the configured pg backend.

**~9.5K LOC removed across 22 files (235 inserts / 9735 deletes).**

### Deleted modules

- `src/pollypm/work/sqlite_service.py` (4205 LOC) — the legacy work service.
- `src/pollypm/work/service_sync.py`
- `src/pollypm/work/service_dependency_manager.py`
- `src/pollypm/work/service_worker_session_manager.py`
- `src/pollypm/work/service_queries.py`
- `src/pollypm/work/service_dependencies.py`
- `src/pollypm/work/service_transition_manager.py`
- `src/pollypm/work/service_notifications.py`
- `src/pollypm/store/sqlalchemy_store.py` (928 LOC)
- `src/pollypm/store/event_buffer.py` (495 LOC) — only used by `SQLAlchemyStore`.

### Factory

`create_work_service` is now pg-only. `db_path` / `project_path` / `sync_manager` kept as tolerated keyword args so the ~30 callers that still pass them work without behaviour change. `_resolve_backend` always returns `"postgres"`.

### Re-routed construction sites

- `doctor.py:3272` (scheduler-cadence) -> `get_store(load_config())`
- `cockpit_inbox_items.py` x3 -> dead-code-deleted (pg branch already canonical via `cockpit_pg_aggregates`)
- `first_shipped.py` -> `get_store(load_config())`
- `plan_review_emit.py` x3 -> `get_store(load_config())`
- `advisor_tick.py` -> `get_store(config)`
- `audit_watchdog.py` x2 -> `get_store(resolved_config)`, sqlite fallback for config-load failure dropped

### Other touches

- `inbox_cli.py` / `session_runtime.py` `register_backend("sqlite", ...)` calls dropped (activity-feed projector handled in PR C).
- `storage_contracts.py` ACTIVITY_EVENT / TASK / EXECUTION repointed at pg classes.
- `release_gate.py` emitter import switched to `pollypm.work.pg_service`.

## Test plan

[...]

PR #2004, opened 2026-05-20
test(path-resolution): doubled-path precondition + verification (closes #1972)

### LINKED ISSUE #1972: [regression] doubled-pollypm-path returns despite #1954 — root cause analysis

samhotchkiss:
## Summary

Closing slice of the #1972 design fix. Adds an end-to-end verification test pinning all 5 invariants of the design fix so future regressions blow up loudly:

1. **Lint gate baseline is zero.** ``_BASELINE == {}`` in ``tests/test_no_raw_pollypm_path_joins.py`` — after PR #2003 every raw join in ``src/pollypm/`` was migrated. A non-empty baseline would be a future regression vector hidden inside an "allowed" file.
2. **Runtime preconditions raise loudly.** Both ``SQLiteWorkService.__init__`` and ``audit.log._append_line`` raise ``RuntimeError`` with a stable ``doubled-pollypm-path`` message + the ``#1972`` ref — pinned so a refactor that softens to log-and-continue breaks the contract.
3. **Every typed helper collapses on GLOBAL_CONFIG_DIR.** Iterates the full ``pollypm.projects`` helper set (26 functions) and asserts no result contains ``.pollypm/.pollypm``. A future helper added without resolver-routing fails this test loudly.
4. **Audit emit chain doesn't leak.** Direct reproducer of #1966: ``emit(project_path=GLOBAL_CONFIG_DIR)`` lands at ``GLOBAL_CONFIG_DIR/audit.jsonl``, not the doubled path.
5. **State-DB resolution doesn't leak via the runtime_services boot path** (the ``rail_daemon`` reproducer where ``workspace_root=None`` and ``root_dir = ~/.pollypm``).

## Stack

Final PR of the #1972 stack: #1996 → #2002 → #2003 → this. Merge order accordingly.

## Test plan

- [x] ``pytest tests/test_doubled_path_design_fix_verification.py`` — 6/6 pass.
- [ ] **Post-merge follow-up (Sam):** ``uv tool install --force pollypm`` from the merged main + restart ``pm rail_daemon`` + ``pm cockpit`` + ``pm heartbeat``, then ``pm doctor`` to confirm ``doubled-pollypm-path`` doesn't fire. Per #1972's "recommended immediate actions" §2, this is the last step that closes the loop on the user's machine.

## Refs

Closes #1972. Refs #1810, #1950, #1954, #1966, #1996, #2002, #2003.

### DIFF
--- a/tests/test_no_raw_pollypm_path_joins.py
+++ b/tests/test_no_raw_pollypm_path_joins.py
@@ -35,18 +35,31 @@ The migration PRs (PR 2 + PR 3 under #1972) shrink the baseline
 counts to zero. Once everything's zero, this test enforces "no
 raw joins, ever."
 
-Exclusion list
---------------
-The gate is intentionally strict. The only allowed call sites are:
+Line-scoped allowlist (PR 4 under #1972)
+----------------------------------------
+Two files legitimately need ``.pollypm`` joins to define the
+helpers + the canonical constant. **Previously** the gate excluded
+those files wholesale, so any future operational join hidden inside
+them slipped past the ratchet — Sam's review on PR #2003 / #2004
+found exactly that footgun (3 ops joins outside the helper surface).
 
-* ``src/pollypm/projects.py`` — the helper module itself defines
-  ``_resolve_pollypm_root`` and the typed helpers, all of which
-  construct the ``.pollypm`` segment internally. This is the one
-  legitimate constructor.
-* ``src/pollypm/config.py`` — defines ``GLOBAL_CONFIG_DIR =
-  Path.home() / ".pollypm"``, the canonical source of truth that
-  the resolver and helpers reference. Allowed at the constant
-  definition only; other joins in this file should use the helpers.
+The gate now scopes the allowance per-line, not per-file:
+
+* ``src/pollypm/config.py`` — the **only** allowed line is the
+  module-level ``GLOBAL_CONFIG_DIR = Path.home() / ".pollypm"``
+  constant at module top. Any other join in this file is treated
+  exactly like a join in any other file: counted against the
+  ratchet, fails the gate.
+* ``src/pollypm/projects.py`` — joins are allowed **only** inside
+  the typed-helper surface (``_resolve_pollypm_root`` and the
+  ``project_*`` helpers funnelled through it). Joins anywhere
+  else in the file (module-level, inside an unrelated function,
+  inside a future helper that bypasses the resolver) fail the gate.
+
+The allowlist is keyed off the Python AST: a join is allowed if
+its line is inside a function whose name appears in
+``_ALLOWED_FUNCTIONS[<filename>]``, OR if its line matches one of
+``_ALLOWED_LINE_PATTERNS[<filename>]`` for module-level constants.
 
 Per-line escape hatch
 ---------------------
@@ -59,6 +72,7 @@ where no project root exists).
 
 from __future__ import annotations
 
+import ast
 import re
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


samhotchkiss:
Blocking as a verification PR: the new verification says the #1972 baseline is closed, but it still does not catch raw joins left in excluded files. On the stack head, `rg -n '/\s*"\.pollypm"' src/pollypm` still reports operational joins in `config.py` and `projects.py` outside the helper/constant definitions.\n\nPlease extend the verification so `config.py` is only allowed to define `GLOBAL_CONFIG_DIR`, and `projects.py` is only allowed to construct `.pollypm` inside the typed helper/resolver surface. Otherwise the zero-baseline assertion can pass while new doubled-path risks hide inside excluded files.

samhotchkiss:
@samhotchkiss addressed in `881b6e6bc` (force-pushed to `test/doubled-path-precondition`, rebased on the updated `chore/migrate-raw-joins-batch-2`).

Replaced the whole-file exclusion in `tests/test_no_raw_pollypm_path_joins.py` with an AST-driven line-scoped allowlist:

- **`_ALLOWED_FUNCTIONS[<filename>]`** — set of function names. An offending line is allowed only if the AST places it inside one of those functions' span.
- **`_ALLOWED_LINE_PATTERNS[<filename>]`** — compiled regexes matched against the full line text. Only valid for joins outside any function (module-level constants).

`config.py` allowlists ZERO functions and exactly one module-level pattern: the `GLOBAL_CONFIG_DIR = Path.home() / ".pollypm"` constant. Any join inside any function in `config.py` fails the gate — that's the shape of the leak you called out.

`projects.py` allowlists the resolver (`_resolve_pollypm_root`) plus every `project_*` typed helper that chains through it. The `PROJECT_INSTRUCTIONS_TEMPLATE` module-level constant (which points at a path in the package install tree, not a project root) gets an explicit line-pattern allowance. A future helper that bypasses the resolver and joins `.pollypm` raw lands outside the allowlist and fails.

Also added AST-driven docstring detection so prose like `Path.home() / ".pollypm"` inside `global_pollypm_dir.__doc__` doesn't count as an offender (previously hidden by the whole-file skip).

Regression tests added (5 new, all in `tests/test_no_raw_pollypm_path_joins.py`):

- `test_config_py_allows_only_global_config_dir_constant` — a `base_dir = root / ".pollypm"` inside a function in a fake `config.py` fails; `GLOBAL_CONFIG_DIR` on its own does not.
- `test_config_py_module_level_non_global_join_fails` — closes the "rename the constant, keep the join" footgun.
- `test_projects_py_join_outside_helper_function_fails` — the exact shape of `projects.py:796` before PR #2003 migrated it.
- `test_projects_py_join_in_new_unlisted_helper_fails` — guards against drift where someone adds a `project_*` helper that looks typed but skips the resolver.
- `test_projects_py_docstring_join_does_not_trip_gate` — pins the docstring carve-out.

Verified per the spec you asked for in #2004:
- Real-codebase `_current_counts()` returns `{}` after PR #2003's migration.
- Deliberately inserted `LEAKY_TEST_DIR = Path.home() / ".pollypm" / "x"` at module scope in `config.py`; gate caught it (1 offender at line 33). Reverted before commit.
- All 9 tests in the gate file pass locally.
- `test_lint_gate_baseline_is_zero` (in `test_doubled_path_design_fix_verification.py`) still passes — `_BASELINE` is still `{}` because the migrated joins were inside the wholesale-skipped files; the line-scoping is a coverage tightening, not a baseline reset.

samhotchkiss:
Re-review blocker: this stack still inherits the raw source join from #2003 at `src/pollypm/state_cache/refresher.py:93`:

```py
return Path.home() / ".pollypm" / "audit"
```

The line-scoped lint changes in this PR are a good direction, but this PR cannot land while its base still violates the now-zero raw-join contract. Fix #2003 first, then re-run/refresh this stack.

I did not merge this PR. The local targeted pytest run for `tests/test_no_raw_pollypm_path_joins.py tests/test_doubled_path_design_fix_verification.py` hung without output after several minutes in this environment, so I stopped it rather than leaving it running.

samhotchkiss:
@samhotchkiss follow-up: rebased on the updated `chore/migrate-raw-joins-batch-2` (PR #2003 new head `56f9067c9`).

**New head:** `e825d588e`.

The previous comment (`881b6e6bc`) addressed the line-scoping review item. After PR #2003 was rebased on the updated #2002 today, that branch picked up `state_cache/refresher.py` from #2000 (which has an import-cycle audit-dir fallback at `Path.home() / ".pollypm" / "audit"`). The line-scoped allowlist gate from this PR caught it as a regression on the rebased stack — exactly the class of leak you asked the gate to catch, just one we now had to action ourselves before the gate would go green.

Fix landed in #2003 (`56f9067c9`) as a per-line `# noqa: pollypm-path-join` with an inline comment explaining the import-cycle rationale (routing through the typed helper would re-introduce the cycle the lazy `from pollypm.config import ...` guards against).

After rebasing this PR on the updated #2003:
- All 9 lint-gate tests pass (4 original ratchet/pragma tests + 5 new line-scoped regression tests).
- `_current_counts()` returns `{}` against the real `src/pollypm/` tree.
- `test_lint_gate_baseline_is_zero` in `test_doubled_path_design_fix_verification.py` still passes.
- Gate correctly rejected the new `refresher.py` join BEFORE the noqa landed (proving the line-scoped allowlist + new-file-not-in-baseline path actually fires under live conditions).

GitHub now reports `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` on both PRs.

samhotchkiss:
@samhotchkiss — addressed in #2003's commit `56f9067c9` (after your re-review at 16:55Z). The `refresher.py:93` raw join now carries an inline-justified `# noqa: pollypm-path-join` documenting the import-cycle reason (state_cache → audit → state_cache). #2003 should now satisfy the zero-baseline contract; this stack is unblocked.

samhotchkiss:
Re-review status: the line-scoped allowlist changes look good, and I verified `_BASELINE == {}` plus `_current_counts() == {}` on this head. The focused pytest command for the lint file hung in this environment, but direct helper execution and `compileall` passed.

This remains blocked by the stack below it: #2003 is now retargeted to `main` and currently `DIRTY`, so this PR cannot land into `main` yet. Once #2003 is rebased/merged, this PR should be retargeted to `main` and rechecked.

One additional sequencing note: this PR adds `tests/test_doubled_path_design_fix_verification.py`, which still imports `SQLiteWorkService` for the doubled-path precondition check. That is fine against current `main`, but #1980 will need to update/remove this new test too before the sqlite ripout can merge.

PR #2006, opened 2026-05-20
fix(reaper): demote reaped tasks + escalate on 3rd reap (closes #1999)

### LINKED ISSUE #1999: [recovery-cascade] worker.session_reaped on non-terminal task: marker cleared but task stays stuck
## Symptom

Worker marker reaper (`worker_marker_reaper`) detects that a worker's tmux window has disappeared while the task is still `in_progress` / `review` / `queued`, emits `worker.session_reaped` with `reason="tmux window missing for non-terminal task (status='<status>')"`, and deletes the marker file.

But the task itself is NOT re-queued or escalated. Result: the task sits in `in_progress` indefinitely with no worker, no progress, and no recovery action.

## Evidence

Across 12 active projects, ~20 events match this reason over the last 2 days:

```
10× tmux window missing for non-terminal task (status='in_progress')
 5× tmux window missing for non-terminal task (status='review')
 5× tmux window missing for non-terminal task (status='queued')
```

Repeat offender: `itsalive/48` was reaped 3 times back-to-back (review → review → in_progress), suggesting the watchdog re-spawned the worker but the spawn died again.

Sample event:
```json
{
  "ts": "2026-05-18T20:32:36+00:00",
  "event": "worker.session_reaped",
  "subject": "itsalive/48",
  "actor": "worker_marker_reaper",
  "status": "warn",
  "metadata": {
    "window_name": "task-itsalive-48",
    "reason": "tmux window missing for non-terminal task (status='in_progress')"
  }
}
```

## Why this is novel

Distinct from #1974 (advisory watchdog escalation tone) and #1810 (worker churn meta-bug already closed in 0a18a4f). The latter is about churn detection; this is about *what to do* when a non-terminal task's worker dies. The reaper currently only cleans up the marker file — it doesn't surface a recovery hint.

## Recommended fix shape

When `worker_marker_reaper` reaps a marker because the tmux window is missing AND the task is non-terminal:

1. Emit a `task.worker_died` audit event (separate from `worker.session_reaped` which is the marker-cleanup signal).
2. Decide:
   - `status='in_progress'` → demote to `queued` so the next claim cycle picks it up; OR
   - `status='in_progress'` and worker_died count >= 3 → escalate via the watchdog finding pipeline as a new finding type `worker_died_repeatedly`.
3. Make the cockpit Worker pane surface "tmux window missing — last reap N min ago" so an operator can see what happened.

Note (`feedback_no_manual_heartbeat`): the fix is in the recovery loop, not manual `pm task` action. The reaper is the right place; it just needs to do more than delete a marker.

## V1 RC scope

P2. Symptom is silent task stalls visible only via audit grep. Should land before V1 GA but not blocking RC3.

samhotchkiss:
## Summary

Worker-marker reaper now demotes reaped tasks back to `queued` and escalates after three consecutive reaps — closes the #1999 silent-stall regression. Field evidence: 20 `worker.session_reaped` events over 2 days with `itsalive/48` reaped 3× in a row with no recovery signal.

- New `work_tasks.reap_count int NOT NULL DEFAULT 0` column (migration 0004 + initial-schema entry, both idempotent). The counter persists across cockpit restarts.
- `bump_reap_count_and_demote` does one atomic UPDATE: increments `reap_count`, sets `work_status='queued'`, clears `assignee`, returns the post-increment count via RETURNING. Guards against clobbering already-terminal tasks.
- Sam-approved 3-strike policy in the reaper: silent demote on reaps 1 + 2; on reap 3+ also create a `notify`-labelled inbox task addressed to the user (`Task <id> reaped N times — likely systemic issue`), deduped on `reap_repeat:<project>/<task>`.
- Demote + escalate fires only on the `tmux window missing for non-terminal task` branch — terminal-task and missing-row reaps still leave the task untouched.

The existing audit emit and watchdog `worker_session_dead_loop` rule remain in place; this PR adds the persistent counter + operator-visible escalation that the issue called out as missing.

## Test plan

- [x] Added `tests/test_pg_reaper_demote_and_escalate.py` — pg-backed regression suite covering: 1st reap demotes silently, 2nd reap still silent, 3rd reap demotes AND emits the inbox escalation, terminal tasks are not clobbered, missing rows return `None`, non-`window_missing` reasons short-circuit cleanly.
- [x] Migration is idempotent (`ADD COLUMN IF NOT EXISTS`).
- [x] Smoke import of `pollypm.supervisor` + reaper module passes.

### DIFF
--- a/src/pollypm/storage/pg_schema.py
+++ b/src/pollypm/storage/pg_schema.py
@@ -328,6 +328,7 @@ CREATE TABLE IF NOT EXISTS work_tasks (
     kind                    text NOT NULL DEFAULT 'legacy',
     roles                   jsonb NOT NULL DEFAULT '{}'::jsonb,
     external_refs           jsonb NOT NULL DEFAULT '{}'::jsonb,
+    reap_count              int NOT NULL DEFAULT 0,
     created_at              timestamptz NOT NULL,
     created_by              text NOT NULL,
     updated_at              timestamptz NOT NULL,
@@ -748,6 +749,32 @@ ALTER TABLE IF EXISTS tier4_promotion_state
 
 
 # --------------------------------------------------------------------- #
+# 0005 — work_tasks.reap_count counter for #1999.
+# --------------------------------------------------------------------- #
+#
+# #1999: ``worker_marker_reaper`` deletes a stale fresh-launch marker
+# whenever the tmux window has vanished for a non-terminal task, but it
+# never demoted the task back to ``queued`` or surfaced repeat reaps to
+# the operator. Sam-approved policy:
+#
+# * 1st + 2nd reap on the same task → demote silently
+# * 3rd+ reap on the same task → demote AND escalate to the inbox
+#
+# The counter must persist across cockpit restarts (Sam may restart
+# between reaps), so it lives as a column on ``work_tasks``. The
+# reaper UPDATE bumps + reads it atomically via RETURNING.
+#
+# Migration is idempotent: ``ADD COLUMN IF NOT EXISTS`` is a no-op when
+# the column is already present (fresh installs pick it up from
+# ``INITIAL_SCHEMA_DDL``; deployed installs gain it via this migration).
+
+_MIGRATION_0005_WORK_TASKS_REAP_COUNT = """
+ALTER TABLE IF EXISTS work_tasks
+    ADD COLUMN IF NOT EXISTS reap_count int NOT NULL DEFAULT 0;
+"""
+
+
+# --------------------------------------------------------------------- #
 # Migration list — forward-only, append-only.
 # --------------------------------------------------------------------- #
 
@@ -759,6 +786,7 @@ MIGRATIONS: list[tuple[int, str, str]] = [
     (2, "0002_alert_dedupe_tuple", _MIGRATION_0002_ALERT_DEDUPE),
     (3, "0003_account_columns_text", _MIGRATION_0003_ACCOUNT_COLUMNS_TEXT),
     (4, "0004_tier4_terminal_handoff", _MIGRATION_0004_TIER4_TERMINAL_HANDOFF),
+    (5, "0005_work_tasks_reap_count", _MIGRATION_0005_WORK_TASKS_REAP_COUNT),
 ]
 
 
--- a/src/pollypm/storage/work_task_state.py
+++ b/src/pollypm/storage/work_task_state.py
@@ -87,17 +87,19 @@ def bump_reap_count_and_demote(
 
     Called by :func:`pollypm.work.worker_marker_reaper._classify_marker`
     when a fresh-launch marker is reaped because the tmux window has
-    vanished while the task is still non-terminal (#1999). The single
-    UPDATE bumps ``reap_count`` and demotes the task in one round-trip,
-    returning the post-increment count so the caller can decide whether
-    to escalate (3rd+ reap → inbox notify).
+    vanished while the task is in one of the *active/retry* statuses
+    (#1999). The single UPDATE bumps ``reap_count`` and demotes the
+    task in one round-trip, returning the post-increment count so the
+    caller can decide whether to escalate (3rd+ reap → inbox notify).
 
-    The update is conservative: it only fires when the task is still in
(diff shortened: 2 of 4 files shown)


samhotchkiss:
Blocking for merge ordering: this branch and #2007 both add Postgres migration `0004` in `src/pollypm/storage/pg_schema.py`. A merge simulation between the heads conflicts in that file, and even without the textual conflict only one migration can own version 4.\n\nPlease rebase after whichever migration lands first and renumber this migration (likely `0005_work_tasks_reap_count` if #2007 lands first), keeping both schema additions in `INITIAL_SCHEMA_DDL` and the migration list in order.

samhotchkiss:
@samhotchkiss — rebased on current `main` (which now has #2007 merged with v4) and renumbered this migration to **v5**.

**State:**
- New head SHA: `39db80021`
- Base: `main`
- Migration list: `[1, 2, 3, 4 (tier4_terminal_handoff), 5 (work_tasks_reap_count)]` — verified in order, no gaps, no duplicates
- Both schema additions present in `INITIAL_SCHEMA_DDL` for fresh installs (verified via direct import)

Rename mechanics: `_MIGRATION_0004_WORK_TASKS_REAP_COUNT` → `_MIGRATION_0005_WORK_TASKS_REAP_COUNT`, list entry `(4, "0004_work_tasks_reap_count", …)` → `(5, "0005_work_tasks_reap_count", …)`. The schema change itself is unchanged — same column on the same table.

samhotchkiss:
Re-review blocker: `bump_reap_count_and_demote()` currently demotes too broadly. The docstring says this should only affect non-terminal active/retry statuses (`in_progress`, `review`, `queued`, `rework`), but the SQL guard is:

```sql
AND work_status NOT IN ('done', 'cancelled', 'abandoned')
```

Combined with `_classify_marker()` marking any non-terminal row with a missing tmux window as `window_missing_for_non_terminal=True`, a stale marker can move `blocked`, `on_hold`, or `draft` tasks back to `queued` and clear `assignee`. That can undo intentional user/operator state just because the corresponding tmux window is absent.

Please restrict the UPDATE/classification to the intended statuses, or explicitly justify and test the broader behavior. I did not merge this PR. The local PG regression test command (`tests/test_pg_reaper_demote_and_escalate.py -q -x`) hung without output after several minutes in this environment, so I stopped it.

samhotchkiss:
@samhotchkiss — addressed the re-review blocker in 084e97260.

**SQL guard, before/after** (`src/pollypm/storage/work_task_state.py::bump_reap_count_and_demote`):

```diff
-  AND work_status NOT IN ('done', 'cancelled', 'abandoned')
+  AND work_status IN ('in_progress', 'review', 'queued', 'rework')
```

Explicit allowlist. `blocked` / `on_hold` / `draft` are now intentionally excluded, so a stale marker can't undo an operator hold or push a draft into the queue. Docstring rewritten to drop the misleading "non-terminal" framing.

**`_classify_marker()` narrowing** (`src/pollypm/work/worker_marker_reaper.py`):

New `WORKER_MARKER_DEMOTE_ALLOWED_STATUSES = {'in_progress','review','queued','rework'}` constant kept in lockstep with the SQL guard. `window_missing_for_non_terminal` now flips to `True` only when the task's status is in that allowlist — so callers downstream (the `_demote_and_maybe_escalate` flow) never see the flag for non-active statuses, even if the tmux window is missing.

**Regression coverage** added under `TestDemoteAllowlistProtectsNonActiveStatuses` in `tests/test_pg_reaper_demote_and_escalate.py`:

- `test_blocked_task_with_stale_marker_does_not_demote`
- `test_on_hold_task_with_stale_marker_does_not_demote`
- `test_draft_task_with_stale_marker_does_not_demote`
- `test_in_progress_task_with_stale_marker_still_demotes`

Each non-active test asserts: `count is None`, `work_status` unchanged, `assignee` preserved. The `in_progress` test asserts the existing happy path (count == 1, demoted to `queued`, assignee cleared) still fires.

I reproduced your hang locally — `pg_work_service` fixture stalls during pool setup. Collection succeeds (13 tests). Letting CI execute the suite.

New head: `084e97260`.

Issue #1664, opened 2026-05-18
[design] Move A: in-process project-state cache with epoch-driven invalidation


samhotchkiss:
## Context

Parent: #1634 (rail-perf meta).

The 2026-05-18 perf review on #1634 traced the user-felt 12s operator-dashboard cold load and the stuttery rail to **per-project sqlite-open fanout, called from multiple surfaces, with no shared cache**. Conservative count per rail rebuild on a 12-project workspace: ~60-100 sqlite opens. The wedges to date (#1605, #1608, #1630) moved the work off the main thread but did not reduce the wall-clock cost — the profile shows ~50% of asyncio-task time still parked on the rlock handing results back from the worker thread.

The review proposed three architectural moves, prioritized A → C → B. **Move A** is the smallest scope and biggest user-felt win:

> In-process project-state cache with epoch-driven invalidation. ~2 days. Kills 60-100 sqlite opens per refresh.

This issue scopes Move A in implementation detail. The output is a design, not code.

---

## 1. What lives in the cache

One entry per `project_key`, keyed off the workspace config. Per-entry payload (derived from reading `dashboard/operator_view.py`, `dashboard/categorization.py`, `cockpit_inbox.py`, and `cockpit_rail.py`):

```
ProjectStateCacheEntry:
  project_key:           str
  project_path:          Path                  # for project_path-aware queries
  tracked:               bool                  # from project config
  db_paths:              tuple[Path, ...]      # canonical → legacy order

  # ── categorization output ──────────────────────────────────
  state:                 ProjectState          # WAITING | WORKING | IDLE | PAUSED
  glyph:                 str                   # already in _GLYPHS, denormalized
  detail:                str                   # why_waiting() / what_working() / "Quiet"

  # ── rail rollup output ─────────────────────────────────────
  rail_state:            ProjectRailState      # RED | YELLOW | GREEN | WORKING | NONE
  rail_badge:            str | None
  rail_sort_rank:        int
  rail_reason:           str
  approvals_pending:     int                   # #1390 / #1426 — drives "(N⚠)" suffix
  plan_blocked:          bool                  # enforce_plan gate

  # ── awaits-user list (the expensive one) ───────────────────
  awaits_user_count:     int                   # rail badge consumes this length
  awaits_user_items:     tuple[InboxEntry, ...]  # dashboard "Waiting on you" rows

  # ── what_working() inputs (only needed if state == WORKING) ──
  working_agent_name:    str
  working_task_title:    str

[...]

PR #2016, opened 2026-05-20
feat(state-cache): PR 2 route 2 hottest call sites + divergence sampler (refs #1664)

samhotchkiss:
## Summary

Move A PR 2 of 4. Per [docs/design/move-a-state-cache.md §6.2](https://github.com/samhotchkiss/pollypm/blob/main/docs/design/move-a-state-cache.md#62-pr-2--route-the-two-hottest-call-sites), routes the two hottest hot-path helpers through the PR 1 in-process state cache when `POLLYPM_STATE_CACHE=1`:

- **`pm_inbox_awaits_user_list`** (`src/pollypm/cockpit_inbox.py`) — cache fast-path concatenates each entry's `awaits_user_items`; cold cache and flag-off fall through unchanged.
- **`project_state_map_from_config`** (`src/pollypm/dashboard/operator_view.py`) — cache fast-path returns `{key: entry.state for ...}`; falls through when any tracked project is missing from the snapshot.

Both call sites wrap the cache read in a `(entry := cache.get(...)) is not None` check so the §6.2 "Flag off → unchanged behavior" contract holds.

**Divergence sampler.** A 1-in-50 sampled call (per-site counter, configurable via `DivergenceCounter(rate=...)`) runs BOTH the cached and direct paths and emits a `WARN` log line on mismatch — the §8 acceptance gate ("0 WARN lines for 24h") reads this. Comparators are tolerant of trivial ordering differences (inbox items) and enum-vs-string typing (state maps) but strict on identity sets.

**Real refresh function.** PR 1's `stub_refresh_fn` is replaced with `refresh_impl.build_refresh_fn`, which calls the DIRECT (`_pm_inbox_awaits_user_list_uncached`) underlying compute — not the cache-routed public helper. The `_uncached` suffix is now the explicit "do not route through cache; refresher feeds the cache from this" contract. Without that guard the refresher would feed off its own snapshot in a feedback loop.

Closes part of #1664 (PR 2 of Move A; PR 3 routes the rest, PR 4 flips the default).

## Test plan

- [x] 15 new tests in `tests/test_state_cache_parity.py` — parity for both call sites in both flag modes, divergence sampler fires WARN on injected mismatch (both call sites) and stays silent on agreement, three-surfaces-one-predicate invariant from `cockpit_inbox.py:268-295` holds in both modes.
- [x] All 53 PR 1 tests still pass (`test_state_cache.py`, `test_state_cache_entry.py`, `test_state_cache_env_flag.py`, `test_state_cache_refresher.py`).
- [x] Related cockpit/rail tests still pass (`test_cockpit_rail_categorization_cache.py`, `test_operator_dashboard_pilot.py`).
- [x] Smoke import test confirms no circular-import regressions.

## Rollback

`POLLYPM_STATE_CACHE=0` (the default) + cockpit restart returns the system to pre-PR-2 behavior. The wrapper `if (entry := cache.get(...)) is not None:` falls through to the existing direct path with no side effects.

## Notes

- Flag stays **OFF** by default — PR 4 will flip after divergence-sampler telemetry is green for 24 h.
- Rail-rollup fields on the entry (`rail_state`, `rail_badge`, `rail_sort_rank`, `rail_reason`, `approvals_pending`) are populated by the new refresh function even though PR 2 doesn't consume them — PR 3 will be pure wiring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,17 @@ Added, Changed, and Removed.
   Plan card no longer reads anonymously on the very surface that
   invites the user to start a plan. #1540 follow-up.
 
+### Fixed
+- `state_cache` production singleton now constructs
+  `StateCacheRefresher` with a config-backed `project_keys` provider
+  (`lambda: list(load_config().projects.keys())`), so the §9.1 startup
+  full-refresh actually enqueues per-project refreshes. Without the
+  provider, `_initial_full_refresh()` saw `[]` and the cache stayed
+  empty until per-project audit events arrived — meaning workspace-
+  scoped invalidations on a cold cache were no-ops because
+  `ProjectStateCache.invalidate(None)` only iterates known entries.
+  PR #2016 review blocker.
+
 ## [1.0.0] - 2026-04-20
 
 ### Added
--- a/src/pollypm/cockpit_inbox.py
+++ b/src/pollypm/cockpit_inbox.py
@@ -42,6 +42,11 @@ from pollypm.cockpit_worker_identity import (
 )
 from pollypm.heartbeats.snapshots import read_recent_heartbeat_snapshot
 from pollypm.inbox_message_refs import row_project_refs as _row_project_refs
+from pollypm.state_cache.divergence import (
+    DivergenceCounter as _DivergenceCounter,
+    compare_awaits_user_lists as _compare_awaits_user_lists,
+    log_divergence as _log_divergence,
+)
 
 if TYPE_CHECKING:
     from pollypm.inbox.kind import InboxItemKind
@@ -154,6 +159,13 @@ _AWAITS_USER_TTL_SECONDS = 1.0
 _AWAITS_USER_CACHE: dict[int, tuple[float, tuple[object, ...]]] = {}
 
 
+# Move A PR 2 — divergence sampler for the cache-routed fast path
+# (``docs/design/move-a-state-cache.md`` §6.2 last bullet). One
+# counter per routed call site so the sampling rate is local; the
+# busy site doesn't borrow samples from a quiet site.
+_AWAITS_USER_DIVERGENCE_COUNTER = _DivergenceCounter()
+
+
 def pm_inbox_awaits_user_list(config) -> list[object]:
     """Return every inbox entry across the config that ``awaits_user``.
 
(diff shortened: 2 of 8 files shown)


samhotchkiss:

The production singleton wires the real `refresh_fn`, but it still constructs `StateCacheRefresher(cache)` without a `project_keys` provider. `StateCacheRefresher.start()` calls `_initial_full_refresh()`, which reads `_candidate_keys()`, and provider-less refreshers return `[]`. That means the startup full-refresh documented in the design and PR body is a no-op, and workspace-scoped invalidations on an initially empty cache also enqueue nothing because `ProjectStateCache.invalidate(None)` only uses already-known entries/versions.

Repro from this PR head:

```py
import os
os.environ["POLLYPM_STATE_CACHE"] = "1"
from pollypm.state_cache import get_cache, get_refresher, reset_for_test
cache = get_cache()
refresher = get_refresher()
print(cache.snapshot())                         # {}
print(getattr(refresher, "_project_keys", None)) # None
reset_for_test()
```

So with `POLLYPM_STATE_CACHE=1`, the cache can stay empty after startup and never serve the routed hot paths unless every project first happens to receive a project-specific invalidating audit event. Please pass a config-backed project-key provider into the refresher, and add an integration test for `get_cache()` proving the initial full refresh installs entries for configured projects.

samhotchkiss:
@samhotchkiss — Codex blocker fixed in 76753ce.

The production singleton in `pollypm.state_cache.get_cache()` now constructs `StateCacheRefresher(cache, project_keys=_project_keys_provider)` where `_project_keys_provider` returns `list(load_config(DEFAULT_CONFIG_PATH).projects.keys())`. Both `load_config` and `projects` access are wrapped in defensive try/except — failures degrade to `[]` (matches the PR 1 provider-less behavior), so a broken config keeps the cache empty rather than crashing the cockpit.

With the fix, the §9.1 startup full-refresh actually enqueues per-project refreshes for every configured project at boot, so workspace-scoped `invalidate(None)` calls have entries to iterate.

Verification against the exact repro from the review:

```py
>>> import os; os.environ["POLLYPM_STATE_CACHE"] = "1"
>>> from pollypm.state_cache import get_cache, get_refresher, reset_for_test
>>> cache = get_cache()
>>> refresher = get_refresher()
>>> sorted(cache.snapshot().keys())
['bikepath', 'booktalk', 'coffeeboardnm', 'health_coach', 'itsalive', 'media', 'polly_remote', 'pollypm', 'russell', 'samblog', 'savethenovel', 'smoketest']
>>> refresher._project_keys is not None
True
>>> reset_for_test()
```

Integration tests added to `tests/test_state_cache_env_flag.py`:

- `test_singleton_wires_project_keys_provider_for_initial_refresh` — monkeypatches `pollypm.config.load_config` to return a fake config with three projects, then asserts the refresher has a non-None `_project_keys` provider AND `cache.snapshot()` contains entries for all three projects after `get_cache()` returns. This is the integration test you asked for.
- `test_singleton_provider_degrades_gracefully_on_config_failure` — `load_config` raises; provider is still wired, returns `[]`, cache stays empty.

Both pass locally (145s; system was under load). CHANGELOG updated under `Fixed`.

PR #2020, opened 2026-05-20
fix(watchdog): dedupe escalations by root_cause_hash not subject (closes #2015)

### LINKED ISSUE #2015: [watchdog] throttle dedupes by subject — finding-body duplicates leak through
**Surfaced by samblog architect at 13:30PT today** after the recovery cascade restart unblocked it.

**Problem:** `audit_watchdog` throttles escalation dispatch at 30-min per `(project, finding_type, subject)` triple. But the subject changes every time the watchdog mints a new draft task. So when the same root-cause (e.g. "TIER HANDOFF — what structural change do you want to apply?") creates samblog/32, samblog/33, samblog/34, samblog/35 — each is a unique subject, so the throttle doesn't fire across them. Sam gets multiple inbox entries asking the same question with the same evidence (queued_subjects [21,26,29,31], same queued_last_updated timestamps).

**Architect's recommended fix:** dedupe by `root_cause_hash` or finding-body, not by subject. The architect itself handled this by cancelling /33 and /34 as duplicates of /35, but the watchdog should dedupe at emission time.

**Concrete evidence (this morning):** samblog/32, /33, /34, /35 all emit the same TIER HANDOFF finding with identical evidence. /33 and /34 already cancelled as dupes.

**Pointer:** `src/pollypm/audit/watchdog.py` — the dispatch throttle key construction. Currently `(project, rule, subject)`; should be `(project, rule, root_cause_hash)` or similar.

samhotchkiss:
## Summary

- Watchdog dispatch throttle was keyed by `(project, finding_type, subject)`. Same root-cause spawning sibling drafts (samblog/32, /33, /34, /35) bypassed the throttle because each task carries a different subject, so the operator saw four inbox cards asking the same question.
- Added `dispatch_dedup_hash(finding)` in `src/pollypm/audit/watchdog.py` — mirrors `audit/tier4.py:root_cause_hash` but excludes `subject` (rule + project + canonical evidence JSON only).
- Threaded `dedup_hash` through `emit_escalation_dispatched` / `emit_operator_dispatched` (stored in event metadata) and `was_recently_dispatched` / `was_recently_operator_dispatched` (matches on hash, falls back to subject match for pre-#2015 rows so old throttle windows survive the rollout).
- Cadence call sites (`_maybe_dispatch_to_architect`, `_maybe_dispatch_to_operator`) compute and thread the hash.

## Why not fold into `root_cause_hash`?

`audit/tier4.py:root_cause_hash` is a stored DB key in `tier4_promotion_state`. Mutating its inputs would invalidate every persisted row mid-RC. The new helper coexists; tier-4 promotion accounting keeps the subject-bearing hash.

## Surfaced by

samblog architect, acting on tonight's TIER HANDOFF cascade. It emitted the same finding-body four times and the cards landed in the operator inbox with identical `queued_subjects` evidence. The architect manually cancelled /33 and /34 as duplicates of /35, then filed #2015.

## Test plan

- [x] `tests/test_audit_watchdog.py` — six new regression cases (hash symmetry across sibling subjects, hash sensitivity to rule / evidence drift, throttle collapses the samblog/32-35 burst, legacy-row subject fallback, two end-to-end cadence drives for architect + operator legs).
- [x] Direct-import smoke (14/14 + 5/5 backward-compat assertions) — used because the full `tests/test_audit_watchdog.py` suite-level run hangs on this developer machine due to the home-write-guard fixture snapshotting `~/.pollypm/` between tests against a live cockpit. Each new test passes individually under the real conftest.
- [ ] Post-merge: watch the samblog architect cascade and confirm a single `watchdog.escalation_dispatched` row per finding-body even when multiple sibling drafts trigger.

Closes #2015.

### DIFF
--- a/src/pollypm/audit/watchdog.py
+++ b/src/pollypm/audit/watchdog.py
@@ -3036,15 +3036,38 @@ def dispatch_dedup_hash(finding: Finding) -> str:
     ``subject`` / ``message`` / ``recommendation`` (the latter two are
     cosmetic copy).
 
+    **Fallback for detectors without ``evidence`` (PR #2020 review).**
+    Most detector paths populate ``metadata`` but not ``evidence``
+    (``stuck_draft`` being the canonical case). Collapsing on
+    rule+project alone would over-dedupe genuinely different findings
+    that happen to share a rule. So:
+
+    1. If ``finding.evidence`` is non-empty, hash rule + project +
+       canonical evidence JSON (the intended #2015 behavior — collapse
+       sibling subjects whose bodies are identical).
+    2. Else if ``finding.metadata`` carries a stable root-cause key
+       (``root_cause_hash`` or ``dedup_key``), hash that instead. This
+       lets a detector opt into root-cause dedup without copying its
+       evidence shape — used by tier-4 callers that already computed
+       a stable hash upstream.
+    3. Else fall back to including ``subject`` in the hash — i.e.
+       legacy single-finding-per-subject behavior. With nothing
+       structured to compare, subject is the only signal that
+       distinguishes ``demo/1`` from ``demo/2``, and we must not
+       collapse them.
+
     Stable across:
         - Different ``subject`` (the whole point — same root cause
-          across sibling draft tasks).
+          across sibling draft tasks), **iff** ``evidence`` is
+          populated (path 1) or a metadata key is provided (path 2).
         - Different ``message`` / ``recommendation`` text.
         - Re-ordering of keys inside ``evidence``.
 
     Sensitive to:
         - Different ``rule`` or ``project``.
         - Any structural change to the ``evidence`` payload.
+        - ``subject`` when neither ``evidence`` nor a metadata
+          dedup key is present (fallback path 3).
 
     16-hex-char output matches :func:`root_cause_hash` for symmetry;
     collision resistance is fine for the cardinality the throttle sees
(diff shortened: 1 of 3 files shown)


samhotchkiss:

`dispatch_dedup_hash()` can over-dedupe unrelated production findings. It hashes only `rule`, `project`, and `finding.evidence`, but several dispatchable detector paths do not populate `evidence` at all. For those real findings, different subjects collapse to the same hash just because they share rule/project, so the new throttle can suppress distinct watchdog dispatches.

Repro from this PR head:

```py
from pollypm.audit.watchdog import Finding, RULE_STUCK_DRAFT, dispatch_dedup_hash
for subject in ["demo/1", "demo/2"]:
    print(subject, dispatch_dedup_hash(Finding(
        rule=RULE_STUCK_DRAFT,
        project="demo",
        subject=subject,
    )))
# demo/1 d14d2be490dfea41
# demo/2 d14d2be490dfea41
```

The tests create synthetic `evidence` for `stuck_draft`, but the actual stuck-draft detector currently sets `metadata`, not `evidence`, so this regression is not covered. Please either populate stable root-cause evidence in the detector paths that use this throttle or fall back to subject-keyed behavior when no structured evidence is present. Add a regression using findings produced by the real detector, not hand-built findings with extra evidence.

samhotchkiss:
@samhotchkiss thanks for the catch — Codex was right, the original hash over-deduped any detector path that doesn't populate `evidence` (which includes the real `_detect_stuck_drafts` — it sets `metadata`, not `evidence`).

Pushed `d15cd9500` to this PR head. Three-path `dispatch_dedup_hash`:

1. **`evidence` non-empty** → existing #2015 behavior (rule + project + canonical evidence JSON, subject-independent). Sibling-subject collapse for stuck_draft-with-evidence still works.
2. **`evidence` empty AND `metadata` has `root_cause_hash` / `dedup_key`** → hash that key. Lets a detector opt into root-cause dedup without restructuring as evidence.
3. **Both empty** → include `subject`. Legacy single-finding-per-subject behavior. With nothing structured to compare, subject is the only signal that distinguishes findings — so we MUST NOT collapse `demo/1` and `demo/2`.

Canonical strings get a path-tag prefix (`v1|` / `meta|` / `subj|`) so the three paths can't alias each other. Hash output stays 16 hex chars.

New regression tests in `tests/test_audit_watchdog.py`:

- `test_dispatch_dedup_hash_no_evidence_falls_back_to_subject` — Codex's exact repro: same rule+project, no evidence, different subjects → distinct hashes.
- `test_dispatch_dedup_hash_same_subject_no_evidence_stable` — same subject + no evidence → stable hash across cosmetic copy changes.
- `test_dispatch_dedup_hash_evidence_collapses_across_subjects` — pins the #2015 contract.
- `test_dispatch_dedup_hash_metadata_root_cause_key_collapses` — opt-in path 2.
- `test_dispatch_dedup_hash_real_stuck_draft_detector_distinguishes_subjects` — drives the **real** `_detect_stuck_drafts` (not hand-built findings) and asserts no over-dedup. This is the "use the actual detector" regression Codex specifically requested.

Re-running Codex's repro post-fix:

```
demo/1 f3abd6d8d5325984
demo/2 e916d318f5a030b8
```

Distinct hashes — the no-evidence collision is gone, and #2015 sibling-subject collapse for evidence-bearing findings still holds (verified by `test_dispatch_dedup_hash_evidence_collapses_across_subjects`).

PR #2014, opened 2026-05-20
polish(tmux-ui): unified footer status bar (refs #1988)

samhotchkiss:
## Summary

Ships the formatter half of the 2026-05-20 audit's "unified status bar" recommendation as a standalone, side-effect-free helper.

- New leaf module ``src/pollypm/cockpit_footer_status.py`` exposing ``render_footer_status(*, project_count, agent_count, inbox_count, alert=None, width=80) -> str``.
- Output shape: ``12 projects · 38 agents · 23 inbox  |  ⚠ heartbeat 26h offline``.
- Per-state color cues route through ``pollypm.cockpit_theme.State`` (counts → ``MUTED``; inbox flips to ``WAITING`` amber when > 0; alert chunk → ``BLOCKED`` red; separators → ``IDLE`` slate).
- Truncation cascade: full-labels → compact-counts (``12·38·23``) → alert-tail-ellipsis → alert-only layout (when counts + separator can't leave 4 chars of alert body) → empty for ``width <= 0``.
- 11 test functions pinning format, colors, plural noun resolution, width-budget invariants, and the layout-promotion ladder.

**Wiring is intentionally NOT in this PR.** The audit doc flagged the rail's footer rendering loop as the higher-risk surface to touch last; landing the pure formatter + tests independently lets future follow-ups adopt it from any candidate site (``PollyCockpitApp._update_hint``, ``cockpit_rail._format_event_ticker``, headless ``pm rail``) without re-deciding the format or risking the live cockpit Sam is monitoring.

## Test plan

- [x] Direct test runner executes all 11 assertions green: format, color cues, inbox amber/muted toggle, singular/plural nouns, compact-counts activation, ellipsis truncation, width budget for w∈{20,32,50,80,120}, alert-only promotion, ``width<=0`` no-render.
- [x] Module imports clean (no Textual / Rich deps; only stdlib + ``cockpit_theme``).
- [ ] CI pytest run.
- [ ] Future PR wires this into ``PollyCockpitApp._update_hint`` once Sam is at the keyboard to watch the rebuild.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1988, opened 2026-05-20
polish(tmux-ui): consolidate duplicated hex literals into cockpit_theme.py


samhotchkiss:
## Problem

Today the cockpit UI ships the same hex color literals duplicated across multiple modules with no shared source of truth:

- ``#f0c45a`` (amber — waiting / unread) — used in ``cockpit_rail_item``, ``cockpit_ui_inbox_format``, ``cockpit_activity``, ``cockpit_ui`` dashboard CSS
- ``#3ddc84`` (green — working / done) — same set
- ``#ff5f6d`` (red — blocked / dead) — but the inbox plan-review row drifted to ``#ff6b5b``, almost certainly accidental
- ``#5b8aff`` (blue — info / nav highlight), ``#4a5568`` (slate — idle), ``#6b7a88`` (warm gray — muted), ``#f0a030`` (orange — attention)

And the same glyph vocabulary is hard-coded across the same modules:

- ``◆`` — "attention" / unread inbox / project-yellow rollup / dashboard "needs attention"
- ``•`` — live / writing
- ``○`` — idle
- ``▲`` — alert
- ``▶`` — decision / plan review
- ``◇`` — paused
- ``◉`` — review
- ``♥``/``♡`` — pulse
- ``⚠`` / ``✕`` / ``✎`` — session work state
- ``◜◝◞◟`` — arc spinner

This makes palette/glyph audits a multi-file grep, and silent drift (the ``#ff6b5b`` vs ``#ff5f6d`` red) is easy.

## Proposal

Introduce a leaf module ``cockpit_theme.py`` exposing two namespaces:

- ``State`` — semantic state → hex color (WAITING / WORKING / BLOCKED / DONE / IDLE / INFO / MUTED / ATTENTION / HEADING / LABEL)
- ``Glyph`` — semantic role → glyph char (ATTENTION ◆ / LIVE • / IDLE ○ / BLOCKED ▲ / DECISION ▶ / WAITING ◇ / REVIEW ◉ / PULSE_ON ♥ / PULSE_OFF ♡ / STUCK ⚠ / EXITED ✕ / REVIEWING ✎ / SPINNER ◜◝◞◟)

Hex values are the ones already in the cockpit — this is rename-only, not a redesign. The rail's existing ``PALETTE`` (tuple-RGB, for raw rail renderer) stays; ``hex_to_rgb()`` helper bridges.

## Migration plan

[...]

PR #2027, opened 2026-05-21
polish(tmux-ui): wire render_footer_status into live cockpit footer (refs #1988)

samhotchkiss:
## Summary

- #2014 shipped `render_footer_status()` (11 tests + 100 parametrized cases) but deferred the wiring.
- This PR wires `PollyCockpitApp._update_hint` to call the helper with live counts (project/agent/inbox + heartbeat-staleness alert).
- Wrapped in try/except with a backstop fall-through to the legacy inline string build — footer never crashes the cockpit.

## Test plan

- [x] Helper invocation pulls live counts from supervisor + inbox
- [x] Exception in helper → log + fall back to legacy path
- [x] Width-budget invariant from #2014 still holds end-to-end

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -27,6 +27,7 @@ import os
 import resource
 import webbrowser
 from collections import OrderedDict
+from dataclasses import dataclass
 from pathlib import Path
 import subprocess
 import time
@@ -368,6 +369,26 @@ def _wrap_with_hanging_indent(
     if current:
         lines.append(f"{indent}{open_tag}{current}{close_tag}")
     return lines
+
+
+@dataclass(slots=True, frozen=True)
+class FooterStateSnapshot:
+    """Precomputed inputs for the unified footer status bar.
+
+    Resolved off-thread by :meth:`PollyCockpitApp._resolve_footer_state`
+    alongside the rail's ``build_items`` worker, then cached on the
+    app instance so :meth:`PollyCockpitApp._update_hint` can render
+    the footer as a pure formatter without any synchronous supervisor
+    load / inbox fanout / heartbeat probe on the UI thread (#2027).
+
+    Fields mirror the four inputs of
+    :func:`pollypm.cockpit_footer_status.render_footer_status`.
+    """
+
+    project_count: int
+    agent_count: int
+    inbox_count: int
+    alert: str | None
 
 
 class PollyCockpitApp(App[None]):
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Blocking performance regression in the footer wiring: `_update_hint()` now calls `_compose_unified_footer_status()` on every rail item rebuild, and that synchronously does all of this on the UI path:

- `self.router._load_supervisor()`
- counts sessions/projects from the loaded config
- imports and calls `pm_inbox_awaits_user_list(config)`
- probes heartbeat state

`_update_hint()` is invoked from the rail render path after `_update_ticker()`, so this adds an inbox fanout / cache warmup dependency to routine navigation and repaint. That is the same surface where we are already chasing 10+ second click-to-content delays. The legacy hint only did the heartbeat check; it did not synchronously recalculate the awaits-user list for footer text.

Please make the footer consume counts that are already computed for the rail/dashboard, cache this with a bounded TTL/background refresh, or otherwise keep `_update_hint()` as a pure/cheap formatter over precomputed state. Add a regression test that proves `_update_hint()` does not call `pm_inbox_awaits_user_list()` on the hot path.

samhotchkiss:
Addressed in 4c513e9f7. ``_update_hint`` is now a pure formatter over a precomputed ``FooterStateSnapshot``.

**Approach (Option A + B hybrid, minimal new surface):**
- New ``_resolve_footer_state`` runs OFF the UI thread inside the same ``build_items`` worker the rail already runs every refresh (``_refresh_rows_worker``). It does the ``_load_supervisor`` + ``pm_inbox_awaits_user_list`` + heartbeat probe + the dict.len math, packs the result into a frozen ``FooterStateSnapshot``, and stashes it on ``self._footer_state`` via ``call_from_thread``.
- The shared 1s TTL cache on ``pm_inbox_awaits_user_list`` collapses this with the rail badge's own ``_inbox_count`` lookup (``core_rail_items._inbox_count``) so the marginal cost off-thread is one heartbeat probe + a couple of dict-len reads.
- ``_update_hint`` is now ``_format_unified_footer_status(snapshot)``: read width off the live ``hint`` widget, call ``render_footer_status``, done. No supervisor load, no inbox fanout, no heartbeat probe on the UI thread.
- First paint (before the worker has populated ``_footer_state``) renders the legacy hint instead of resolving inputs synchronously — the next ``_apply_built_items`` repaints with the unified bar within ~1 tick (0.8s).
- The legacy backstop still wraps the formatter, and ``_route_status_hint`` still wins.

**Regression test:** ``test_update_hint_does_not_call_pm_inbox_awaits_user_list_on_hot_path`` in ``tests/test_cockpit_footer_status_wiring.py`` monkeypatches ``pollypm.cockpit_inbox.pm_inbox_awaits_user_list`` with a tracking stub, seeds a ``FooterStateSnapshot``, then re-enters ``_update_hint`` 25 times (simulating a j/k click burst) and asserts the fanout call count stays at 0. Plus ``test_update_hint_renders_legacy_when_snapshot_not_yet_populated`` for the first-paint path and ``test_resolve_footer_state_returns_snapshot_with_expected_counts`` for the off-thread resolver semantics.

All 7 wiring tests + 115 leaf footer-status tests pass. Pre-existing ``tests/test_cockpit.py`` internal-error and 2 unrelated failures are not introduced by this commit (verified on origin/HEAD before stash-pop).

PR #2029, opened 2026-05-21
feat(state-cache): PR 4 flip POLLYPM_STATE_CACHE default to ON (closes #1664)

### LINKED ISSUE #1664: [design] Move A: in-process project-state cache with epoch-driven invalidation

samhotchkiss:
## Summary

Move A PR 4 — flips the `POLLYPM_STATE_CACHE` env-flag default
from OFF to **ON**, closing #1664. The env var stays as a
kill-switch for one release per design §6.4.

**Stacked on PR #2026** (feat/move-a-pr3-route-remaining-sites).
Merge that one first.

This is PR 4 of 4 in the Move A sequence:
- PR 1 #2000: cache infra + env flag (OFF by default)
- PR 2 #2016: routed the 2 hottest call sites + divergence sampler
- PR 3 #2026: routed the remaining 5 call sites + removed the 2s TTL
- **PR 4 (this PR)**: flips the default to ON

### Flag default flip (`state_cache/__init__.py`)
- `is_enabled()` now returns True by default. Kill-switch values
  (`0` / `false` / `no` / `off`) still disable.
- Unset / empty / unknown → enabled.

### Divergence sampler no-op (`state_cache/divergence.py`)
- `DivergenceCounter.should_sample()` returns False whenever the
  cache is authoritative (kill-switch NOT set). The
  parity-debugging window closed with this flip; the sampler only
  runs when an operator has explicitly flipped the kill-switch
  (i.e. they're triaging a suspected cache vs direct mismatch).
- New test-only `always=True` constructor escape hatch lets PR 2/3
  parity tests keep injecting mismatches and asserting the WARN
  line lands.

### Design doc (`docs/design/move-a-state-cache.md`)
- Status flipped from **proposed** to **implemented**.
- PR references added (#2000 / #2016 / #2026 / #2027).
- §6.4 ✅-marks the shipped items and ⏳-marks the "remove the
  shim after 14d" follow-up.

### Tests
- `tests/test_state_cache_env_flag.py`:
  - `test_flag_default_is_on` pins the new default.
  - `test_kill_switch_values_disable_flag` pins the rollback path.
  - `test_get_cache_returns_real_cache_by_default` confirms the
    real cache + refresher come up on a flag-unset cold start.
  - New `TestPr4DivergenceSamplerNoop` class (4 tests) pinning the
    sampler-is-noop-when-authoritative contract + the `always=True`
    test escape hatch.
- `tests/test_state_cache_parity.py`:
  - Every `monkeypatch.delenv(POLLYPM_STATE_CACHE)` rewritten to
    `monkeypatch.setenv("POLLYPM_STATE_CACHE", "0")` so "flag off
    → direct path" assertions still hold under PR 4's default.
  - `DivergenceCounter(rate=1)` constructions get `always=True` so
    the mismatch-injection tests still fire.

## Rollback path

Revert this PR. PRs 1-3 stay in tree and dormant — every routed
call site preserves its flag-off fall-through, so an emergency
rollback is a process restart with `POLLYPM_STATE_CACHE=0` set.

## Test plan

- [ ] `pytest tests/test_state_cache_env_flag.py tests/test_state_cache_parity.py tests/test_cockpit_rail_categorization_cache.py`
- [ ] Cold cockpit boot with no env var set — confirm cache + refresher come up.
- [ ] Cold cockpit boot with `POLLYPM_STATE_CACHE=0` — confirm rail / dashboard / inbox render unchanged (direct paths active).
- [ ] 24h production observation with no divergence WARN lines (sampler is silent by design now; what we want to confirm is no rail / dashboard regressions in the wild).

Closes #1664.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,16 @@ Added, Changed, and Removed.
   PR #2016 review blocker.
 
 ### Changed
+- State-cache routing (Move A PR 4, closes #1664) flips the
+  `POLLYPM_STATE_CACHE` env-flag default from OFF to **ON**. The
+  cache is now the authoritative source for the 7 hot-path call
+  sites routed in PRs 2 and 3. The env var stays as a kill-switch
+  for one release: `POLLYPM_STATE_CACHE=0` (or `false` / `no` /
+  `off`) falls back to the direct DB paths the routed call sites
+  preserved. Per design §6.4, the parity-debugging window closed
+  with this flip — the divergence sampler is now a no-op when the
+  cache is authoritative (it only runs when an operator explicitly
+  sets the kill-switch).
 - State-cache routing (Move A PR 3, refs #1664) extends the
   `POLLYPM_STATE_CACHE=1` fast path to the remaining 5 hot-path call
   sites from `docs/design/move-a-state-cache.md` §5. With the flag
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,8 +100,11 @@ Added, Changed, and Removed.
   no `heartbeat.*` invalidation, so any prefetch could only serve a
   stale snapshot. Bulk heartbeat prefetch is deferred to #2050 (once
   the refresher subscribes to heartbeat audit events). Each call
-  site keeps a flag-off fall-through, so behaviour is unchanged
-  until PR 4 flips the default. Flag still defaults OFF.
+  site keeps a fall-through gate (config-identity mismatch, partial
+  cache, workspace-root inbox, actionable rail alerts). As of PR
+  #2029 (PR 4), `POLLYPM_STATE_CACHE` defaults ON; set to 0 to
+  disable. The cache is authoritative per identity-stamped
+  snapshots; rollback path is the kill-switch.
 
 ### Removed
 - Rail-side 2s TTL on `CockpitRouter._project_categorizations`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,13 +76,12 @@ Added, Changed, and Removed.
 - State-cache routing (Move A PR 4, closes #1664) flips the
   `POLLYPM_STATE_CACHE` env-flag default from OFF to **ON**. The
   cache is now the authoritative source for the 7 hot-path call
-  sites routed in PRs 2 and 3. The env var stays as a kill-switch
-  for one release: `POLLYPM_STATE_CACHE=0` (or `false` / `no` /
-  `off`) falls back to the direct DB paths the routed call sites
-  preserved. Per design §6.4, the parity-debugging window closed
-  with this flip — the divergence sampler is now a no-op when the
-  cache is authoritative (it only runs when an operator explicitly
-  sets the kill-switch).
+  sites routed in PRs 2 and 3. After PR4 lands, there is no
+  runtime parity sampler. The cache is authoritative by default.
+  Rollback path is `POLLYPM_STATE_CACHE=0` which makes every
+  routed call site fall through to the direct facade. Parity is
+  covered by tests; runtime divergence telemetry is deferred to
+  future observability work.
 - State-cache routing (Move A PR 3, refs #1664) extends the
   `POLLYPM_STATE_CACHE=1` fast path to the remaining 5 hot-path call
   sites from `docs/design/move-a-state-cache.md` §5. With the flag
--- a/docs/design/move-a-state-cache.md
+++ b/docs/design/move-a-state-cache.md
@@ -9,8 +9,17 @@ The shipped Move A behavior diverges from the original design in three places. T
 - `_maybe_cache_route_awaits_user` / `_maybe_cache_count_awaits_user` decline when the workspace-root inbox has any open message (via `has_workspace_root_open_messages` probe). Workspace-root inbox isn't yet represented in cache entries. Tracked in [#2051](https://github.com/samhotchkiss/pollypm/issues/2051).
 
 Design document for issue [#1664](https://github.com/samhotchkiss/pollypm/issues/1664).
-Status: **proposed** (design only; implementation deferred to post-RC).
+Status: **implemented** (2026-05-20) — PR 1 #2000, PR 2 #2016, PR 3 #2026, PR 4 #2027.
 Authors: Sam, with research from the 2026-05-18 rail-perf review on [#1634](https://github.com/samhotchkiss/pollypm/issues/1634).
+
+> **Implementation note (2026-05-20):** The four-PR rollout from §6
+> landed over the 2026-05-19 → 2026-05-20 window. `POLLYPM_STATE_CACHE`
+> now defaults **ON**; the env var stays as a kill-switch
+> (`POLLYPM_STATE_CACHE=0` falls back to direct paths for one release).
+> The 2s `_project_categorizations` TTL was removed (§9.5). The
+> divergence sampler from PR 2 is silent when the cache is
+> authoritative (kill-switch not set) — the parity-debugging window
+> closed with PR 4.
 
 > **Note (2026-05-19, pre-PR-1):** This doc was written 2026-05-18, before
 > the postgres cutover (#1737) completed. Where the prose says "sqlite open"
@@ -420,10 +429,14 @@ Effort: ~0.5d.
 
 Effort: ~0.5d.
 
-- `POLLYPM_STATE_CACHE` default flips to **on**.
-- Leave the env var as a kill-switch for one release.
+- `POLLYPM_STATE_CACHE` default flips to **on**. ✅ Shipped #2027.
+- Leave the env var as a kill-switch for one release. ✅ Kept.
 - After two weeks of green production telemetry, remove the shim path and
-  the direct-DB fallbacks in the routed call sites.
+  the direct-DB fallbacks in the routed call sites. ⏳ Deferred — tracked
+  as a follow-up after 14-day production observation.
+- Divergence sampler from PR 2 is no-op when cache is authoritative
+  (kill-switch not set). The sampler only emits parity warnings if an
+  operator explicitly flips the kill-switch. ✅ Shipped #2027.
 
 ### 6.5 Rollback
 
--- a/docs/design/move-a-state-cache.md
+++ b/docs/design/move-a-state-cache.md
@@ -444,26 +444,27 @@ Effort: ~0.5d.
 
 - During PRs 1-3: `POLLYPM_STATE_CACHE=0` + cockpit restart returns the
   system to pre-change behavior.
-- After PR 4: rollback is a revert of PR 4 only. PRs 1-3 stay landed and
-  dormant.
+- After PR 4: rollback path is set `POLLYPM_STATE_CACHE=0` in the
+  deployment env + restart pm daemon for one release. Reverting the PR
+  is the heavier escape hatch if config flag rollback isn't sufficient.
 
 ---
 
 ## 7. Risks
 
-> Superseded by PR #2029 (PR4): runtime divergence sampler removed; rollback is the kill-switch.
+> **Superseded by PR #2029 (PR4):** runtime divergence sampler removed; rollback is the kill-switch; risk rows referencing the sampler/PR2-telemetry are historical only — see post-PR4 contract above.
 
 | Risk | Detection | Mitigation |
 |---|---|---|
-| Cached data goes stale (audit-log event dropped, write path bypasses emit) | PR 2's divergence sampler logs `WARN`; `state_epoch.mtime()` advancing without a matching audit event triggers a forced refresh on the next tick | `MAX_STALE_AGE_S=30s` ceiling enforced by §4.3 mtime tiebreaker |
+| Cached data goes stale (audit-log event dropped, write path bypasses emit) *(historical — PR4 removed the sampler)* | PR 2's divergence sampler logs `WARN`; `state_epoch.mtime()` advancing without a matching audit event triggers a forced refresh on the next tick | `MAX_STALE_AGE_S=30s` ceiling enforced by §4.3 mtime tiebreaker |
 | Audit log tailing falls behind under burst (1000s of events in <1s) | New metric `cache.lag_events` (queue depth); alert if sustained >100 | Tail thread coalesces events per project — N events for one project collapse to one invalidation; bounded queue |
-| Cache mismatch on dual-DB legacy/canonical split (#1542 split-brain) | Parity sampler runs both paths through `_open_work_service`'s canonical→legacy walk and confirms the same DB wins | Cache MUST mirror `_open_work_service`'s "first non-empty wins" walk exactly — refresher calls that helper, does not re-implement it |
+| Cache mismatch on dual-DB legacy/canonical split (#1542 split-brain) *(historical — PR4 removed the sampler)* | Parity sampler runs both paths through `_open_work_service`'s canonical→legacy walk and confirms the same DB wins | Cache MUST mirror `_open_work_service`'s "first non-empty wins" walk exactly — refresher calls that helper, does not re-implement it |
 | Memory growth from holding tuples of inbox items | Per-entry size ~10-50KB; 12 projects → ~600KB ceiling; track via `sys.getsizeof` in tests | If memory becomes real (it should not), TTL out paused-project `awaits_user_items` after 5min idle |
 | Thread-safety bug in entry write / read | RLock around all writes; frozen dataclass entries; reads are atomic dict gets | Hypothesis test in `tests/test_project_state_cache_concurrency.py`: N invalidators + N readers, assert no exceptions and no torn reads |
 | Test-suite breakage (~15 files monkeypatch `pm_inbox_awaits_user_list` etc.) | CI | Expected breakage; tests get a `seed_project_state_cache(cache, {...})` fixture instead of monkeypatching the helper. Per [`feedback_format_string_test_grep`](../../docs/conventions.md), grep for the legacy helper names before changing |
 | Multi-process consistency (cockpit + rail_daemon + heartbeat each hold their own cache) | n/a in v1 — Move A is single-process | §9.6 Open Question: defer to Move B if needed |
 | Refresher thread starves the UI under burst | Refresher runs in a dedicated worker thread with bounded work-per-tick; UI reads are dict gets and never block | Profile the refresher tick; cap at e.g. 4 project refreshes per 100ms |
-| Cache lookup returns `None` when caller expected a hit (cold project) | Every call site has a fall-through to the existing direct-DB path during PRs 1-3 | Remove fall-throughs only in PR 4, after parity sampler has been silent for 14 days |
+| Cache lookup returns `None` when caller expected a hit (cold project) *(historical — PR4 removed the sampler)* | Every call site has a fall-through to the existing direct-DB path during PRs 1-3 | Remove fall-throughs only in PR 4, after parity sampler has been silent for 14 days |
 
 ### 7.1 Staleness windows
 
@@ -497,9 +498,11 @@ Pass conditions for "Move A is done":
 3. **Rail rebuild.** `_project_tasks_for_rollup` is no longer called from
    `build_items`. The rail-refresh worker reads only from `cache.snapshot()`
    for project-state data.
-4. **Parity.** 24 hours of production telemetry with PR 2's divergence
-   sampler shows **0** WARN lines (cached and uncached paths agree on
-   every sampled call).
+4. **Parity.** Acceptance criteria for PR #2029:
+   `tests/test_state_cache_parity.py` +
+   `tests/test_state_cache_config_identity.py` pass on a real
+   `~/.pollypm/` workspace with the cache default-on; no runtime
+   divergence telemetry required.
 5. **Test coverage.** New tests pass — `test_project_state_cache_parity.py`,
    `test_project_state_cache_invalidation.py`,
    `test_project_state_cache_concurrency.py`.
--- a/docs/design/move-a-state-cache.md
+++ b/docs/design/move-a-state-cache.md
@@ -9,7 +9,7 @@ The shipped Move A behavior diverges from the original design in three places. T
 - `_maybe_cache_route_awaits_user` / `_maybe_cache_count_awaits_user` decline when the workspace-root inbox has any open message (via `has_workspace_root_open_messages` probe). Workspace-root inbox isn't yet represented in cache entries. Tracked in [#2051](https://github.com/samhotchkiss/pollypm/issues/2051).
 
 Design document for issue [#1664](https://github.com/samhotchkiss/pollypm/issues/1664).
-Status: **implemented** (2026-05-20) — PR 1 #2000, PR 2 #2016, PR 3 #2026, PR 4 #2027.
+Status: **implemented** (2026-05-20) — PR 1 #2000, PR 2 #2016, PR 3 #2026, PR 4 #2029.
 Authors: Sam, with research from the 2026-05-18 rail-perf review on [#1634](https://github.com/samhotchkiss/pollypm/issues/1634).
 
 > **Implementation note (2026-05-20):** The four-PR rollout from §6
@@ -429,14 +429,16 @@ Effort: ~0.5d.
 
 Effort: ~0.5d.
 
-- `POLLYPM_STATE_CACHE` default flips to **on**. ✅ Shipped #2027.
+- `POLLYPM_STATE_CACHE` default flips to **on**. ✅ Shipped #2029.
 - Leave the env var as a kill-switch for one release. ✅ Kept.
 - After two weeks of green production telemetry, remove the shim path and
   the direct-DB fallbacks in the routed call sites. ⏳ Deferred — tracked
   as a follow-up after 14-day production observation.
-- Divergence sampler from PR 2 is no-op when cache is authoritative
-  (kill-switch not set). The sampler only emits parity warnings if an
-  operator explicitly flips the kill-switch. ✅ Shipped #2027.
+- After PR4 lands, there is no runtime parity sampler. The cache is
+  authoritative by default. Rollback path is `POLLYPM_STATE_CACHE=0`
+  which makes every routed call site fall through to the direct facade.
+  Parity is covered by tests; runtime divergence telemetry is deferred
+  to future observability work. ✅ Shipped #2029.
 
 ### 6.5 Rollback
 
@@ -448,6 +450,8 @@ Effort: ~0.5d.
 ---
 
 ## 7. Risks
+
+> Superseded by PR #2029 (PR4): runtime divergence sampler removed; rollback is the kill-switch.
 
 | Risk | Detection | Mitigation |
 |---|---|---|
--- a/src/pollypm/cockpit_inbox.py
+++ b/src/pollypm/cockpit_inbox.py
@@ -162,10 +162,11 @@ _AWAITS_USER_TTL_SECONDS = 1.0
 _AWAITS_USER_CACHE: dict[int, tuple[float, tuple[object, ...]]] = {}
 
 
-# Move A PR 2 — divergence sampler for the cache-routed fast path
-# (``docs/design/move-a-state-cache.md`` §6.2 last bullet). One
-# counter per routed call site so the sampling rate is local; the
-# busy site doesn't borrow samples from a quiet site.
+# Move A PR 4 — cache fall-through gate for the cache-routed fast path
+# (``docs/design/move-a-state-cache.md`` §6.2). No runtime sampling
+# (deferred per #2050). Counter retained as a historical test-only
+# helper; production fall-through is gated by config-identity + partial
+# cache + workspace-root + actionable-rail-alerts checks.
 _AWAITS_USER_DIVERGENCE_COUNTER = _DivergenceCounter()
 
 
--- a/src/pollypm/cockpit_inbox.py
+++ b/src/pollypm/cockpit_inbox.py
@@ -191,13 +191,12 @@ def pm_inbox_awaits_user_list(config) -> list[object]:
     instead of each running their own (~2 pg queries + the plan-review
     filter). Callers receive a fresh ``list`` copy so mutation is safe.
 
-    Move A PR 2 (#1664): when ``POLLYPM_STATE_CACHE=1`` is set AND the
+    Move A PR 2 (#1664): when ``POLLYPM_STATE_CACHE`` is enabled AND the
     in-process cache has at least one populated entry, this function
     short-circuits to concatenating each entry's ``awaits_user_items``
-    and skips the workspace-wide pg sweep entirely. The flag is OFF
-    by default; PR 4 will flip it after divergence-sampler telemetry
-    is green. A 1-in-N sampler runs both paths and logs a WARN on
-    mismatch (see :mod:`pollypm.state_cache.divergence`).
+    and skips the workspace-wide pg sweep entirely. Cache is authoritative
+    by default. ``POLLYPM_STATE_CACHE=0`` forces fall-through. No runtime
+    sampler.
     """
     cached_result = _maybe_cache_route_awaits_user(config)
     if cached_result is not None:
@@ -296,9 +295,8 @@ def _maybe_cache_route_awaits_user(config) -> list[object] | None:
     * Any unexpected exception (defensive — broken cache must never
       crash the rail badge).
 
-    On a hit the divergence sampler MAY also run the direct path and
-    log a WARN if the two disagree. The sampler runs at most 1-in-N
-    so the cache fast-path stays a single dict scan in the common case.
+    Cache lookup; fall through to direct path on cache miss / cache
+    disabled / config-identity mismatch / workspace-root inbox row.
     """
 
     try:
--- a/src/pollypm/dashboard/operator_view.py
+++ b/src/pollypm/dashboard/operator_view.py
@@ -544,21 +544,20 @@ def _scan_to_state(
     )
 
 
-# Move A PR 2 — divergence sampler for the cache-routed fast path
-# (``docs/design/move-a-state-cache.md`` §6.2 last bullet). Per-call-site
-# counter so a busy site doesn't borrow samples from a quiet one.
+# Move A PR 4 — cache-routed fast path for the dashboard state map
+# (``docs/design/move-a-state-cache.md`` §6.2). Cache is authoritative
+# by default; ``POLLYPM_STATE_CACHE=0`` forces fall-through. Per-call-site
+# counter retained for the kill-switch path so a busy site doesn't borrow
+# samples from a quiet one.
 _STATE_MAP_DIVERGENCE_COUNTER = _DivergenceCounter()
 
 
 def _maybe_cache_route_state_map(config) -> dict[str, ProjectState] | None:
     """Return the cache-routed state map, or ``None`` to fall through.
 
-    Returns ``None`` when the env flag is off, when the cache is cold
-    (no entries yet — letting the direct path warm it via the
-    refresher), when the cache lacks a state for any tracked project
-    (the direct path is needed to fill the gap), or on any
-    unexpected exception. The 1-in-N divergence sampler then runs
-    both paths and logs a WARN on mismatch.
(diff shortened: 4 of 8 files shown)


samhotchkiss:
Holding this PR for now because its base PR (#2026) is blocked on cache correctness/performance issues.

This branch flips `POLLYPM_STATE_CACHE` default-on, so it should not land until PR3’s partial-cache count behavior and heartbeat-prefetch implementation are corrected and merged. Otherwise the default would make the undercount/per-project Supervisor-init path authoritative.

Also, once #2026 lands via squash into `main`, please restack/retarget this PR onto `main`; merging it into the PR3 branch would not land the default flip on `main`.

samhotchkiss:
Still blocked by the base PR (#2026).

I re-reviewed the stack today. Since #2026 still needs to remove the `StateStore` / `state_db` heartbeat prefetch from `state_cache.refresh_impl`, this default-on PR should not merge yet. Once #2026 is corrected and merged, please retarget/rebase this PR onto `main` so the default flip can be reviewed against the final pg-only cache path.

samhotchkiss:
Rebased onto current main + retargeted base from `feat/move-a-pr3-route-remaining-sites` to `main`.

- Rebase: `git rebase origin/main` — PR3 commit (already in main as 76e368b74) replayed with conflicts in CHANGELOG.md, cockpit_inbox.py, cockpit_rail.py, dashboard/operator_view.py, state_cache/refresh_impl.py, test_state_cache_parity.py. Resolved via `git rebase --skip` (its content is identical to main's).
- Force-with-lease push: `e46fbf357...e207cec0c`
- Final HEAD: `e207cec0c`
- Diff vs main: only the PR4 flip-default commit — 6 files (`CHANGELOG.md`, `docs/design/move-a-state-cache.md`, `src/pollypm/state_cache/__init__.py`, `src/pollypm/state_cache/divergence.py`, `tests/test_state_cache_env_flag.py`, `tests/test_state_cache_parity.py`), +219/-29.
- Tests: `pytest tests/test_state_cache*.py` → **117 passed**.
- Mergeable: `MERGEABLE/CLEAN` against main.

samhotchkiss:

I fetched latest remote state, checked PR comments and GraphQL review threads (none), verified the base PR #2026 is merged and this PR merges cleanly onto current `main`, reviewed the cache/default-on invariant, and ran targeted tests:

- `uv run --with pytest python -m pytest tests/test_state_cache_env_flag.py tests/test_state_cache_parity.py tests/test_state_cache_config_identity.py tests/test_cockpit_rail_categorization_cache.py -v` -> 83 passed

Required before merge:

1. The PR's divergence-sampler contract is not enforced by the production routed helpers. `DivergenceCounter.should_sample()` now only samples when `_kill_switch_set()` is true (`src/pollypm/state_cache/divergence.py:91-98`), and the PR body/docs/tests say this preserves parity telemetry when an operator sets `POLLYPM_STATE_CACHE=0`. But every production cache route returns before reaching the sampler when the kill-switch is set, for example `cockpit_inbox._maybe_cache_route_awaits_user()` returns at `src/pollypm/cockpit_inbox.py:304-309` before the sampler at `:365`, and `operator_view._maybe_cache_route_state_map()` returns at `src/pollypm/dashboard/operator_view.py:564-569` before the sampler at `:621`. The same pattern applies to the other routed cache helpers.

   Please make this invariant explicit one way or the other: either remove the unreachable kill-switch sampling branch/tests/docs and state that PR4 disables divergence sampling entirely once the cache is authoritative, or add routed-helper coverage and implementation that actually compares cache vs direct while in debug/kill-switch mode without making the cache authoritative. As written, the tests only prove the standalone counter behavior, not the operational behavior they claim to protect.

2. The in-code architecture docs still contradict the default-on behavior in the modules this PR changes. `src/pollypm/state_cache/__init__.py:6-21` still says no call sites are wired, `POLLYPM_STATE_CACHE` defaults off, and the real cache starts only when the flag is on. `src/pollypm/state_cache/divergence.py:1-15` still describes every routed call site as sampling 1-in-N calls and running both paths. These comments describe the system invariant operators/reviewers will rely on during rollback and incident triage, so they need to be updated or removed as part of the default flip.

The functional cache guards from #2026 still look directionally sound: cold/partial cache, config-identity mismatch, workspace-root inbox rows, and actionable rail alerts all fall through to the direct path. The blocker is that this PR leaves the new default-on/kill-switch/telemetry contract unclear and partly untested at the production call-site level.

samhotchkiss:
## Codex blockers addressed in d9832af5d

**Fix 1 — Unreachable divergence sampler branch.**

`src/pollypm/state_cache/divergence.py:82-98` (was): `should_sample()` only sampled when `_kill_switch_set()` was true. But all five routed call sites (cockpit_inbox list/count, operator_view, cockpit_rail rollups, project_state_map_from_config) guard on `is_enabled()` and return `None` BEFORE reaching the sampler when the kill-switch is set — making the branch unreachable in production.

- Removed the `_kill_switch_set()` check in `should_sample()` (and the `_kill_switch_set()` helper itself). `should_sample()` is now deterministic-off in production; `always=True` is preserved as the test escape hatch so parity tests can still drive the comparison helpers.
- `DivergenceCounter`, `compare_awaits_user_lists`, `compare_state_maps`, `log_divergence` are all retained — they're used by the parity test suite for bulk equivalence checks.

**Fix 2 — Stale docstrings.**

- `src/pollypm/state_cache/__init__.py:1-31` — updated from "no call sites wired / defaults off" to the post-PR4 contract: defaults ON, five call sites wired, kill-switch fall-through, cache authoritative, no in-process sampling.
- `src/pollypm/state_cache/divergence.py:1-19` — updated from "each routed call samples 1-in-N and runs both paths" to "historical / no-op; parity is verified by tests, not runtime sampling." Links #2029.
- `DivergenceCounter` class docstring (`divergence.py:46-66`) updated to reflect deterministic-off behavior.

**Tests.**

`tests/test_state_cache_env_flag.py`:
- Renamed `TestPr4DivergenceSamplerNoop` → `TestDivergenceSamplerNoop` with updated docstring.
- Replaced `test_sampler_runs_when_killswitch_set` (which asserted `[F,F,T,F,F,T]` under kill-switch — now impossible) with `test_sampler_is_silent_when_killswitch_set` (asserts deterministic-off under kill-switch too, since routes short-circuit before the sampler is reached).
- Renamed `test_always_override_ignores_killswitch` → `test_always_override_drives_legacy_cadence`; assertion unchanged.
- Tests 1, 2 (`silent_by_default`, `silent_when_flag_explicitly_on`) unchanged — behavior is preserved.

**Verification.** `pytest --noconftest tests/test_state_cache*.py` → 117 passed.

samhotchkiss:

I fetched latest remote state after #2047 merged, checked PR comments and GraphQL review threads (none), locally merged #2029 onto latest `origin/main`, reviewed the state-cache default-on contract, and ran targeted verification:

- `uv run --with pytest python -m pytest tests/test_state_cache_env_flag.py tests/test_state_cache_parity.py tests/test_state_cache_config_identity.py tests/test_cockpit_rail_categorization_cache.py tests/test_state_cache_no_sqlite_imports.py -v` -> 86 passed

The code change in `DivergenceCounter.should_sample()` now makes the sampler deterministic-off in production unless `always=True`, and production never sets `always=True`. That resolves the previous unreachable kill-switch sampling branch, but the documented operator contract is still split across old and new stories.

Required before merge:

1. Update the user/operator docs to match the actual sampler behavior. `CHANGELOG.md:76-85` and `docs/design/move-a-state-cache.md:437-439` still say the divergence sampler runs when an operator sets `POLLYPM_STATE_CACHE=0`. It does not: `src/pollypm/state_cache/divergence.py:82-97` returns `False` unless `always=True`, and routed call sites return before sampling when `is_enabled()` is false. Please state the real contract: after PR4 there is no runtime parity sampler; rollback is the kill-switch/direct path, and parity is covered by tests or whatever follow-up observability replaces the old WARN-line window.

2. Clean up stale design references that would mislead rollback/incident triage. `docs/design/move-a-state-cache.md:12`, `:432`, and `:439` mark PR4 as `#2027`; this PR is `#2029` and #2027 is unrelated. The risk/acceptance sections still lean on PR2 sampler telemetry too (`docs/design/move-a-state-cache.md:454-456`, `:462`), so they need the same post-PR4 correction or an explicit superseded note.

3. Update the production call-site docs/comments that still describe the pre-flip invariant. Examples: `src/pollypm/cockpit_inbox.py:194-200` says the flag is off by default and PR4 will flip it after sampler telemetry; `src/pollypm/cockpit_inbox.py:299-301` and `src/pollypm/dashboard/operator_view.py:547-561` still describe an active 1-in-N sampler. Those comments sit on the hot-path helpers operators will inspect during cache incidents, so they need to say default-on + kill-switch fall-through + no runtime sampling. The test module header has the same stale “PR1 ships cache OFF by default” story at `tests/test_state_cache_env_flag.py:1-10`.

4. Minor code hygiene while touching the flag plumbing: `src/pollypm/state_cache/__init__.py:88-89` leaves `_flag_truthy()` unused after the default flip. Please remove it or use it so the flag parser has one clear source of truth.

The important invariant this PR touches is now: cache is authoritative by default, every routed helper must still fail open to the direct path under `POLLYPM_STATE_CACHE=0`, and there is no runtime compare-against-direct safety net. The tests I ran support the functional pieces of that invariant, but the docs/comments still describe a different operational contract.

samhotchkiss:
Codex round-3 cleanup pushed as `e339c4850` — cosmetic / doc-only, no behavior change.

**Fix 1 — CHANGELOG + design said the sampler runs on the kill-switch.**
- `CHANGELOG.md:76-84` — rewrote the Move A PR 4 entry to the post-PR4 contract: cache authoritative by default, `POLLYPM_STATE_CACHE=0` falls through to the direct facade, parity covered by tests, runtime telemetry deferred.
- `docs/design/move-a-state-cache.md:437-441` — same rewrite under §6.4.

**Fix 2 — Design doc referenced wrong PR number + leaned on PR2 sampler telemetry.**
- `docs/design/move-a-state-cache.md:12` — `PR 4 #2027` → `#2029`.
- `docs/design/move-a-state-cache.md:432` — `Shipped #2027` → `#2029`.
- `docs/design/move-a-state-cache.md:441` — `Shipped #2027` → `#2029`.
- `docs/design/move-a-state-cache.md:454` (new line) — added `> Superseded by PR #2029 (PR4): runtime divergence sampler removed; rollback is the kill-switch.` immediately under the §7 Risks header so the table's stale PR2-sampler references read as historical.

**Fix 3 — Source-code comments stale.**
- `src/pollypm/cockpit_inbox.py:194-199` — `pm_inbox_awaits_user_list` docstring no longer claims the flag is OFF by default or that PR4 is pending sampler telemetry.
- `src/pollypm/cockpit_inbox.py:298-300` — `_maybe_cache_route_awaits_user` docstring replaced the 1-in-N sampler paragraph with: "Cache lookup; fall through to direct path on cache miss / cache disabled / config-identity mismatch / workspace-root inbox row."
- `src/pollypm/dashboard/operator_view.py:547-560` — same stale-sampler description on `_maybe_cache_route_state_map`. Comment now reflects the post-PR4 contract. Note: `_STATE_MAP_DIVERGENCE_COUNTER` was retained because it is still referenced by the live kill-switch sampler at line ~618 — removing it would break runtime.
- `tests/test_state_cache_env_flag.py:1-12` — module docstring updated to "Post-PR4: cache defaults ON. These tests pin the kill-switch (`POLLYPM_STATE_CACHE=0`) fall-through path and the env-flag parser."

**Fix 4 — `_flag_truthy` removed.**
- `src/pollypm/state_cache/__init__.py:88-89` — `grep -rn "_flag_truthy" src/ tests/` returned only the definition itself and the helper is not in `__all__`. Removed.

Re-applied `needs-review`.

samhotchkiss:
Still blocking after `e339c485`; not merging this run.

I fetched latest remote state, checked PR comments and GraphQL review threads (none open), locally merged #2029 onto current `origin/main`, and re-ran the state-cache regression slice:

- `uv run --with pytest python -m pytest tests/test_state_cache*.py -v` — 117 passed.
- `git diff --cached --check` — clean.

The executable behavior is in better shape: default-on, kill-switch fall-through, no sqlite/runtime legacy imports in `refresh_impl`, and the production `DivergenceCounter.should_sample()` no-op are all covered by the tests above. The blocker is still the same invariant clarity issue: the PR makes the cache authoritative by default and removes runtime parity sampling, but several source/docs/test comments still describe the old sampler/default-off/rollback contract.

Required before merge:

1. Finish cleaning the production comments around the retained counters. `src/pollypm/cockpit_inbox.py:165-169` still says this is a divergence sampler with local 1-in-N sampling, and `src/pollypm/dashboard/operator_view.py:547-552` says the counter is retained for the kill-switch path. That is no longer true: `DivergenceCounter.should_sample()` is deterministic-off unless `always=True`, and kill-switch routing returns before the sampler is reached. `src/pollypm/state_cache/divergence.py:39-41` still references the old §8 “0 WARN lines for 24h” gate, and `src/pollypm/state_cache/divergence.py:209-215` still says PR4 will promote the WARN line to an alert. These comments need to be historical/test-only, not active operational guidance.

2. Fix the remaining default-off/source comment drift. `src/pollypm/dashboard/operator_view.py:784-788` still says `POLLYPM_STATE_CACHE=1`, flag OFF by default, and PR4 flips after divergence-sampler telemetry is green. This PR is PR4; after merge, that comment would be wrong on the hot-path helper operators inspect during cache incidents.

3. Update the design doc’s operational sections, not just the top note. `docs/design/move-a-state-cache.md:445-448` says after PR4 rollback is reverting PR4 only, while the new PR4 contract and code say rollback is `POLLYPM_STATE_CACHE=0` plus restart for one release. `docs/design/move-a-state-cache.md:500-502` still requires 24h of production telemetry from the PR2 sampler, and risk rows such as `docs/design/move-a-state-cache.md:458`, `:460`, and `:466` still describe the removed sampler / removing fall-throughs after PR4. A single “superseded” block at `:454` is too easy to miss; please rewrite or explicitly mark those rows as historical so the recovery path and acceptance criteria are not split.

4. Clean up the changelog/test docs that still claim the old runtime sampler/default-off story. `CHANGELOG.md:103-104` now sits in the same release section as the PR4 default-on entry but still says behavior is unchanged until PR4 and “Flag still defaults OFF.” `tests/test_state_cache_parity.py:1-11` and `tests/test_state_cache_parity.py:373-374` still frame the sampler WARN as PR4 telemetry; those tests now use `always=True` to exercise historical/test-only helpers, not runtime sampling.

The invariant to preserve is clear now: cache is authoritative by default; every routed helper must keep `POLLYPM_STATE_CACHE=0` fall-through; there is no runtime compare-against-direct safety net. The code/tests mostly express that, but the surrounding contract text still tells two different stories.

samhotchkiss:
Round-4 doc/comment drift cleanup (commit 9e2c2d3a3):

**Fix 1 — Production source comments rewritten to reflect post-PR4 contract:**
- `src/pollypm/cockpit_inbox.py:165-169` — "divergence sampler" → cache fall-through gate, no runtime sampling (deferred per #2050)
- `src/pollypm/dashboard/operator_view.py:547-552` — counter now described as historical test-only helper; production is cache + kill-switch + identity fall-throughs
- `src/pollypm/state_cache/divergence.py:39-41` — removed "§8 0 WARN lines for 24h"; now "Historical / test-only. Production no-ops as of PR4 (#2029)"
- `src/pollypm/state_cache/divergence.py:209-215` — removed "PR 4 will promote WARN to alert"; now "Pre-PR4 design rationale; superseded"

**Fix 2 — operator_view.py:784-788 default-off drift:** rewrote to state `POLLYPM_STATE_CACHE` defaults ON as of PR #2029, set to 0 for incident rollback, with the per-route fall-through gates listed.

**Fix 3 — `docs/design/move-a-state-cache.md`:**
- `:445-448` rollback path — kill-switch as primary, PR revert as heavier escape hatch
- `:454` superseded note expanded with explicit "risk rows are historical only" pointer
- `:458, :460, :466` — inline `(historical — PR4 removed the sampler)` annotations on each affected risk row
- `:500-502` acceptance criteria #4 — replaced "24h sampler telemetry" with parity-test contract on a real `~/.pollypm/` workspace

**Fix 4 — CHANGELOG + tests:**
- `CHANGELOG.md:103-104` — merged into PR4 default-on entry; "flag still defaults OFF" tail removed
- `tests/test_state_cache_parity.py:1-11` module docstring — sampler/WARN/PR4-telemetry language removed; describes `always=True` as test-only helper
- `:373-374` `TestDivergenceSampler` docstring — same correction (test-only helper, not runtime sampling)

6 files changed, 42 insertions(+), 31 deletions(-). Cosmetic only — no runtime behaviour change.

PR #2030, opened 2026-05-21
feat(doctor): --fix action for doubled-pollypm-path moves legacy artifacts to .bak

samhotchkiss:
## Summary

`pm doctor` already warns about the legacy `~/.pollypm/.pollypm/` doubled-path tree that pre-#1972 config renders stranded on disk. This adds the `--fix` hook so the operator can drain the historical artifacts via the normal doctor flow instead of running the suggested `mv` command by hand. Refs #1972.

- **`pollypm.doctor_cleanup_doubled_path`** (new helper) — moves the doubled tree to `~/.pollypm/.pollypm.bak-YYYYMMDD-HHMMSS/`. **Always MOVES, never deletes** — operator can `rm -rf` the backup once verified.
- **Safety guards**: refuse if target missing / is a symlink / is a file / backup path already exists / direct children look like a real config root (`audit/`, `agent_homes/`, `plugins/`, `state.db`, `pollypm.toml`, ...). Symlinks are never followed during the scan.
- **Wiring**: `check_doubled_pollypm_path` now exposes `fixable=True` + `fix_fn=...` so the standard `pm doctor --fix` machinery (including the `verify_fix_results` cross-check) picks it up. No new CLI subcommand — the existing per-check `fix_fn` pattern is the right grain.
- **Audit**: emits `cockpit.doubled_path_artifacts_cleaned` (with moved-file count, size, backup path) on successful move; no-op on refusal/dry-run.
- **Dry-run**: helper accepts `dry_run=True` for a non-mutating scan + summary; tests assert filesystem byte-identity before/after.

## Test plan

- [x] Unit tests for the helper: parametrized refusal across every name in `ANOMALOUS_DIRECT_CHILDREN`, refusal on missing/symlink/file targets, refusal on pre-existing backup, normal move, dry-run non-mutation, audit-event emission.
- [x] Wiring tests for `check_doubled_pollypm_path`: `fix_fn` is present and `fixable=True` when the warning fires; end-to-end fix relocates the tree and the post-fix re-check passes; anomalous contents propagate the helper's refusal.
- [x] Existing baseline tests in `tests/test_doctor_dual_db_health.py` still pass (`warns_with_files`, `warns_when_empty`, `pass_when_absent`) — verified by direct invocation.
- [x] Ruff clean on new files.
- [ ] **Operator step (Sam runs this himself, not in CI)**: `pm doctor --fix` on the live `~/.pollypm/.pollypm/` to drain the real artifacts.

### DIFF
--- a/src/pollypm/doctor.py
+++ b/src/pollypm/doctor.py
@@ -1735,6 +1735,26 @@ def check_doubled_pollypm_path() -> CheckResult:
         }
     )
 
+    # The ``--fix`` handler MOVES (never deletes) the doubled tree to
+    # a timestamped sibling. Helper carries the safety guards (refuses
+    # if direct children look like a real config root; refuses on
+    # symlink target; never follows symlinks) so the rule is testable
+    # in isolation. See :mod:`pollypm.doctor_cleanup_doubled_path`.
+    def _fix() -> tuple[bool, str]:
+        from pollypm.doctor_cleanup_doubled_path import (
+            cleanup_doubled_path,
+            emit_cleanup_audit_event,
+            format_summary,
+        )
+
+        plan = cleanup_doubled_path(doubled)
+        if plan.refused_reason is not None:
+            return (False, plan.refused_reason)
+        if not plan.moved:
+            return (False, format_summary(plan))
+        emit_cleanup_audit_event(plan)
+        return (True, format_summary(plan))
+
     # Empty-directory case — still flag (the path should not exist at
     # all) but report it as a tiny, informational warn.
     if file_count == 0:
@@ -1748,9 +1768,12 @@ def check_doubled_pollypm_path() -> CheckResult:
             fix=(
                 f"Remove the empty directory —\n"
                 f"  rmdir {doubled}\n"
+                "Or run:  pm doctor --fix   # moves it to a .bak sibling\n"
                 "Recheck: pm doctor"
             ),
             severity="warning",
+            fixable=True,
+            fix_fn=_fix,
             data=data,
         )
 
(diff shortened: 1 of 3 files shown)


samhotchkiss:
Blocking issue in the doubled-path cleanup guard: the anomaly list refuses a case the helper itself says should be safe to move.

`ANOMALOUS_DIRECT_CHILDREN` includes `"pollypm.toml"`, but the comment immediately above the set says the config TOML can legitimately appear under the doubled path from the original bad renders and is safe to relocate as part of the cleanup. With the current code, `pm doctor --fix` will refuse to clean a common legacy artifact shape (`~/.pollypm/.pollypm/pollypm.toml`) instead of moving it to the backup sibling.

Please either remove `pollypm.toml` from the anomalous set and add a test proving that doubled-path TOMLs are moved, or update the design/comment/tests to explain why that file is now considered live-state and should block cleanup. As written, the implementation and safety rationale disagree, and the user-facing fix will no-op on a plausible artifact.

samhotchkiss:
Fixed in 6dd8b55f2.

Removed `"pollypm.toml"` from `ANOMALOUS_DIRECT_CHILDREN` in `src/pollypm/doctor_cleanup_doubled_path.py:57` — the comment above the set was the correct rationale (doubled-path TOMLs are legacy artifacts from pre-#1972 bad renders; the real config lives at `~/.pollypm/pollypm.toml`, one level up). The implementation now matches the comment.

Added regression test `test_cleanup_moves_doubled_path_pollypm_toml` in `tests/test_doctor_cleanup_doubled_path.py` that plants `~/.pollypm/.pollypm/pollypm.toml`, runs the helper, and asserts:
- no refusal,
- the doubled tree is gone,
- the TOML is present in the `.pollypm.bak-YYYYMMDD-HHMMSS/` sibling with contents intact.

The existing parametrized `test_cleanup_refuses_when_real_config_subdirs_present` iterates the set so it automatically dropped `pollypm.toml` coverage — also cleaned up its now-dead `.toml` branch in the test planter. All 18 tests in the file pass.

Issue #2023, opened 2026-05-21
[watchdog] role_session_missing fires 456x for worker role with structurally no-op self-heal


samhotchkiss:
## Symptom

`pollypm.jsonl` in last 24h: **456 `role_session_missing` findings**, **0 `audit.worker_lane_spawned` events**. Three queued tasks (pollypm/36, /109, /113) each fire the finding 152 times in 14h — every ~5.5min — with no recovery action ever taken.

Sample:
```
{ event:"audit.finding", subject:"pollypm/36",
  metadata:{
    rule:"role_session_missing",
    role:"worker",  status:"queued",
    expected_window:"worker-pollypm",
    message:"Task pollypm/36 is at status=queued but no 'worker-pollypm'
             window exists in the storage-closet — the role agent is not running.",
    recommendation:"Spawn a `worker` for pollypm ..."
  }
}
```

Same pattern on **samblog (376/24h), media (153/24h)**.

## Root cause

The detector ↔ self-heal contract is inconsistent for `role == "worker"`:

1. **Detector** (`src/pollypm/audit/watchdog.py:1444-1530`) fires for `queued` tasks falling through to default role=`"worker"` (lines 1476-1482) when:
   - The project is **not** at `max_parallel_workers` cap (back-pressure suppression only applies *at* cap, line 1492), AND
   - No `worker-<project>` window exists in the storage closet.

2. **Self-heal** (`src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py:1475-1481`) for `role == "worker"` is a **structural no-op**:
   ```python
   if role == "worker":
       # Per-task workers are provisioned automatically by ``pm task
       # claim``; spawning a long-running ``worker-<project>`` would
       # leak memory ...
       return counters
   ```

Net result: the detector fires every cycle, the self-heal returns `{worker_lane_spawned: 0}`, and the finding goes to escalation (`watchdog.escalation_dispatched` at 806 events/24h on pollypm) where the architect / operator gets pinged about an unactionable item.

## Why this is novel

- Existing #1737 added back-pressure suppression *only* for the at-cap case.
- The legitimate case the detector wants to catch (worker session crashed, task still in_progress) is **valid** for `in_progress` tasks — but for `queued` tasks the per-task spawn-on-claim path is the recovery, not a long-running `worker-<project>` window.

## Suggested fix

Skip the finding when `role_used == "worker"` AND `status_value == "queued"` AND no per-task worker window already exists, since:
[...]

PR #2032, opened 2026-05-21
feat(audit): per-project audit.jsonl rotation + retention (refs #2023)

samhotchkiss:
## Summary

- Per-project `audit.jsonl` files (and the central-tail mirror) now rotate when they exceed a configured size — coffeeboardnm was logging 5550 `stuck_draft` events / 24h; samblog was past 10 MB. Without rotation those grow forever.
- New `[audit]` config section: `rotate_size_mb` (50), `retention_count` (4), `disable_rotation` (false). All knobs route through `_resolve_rotation_policy` with module-default fallbacks, so a fat-fingered config never silently kills rotation.
- Atomic rotation via `os.rename` -> gzip into `.gz.partial` -> rename to `.gz`, then unlink the in-flight `.rotating` file. Bump-suffix loop checks BOTH names so two rotations in the same second cannot overwrite an archive.
- Rotation never blocks the append. Every IO failure inside `_maybe_rotate` is caught + logged; the outer `_append_line` wraps the call in a broad try so a housekeeping bug can never cause audit data loss.

## Rotation flow

1. `os.stat` live file; skip if under threshold.
2. `os.rename(audit.jsonl, audit.jsonl.<ts>[.bump].rotating)` — atomic, no half-state.
3. `gzip` into `<final>.gz.partial`.
4. `os.rename` partial -> `.gz`.
5. unlink the `.rotating` file.
6. prune `.gz` archives beyond `retention_count` (oldest by mtime).

## Defaults + rationale

- **50 MB threshold** — large enough for real incident scrollback in the live file; small enough that 5550 events/day cannot fill a disk before rotation kicks in. ~150 bytes/event * 5550/day -> ~830 KB/day, so a normal project rotates a few times a year while a runaway loop self-bounds in hours.
- **4 retention** — 50 MB live + 4 * ~10 MB gz = ~90 MB ceiling per project, well-bounded for a workstation tool while keeping enough history to reconstruct a multi-day incident.
- **disable_rotation escape hatch** — for incident debugging that needs a continuous file. Env var `POLLYPM_AUDIT_DISABLE_ROTATION=1` exists for tests / CI without mutating `~/.pollypm/pollypm.toml`.

## Test plan

- [x] `tests/test_audit_log.py` — 19/19 pass (5 new rotation tests; full suite green without conftest):
  - rotation fires past threshold, live file truncated
  - gz archive contains readable JSONL chunk of pre-rotation events
  - retention=N caps surviving archives at N over many rotations
  - rotation failure (forced `os.rename` boom) doesn't break append
  - `POLLYPM_AUDIT_DISABLE_ROTATION=1` short-circuits rotation
- [x] `tests/test_audit_watchdog.py` + `tests/test_contract_audit.py` + `tests/test_audit_log_docs.py` — 134/134 pass
- [x] `tests/test_no_raw_pollypm_path_joins.py` — 9/9 pass (doubled-path guard still firing)
- [x] `tests/test_config.py` — 23 pass / 1 fail (pre-existing storage-default failure, unrelated)
- [ ] Manual: drop into samblog with this build; confirm `audit.jsonl` rotates the next time it crosses 50 MB and `.gz` is readable via `zcat`.

Refs #2023.

### DIFF
--- a/src/pollypm/audit/log.py
+++ b/src/pollypm/audit/log.py
@@ -778,27 +778,99 @@ def _iter_log_lines(path: Path) -> Iterable[dict[str, Any]]:
     mid-write) can leave a truncated final line. Skip those rather
     than crashing the reader — the heartbeat must keep working
     even when the log has a junk tail.
+
+    Routes ``.gz`` archives through :func:`gzip.open` in text mode so
+    callers walking a rotation chain (live ``.jsonl`` + ``.gz``
+    siblings) get the same line-iteration shape regardless of file
+    format. See :func:`_walk_log_chain` for the chain producer.
     """
     if not path.exists():
         return
     try:
-        with open(path, "r", encoding="utf-8") as fh:
-            for line in fh:
-                stripped = line.strip()
-                if not stripped:
-                    continue
-                try:
-                    obj = json.loads(stripped)
-                except json.JSONDecodeError:
-                    # Truncated / corrupt line — skip silently. We
-                    # do not log here because a single bad tail line
-                    # would otherwise spam the error log on every
-                    # heartbeat tick.
-                    continue
-                if isinstance(obj, dict):
-                    yield obj
+        if path.suffix == ".gz":
+            fh = gzip.open(path, "rt", encoding="utf-8")
+        else:
+            fh = open(path, "r", encoding="utf-8")
     except OSError as exc:
         logger.warning("audit.read: open failed for %s: %s", path, exc)
+        return
+    try:
+        for line in fh:
+            stripped = line.strip()
+            if not stripped:
+                continue
+            try:
+                obj = json.loads(stripped)
+            except json.JSONDecodeError:
+                # Truncated / corrupt line — skip silently. We
+                # do not log here because a single bad tail line
+                # would otherwise spam the error log on every
+                # heartbeat tick.
+                continue
+            if isinstance(obj, dict):
+                yield obj
+    except OSError as exc:
+        logger.warning("audit.read: read failed for %s: %s", path, exc)
+    finally:
+        try:
+            fh.close()
+        except Exception:  # noqa: BLE001 — close-time errors are noise
+            pass
+
+
+def _archive_sort_key(path: Path) -> tuple[float, str]:
+    """Sort key for ``audit.jsonl.<ts>[.bump].gz`` archives.
+
+    We want newest-first iteration. Use mtime as primary because the
+    rotation timestamp is embedded in the filename but tests + edge
+    cases can leave clock-skew; fall back to filename for stable
+    ordering within the same mtime second.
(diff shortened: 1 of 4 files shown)


samhotchkiss:
Blocking issue: rotation hides historical audit events from the existing `read_events()` API.

`_maybe_rotate()` moves `audit.jsonl` into `audit.jsonl.<ts>.gz`, but `read_events()` still reads only the live `.jsonl` file (or the central live tail fallback). It never walks the new `.gz` siblings. That changes behavior for existing consumers like the watchdog, SSE, morning briefing, and doctor checks: once a file rotates, recent events that landed in the rotated chunk disappear from `read_events(..., since=..., event=...)` until enough new events accumulate in the live file.

This is especially risky for watchdog/escalation dedupe windows because a rotation can make the prior `watchdog.escalation_dispatched` row invisible and allow duplicate dispatches.

Please make `read_events()` rotation-aware too: walk live first plus `.gz` archives newest-first, apply `since`/`event` filters across the chain, and stop once the requested `limit` / time window is satisfied. Add a regression test where an event exists only in `audit.jsonl.<ts>.gz` and `read_events()` still returns it.

samhotchkiss:
Fixed in a997cd412 (pushed to `feat/audit-log-rotation`).

`read_events()` is now rotation-aware:

- New `_walk_log_chain(live)` yields the live `.jsonl` first then `.<ts>[.bump].gz` siblings sorted newest-first by mtime (with filename as the in-second tiebreak).
- `_iter_log_lines()` now opens `.gz` paths through `gzip.open(..., "rt")` so callers iterate the same JSON-line shape regardless of file format.
- `read_events()` walks the chain, applies `event` / `since` filters per file, then re-sorts the combined results globally by `ts` so cross-file ordering stays chronological even under mtime / clock skew.

Helpers duplicated (intentionally) from PR #2036's `cli_features.audit` walker — `audit/log.py` is stdlib-only to stay import-cycle-safe, and pulling in `cli_features` (typer) would break that contract.

Regression test (the one you asked for, plus two siblings):

- `test_read_events_walks_gz_archives_after_rotation` — seeds a `watchdog.escalation_dispatched` event into `audit.jsonl.<ts>.gz`, leaves the live `audit.jsonl` empty, calls `read_events(event="watchdog.escalation_dispatched", since=<old_iso>)` and asserts the archived row is returned (with metadata intact).
- `test_read_events_chains_live_and_gz_in_chronological_order` — events split across `.gz` + live return in `ts` order, not file order.
- `test_read_events_walks_gz_archives_for_central_tail` — central-tail fallback (no `project_path`) also walks `.gz` archives, so workspace-level consumers (SSE, doctor) see rotated history too.

`pytest --noconftest tests/test_audit_log.py -v` → 22/22 passing.

Issue #2035, opened 2026-05-21
[tests] _snapshot_guarded_dirs autouse fixture hangs pytest for minutes on long-lived dev machines


samhotchkiss:
## Summary

Every pytest invocation runs the `_pollypm_home_write_guard` autouse fixture in `tests/conftest.py`, which snapshots a recursive file-set of every guarded subdirectory under `~/.pollypm/` BEFORE and AFTER each test. On any dev machine that has been driving PollyPM for a while, `~/.pollypm/snapshots/` accumulates millions of files (operator checkpoints, agent snapshots, history). Every test pays an O(snapshots-tree) walk twice.

## Measured baseline (this workstation)

```
~/.pollypm/snapshots         : 1,395,189 files
~/.pollypm/homes             :    25,599 files
~/.pollypm/artifacts         :    15,866 files
~/.pollypm/transcripts       :       416 files
~/.pollypm/agent_homes       :       412 files
```

The `_GUARD_SNAPSHOT_FILE_CAP = 5000` safety net exists, but `Path.rglob('*')` still materialises every dir entry up to the 5000th *file* hit, with an extra `is_file()` stat syscall per entry. On a 1.4M-file `snapshots/` tree this is minutes per test, twice (pre + post). A 1000-file synthetic tree runs the current impl in ~38 ms; scaled to the real shape this is the long-reported "pytest hangs for ages" symptom.

Every overnight agent has flagged this as blocking test verification.

## Chosen fix

Option A + Option B combined:

1. **Metadata-only snapshot via `os.scandir`** — record `(rel_path, kind, mtime, size)` tuples instead of `Path` objects. Single stat per entry; no `Path.__init__` overhead; set-diff stays correct because tuples are hashable.
2. **Shallow-scan for known append-only dirs** — `snapshots/`, `transcripts/`, `homes/`, `agent_homes/`, `worktrees/` snapshot only top-level children. The #1902 leak shape (a fresh `operator/` checkpoint dir appearing under `artifacts/`) is still caught: deep dirs that are not append-only (`artifacts/`, `checkpoints/`, `dossier/`) still recurse, and a new file deep inside an append-only dir surfaces because its parent's mtime / a new top-level entry will diff.
3. Reconstruct an absolute `Path` from the diff tuple + subdir root before passing to `_looks_test_induced` / `_file_written_during_test`, preserving the existing token / mtime attribution.

## Acceptance

[...]

PR #2039, opened 2026-05-21
feat(cli): pm sessions health summary subcommand (refs #2012)

samhotchkiss:
## Summary
- New top-level `pm sessions [--health] [--json]` command (refs #2012) — lists every configured session with its live tmux + heartbeat state, the Lever-2 `session.auth_token` trust signal, and (with `--health`) the pane current command + pid + log mtime.
- Read-only: no Supervisor spin-up; uses the existing `TmuxClient.list_windows` against the storage-closet session + the backend-aware `latest_heartbeat` facades (pg via `pollypm.storage.pg_heartbeats`, sqlite via `StateStore.latest_heartbeat`).
- `--json` emits one JSON object per line (machine-readable). Disabled sessions are filtered out.

## Status thresholds
- `healthy` — heartbeat <= 5 min old AND tmux window present (matches the cockpit-inbox "stuck pane" cutoff).
- `stale` — heartbeat older than 5 min (window present).
- `missing` — configured `window_name` not in tmux (wins over `stale`).
- `unknown` — no heartbeat row recorded yet (window present).

## Invocation examples
```
pm sessions
pm sessions --health
pm sessions --json | jq 'select(.token == "missing")'
```

## Test plan
- [x] 17 unit tests in `tests/test_sessions_health_cli.py` (classifier thresholds, humanize, token classification, end-to-end CLI text/json/health, disabled-session filtering). Validated via `pytest --noconftest tests/test_sessions_health_cli.py` (0.12s, all pass).
- [x] Direct Python CliRunner smoke against `pollypm.cli.app` exercises every status bucket (healthy / stale / missing / unknown) and confirms `TOKEN ok|missing` matches `session.auth_token` presence.

The local autouse home-write-guard fixture (#2035 / PR #2037) is still the bottleneck for a full conftest-driven pytest run; tests verified with `--noconftest` since this CLI does not write to `~/.pollypm/` at all.

### DIFF
--- a/src/pollypm/cli.py
+++ b/src/pollypm/cli.py
@@ -59,6 +59,7 @@ from pollypm.cli_features.migrate import register_migrate_commands
 from pollypm.cli_features.projects import register_project_commands
 from pollypm.cli_features.send_up import register_send_up_commands
 from pollypm.cli_features.session_runtime import register_session_runtime_commands
+from pollypm.cli_features.sessions_health import register_sessions_health_command
 from pollypm.cli_features.storage import register_storage_commands
 from pollypm.cli_features.ui import register_ui_commands
 from pollypm.cli_features.update import register_update_commands
@@ -233,6 +234,7 @@ register_storage_commands(app)
 register_worker_commands(app)
 register_tier4_commands(app)
 register_session_runtime_commands(app, helpers=sys.modules[__name__])
+register_sessions_health_command(app)
 register_send_up_commands(app, helpers=sys.modules[__name__])
 register_web_api_commands(app)
 
--- a/src/pollypm/cli_features/sessions_health.py
+++ b/src/pollypm/cli_features/sessions_health.py
@@ -110,25 +110,15 @@ def _classify_status(
 def _latest_heartbeat(config, session_name: str):
     """Return the most-recent heartbeat record for ``session_name``.
 
-    Backend-aware (#1737 / #1811): postgres installs read through the
-    pg facade; sqlite installs fall back to the legacy ``StateStore``.
-    Returns ``None`` when no heartbeat row exists OR the read fails —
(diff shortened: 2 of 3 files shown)


samhotchkiss:
Blocking issue from review: this adds a new runtime path back through the legacy store.

`src/pollypm/cli_features/sessions_health.py::_latest_heartbeat()` dispatches on `is_pg_backend(config)` and, for non-pg configs, falls back to `pollypm.store.get_store()` / legacy `StateStore`; the docstring also explicitly says “sqlite installs fall back.” We’ve been enforcing the no-sqlite-runtime direction, so new code should not add another backend-dispatch branch into the legacy store.

Please make this command read through the Postgres heartbeat facade only and degrade to `unknown`/`None` on lookup failure. Also remove the sqlite wording and any tests that encode the fallback.

samhotchkiss:
Addressed in cfd76676ff538c8e92c8ebdf178b7f27d573f094.

- `src/pollypm/cli_features/sessions_health.py::_latest_heartbeat()` no longer dispatches on `is_pg_backend(config)`. The function reads through `pollypm.storage.pg_heartbeats.latest_heartbeat` only and returns `None` on any exception — downstream `_classify_status` already maps that to `unknown`.
- Docstring rewritten: sqlite wording removed; states pg-only + degrades to `None` on failure.
- No existing tests encoded the sqlite fallback contract (all CLI tests already monkeypatch `_latest_heartbeat` directly, so nothing to delete).
- Added `TestLatestHeartbeatPgOnly::test_pg_lookup_failure_degrades_to_unknown` — asserts a raising pg facade surfaces as `status="unknown"` end-to-end with no crash.
- `pytest --noconftest tests/test_sessions_health_cli.py -v` → 18 passed.

PR #2040, opened 2026-05-21
feat(cli): pm storage report + prune for ~/.pollypm/ bloat visibility

samhotchkiss:
## Summary
- `pm storage report` — tabular per-subdir breakdown of `~/.pollypm/` (file count, bytes, oldest/newest mtime, NOTES). Flags unbounded growth, orphaned worktrees, and audit rotation. `--sort=bytes|files|mtime`, `--json`.
- `pm storage prune <target> --older-than <spec>` — targeted cleanup for snapshots / transcripts / worktrees (orphans only) / homes (completed-agent only). Dry-run by default; `--yes` to actually delete (no interactive prompt — works via tmux send-keys).
- `os.scandir`-based scans with a 50k cap on `snapshots/` so the report stays fast even on already-broken installs (the exact failure mode PR #2037 surfaced: 1.4M snapshot files wedging the test suite).

## Why now
PR #2037 caught `~/.pollypm/snapshots/` silently grown to 1.4M files / 38 GB. There was no surface that would have shown Sam *which* subtree was growing unbounded — `du -sh` was itself too slow on the broken state to be useful preventatively. This is that surface, plus targeted cleanup.

## Prune-safety guarantees
1. `--dry-run` never modifies the filesystem.
2. `--yes` required for destructive runs; no interactive prompt.
3. Without `--dry-run` or `--yes`, the command refuses (exit 2).
4. Files newer than `--older-than` never touched.
5. Worktrees / homes whose agent is in-progress never touched (work-service lookup, degrades to "no known live agents" when unreachable).

## Test plan
- [x] `tests/test_cli_storage_report.py` — 47 cases, all passing locally.
- [x] Helpers: byte/mtime formatting, `--older-than` parsing (positive units, malformed, negative, zero).
- [x] `scan_pollypm_home`: counts, bytes, mtime, symlink containment, orphan flag, audit rotation flag, cap-hit unbounded flag.
- [x] CLI: human + JSON report, sort modes, missing-home graceful exit, prune refuse-without-flags, dry-run preserves files, `--yes` deletes only old, threshold filtering, all four targets, JSON shape, bad-target rejection.
- [x] Existing `tests/test_bootstrap_pg_cli.py` (10 cases) + `tests/test_cli_reference.py` / `test_cli_help_examples.py` (39 cases) still pass — `register_storage_commands` wiring + `storage_app` help refresh haven't broken adjacent surfaces.
- [x] Smoke: `pm storage report --home <tmp>` and `pm storage prune snapshots --older-than 1d --dry-run` produce the expected tabular / sample output.

## Files
- `src/pollypm/cli_features/storage.py` — extends the existing `storage_app` with `report` + `prune` commands (alongside `migrate-to-pg`).
- `tests/test_cli_storage_report.py` — new test module.

### DIFF
--- a/src/pollypm/cli_features/storage.py
+++ b/src/pollypm/cli_features/storage.py
@@ -36,11 +36,17 @@ Contract
 
 from __future__ import annotations
 
+import json
+import os
 import platform
+import re
 import shutil
 import subprocess
-from dataclasses import dataclass
+import time
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
 from pathlib import Path
+from typing import Iterable
 
 import typer
 
@@ -51,6 +57,7 @@ from pollypm.config import DEFAULT_CONFIG_PATH, load_config, resolve_config_path
 __all__ = [
     "bootstrap_pg",
     "register_storage_commands",
+    "scan_pollypm_home",
     "storage_app",
 ]
 
@@ -573,11 +580,15 @@ def bootstrap_pg(
 
 storage_app = typer.Typer(
     help=help_with_examples(
-        "Storage backend tooling (sqlite ↔ postgres migration).",
+        "Storage backend tooling (migration + ~/.pollypm/ disk usage).",
         [
             (
-                "pm storage migrate-to-pg",
-                "preview the sqlite → postgres data migration (dry-run)",
+                "pm storage report",
+                "show ~/.pollypm/ disk usage by subdir (bytes, files, mtime)",
+            ),
+            (
+                "pm storage prune snapshots --older-than 14d --dry-run",
+                "preview snapshot pruning without deleting anything",
             ),
             (
                 "pm storage migrate-to-pg --commit --yes",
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Blocking issue: `pm storage prune worktrees/homes --yes` can delete live agent directories when live-agent detection fails or when task statuses are enums.

`_count_live_agent_worktrees()` returns `set()` on config/work-service/listing failures, and both prune paths interpret that empty set as “nothing is live.” That means a machine with an unreachable work service can treat every stale `worktrees/agent-*` or `homes/agent-*` dir as a delete candidate. This also contradicts the `_iter_prune_candidates_homes()` docstring, which says an unreachable service gets nothing pruned.

There is a second live-detection miss here: `status not in ("in_progress", "claimed", "assigned")` does not normalize `WorkStatus` enum values, so active tasks represented as `WorkStatus.IN_PROGRESS` are ignored and their dirs can be pruned.

Please distinguish “live set unknown” from “known empty live set” and refuse/skip worktree/home pruning when unknown. Also normalize statuses with `getattr(status, "value", status)` or equivalent, and add tests for service-unreachable and `WorkStatus` enum cases before this ships.

samhotchkiss:
Addressed in 678a5be32.

Fixes:

1. `_count_live_agent_worktrees()` return contract is now `Optional[set[str]]` — `None` = live set unknown (config / work-service / listing failed), `set()` = service reachable but no active agents, populated set = real live agents.
2. `_iter_prune_candidates_worktrees` and `_iter_prune_candidates_homes` raise a new `_LiveSetUnknownError` sentinel when the live set is `None`. The CLI surface (`storage_prune`) catches it, prints `live-agent detection failed — refusing to prune worktrees/homes`, and exits with code 3. An unreachable work service can no longer cause live `agent-*` dirs to be deleted.
3. Status normalization via `getattr(status, "value", status)` so `WorkStatus.IN_PROGRESS` (and the other enum members) are recognised by the filter — previously the comparison was against raw enum members and never matched.
4. `scan_pollypm_home` (visibility-only path, never deletes) collapses `None` to `set()` — over-flagging the NOTES badge is the safe direction there.

New tests (all passing, 50/50 in `tests/test_cli_storage_report.py`):

- `test_prune_worktrees_refuses_when_live_set_unknown` — monkeypatches `_count_live_agent_worktrees` to `None`; asserts exit code ≠ 0, "refusing to prune" in output, and that the stale dir is NOT deleted.
- `test_prune_homes_refuses_when_live_set_unknown` — same shape for the homes path.
- `test_prune_normalizes_WorkStatus_enum` — builds a fake task list containing `WorkStatus.IN_PROGRESS` and `WorkStatus.DONE`; asserts `agent-live` is in the live set, `agent-completed` is not, and the IN_PROGRESS agent's stale worktree is NOT a prune candidate.

Two pre-existing tests (`test_prune_worktrees_skips_fresh`, `test_prune_homes_completed_agents`) were updated to monkeypatch the detector to `set()` (known-empty live set) — they were previously relying on the unsafe `unknown == empty` fallback that this fix removes.

PR #2044, opened 2026-05-22
feat(chat-api): P1 transcript readers + surface registry

samhotchkiss:
## Summary

Implements **P1** of the chat-endpoints spec (`~/Desktop/pollypm-chat-endpoints-spec.md` §6) — the data layer that powers the HTTP chat endpoints that ship in P2. **No HTTP routes** in this PR.

New package `src/pollypm/web_api/chat/`:

- **`envelope.py`** — `MessageEnvelope` dataclass with 8 type discriminators (`text`, `tool_use`, `tool_result`, `thinking`, `ask_user`, `file`, `subagent_spawn`, `subagent_result`, `system_event`) per spec §3.
- **`transcripts.py`** — parses the normalized `events.jsonl` archive emitted by `TranscriptIngestor` into envelopes. Drops `token_usage` / `session_state` (not chat messages). Detects subagent pairs via Task `tool_use` + `task-notification` result blocks. Resolves the right archive subdir by `cwd` match → most-recent mtime fallback.
- **`tmux_capture.py`** — fallback reader using `tmux capture-pane -p -S -3000` when the archive is missing or >60s stale (spec §4.7). Synthesizes deterministic `cap_<hex>` envelope ids from `blake2b(session_name : line_index : content)`. Strips ANSI escapes.
- **`registry.py`** — enumerates all four chat surface types (operator/architect/advisor/worker) into `ChatSurface` dataclasses with `session_name`, `surface_type`, `persona`, `project`, `task_id`, tmux window state, and resolved transcript path. Worker enumeration reads `WorkerSession` rows from work-service. Tmux probing is one `list_windows` call cached per request.

## Subagent linking decision

Implemented **detection** (spawn + result envelopes paired via `tool_use_id` as `subagent_id`, with `task-notification` block fields surfaced in result metadata). **Deferred** the recursive inline `subagent_transcript[]` payload that the spec's `?include_subagents=true` flag requires — that needs `TranscriptIngestor` to preserve a clean `output-file` reference per Task `tool_result`, which it currently keeps inside the raw payload but not as a first-class field. P1 spec calls out: "decide if subagent linking is feasible in P1 or needs an ingestor enhancement (file a follow-up issue if so, don't try to land both in P1)." Filing a follow-up issue rather than touching the ingestor in this PR.

## Spec gaps noticed while implementing

1. The spec assumes `<project>/.pollypm/transcripts/<session_name>/events.jsonl` exists with `session_name` matching the tmux session name. In practice the ingestor names subdirs by the provider-internal session UUID (Claude `sessionId` / Codex `session_meta.id`), not the tmux session_name. `resolve_transcript_path` works around this by preferring `cwd` match then most-recent mtime.
2. Codex tool-call payload shape isn't pinned in the spec; the parser renders a generic `[<tool_name>]` summary and preserves the raw payload in metadata. Deeper Codex tool parsing can land in P3 alongside the send-side Codex work.
3. AskUserQuestion `selections_no_question` / `selections_invalid` errors (spec §2.3) belong on the POST endpoint in P3 — not the read path.

## Tests shipped

**111 cases**, all green (well above the ≥40 case goal):

- `tests/test_chat_transcripts.py` — 54 cases. All 8 envelope types, Claude + Codex shapes, stale detection, cwd-match resolution, malformed-line tolerance, envelope id stability.
- `tests/test_chat_registry.py` — 31 cases. All 4 surface types, persona fallback, tmux probing, disabled / non-chat filtering, deterministic sort order, `to_dict` serialization.
- `tests/test_chat_tmux_capture.py` — 26 cases. ANSI stripping, deterministic id synthesis, empty / error handling, capture-depth wiring.

Run: `uv run --with pytest python -m pytest tests/test_chat_*.py --noconftest` → `111 passed in 1.40s`.

## Constraints honoured

- No ingestor changes (spec gap flagged for follow-up issue)
- No HTTP routes (P2)
- No `StateStore` imports
- No new pg paths
- v1 RC stability — pure read-side, only adds new files

## Test plan

- [x] Unit tests for every MessageEnvelope discriminator
- [x] Unit tests for surface enumeration across all 4 types
- [x] Unit tests for tmux capture fallback
- [x] Manual: package imports cleanly (`from pollypm.web_api.chat import ...`)
- [ ] Integration: wired into FastAPI in P2 (next PR)

### DIFF
--- a/src/pollypm/web_api/chat/__init__.py
+++ b/src/pollypm/web_api/chat/__init__.py
@@ -3,9 +3,10 @@
 P1 ships the read-side primitives the chat endpoints (P2) compose:
 
 - :mod:`envelope` — the uniform :class:`MessageEnvelope` dataclass and
-  the eight type discriminators (``text``, ``tool_use``, ``tool_result``,
-  ``thinking``, ``ask_user``, ``file``, ``subagent_spawn``,
-  ``subagent_result``, ``system_event``) per spec §3.
+  the type discriminators (``text``, ``tool_use``, ``tool_result``,
+  ``ask_user``, ``file``, ``subagent_spawn``, ``subagent_result``,
+  ``system_event``) per spec §3. ``thinking`` is deferred — the
+  ingestor doesn't preserve provider thinking blocks yet.
 - :mod:`registry` — enumerates every live chat surface (operator,
   architect, advisor, worker) into :class:`ChatSurface` dataclasses,
   resolving the tmux window + transcript path for each.
--- a/src/pollypm/web_api/chat/__init__.py
+++ b/src/pollypm/web_api/chat/__init__.py
@@ -1,0 +1,63 @@
+"""Chat-surface data layer for the PollyPM Web API.
+
+P1 ships the read-side primitives the chat endpoints (P2) compose:
+
+- :mod:`envelope` — the uniform :class:`MessageEnvelope` dataclass and
+  the eight type discriminators (``text``, ``tool_use``, ``tool_result``,
+  ``thinking``, ``ask_user``, ``file``, ``subagent_spawn``,
+  ``subagent_result``, ``system_event``) per spec §3.
+- :mod:`registry` — enumerates every live chat surface (operator,
+  architect, advisor, worker) into :class:`ChatSurface` dataclasses,
+  resolving the tmux window + transcript path for each.
+- :mod:`transcripts` — parses the normalized ``events.jsonl`` archive
+  emitted by :mod:`pollypm.transcript_ingest` into envelopes.
+- :mod:`tmux_capture` — fallback reader using ``tmux capture-pane`` when
+  the archive is absent or stale (>60s).
+
+No HTTP routes here — P2 (#TBD) wires these primitives into FastAPI.
+Per the chat-endpoints spec (``~/Desktop/pollypm-chat-endpoints-spec.md``)
+§6, the layered split keeps the parser unit-testable and lets the
+router be a thin adapter.
+"""
+
+from __future__ import annotations
+
+from pollypm.web_api.chat.envelope import (
+    MessageEnvelope,
+    MessageRole,
+    MessageType,
+)
+from pollypm.web_api.chat.registry import (
+    ChatSurface,
+    SurfaceType,
+    enumerate_chat_surfaces,
+    enumerate_config_surfaces,
+    enumerate_worker_surfaces,
+)
+from pollypm.web_api.chat.tmux_capture import (
+    capture_envelopes,
+    synthesize_capture_id,
+)
+from pollypm.web_api.chat.transcripts import (
+    STALE_THRESHOLD_SECONDS,
+    is_archive_stale,
+    parse_events_jsonl,
+    resolve_transcript_path,
+)
+
+__all__ = [
+    "STALE_THRESHOLD_SECONDS",
+    "ChatSurface",
+    "MessageEnvelope",
+    "MessageRole",
+    "MessageType",
+    "SurfaceType",
+    "capture_envelopes",
+    "enumerate_chat_surfaces",
+    "enumerate_config_surfaces",
+    "enumerate_worker_surfaces",
+    "is_archive_stale",
+    "parse_events_jsonl",
+    "resolve_transcript_path",
+    "synthesize_capture_id",
+]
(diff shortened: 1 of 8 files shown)


samhotchkiss:
Blocking review for now. The read-side shape is close, and the targeted tests pass, but this PR weakens the core chat identity invariant: `session_name` must resolve to the exact surface transcript, not a best-effort project transcript.

Required changes:

- `src/pollypm/web_api/chat/transcripts.py:87` (`resolve_transcript_path`) currently scans `<project>/.pollypm/transcripts/*/events.jsonl` and falls back from cwd match to freshest mtime. That can attach the wrong transcript whenever two surfaces share a project/cwd (operator + architect, restarted sessions, or multiple active workers under the same project). `registry.py:214` and `registry.py:279` then expose that path as if it belongs to the requested surface. Please replace this with a real `session_name -> session_id/events.jsonl` mapping, or return an explicit unresolved/`None` state instead of falling back to another surface's transcript. Add a regression with two transcripts for the same project/cwd and prove the registry does not cross-select.
- Avoid doing a fresh transcript-root scan per surface in `enumerate_chat_surfaces`. The ingestor already had to add scan caching for large transcript trees; the API path should not reintroduce repeated unbounded tail reads. Build the mapping once per project/request, or consume an indexed source of truth.
- `MessageType.THINKING` and the `include_thinking` flag are documented as supported (`src/pollypm/web_api/chat/transcripts.py:37`, `parse_events_jsonl(... include_thinking=...)`), but `_event_to_envelopes` never emits `MessageType.THINKING` and the flag is unused. Either implement the thinking event path and tests, or remove it from the P1 contract until the ingestor actually preserves it.

Verification I ran:

- `uv run --with pytest python -m pytest tests/test_chat_*.py --noconftest` -> `119 passed, 1 xfailed, 1 xpassed`
- `uv run --with ruff ruff check src/pollypm/web_api/chat tests/test_chat_registry.py tests/test_chat_tmux_capture.py tests/test_chat_transcripts.py` -> passed

samhotchkiss:
All three review blockers addressed in commit 2213ae28a (pushed to `feat/chat-api-p1-readers`).

**Blocker 1 — session_name → events.jsonl cross-attachment**
- `src/pollypm/web_api/chat/transcripts.py:87-216` — `resolve_transcript_path` rewritten on top of `build_session_index` + `lookup_transcript_path`. The index reads the first event of each `<session_id>/events.jsonl` to extract the `(session_id, cwd, account_name, provider)` fingerprint. Lookup filters by cwd + account + provider; unresolved → `None` (no freshest-mtime fallback to another surface's transcript).
- `src/pollypm/web_api/chat/registry.py:254` (config surfaces) and `registry.py:327` (worker surfaces) now call `lookup_transcript_path` with the surface fingerprint (`session.account`, `str(session.provider)` for config; worker `provider` for workers).
- Confirmed `.ingestion-state.json` shape via `src/pollypm/transcript_ingest.py:122-167`: cursor entries store `{offset, session_id, cwd, model_name, mtime}` per source path. Picked option (b) over (a) because the cursor file has no `account_name` and we needed that field to disambiguate operator+architect sharing the same project root.
- Regression test: `tests/test_chat_registry.py::test_two_surfaces_sharing_cwd_resolve_to_correct_transcripts` — plants two events.jsonls under the same project/cwd (operator on `claude_main`, architect on `claude_alt`); the architect transcript is the freshest mtime, which would have cross-attached under the prior fallback. New code resolves each surface to its OWN events.jsonl. Symmetric assertion confirms no cross-attachment.

**Blocker 2 — repeated scans per surface**
- `src/pollypm/web_api/chat/registry.py:206` — new `_build_session_index(project_root, cache)` helper memoizes the `build_session_index` result keyed by transcripts root. `enumerate_chat_surfaces` (`registry.py:160`) constructs a request-scoped `session_index_cache: dict[Path, list]` and threads it through both `enumerate_config_surfaces` and `enumerate_worker_surfaces`, so each project's transcripts dir is walked at most once per request. No module-level cache.

**Blocker 3 — THINKING contract removed**
- `src/pollypm/web_api/chat/envelope.py:30` — `MessageType.THINKING` removed; docstring explains the deferral.
- `src/pollypm/web_api/chat/transcripts.py:205` — `parse_events_jsonl` no longer accepts `include_thinking`; `_event_to_envelopes` signature updated to match.
- Module docstring + `pollypm.web_api.chat.__init__` doc string updated.
- Tests updated; new follow-up issue #2048 tracks the ingestor-side preservation arc.

**Verification**
- `uv run --with pytest python -m pytest tests/test_chat_*.py --noconftest` → 122 passed, 1 xfailed, 1 xpassed
- `uv run --with ruff ruff check src/pollypm/web_api/chat tests/test_chat_registry.py tests/test_chat_transcripts.py` → passes
- `needs-review` label re-applied.

Follow-up issue for THINKING: #2048

Issue #10, opened 2026-04-10
Inbox routing and PM/PA handoff


samhotchkiss:
## Goal
Implement the PM/PA routing rules so the operator (Polly) can triage inbox items and route execution to workers.

## Scope
- PM triages items from open/
- PM routes to PA for execution-only tasks
- PM keeps policy/scope/priority questions
- PA writes handoff.json when taking or returning ownership
- Worker replies surface through PA before updating thread
- Wire into operator prompt so Polly monitors open/

## Acceptance Criteria
- [ ] PM can triage and route inbox items
- [ ] Handoff between PM and PA is tracked in handoff.json
- [ ] Routing follows the rules: PM keeps policy questions, routes execution to PA
- [ ] Worker replies update the thread through PA
- [ ] Operator prompt includes inbox awareness

## Test Instructions
1. Create an inbox item, verify Polly picks it up
2. Route to PA, verify handoff.json updates
3. PA completes work, verify thread returns to PM
4. Test a policy question stays with PM

## Dependencies
- Depends on inbox file layout and state machine

samhotchkiss:
Verified implemented: Inbox routing — triage_inbox_item, route_inbox_thread, transition_inbox_thread, handoff_inbox_thread all implemented in service_api

PR #2054, opened 2026-05-22
feat(api): GET /api/v1/storage — Phase 2 storage report

samhotchkiss:
## Summary

- Phase 2 surface #10: read-only storage report. Mirrors `pm storage report --json` (#2040), reusing the CLI's `scan_pollypm_home` helper directly.
- New router `src/pollypm/web_api/routes/storage.py`:
  - `GET /api/v1/storage` — whole-home report (every canonical subdir + config files + totals)
  - `GET /api/v1/storage/{subdir}` — single canonical-subdir slice (restricted to the known subdir list; arbitrary paths are intentionally rejected)
- New Pydantic models: `StorageReport`, `StorageEntry`, `StorageConfigFiles`.
- Prune endpoint deferred per spec §12.2 — disk-deletion stays CLI-only.
- OpenAPI `docs/api/openapi.yaml` updated with paths, schemas, and a new `Storage` tag.

Per spec `~/Desktop/pollypm-phase2-endpoints-spec.md` §0–§2 + §12.

## Test plan

- [x] `pytest --noconftest tests/test_storage_endpoint.py -v --timeout=120` (7 tests, all pass)
  - Happy path with synthetic `~/.pollypm/`
  - Per-subdir surfacing (file count + bytes per canonical subdir)
  - Cap-hit annotation (`_SCAN_FILE_CAP` patched low, snapshots/ exceeds cap)
  - Auth required (no bearer → 401)
  - Missing home dir → 200 with empty report (not 500)
  - Subdir detail endpoint
  - Unknown subdir → 404 with hint listing valid names
- [x] OpenAPI YAML validates as OpenAPI 3.1 (`tests/web_api/test_openapi_conformance.py::test_contract_yaml_is_valid_openapi_31`)
- [x] Existing chat-endpoint regression suite still green (42/42)


src/pollypm/web_api/routes/storage.py
```diff
@@ -0,0 +1,194 @@
...
+from __future__ import annotations
+
+import logging
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any
+
+from fastapi import APIRouter
+
+from pollypm.cli_features.storage import (
```
samhotchkiss: Blocking on the module boundary here. `web_api` is another frontend surface, so it should not import the CLI feature module or its private `_HOME_SUBDIRS` constant directly. This pulls the whole Typer/migration/prune CLI module into `pm serve` and makes the canonical storage-report invariant live under a presentation layer. Please extract the read-only scanner/types/canonical subdir list into a neutral helper/facade module (for example under `pollypm.storage` or a small `pollypm.storage_report` module) and have both `pm storage report` and this route depend on that. That keeps the single source of truth without coupling API runtime to CLI implementation details.

src/pollypm/web_api/routes/storage.py
```diff
@@ -0,0 +1,194 @@
...
+            hint="Check filesystem permissions on ~/.pollypm/ and retry.",
+        ) from exc
+    for row in report.rows:
+        if row.name == subdir:
+            return _entry_from_scan(row)
+    # ``scan_pollypm_home`` always emits a row per ``_HOME_SUBDIRS``
+    # entry (empty DirScan when the dir doesn't exist on disk), so
+    # this branch is only reached if the canonical list and the scan
+    # diverge — that's a programming error worth surfacing.
+    raise not_found(
```
samhotchkiss: This fallback becomes a user-facing regression for a fresh install or alternate `base_dir` whose home directory does not exist yet. `GET /api/v1/storage` correctly treats that as normal and returns 200 with empty totals, but `GET /api/v1/storage/snapshots` returns this 404 for a valid canonical subdir with the misleading hint `This indicates a bug in scan_pollypm_home`. I confirmed that behavior locally. The subdir endpoint should preserve the same missing-home invariant as the top-level report, e.g. return an empty `StorageEntry(name=subdir, files=0, bytes=0, ...)` for a valid subdir, and add a regression test for valid subdir + missing home.

samhotchkiss (commented):

Verified locally:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_storage_endpoint.py -v --timeout=120` -> 7 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py::test_contract_yaml_is_valid_openapi_31 tests/web_api/test_openapi_conformance.py::test_implementation_openapi_validates_as_31 -v` -> 2 passed

src/pollypm/web_api/routes/storage.py
samhotchkiss: Fixed in d7c99d4. Extracted the canonical scanner (HOME_SUBDIRS, DirScan/HomeReport, scan_pollypm_home + all walk helpers) into the new neutral `src/pollypm/storage_report.py` module. `web_api/routes/storage.py:42-47` now imports from `pollypm.storage_report` (no more cli_features dependency). `cli_features/storage.py:869-893` re-imports the same names under the original `_`-prefixed identifiers so the prune paths + existing tests keep working. Module-boundary regression test: `tests/test_storage_endpoint.py::test_storage_route_does_not_import_cli_features` greps the route source to lock in the boundary.

src/pollypm/web_api/routes/storage.py
samhotchkiss: Fixed in d7c99d4. `web_api/routes/storage.py:186-194` now mirrors the top-level endpoint's missing-home invariant — when the canonical subdir validates but `scan_pollypm_home` produces no row (the expected cause is a missing/not-yet-provisioned home), the route returns 200 with a zeroed StorageEntry (`_entry_from_scan(DirScan(name=subdir))`) instead of 404. Removed the misleading "bug in scan_pollypm_home" hint. Regression test: `tests/test_storage_endpoint.py::test_storage_subdir_returns_empty_when_home_missing` iterates every canonical subdir against a wiped home and asserts 200 + empty entry.

PR #2055, opened 2026-05-22
feat(api): GET /api/v1/heartbeats[/{session}] — Phase 2 heartbeat reads

samhotchkiss:
## Summary

Phase 2 §11 surface — read-side heartbeats endpoints (small focused PR; SSE stream deferred to follow-up).

- `GET /api/v1/heartbeats` — one `HeartbeatLatest` row per enabled session; classifies `healthy` / `stale` / `initializing` / `unknown` using the same 5-minute threshold as `pm sessions` so the API agrees with the CLI summary verbatim.
- `GET /api/v1/heartbeats/{session_name}` — latest + recent-tick history (`?limit` defaults 20, clamps at 200 per §11.3 rather than 422'ing).
- All reads go through the canonical `pollypm.storage.pg_heartbeats` facade — pg-only paths, no parallel reader.

### Design choices (worth a look during review)

- Facade outages -> `503 service_unavailable` (callers can retry; we deliberately do NOT collapse a pg-pool outage to an empty list because that would tell the cockpit "every session is initializing").
- Configured-but-silent session -> 200 with `status="initializing"` (not 404) so a brand-new session isn't conflated with one that's missing.
- Unknown session (not in config, not in ledger) -> 404 `not_found`.
- Ledger-only ghost session (e.g. reaped worker) -> 200 with `role`/`project` null.
- `?limit` clamps silently at 200 (per spec §11.3) with `clamped: true` in the response.

## Spec gaps flagged

- §11.1 row 4: SSE stream endpoint is deliberately deferred to a separate follow-up PR (per mission brief).
- The spec calls out `last_verdict` / `last_reason` fields on `HeartbeatLatest`, but the current `pg_heartbeats` facade + `HeartbeatRecord` dataclass don't carry a verdict column — those are derived by the supervisor's classifier, not stored in the heartbeats ledger. I omitted those two fields rather than fabricate them; if Sam wants them, they'd need to surface from the supervisor cascade (separate work).

## Test plan

- [x] `pytest --noconftest tests/test_heartbeats_endpoint.py -v --timeout=120` — 17 passed in 1.6s
- [x] `tests/web_api/test_openapi_conformance.py::test_contract_yaml_is_valid_openapi_31` passes (OpenAPI doc still validates as 3.1)
- [x] `tests/web_api/test_openapi_conformance.py::test_implementation_openapi_validates_as_31` passes (auto-generated doc still valid)
- [x] `tests/web_api/test_openapi_conformance.py::test_contract_lists_phase_1_paths` passes (no regression to existing paths)

### Test cases covered (17)

- Happy paths: list + detail
- Status classification: healthy / stale / initializing
- Empty list when no sessions configured
- Disabled sessions filtered out
- 404 on truly-unknown session names
- Configured-but-silent session returns initializing (not 404)
- Ghost session in ledger but not config returns 200 with null role/project
- `?limit` clamping at 200
- Facade outage -> 503 on both list + detail endpoints
- Bearer auth required (list + detail)
- Wrong token -> 401 invalid_token
- `?limit=0` -> 422 validation_error

### DIFF
--- a/src/pollypm/web_api/routes/heartbeats.py
+++ b/src/pollypm/web_api/routes/heartbeats.py
@@ -267,32 +267,48 @@ def list_heartbeats_endpoint(config: ConfigDep) -> HeartbeatsListResponse:
     caller can retry; we don't paper over a pg-pool outage with an
     empty list because that would tell the cockpit "every session is
     initializing" which is dangerously misleading.
-    """
-    from pollypm.storage.pg_heartbeats import latest_heartbeat as _latest
 
-    rows: list[HeartbeatLatest] = []
-    for session_name in _configured_session_names(config):
-        try:
-            record = _latest(session_name, config=config)
-        except Exception as exc:  # noqa: BLE001 — typed translation below
-            logger.warning(
-                "heartbeats: latest_heartbeat(%s) raised; surfacing 503",
-                session_name,
-                exc_info=True,
-            )
-            raise service_unavailable(
-                f"heartbeats ledger unavailable while reading session "
-                f"{session_name!r}: {exc.__class__.__name__}",
-                hint=(
-                    "The Postgres heartbeats facade is unreachable. "
-                    "Retry shortly; check `pm doctor` and pg pool health."
-                ),
-            ) from exc
-        rows.append(_row_to_latest(
+    Performance contract (pinned by
+    ``test_list_heartbeats_uses_bulk_query``): exactly ONE
+    ``pg_heartbeats`` query for N sessions, not N. The previous loop
+    was an N+1 read that hurt dashboard polling at session counts
+    >~10. See PR #2055 round 1.
+    """
+    from pollypm.storage.pg_heartbeats import (
+        latest_heartbeats_bulk as _latest_bulk,
+    )
+
+    session_names = _configured_session_names(config)
+    try:
+        records_by_name = _latest_bulk(session_names, config=config)
+    except Exception as exc:  # noqa: BLE001 — typed translation below
+        logger.warning(
+            "heartbeats: latest_heartbeats_bulk(%d sessions) raised; "
+            "surfacing 503",
+            len(session_names),
+            exc_info=True,
+        )
+        raise service_unavailable(
+            f"heartbeats ledger unavailable while listing "
+            f"{len(session_names)} session(s): {exc.__class__.__name__}",
+            hint=(
+                "The Postgres heartbeats facade is unreachable. "
+                "Retry shortly; check `pm doctor` and pg pool health."
+            ),
+        ) from exc
+
+    rows = [
+        _row_to_latest(
             config=config,
             session_name=session_name,
-            row=record,
-        ))
+            # Sessions missing from the bulk result have never reported
+            # in — map to None so ``_row_to_latest`` produces an
+            # ``initializing`` row (matches the previous per-session
+            # ``latest_heartbeat() -> None`` behaviour).
+            row=records_by_name.get(session_name),
+        )
+        for session_name in session_names
+    ]
     return HeartbeatsListResponse(heartbeats=rows)
 
 
--- a/src/pollypm/web_api/routes/heartbeats.py
+++ b/src/pollypm/web_api/routes/heartbeats.py
@@ -1,0 +1,409 @@
+"""Heartbeats GET endpoints (Phase 2 — §11 of the endpoints spec).
+
+Implements:
+
+- ``GET /api/v1/heartbeats`` — latest heartbeat row per configured
+  session.
+- ``GET /api/v1/heartbeats/{session_name}`` — single-session detail
+  with the recent-tick history attached.
+
+Reads-only — pg-only paths via the canonical
+:mod:`pollypm.storage.pg_heartbeats` facade so the surface composes
+with the cockpit / ``pm sessions`` / supervisor without becoming a
+second reader implementation. Phase 2 §11.1 also lists an SSE stream
+endpoint; that ships separately (see follow-up) so this PR stays
+focused on the read-side surface.
+
+The route classifies each session into ``healthy`` / ``stale`` /
+``unknown`` / ``initializing`` using the same 5-minute staleness
+threshold as :mod:`pollypm.cli_features.sessions_health` so the API
+agrees with the CLI summary verbatim.
+"""
+
+from __future__ import annotations
+
+import logging
+from datetime import UTC, datetime
+from typing import Annotated
+
+from fastapi import APIRouter, Query
+from pydantic import BaseModel
+
+from pollypm.web_api.errors import not_found, service_unavailable
+from pollypm.web_api.routes._deps import ConfigDep
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(tags=["Heartbeats"])
+
+
+# ---------------------------------------------------------------------------
+# Classification thresholds (kept in sync with sessions_health.py — picked
+# once, used everywhere)
+# ---------------------------------------------------------------------------
+
+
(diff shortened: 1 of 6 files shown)


src/pollypm/web_api/routes/heartbeats.py
```diff
@@ -0,0 +1,409 @@
...
+
+    Backing-store outages bubble as 503 ``service_unavailable`` so the
+    caller can retry; we don't paper over a pg-pool outage with an
+    empty list because that would tell the cockpit "every session is
+    initializing" which is dangerously misleading.
+    """
+    from pollypm.storage.pg_heartbeats import latest_heartbeat as _latest
+
+    rows: list[HeartbeatLatest] = []
+    for session_name in _configured_session_names(config):
```
samhotchkiss: Blocking on the N+1 storage reads in this list endpoint. `GET /api/v1/heartbeats` is a dashboard/polling-style read surface, but this loop opens a facade read for every configured session. With enough sessions, a single UI refresh becomes N Postgres round trips and N pool checkouts, and a slow/failing read on one row holds the whole response. Please add/use a bulk `pg_heartbeats` facade (for example one query over the configured session names using `DISTINCT ON (session_name)` or a window function) and have the route map missing rows to `initializing`. Add a regression test that asserts the list endpoint does one facade call/query for multiple sessions so this contract stays pinned.

samhotchkiss (commented):

Verified locally:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_heartbeats_endpoint.py -v --timeout=120` -> 17 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py::test_contract_yaml_is_valid_openapi_31 tests/web_api/test_openapi_conformance.py::test_implementation_openapi_validates_as_31 tests/web_api/test_openapi_conformance.py::test_contract_lists_phase_1_paths -v` -> 3 passed

src/pollypm/web_api/routes/heartbeats.py
samhotchkiss: Round 1 fix shipped at 1982222721483e97d9899354b0add38e7621ea3f.

Added pg_heartbeats.latest_heartbeats_bulk(session_names) — single query, DISTINCT ON (session_name) ORDER BY session_name, id DESC. The list endpoint now calls it exactly once; missing sessions map to status="initializing" so the contract for never-reported sessions is unchanged. Empty input short-circuits without issuing a query.

Tests:
- tests/test_heartbeats_endpoint.py::test_list_heartbeats_uses_bulk_query — adds 5 extra configured sessions and asserts exactly ONE bulk-helper invocation that contains every enabled session (regression pin for the N+1).
- tests/test_heartbeats_endpoint.py::test_list_heartbeats_missing_sessions_show_initializing — half reported, half silent; asserts the silent half land as status="initializing" with last_tick_ts=null.
- tests/test_pg_heartbeats.py::test_pg_heartbeats_latest_bulk_returns_one_row_per_session — real pg, DISTINCT ON picks latest per session, gamma absent (not None).
- tests/test_pg_heartbeats.py::test_pg_heartbeats_latest_bulk_empty_input_no_query.

Existing list tests migrated to the bulk stub. 19/19 heartbeat endpoint tests green; 6/6 pg_heartbeats tests green; 8/8 openapi conformance green.

samhotchkiss:

The N+1 review thread is addressed: the list endpoint now calls `pg_heartbeats.latest_heartbeats_bulk()` once, missing rows still map to `initializing`, and the new tests pin that behavior. The remaining blocker is a PollyPM architecture invariant regression around session health classification.

`src/pollypm/web_api/routes/heartbeats.py:46-49` adds a new `_STALE_HEARTBEAT_SECONDS`, `:127-142` reimplements age parsing, and `:145-164` reimplements heartbeat status classification. Current `main` already has `src/pollypm/session_health.py` as the explicit single source of truth for this contract; `cli_features.sessions_health` and `web_api.routes.sessions_admin` import from it so the API and CLI cannot drift. This PR reintroduces the drift risk in a new API surface while the file comments say it should match `pm sessions`.

Please route the heartbeat GET endpoint through the shared helper layer instead of duplicating the predicate. A small shape would be: import `age_seconds` and `classify_status` from `pollypm.session_health`; keep only the endpoint-specific mapping of “no ledger row” to `initializing`; for rows, compute age with the shared `age_seconds()` and classify with `classify_status(window_present=True, age_seconds=age)` so the `healthy` / `stale` / `unknown` threshold stays owned by `session_health.py`. Add or adjust a regression so changing the shared threshold/classifier changes this endpoint behavior too, not a private copy.

Verification run on PR head:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_heartbeats_endpoint.py -q --timeout=120` -> 19 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 8 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_pg_heartbeats.py -q --timeout=120` -> 6 passed

samhotchkiss:
Round 2 fix pushed at fadfdff8d.

**Architecture invariant restored.** `src/pollypm/web_api/routes/heartbeats.py` no longer carries a private `_STALE_HEARTBEAT_SECONDS`, local age parser, or local classifier. It now imports `pollypm.session_health` (`heartbeats.py:31`), calls `session_health.age_seconds()` for row-age computation (`heartbeats.py:195`), and delegates the `healthy`/`stale`/`unknown` branch to `session_health.classify_status(window_present=True, age_seconds=age)` (`heartbeats.py:149-152`). The only endpoint-specific layer left is the `initializing` override for "no ledger row yet" — which is route policy, not classification (matches your suggested shape).

Resolved through the `session_health` module attribute rather than a bound name so a monkeypatch on `pollypm.session_health.classify_status` flows through the route verbatim.

**Regression pin.** New test `test_heartbeat_status_uses_shared_classify` in `tests/test_heartbeats_endpoint.py` patches `session_health.classify_status` to a sentinel and asserts the endpoint surfaces it — proves the route reads through the shared helper. Also asserts the `initializing` override still wins for rows-absent sessions, and that the route called the shared classifier with the storage-only contract (`window_present=True`).

**Verification run on PR head fadfdff8d:**
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_heartbeats_endpoint.py -x --timeout=120` → 20 passed (19 round-1 + 1 new)
- `uv run --extra dev --extra test --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py tests/test_sessions_health_cli.py -x --timeout=120` → 26 passed

Re-labelled `needs-review`.

samhotchkiss:

I rechecked this branch after the latest `origin/main` updates, including #2054 and #2063. The earlier N+1/read-side drift concerns appear addressed by the pushed fixes and replies: the list endpoint uses the bulk heartbeat facade, missing rows still map to `initializing`, and heartbeat status delegates through `pollypm.session_health`.

The current blocker is mergeability. `git merge-tree --write-tree origin/main origin/feat/api-heartbeats-get` now reports a content conflict in `docs/api/openapi.yaml` while auto-merging `src/pollypm/web_api/app.py`. Please rebase onto current `main` and resolve the OpenAPI conflict while preserving the already-merged storage and project-write paths/schemas.

After rebasing, rerun the heartbeat endpoint tests and OpenAPI conformance on the merged tree, not just the branch tip:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_heartbeats_endpoint.py -v --timeout=120`

samhotchkiss:
Rebased onto current main (0112bc274). Kept all existing surface tags/paths/schemas + Heartbeats entries. Tests pass.

PR #2056, opened 2026-05-22
feat(api): GET /api/v1/config[/projects/{key}] — Phase 2 read-only config (secrets redacted)

samhotchkiss:
## Summary

Implements Phase 2 surface §11 / §13 of `~/Desktop/pollypm-phase2-endpoints-spec.md`:

- `GET /api/v1/config` — full `PollyPMConfig` snapshot, JSON-safe, with credential-shaped fields redacted to `"***"`.
- `GET /api/v1/config/projects/{key}` — single-project view of the `[projects.<key>]` block (404 on unknown key).

Mutation is deferred to Phase 3 per spec — config edits stay TOML-first.

## Redaction

Field names are checked case-insensitively against a substring list (`auth`, `token`, `secret`, `password`, `passwd`, `api_key`, `apikey`, `credential`, `private_key`). The walker recurses through every nested dict + dataclass, so account `env` entries like `OPENAI_API_KEY` redact at depth-3 too. Empty / `None` values stay as-is so the operator can distinguish "field unset" from "field hidden" (matches `SessionConfig.auth_token=""` legacy semantics).

## Test plan

- [x] `pytest --noconftest tests/test_config_endpoint.py -v --timeout=120` — 11 / 11 pass
- [x] `pytest tests/web_api/test_openapi_conformance.py` — 5 / 5 pass (yaml + impl conformance hold)
- [x] Manual smoke: `GET /api/v1/config` returns redacted session tokens + account env api keys
- [x] Unknown project key → typed 404 `not_found` with hint
- [x] Missing bearer → 401 on both endpoints

### DIFF
--- a/src/pollypm/web_api/routes/config.py
+++ b/src/pollypm/web_api/routes/config.py
@@ -202,6 +202,51 @@ def _value_looks_credential(value: Any) -> bool:
     return any(pattern.match(value) for pattern in _CREDENTIAL_VALUE_PATTERNS)
 
 
+# libpq keyword DSN credential-bearing field names (lowercase). The
+# libpq docs (https://www.postgresql.org/docs/current/libpq-connect.html)
+# list ``password`` and ``passfile`` as the auth-carrying keywords;
+# common short aliases (``passwd``, ``pwd``) are accepted by some
+# wrappers and tools, so we treat them as DSN secrets too.
+_LIBPQ_SECRET_KEYS: frozenset[str] = frozenset(
+    {"password", "passwd", "pwd", "passfile"},
+)
+
+
+# Matches whitespace-delimited libpq keyword=value tokens. libpq's
+# format is ``<keyword>=<value>`` with the value either bare (no
+# whitespace) or single-quoted (backslash escapes allowed inside). We
+# only need the keyword (group 1) for the secret-name check.
+_LIBPQ_TOKEN_RE = re.compile(
+    r"(?:^|\s)([a-zA-Z_][a-zA-Z_0-9]*)\s*=\s*"
+    r"(?:'(?:[^'\\]|\\.)*'|[^\s'][^\s]*)"
+)
+
+
+def _value_looks_libpq_dsn(value: Any) -> bool:
+    """Return True iff ``value`` parses as a libpq keyword DSN that
+    carries a credential-bearing field (``password`` / ``passwd`` /
+    ``pwd`` / ``passfile``).
+
+    Codex round-3 P0 on PR #2056: ``PG_CONN = "host=db.example
+    dbname=pollypm user=alice password=secretpw"`` is a real libpq
+    connection string but the carrier key ``PG_CONN`` doesn't match
+    ``_looks_secret`` and the value isn't URL-shaped, so the prior
+    redactor leaked the raw password.
+
+    The 2-token minimum prevents false positives on trivial
+    ``key=value`` config (e.g. ``mode=production``). The key-name
+    allowlist keeps the check narrow — DSNs that don't carry password
+    fields (``host=db dbname=x user=alice sslmode=require``) pass
+    through unchanged so observability isn't hurt.
+    """
+    if not isinstance(value, str) or not value or "=" not in value:
+        return False
+    tokens = _LIBPQ_TOKEN_RE.findall(value)
+    if len(tokens) < 2:
+        return False
+    return any(token.lower() in _LIBPQ_SECRET_KEYS for token in tokens)
+
+
 def _coerce_value(value: Any) -> Any:
     """Coerce a single config value into a JSON-friendly shape.
 
@@ -236,6 +281,12 @@ def _coerce_value(value: Any) -> Any:
     # PATs, Slack tokens, …). Backstops env entries whose key names
     # don't include any credential keyword (Codex round-2 P0 on PR #2056).
     if _value_looks_credential(value):
+        return REDACTED
+    # libpq keyword DSNs (whitespace-delimited ``host=… password=…``).
+    # The carrier key (e.g. ``PG_CONN``) need not match ``_looks_secret``
+    # and the value isn't URL-shape, so this is the only check that
+    # catches the format. Codex round-3 P0 on PR #2056.
+    if _value_looks_libpq_dsn(value):
         return REDACTED
     return value
 
--- a/src/pollypm/web_api/routes/config.py
+++ b/src/pollypm/web_api/routes/config.py
@@ -1,0 +1,252 @@
+"""Read-only config endpoints (Phase 2, surface §13 of the endpoints spec).
+
+Implements:
+
+- ``GET /api/v1/config`` — the operator's whole :class:`PollyPMConfig`,
+  serialised to a JSON-safe dict with every credential-shaped field
+  redacted to ``"***"``.
+- ``GET /api/v1/config/projects/{key}`` — the same redaction applied to
+  a single ``[projects.<key>]`` block (the per-project config view from
+  ``KnownProject``).
+
+Per ``~/Desktop/pollypm-phase2-endpoints-spec.md`` §13 (Config,
+read-only): mutation is deferred to Phase 3 — config edits should
+remain TOML-first so the user can review diffs.
+
+The handler intentionally does not import any work-service / pg-pool
+plumbing. The config is already loaded by the app factory and handed
+to us via :data:`ConfigDep`; we only need to flatten it into a
+JSON-friendly shape and walk it once to redact secrets.
+"""
+
+from __future__ import annotations
+
+import dataclasses
+from enum import Enum
+from pathlib import Path
+from typing import Any
+
+from fastapi import APIRouter
+from pydantic import BaseModel, Field
+
+from pollypm.web_api.errors import not_found
+from pollypm.web_api.routes._deps import ConfigDep
+
+router = APIRouter(tags=["Config"])
+
+
+# ---------------------------------------------------------------------------
+# Redaction
+# ---------------------------------------------------------------------------
+
+
+# Sentinel string the redactor substitutes for any field whose name
+# matches the credential heuristic. Kept identical to the spec §13.2
+# example so clients can rely on the literal string.
+REDACTED = "***"
+
+
+# Field-name substrings (lowercase) that mark a value as
+# credential-shaped. The endpoint walks every nested dict key and
+# redacts the value when any of these appear as a substring. The list
+# is intentionally broader than just the documented examples
+# (``auth_token``, ``api_key``, ``webhook_secret``) — anything that
+# *smells* credential-shaped should not leave the process even if a
+# future config field slips through review (spec §13 mission note:
+# "any field that smells like a token/key/credential…").
+_SECRET_NAME_SUBSTRINGS: tuple[str, ...] = (
+    "token",
+    "secret",
+    "password",
+    "passwd",
+    "api_key",
+    "apikey",
+    "credential",
+    "private_key",
+    "auth",  # matches auth_token, basic_auth, oauth_*, etc.
+)
+
+
+def _looks_secret(field_name: str) -> bool:
+    """Return True iff ``field_name`` matches the credential heuristic.
+
+    Case-insensitive substring match against
+    :data:`_SECRET_NAME_SUBSTRINGS`. The match is deliberately
+    over-eager — false positives (redacting a non-secret field whose
+    name happens to contain ``"auth"``) are operationally fine; false
+    negatives (leaking a real secret) are not.
+    """
+    lower = field_name.lower()
+    return any(needle in lower for needle in _SECRET_NAME_SUBSTRINGS)
+
+
+def _coerce_value(value: Any) -> Any:
+    """Coerce a single config value into a JSON-friendly shape.
+
+    Handles the non-JSON-native types the config dataclasses use:
+
+    - :class:`pathlib.Path` → ``str``
+    - :class:`enum.Enum` (incl. :class:`enum.StrEnum`) → ``.value``
+    - ``tuple`` → ``list``
+    - dataclasses → ``dataclasses.asdict`` then recursed
+    - dicts / lists → recursed element-wise
+    - everything else returned as-is (FastAPI's JSON encoder handles
+      ``datetime`` / ``bytes`` / etc. if any sneak in later)
+    """
+    if isinstance(value, Path):
+        return str(value)
+    if isinstance(value, Enum):
+        return value.value
+    if dataclasses.is_dataclass(value) and not isinstance(value, type):
+        return _redact_dict(dataclasses.asdict(value))
+    if isinstance(value, dict):
+        return _redact_dict(value)
+    if isinstance(value, (list, tuple)):
+        return [_coerce_value(item) for item in value]
+    return value
+
+
+def _redact_dict(data: dict[Any, Any]) -> dict[str, Any]:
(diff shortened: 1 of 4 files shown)


samhotchkiss (commented):

Verified locally:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_config_endpoint.py -v --timeout=120` -> 11 passed

src/pollypm/web_api/routes/config.py
```diff
@@ -0,0 +1,252 @@
...
+    :class:`SessionConfig` docstring; redacting that to ``"***"`` would
+    misrepresent the on-disk state).
+
+    Nested dicts and dataclasses recurse through :func:`_coerce_value`,
+    so the redaction applies at every depth.
+    """
+    out: dict[str, Any] = {}
+    for raw_key, value in data.items():
+        key = str(raw_key)
+        if _looks_secret(key) and value not in (None, "", b""):
```
samhotchkiss, line 358: Blocking security issue: the redactor only looks at key names, so credential-bearing DSNs/URLs currently pass through unchanged. I verified `_serialise_config()` leaks values such as `storage.url = "postgresql://user:urlpass@example/db"`, `storage.pg.dsn = "postgresql://user:dsnpass@example/db"`, and account env entries like `DATABASE_URL = "postgresql://user:dbpass@example/db"`. Those are live credentials even though the keys are `url`, `dsn`, or `DATABASE_URL`, and this endpoint returns them over HTTP. Please extend the redaction contract to cover DSN/URL-shaped credentials, at minimum `[storage].url`, `[storage.pg].dsn`, and env values whose key is a known DSN/URL carrier (`DATABASE_URL`, `POSTGRES_DSN`, `SENTRY_DSN`, etc.) or whose value parses as a URL with userinfo. Add explicit leak-regression tests for those examples before this ships.

samhotchkiss: Fixed in 8acbbe963: redactor now inspects values for URL+userinfo shape via urllib.parse.urlsplit. Wired into _coerce_value so it applies at every depth (nested dicts, env maps, list items) regardless of key name. New tests at tests/test_config_endpoint.py:407-535 cover [storage].url, [storage.pg].dsn, the three env DSN carriers (DATABASE_URL/POSTGRES_DSN/SENTRY_DSN), plus negative cases (public URLs without userinfo + plain strings pass through). 16/16 tests pass.

samhotchkiss:

The endpoint's load-bearing invariant is: no credential from `PollyPMConfig` can leave the process through `GET /api/v1/config` or `GET /api/v1/config/projects/{key}`. The current redactor still has security gaps:

1. `src/pollypm/web_api/routes/config.py:_looks_secret` only catches `api_key` / `token` / `secret` / etc. by key name, so real account env credentials whose names do not contain those substrings pass through unchanged. I verified `_redact_dict()` currently returns raw values for examples like `accounts.*.env.AWS_ACCESS_KEY_ID = "AKIA..."` and `accounts.*.env.GITHUB_PAT = "ghp_..."`. Since the endpoint exposes the whole account env map, the redaction contract needs to cover credential-shaped env names beyond `OPENAI_API_KEY` before this ships. Add explicit regression tests for these examples.

2. `src/pollypm/web_api/routes/config.py:_value_has_url_userinfo` only redacts URL values with username/password in the netloc. It still leaks DSN credentials carried outside userinfo, e.g. `storage.pg.dsn = "postgresql://db.example/pollypm?user=alice&password=secretpw"`, because the key `dsn` does not match `_looks_secret` and the value has no URL userinfo. The previous review thread asked for `[storage].url`, `[storage.pg].dsn`, and known DSN/URL carriers to be covered; this needs to include common DSN password/query/libpq shapes, not just `scheme://user:pass@host`. Add leak-regression tests for at least a query-param password DSN.

3. `src/pollypm/web_api/routes/config.py:get_config_endpoint` says a config edit between requests shows on the next call, but `src/pollypm/web_api/app.py:create_app` wires `_config_provider` to `lambda: config`, the object loaded at `pm serve` startup. That means `GET /config` serves a stale config until server restart. Either make the dependency/route reload via the existing `load_config` cache when `config.config_path` is available, or change the documented/API contract deliberately. Given the Phase 2 spec's stale-cache edge case, I think this should reload and have a test that edits the TOML after app startup and verifies the next GET reflects it.

Targeted verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_config_endpoint.py -v --timeout=120` -> 16 passed
- One-off redaction probe confirmed the leak examples above still return raw values.

samhotchkiss:
Round-2 fixes pushed in `17e186965` — addressing all three blockers from the 14:07 UTC review.

**Blocker 1 — env credentials with non-keyword names.** Two-layer fix in `src/pollypm/web_api/routes/config.py`:
- `_SECRET_NAME_SUBSTRINGS` (lines 60-83) extended with `access_key`, `pat`, `pwd`, `credentials`, `dsn`, `database_url` (substring + case-insensitive, so `AWS_ACCESS_KEY_ID` → `access_key`, `GITHUB_PAT` → `pat`).
- New `_value_looks_credential` helper (lines 138-149) with anchored regexes for AWS access keys (`^(AKIA|ASIA)[0-9A-Z]{16}$`), GitHub tokens (`^gh[pousr]_[A-Za-z0-9_]{36,}$`, plus `github_pat_`), and Slack tokens. Wired into `_coerce_value` (lines 196-200) alongside the existing URL-userinfo check.

**Blocker 2 — DSN with password in query string.** `_value_has_url_userinfo` in `src/pollypm/web_api/routes/config.py` (lines 99-136) now also runs `parse_qs(parts.query)` and redacts when any param key matches `_URL_QUERY_SECRET_KEYS` = `{password, pwd, passwd, secret, token, api_key, apikey, auth, access_token, auth_token}`. `dsn` and `database_url` were also added to the key-name heuristic as belt-and-suspenders. Regression test in `tests/test_config_endpoint.py` uses the exact `postgresql://db.example/pollypm?user=alice&password=secretpw` example you flagged.

**Blocker 3 — config stale until restart.** `src/pollypm/web_api/app.py` lines 100-128 now wires `_config_provider` to a closure that calls `load_config(config.config_path)` per request. `load_config` already has an mtime-keyed cache (`src/pollypm/config.py:930-940`), so unchanged files are a single `stat` call — no parse cost. Wrapped in try/except: disk read failures (file deleted, malformed TOML mid-edit) fall back to the startup snapshot with a warning log, rather than 500ing. The fallback path also handles in-memory configs without a `config_path` (test fixtures that build `PollyPMConfig` directly). Route docstring updated to document the reload contract.

**Verification.**
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_config_endpoint.py -v --timeout=120` → 23 passed (16 existing + 7 new).
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -v` → 5 passed.
- New tests: `test_aws_access_key_id_is_redacted`, `test_github_pat_is_redacted`, `test_aws_secret_access_key_is_redacted`, `test_value_shape_redaction_for_bare_credential_value`, `test_storage_pg_dsn_with_query_password_is_redacted`, `test_env_dsn_with_query_password_is_redacted`, `test_get_config_reloads_on_disk_edit` (writes TOML, GETs, edits TOML adding `[projects.beta]`, GETs again, asserts new project appears).

Re-added `needs-review` for another pass.

samhotchkiss:

The round-2 fixes cover the previous concrete examples: AWS access key IDs, GitHub PATs, URL userinfo, query-string password DSNs, and reload-on-edit are now tested and passing. The remaining blocker is still in the redaction invariant: common libpq-style DSNs can leak when the carrier key does not itself look secret.

I ran a direct probe against `_redact_dict()` on the new head:

```
query_dsn {'PG_CONN': '***'}
libpq_dsn {'PG_CONN': 'host=db.example dbname=pollypm user=alice password=secretpw'}
```

`PG_CONN = "host=db.example dbname=pollypm user=alice password=secretpw"` is a real DSN credential shape. Because the key does not contain `dsn`, `database_url`, `password`, etc., and `_value_has_url_userinfo()` only handles URL/userinfo and URL query params, the raw password leaves `GET /api/v1/config` unchanged. Please extend the value-shape redactor to cover libpq keyword DSNs / whitespace-delimited `password=...` connection strings, and add regression coverage for an innocuous carrier key like `PG_CONN`.

Docs need a small follow-up too: `docs/api/openapi.yaml` still describes only the old name-substring heuristic (`auth`, `token`, `secret`, `password`, `api_key`, `credential`, `private_key`) and does not mention the value-shape redaction or the newly covered `access_key` / `pat` / `dsn` / `database_url` names. Since this is a security contract, clients/operators should not have to infer the broader behavior from code.

Tests run:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_config_endpoint.py -v --timeout=120` -> 23 passed
- Manual redaction probe: AWS, GitHub PAT, URL userinfo, and query-string DSN redacted; libpq-style `PG_CONN` leaked as shown above.

samhotchkiss:
Round-3 fix pushed in `7fb4c9c02` — addresses the remaining libpq-DSN leak and the openapi docs gap from the 14:25 UTC review.

**Blocker — libpq keyword DSN leaks.** New `_value_looks_libpq_dsn` helper in `src/pollypm/web_api/routes/config.py` (lines 205-247) tokenizes whitespace-delimited `<keyword>=<value>` DSNs (with libpq's single-quoted-value form) and redacts when any keyword is in `_LIBPQ_SECRET_KEYS = {password, passwd, pwd, passfile}`. Wired into `_coerce_value` at line 256 alongside the existing URL-userinfo and vendor-prefix checks. Two narrowness guards keep observability intact:
- 2-token minimum prevents trivial `mode=production` config from being flagged as a DSN.
- Keyword allowlist means safe DSNs (`host=db dbname=x user=alice sslmode=require`) still pass through — only password-bearing ones redact.

Codex's exact probe (`PG_CONN = "host=db.example dbname=pollypm user=alice password=secretpw"`) is now covered by `test_libpq_dsn_with_password_token_is_redacted`.

**Docs gap.** `docs/api/openapi.yaml` `/config` GET description (lines 838-871) now enumerates the full redaction contract: name-substring (with the round-2-expanded list), URL userinfo, URL credential-bearing query params, vendor prefixes (AWS / GitHub / Slack), and libpq keyword DSNs. Operators can now audit the security boundary without grepping the route source. The description explicitly notes "keep this in sync if new shapes are added" so the next round-N reviewer sees it.

**New tests (5, total 28 in file).**
- `test_libpq_dsn_with_password_token_is_redacted` — your exact probe.
- `test_libpq_dsn_without_password_token_not_redacted` — `host=db ... sslmode=require` passes through.
- `test_single_key_value_not_treated_as_dsn` — `mode=production` not flagged (2-token guard).
- `test_libpq_dsn_with_quoted_password` — `password='se cret pw'` (single-quoted with whitespace) redacts.
- `test_openapi_describes_value_shape_redaction` — parses `openapi.yaml` and asserts `/config` description mentions "libpq" and "userinfo".

**Verification.**
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_config_endpoint.py -v --timeout=120` -> 28 passed.
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -v` -> 5 passed.

Re-added `needs-review` for another pass.

samhotchkiss:

1. The round-3 redaction fix now over-redacts ordinary config paths. `_SECRET_NAME_SUBSTRINGS` includes `"pat"` as a raw substring at `src/pollypm/web_api/routes/config.py:59-81`, and `_redact_dict()` applies substring matches to every key at `routes/config.py:308-314`. That makes keys like `path` and `project_path` look secret. I reproduced this on the current head:

```python
_redact_dict({"path": "/tmp/project", "project_path": "/tmp/project", "GITHUB_PAT": "ghp_..."})
# => {"path": "***", "project_path": "***", "GITHUB_PAT": "***"}
```

That breaks the read-only config contract: project paths are not credentials and are core operator-visible config. Please make PAT detection boundary-aware or env-key-specific (for example exact `PAT` / suffix `_PAT` / known provider token names), while keeping `path`, `project.path`, and `config_path` visible. Add regressions that `GITHUB_PAT` redacts but `path` / `project_path` / `config_path` do not.

2. The PR is still not mergeable against current `main`: GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY` and the PR base is still the older `1a5696597...` while current `main` is `11ff7783d...`. Please rebase onto current `main` and preserve both this Config contract and the already-merged Dashboard OpenAPI/app changes.

Targeted verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_config_endpoint.py -q --timeout=120` -> 28 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 5 passed
- One-off redaction probe above -> reproduced `path` / `project_path` over-redaction

samhotchkiss:
Round-4 fix pushed in `6db8865bd` — addresses both blockers from the 14:48 UTC review.

**Blocker 1 — `path` / `project_path` / `config_path` over-redacted by raw `pat` substring.** Replaced the substring-only check in `src/pollypm/web_api/routes/config.py` with three layered name-side checks in `_looks_secret` (lines 80-103):

- `_SECRET_NAME_SUBSTRINGS` (lines 57-72) — `pat` removed.
- `_SECRET_EXACT_NAMES_LOWER` (lines 79-84) — case-insensitive exact match for provider-specific names: `github_pat`, `github_token`.
- `_SECRET_NAME_SUFFIXES` (lines 93-97) — case-insensitive `_pat` suffix match. `_TOKEN` / `_KEY` / `_SECRET` are already covered by the substring list, so the suffix set is intentionally narrow to avoid duplication.

Your exact repro now returns the expected output:
```python
_redact_dict({"path": "/tmp/project", "project_path": "/tmp/project", "GITHUB_PAT": "ghp_..."})
# => {"path": "/tmp/project", "project_path": "/tmp/project", "GITHUB_PAT": "***"}
```

Value-shape detectors (`_value_looks_credential`, `_value_has_url_userinfo`, `_value_looks_libpq_dsn`) are unchanged — they remain the second line of defence.

OpenAPI `/config` description (lines 844-853) now enumerates the new exact/suffix layer alongside the substring list and explicitly notes that `path` / `project_path` / `config_path` stay visible.

**Blocker 2 — rebase onto current main.** Rebased the 5-commit branch onto `11ff7783d` (current `main`). Conflicts on `docs/api/openapi.yaml` (tags block: kept BOTH the Dashboard entry from main AND the Config entry from this PR) and `src/pollypm/web_api/app.py` (both `from … import dashboard_routes` and `from … import config_routes`, plus both `app.include_router(...)` calls). New head: `6db8865bd`.

**New tests (3, file total now 31).**
- `test_path_keys_are_not_redacted` — pins the regression on `path` / `project_path` / `config_path`.
- `test_github_pat_still_redacted_by_name` — uses a non-`ghp_` value so only the exact-name check can carry the redaction (pins that name-side coverage didn't regress).
- `test_provider_pat_suffixes_redacted` — `WHATEVER_PAT`, `GITLAB_PAT` redact via suffix.
- Existing round-2 `test_github_pat_is_redacted` (value `ghp_<36×a>`) still passes — covered by name (exact) AND value-shape (`^ghp_…`).

**Verification.**
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_config_endpoint.py -q --timeout=120` → 31 passed.
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` → 5 passed.

Re-added `needs-review` for another pass.

PR #2057, opened 2026-05-22
feat(api): GET /api/v1/dashboard — Phase 2 dashboard aggregate state

samhotchkiss:
## Summary

Phase 2 §3.1 — ships the read-only `GET /api/v1/dashboard` endpoint that returns the same whole-system snapshot the cockpit rail computes. Small focused PR; SSE streaming (§3.2) and `?async=true` long-op handling (§2.7) are deferred to follow-up PRs.

- New router at `src/pollypm/web_api/routes/dashboard.py` (one GET handler) that wraps `pollypm.dashboard_data.gather()` + `pollypm.web_api.service.list_projects()`. No new business logic — pure adapter.
- Filters per spec: `?project=<key>` (narrows projects + rollups + per-project rows), `?include_token_history=true`, `?include_briefing=true`.
- Daemon-down is normal — returns `200` with `daemon_status: "down"` and empty `active_sessions[]` (spec §3.3); transient gather/list_projects failures surface as `503 service_unavailable`.
- pg-only path: `gather(config, None)` lets the helper open its own pg facades. No `StateStore`.

## Test plan

- [x] 12 FastAPI TestClient cases in `tests/test_dashboard_endpoint.py` covering happy path, daemon-up/down, bearer auth (missing + wrong token), `?project` filter, unknown project 404, `include_token_history`, `include_briefing`, empty-config, gather-503, list_projects-503, account-usage passthrough.
- [x] Run: `pytest --noconftest tests/test_dashboard_endpoint.py -v` — 12 passed in 1.68s.
- [x] OpenAPI conformance suite still green (`tests/web_api/test_openapi_conformance.py` — 4 passed; the 1 pre-existing fixture error is a `--noconftest` artifact unrelated to this PR).
- [ ] Manual: `curl http://127.0.0.1:8765/api/v1/dashboard -H "Authorization: Bearer $(cat ~/.pollypm/api-token)" | jq .rollups` should match `pm` cockpit rail counters.

## OpenAPI

`docs/api/openapi.yaml` adds the `/dashboard` path with full parameter/response docs, a new `Dashboard` tag, and 7 new schemas (`DashboardResponse`, `DashboardRollups`, `DashboardTokens`, `SessionActivity`, `CommitInfo`, `CompletedItem`, `InboxPreview`, `AccountQuotaUsage`).

## Spec notes / ambiguities

- Spec §3.1 lists `pending_plan_reviews` in `rollups` but mentions `stuck_alerts` / `in_progress_tasks` / `review_tasks` as well — kept the four spec-explicit fields (`tracked_count`, `open_inbox_count`, `pending_plan_reviews`, `alert_count`) plus the three cockpit-native 24h counters (`sweep`, `message`, `recovery`). `in_progress_tasks` / `review_tasks` are already discoverable per-project from each `Project.task_counts` map; flagged for spec follow-up if Sam wants them denormalized.
- Spec §3.1 mentions `alerts: [Alert]` from heartbeats — left out of this PR (would require importing heartbeats/types.py and is bigger than the small-PR scope). `rollups.alert_count` carries the aggregate count.

### DIFF
--- a/src/pollypm/web_api/routes/dashboard.py
+++ b/src/pollypm/web_api/routes/dashboard.py
@@ -1,0 +1,398 @@
+"""Dashboard read endpoint (Phase 2 — Q1, §3.1 of the Phase 2 spec).
+
+Implements ``GET /api/v1/dashboard`` per
+``~/Desktop/pollypm-phase2-endpoints-spec.md`` §3.1. The route is a
+thin adapter over :func:`pollypm.dashboard_data.gather` (the helper
+the cockpit rail renders from) plus :func:`pollypm.web_api.service.list_projects`.
+
+Out of scope for this PR:
+
+- ``GET /api/v1/dashboard/stream`` SSE (deferred to a follow-up PR;
+  spec §3.2).
+- ``?async=true`` long-op path (spec §2.7) — sync only for now.
+
+The route's behaviour mirrors the cockpit per spec §3.3 edge cases:
+
+- A missing daemon is a normal operating mode — return ``200`` with
+  ``daemon_status="down"`` instead of 503.
+- A transient ``gather`` failure (pg pool down, broken project)
+  surfaces as ``503 service_unavailable`` so clients retry rather than
+  treat empty data as truth.
+"""
+
+from __future__ import annotations
+
+import logging
+from datetime import datetime, timezone
+from typing import Annotated, Any
+
+from fastapi import APIRouter, Query
+from pydantic import BaseModel, Field
+
+from pollypm.web_api.errors import not_found, service_unavailable
+from pollypm.web_api.models import Project
+from pollypm.web_api.routes._deps import ConfigDep
+from pollypm.web_api.service import list_projects
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(tags=["Dashboard"])
+
+
+# ---------------------------------------------------------------------------
+# Response models (spec §3.1 — mirror the cockpit dataclasses)
+# ---------------------------------------------------------------------------
+
+
+class SessionActivityModel(BaseModel):
+    """Active-session row for the dashboard (cockpit ``SessionActivity``)."""
+
+    name: str
+    role: str
+    project: str
+    project_label: str
+    status: str
+    description: str
+    age_seconds: float
+
+
+class CommitInfoModel(BaseModel):
+    """Recent commit row (cockpit ``CommitInfo``)."""
+
+    hash: str
+    message: str
+    author: str
+    age_seconds: float
+    project: str
+
+
+class CompletedItemModel(BaseModel):
+    """Recently-completed work item (cockpit ``CompletedItem``)."""
+
+    title: str
+    kind: str
+    project: str
+    age_seconds: float
+
+
+class InboxPreviewModel(BaseModel):
+    """Recent inbox preview row (cockpit ``InboxPreview``)."""
+
+    sender: str
+    title: str
+    project: str
+    task_id: str
+    age_seconds: float
+
+
+class AccountQuotaUsageModel(BaseModel):
+    """Per-account quota row (cockpit ``AccountQuotaUsage``)."""
+
+    account_name: str
+    provider: str
+    email: str
+    used_pct: int
+    summary: str
+    severity: str
+    limit_label: str = "limit"
+    reset_at: str = ""
+
+
+class DashboardRollups(BaseModel):
+    """Aggregate counters mirrored from the cockpit rail."""
+
+    tracked_count: int
+    open_inbox_count: int
+    pending_plan_reviews: int
+    alert_count: int
+    sweep_count_24h: int
+    message_count_24h: int
+    recovery_count_24h: int
+
+
+class DashboardTokens(BaseModel):
+    """Token-usage summary."""
+
+    today: int
+    total: int
+
+
+class DashboardResponse(BaseModel):
+    """``GET /api/v1/dashboard`` envelope (spec §3.1)."""
+
+    generated_at: datetime
+    daemon_status: str = Field(
+        description=(
+            "One of 'up' (active sessions present) or 'down' "
+            "(no active sessions / supervisor unreachable). Per spec "
+            "§3.3, a down daemon is a normal operating mode — not a 503."
+        ),
+    )
+    projects: list[Project]
+    rollups: DashboardRollups
+    active_sessions: list[SessionActivityModel]
+    recent_commits: list[CommitInfoModel]
+    completed_items: list[CompletedItemModel]
+    recent_messages: list[InboxPreviewModel]
+    tokens: DashboardTokens
+    account_usages: list[AccountQuotaUsageModel]
+    daily_tokens: list[list[Any]] | None = Field(
+        default=None,
+        description=(
+            "Optional ``[date, tokens]`` history (last 30 days). Only "
+            "populated when ``?include_token_history=true`` since the "
+            "list can be large."
+        ),
+    )
+    briefing: str | None = Field(
+        default=None,
+        description=(
+            "Optional morning-briefing narrative. Only populated when "
+            "``?include_briefing=true`` since the body can be megabytes."
+        ),
+    )
+
+
+# ---------------------------------------------------------------------------
+# Adapters (dataclass -> pydantic wire shape)
+# ---------------------------------------------------------------------------
+
+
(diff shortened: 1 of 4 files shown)


src/pollypm/web_api/routes/dashboard.py
```diff
@@ -0,0 +1,398 @@
...
+    open_inbox_count = sum(p.open_inbox_count for p in projects_view)
+    pending_plan_reviews = sum(
+        1 for p in projects_view if p.pending_plan_review
+    )
+
+    rollups = DashboardRollups(
+        tracked_count=tracked_count,
+        open_inbox_count=open_inbox_count,
+        pending_plan_reviews=pending_plan_reviews,
+        alert_count=int(getattr(data, "alert_count", 0) or 0),
```
samhotchkiss, line 418: Blocking correctness issue for `?project=`. The route documents and tests that the filter narrows projects and rollups, but only `tracked_count`, `open_inbox_count`, and `pending_plan_reviews` are project-scoped. `alert_count`, `sweep_count_24h`, `message_count_24h`, and `recovery_count_24h` stay whole-system. I confirmed a filtered `?project=myproj` response can return only `myproj` in `projects` while `rollups.alert_count=6` and the 24h counters come entirely from other projects. Either compute these through the same project-scoped predicates the cockpit uses, or make the contract explicit that these fields are global and do not claim rollups are narrowed. Add regression coverage for cross-project alert/activity data under `?project=`.

src/pollypm/web_api/routes/dashboard.py
```diff
@@ -0,0 +1,398 @@
...
+        sweep_count_24h=int(getattr(data, "sweep_count_24h", 0) or 0),
+        message_count_24h=int(getattr(data, "message_count_24h", 0) or 0),
+        recovery_count_24h=int(getattr(data, "recovery_count_24h", 0) or 0),
+    )
+
+    daemon_status = "up" if active_sessions else "down"
```
samhotchkiss: This derives `daemon_status` after filtering active sessions by project, which can report `down` even when the daemon is up and active sessions exist for another project. `daemon_status` is a system/daemon health field, not a per-project row count. I reproduced `?project=myproj` returning `daemon_status="down"` with `active_sessions=[]` while the gathered dashboard data had an active `otherproj` session. Please derive daemon status from the unfiltered gather result (or rename/document a project-scoped status separately) and cover the case with a test.

samhotchkiss (commented):

Verified locally:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_dashboard_endpoint.py -v --timeout=120` -> 12 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py::test_contract_yaml_is_valid_openapi_31 tests/web_api/test_openapi_conformance.py::test_implementation_openapi_validates_as_31 -v` -> 2 passed

src/pollypm/web_api/routes/dashboard.py
samhotchkiss: Fixed in 532913f26 — src/pollypm/web_api/routes/dashboard.py:367-380 (the ?project= branch now builds an explicit scoped_fields list enumerating the narrowed fields). Took the contract-explicit fallback per the comment's second option: alert_count + the three *_24h counters stay global because re-scoping them would require new project-aware pg queries against the alerts / events tables, which is outside the v1 RC minimal-diff window. Added DashboardRollups docstring (dashboard.py:101-122), DashboardResponse.scoped_fields field (dashboard.py:152-161), and openapi.yaml prose (docs/api/openapi.yaml:1250-1276, 1322-1331) so the contract is discoverable. Regression test: test_dashboard_endpoint.py::test_dashboard_project_filter_keeps_global_activity_rollups asserts both the unchanged-global behavior AND the scoped_fields membership.

src/pollypm/web_api/routes/dashboard.py
samhotchkiss: Fixed in 532913f26 — src/pollypm/web_api/routes/dashboard.py:355-360. daemon_status is now derived from data.active_sessions (the UNFILTERED gather result) BEFORE the per-project slice that produces the local active_sessions list. Updated the Field description (dashboard.py:124-132) + openapi prose (docs/api/openapi.yaml:1291-1300) to spell out that ?project= does not influence this field. Regression test: test_dashboard_endpoint.py::test_dashboard_daemon_status_reflects_unfiltered_sessions reproduces your scenario (single live session on otherproj, ?project=myproj request) and asserts daemon_status=="up" while active_sessions==[].

samhotchkiss:

The prior `?project=` review comments are mostly reflected in the current branch: `scoped_fields` now documents which fields are narrowed, and `daemon_status` is derived before the per-project slice. I do not think the daemon-status contract is correct yet.

Blocking correctness issue: `src/pollypm/web_api/routes/dashboard.py:get_dashboard_endpoint` sets `daemon_status = "up" if data.active_sessions else "down"` (lines 349-355), but the real `pollypm.dashboard_data.gather()` path appends a `SessionActivity` for every planned launch even when no runtime exists; it sets `status = "unknown"` when the session is not present in `runtime_map` (see `src/pollypm/dashboard_data.py:823-850`). On a normal configured workspace with the supervisor/tmux daemon down or no runtime rows, `data.active_sessions` can still be non-empty, so the API reports `daemon_status: "up"` even though the documented daemon-down case should return `"down"`.

Please derive `daemon_status` from a real runtime-backed/liveness signal rather than list non-emptiness. A small route-local fix could treat only known live runtime statuses as up, but the cleaner contract may be for `dashboard_data.gather()` to expose an explicit daemon/liveness boolean so the cockpit and API do not drift. Add a regression test where `_gather_dashboard` returns one `SessionActivity(status="unknown", ...)` and assert `daemon_status == "down"`; also keep the existing unfiltered cross-project test so `?project=` still cannot mask a genuinely live session in another project.

Docs also still need one cleanup: `docs/api/openapi.yaml:301-309` says `?project=<key>` narrows “rollup counters” to one project, while the new contract intentionally leaves `alert_count` and the three `*_24h` counters global. The schema prose later explains this, but the path-level description is what clients will read first and currently contradicts the fix.

Targeted verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_dashboard_endpoint.py -v --timeout=120` -> 15 passed

samhotchkiss:
Round-2 fixes pushed as 9c4162c99 on `feat/api-dashboard-get`.

**Blocker 1 — `daemon_status` liveness (route-local, option a).**
`src/pollypm/web_api/routes/dashboard.py:349-368` now derives the status from `any(s.status != "unknown" for s in data.active_sessions)` rather than list truthiness. Rationale for picking the narrow option per v1 RC minimal-diff guidance: `gather()` has no existing daemon/liveness boolean to lean on, so introducing one would have rippled into `dashboard_data.gather()` and cockpit callsites — out of scope for an RC fix. The synthetic `status="unknown"` sentinel is the only marker `gather()` uses for "no `runtime_map` entry" (see `src/pollypm/dashboard_data.py:824-825`); every other status (`healthy`, `waiting_on_user`, `needs_followup`, etc.) comes from a real `SessionRuntimeRecord`, so "anything but `unknown`" is the correct live set.

Regression coverage in `tests/test_dashboard_endpoint.py`:
- `test_dashboard_daemon_status_down_when_only_unknown_sessions` — two synthetic `unknown` rows in a configured workspace, asserts `daemon_status == "down"` and that the placeholder rows still pass through `active_sessions` (the cockpit panel shows them).
- `test_dashboard_daemon_status_up_when_mixed_unknown_and_live` — one live `healthy` row alongside an `unknown` placeholder flips it to `"up"`.
- The pre-existing `test_dashboard_daemon_status_reflects_unfiltered_sessions` cross-project test is kept unchanged so `?project=` still cannot mask a live session in another project.

**Blocker 2 — openapi prose.** `docs/api/openapi.yaml:301-314` (path-level `?project=` description) now spells out that only the three project-derived counters narrow, while `alert_count` + the three `*_24h` counters + `daemon_status` stay workspace-wide, and explicitly points clients at `scoped_fields` as the authoritative enumeration.

**Verification (local):**
- `pytest --noconftest tests/test_dashboard_endpoint.py -x --timeout=120` -> 17 passed (was 15; +2 new tests above).
- `pytest tests/web_api/test_openapi_conformance.py -x --timeout=120` -> 5 passed.

Re-labeled `needs-review`.

PR #2058, opened 2026-05-22
feat(api): GET+POST /api/v1/doctor — Phase 2 doctor checks + run/fix

samhotchkiss:
## Summary

Phase 2 surface #5 (doctor). Adds three endpoints behind the existing bearer-auth dep, per `~/Desktop/pollypm-phase2-endpoints-spec.md` §7:

- `GET  /api/v1/doctor/checks` — list registered checks (`name`, `category`, `severity`).
- `GET  /api/v1/doctor/report` — last in-process report; `404 not_found` until a run happens.
- `POST /api/v1/doctor/run` — body `{check?: str, fix: bool}`. `check` omitted ⇒ run all; unknown ⇒ `404`. `fix=true` invokes `apply_fixes` and re-runs the fixed checks so the response reflects post-fix state. Wall-clock budget defaults to 25s (max 120s); overrun ⇒ `504 timeout` per spec §2.7.

Route is a thin adapter over `pollypm.doctor.run_checks` / `apply_fixes` / `_registered_checks` — no new state-machine logic, no on-disk cache (last-report cache is process-local). OpenAPI 3.1 contract updated with six new schemas (`DoctorCheckInfo` / `DoctorChecksResponse` / `DoctorCheckResult` / `DoctorReportResponse` / `DoctorRunRequest` / `DoctorRunResponse`); existing V1 `/doctor` summary surface left untouched.

## Test plan

- [x] `pytest --noconftest tests/test_doctor_endpoint.py -v --timeout=120` — 14/14 pass in 5.4s
- [x] `pytest --noconftest tests/web_api/test_openapi_conformance.py` — 4/4 pass (1 ERROR is the pre-existing `client` fixture from `--noconftest`, not a regression)
- [x] OpenAPI 3.1 validates via `openapi_spec_validator`
- [ ] Manual: `curl /api/v1/doctor/checks` with bearer — returns the live registry
- [ ] Manual: `curl -XPOST /api/v1/doctor/run -d '{"check":"pollypm-home-writable","fix":true}'` — runs + verifies one check

## Spec gaps flagged

- Spec §7.1 hints at `severity` / `description` per check, but `pollypm.doctor.Check` has no docstring field — `description` ships as empty string; a follow-up could plumb richer metadata.
- §7.2 mentions `424 fix_failed` and per-check locks; not implemented — single-flight is achieved implicitly by the in-process `threading.Lock` around the cache, and "fix ran but check still fails" is reflected via the post-fix verify rerun rather than a distinct status code. Worth a follow-up if Sam wants `424` taxonomy.
- Spec §2.7 `?async=true` is deferred to Q13 — sync-only with 504 on overrun matches the §16.3 recommendation.

### DIFF
--- a/src/pollypm/web_api/routes/doctor.py
+++ b/src/pollypm/web_api/routes/doctor.py
@@ -28,7 +28,7 @@ import concurrent.futures
 import logging
 import threading
 import time
-from typing import Annotated, Any
+from typing import Annotated, Any, Callable, TypeVar
 
 from fastapi import APIRouter, Query
 from pydantic import BaseModel, Field
@@ -376,62 +376,140 @@ def _select_checks(name: str | None) -> list[Check]:
     raise _unknown_check(name, [c.name for c in registered])
 
 
+# Shared bounded executor for all doctor work (checks + fixes + verify).
+#
+# Round-3 (Codex on PR #2058) flagged the prior per-request executor as a
+# DoS vector: each timeout abandoned a worker thread, so a client retrying
+# a stuck check could accumulate unbounded background workers. Switching
+# to a single module-level pool caps the number of in-flight (or leaked)
+# doctor workers at ``max_workers``. When all slots are saturated by hung
+# workers, new ``submit`` calls queue — but in practice ``fix=true`` is
+# single-flighted by :data:`_FIX_OPERATION_LOCK` (one fix at a time) and
+# ``fix=false`` runs are read-only / fast, so 2 slots is more than enough
+# for the v1 RC surface. Tuning lives here if we ever need more.
+_DOCTOR_EXECUTOR = concurrent.futures.ThreadPoolExecutor(
+    max_workers=2,
+    thread_name_prefix="doctor",
+)
+
+
+_T = TypeVar("_T")
+
+
 def _run_with_budget(
-    selected: list[Check],
-    *,
-    name: str | None,
-    timeout: float,
-) -> DoctorReport:
-    """Run the selected checks under a real wall-clock budget.
+    fn: Callable[..., _T],
+    *args: Any,
+    budget_s: float,
+    name: str | None = None,
+    **kwargs: Any,
+) -> _T:
+    """Run ``fn(*args, **kwargs)`` under a real wall-clock budget.
 
-    ``run_checks`` itself doesn't honor a timeout — individual checks
-    each have their own short subprocess timeouts (see ``_run_cmd``),
+    Doctor work (``run_checks`` itself, ``apply_fixes``, the post-fix
+    verify rerun) doesn't honor a Python-level timeout — individual
+    checks have their own short subprocess timeouts (see ``_run_cmd``),
     but a pure-Python hang (network call, stuck storage probe, blocking
-    fix path) would otherwise hold the request worker forever.
+    fix on a flaky filesystem) would otherwise hold the request worker
+    forever.
 
-    To bound the HTTP request, we hand ``run_checks`` off to a
-    single-shot :class:`concurrent.futures.ThreadPoolExecutor` and
-    wait at most ``timeout`` seconds for the result. On overrun we
-    raise a typed 504 ``timeout`` immediately — the worker thread is
-    *not* cancelled (Python stdlib has no safe way to interrupt
-    arbitrary blocking code) and is allowed to finish in the
-    background. The HTTP caller sees the 504 within the budget; the
-    leaked thread is the accepted v1 RC tradeoff (Codex round-1 P0
-    on PR #2058).
+    To bound the HTTP request, the callable is dispatched to a shared
+    bounded :class:`concurrent.futures.ThreadPoolExecutor`
+    (:data:`_DOCTOR_EXECUTOR`) and we wait at most ``budget_s`` seconds
+    for the result. On overrun we raise a typed 504 ``timeout``
+    immediately — the worker thread is *not* cancelled (Python stdlib
+    has no safe way to interrupt arbitrary blocking code) and is
+    allowed to finish in the background, eating its slot in the pool
+    until it returns. The HTTP caller sees the 504 within the budget;
+    the leaked thread is the accepted v1 RC tradeoff (Codex round-1 P0
+    on PR #2058). The shared pool bounds the total number of leaks
+    (Codex round-3 P0 on PR #2058).
 
-    ``DoctorThreadLeak`` is intentionally not raised on leak: the
-    leak is silent and best-effort cleanup happens when the thread
-    finally returns.
+    ``name`` is used purely for the 504 message; the caller (the
+    endpoint) supplies the user-facing check name (or ``None`` for
+    "all checks").
     """
     t0 = time.monotonic()
-    executor = concurrent.futures.ThreadPoolExecutor(
-        max_workers=1,
-        thread_name_prefix="doctor-run",
-    )
+    future = _DOCTOR_EXECUTOR.submit(fn, *args, **kwargs)
     try:
-        future = executor.submit(run_checks, selected)
-        try:
-            report = future.result(timeout=timeout)
-        except concurrent.futures.TimeoutError:
-            elapsed = time.monotonic() - t0
-            logger.warning(
-                "doctor: run exceeded %.1fs budget for %s; abandoning worker thread",
-                timeout,
-                name or "all checks",
-            )
-            raise _check_timeout(name, elapsed) from None
-    finally:
-        # ``wait=False`` so we don't block the request on the leaked
-        # worker thread when the budget already fired. If the future
-        # finished cleanly the shutdown is effectively a no-op.
-        executor.shutdown(wait=False)
+        result = future.result(timeout=budget_s)
+    except concurrent.futures.TimeoutError:
+        elapsed = time.monotonic() - t0
+        logger.warning(
+            "doctor: %s exceeded %.1fs budget; abandoning worker thread",
+            name or "run",
+            budget_s,
+        )
+        raise _check_timeout(name, elapsed) from None
 
     elapsed = time.monotonic() - t0
-    if elapsed > timeout:
+    if elapsed > budget_s:
         # Defensive: the future returned just past the deadline. Still
         # surface 504 so the caller sees a consistent contract.
         raise _check_timeout(name, elapsed)
-    return report
+    return result
+
+
+def _run_full_fix_under_budget(
+    selected: list[Check],
+) -> tuple[DoctorReport, list[dict[str, Any]]]:
+    """Run ``run_checks`` → ``apply_fixes`` → verify rerun in one closure.
+
+    Round-3 (Codex on PR #2058): the prior implementation called the
+    initial run under :func:`_run_with_budget` and the verify rerun
+    under :func:`_run_with_budget`, but :func:`apply_fixes` itself was
+    called inline. A hanging fix (stuck subprocess, blocked filesystem
+    on a network mount, jammed worktree git op) would therefore block
+    the HTTP worker past the documented ``timeout_seconds`` ceiling
+    with no 504.
+
+    By packing all three phases into one closure dispatched through
+    :func:`_run_with_budget`, the wall-clock budget covers the full
+    sequence end-to-end. A hang in any phase trips the timeout and
+    surfaces 504 within the budget.
+
+    Returns ``(merged_report, fixes_applied)``. ``fixes_applied`` is
+    the wire-shape list of attempted fixes (matches the JSON we ship
+    back to clients).
+    """
+    report = run_checks(selected)
+
+    # ``apply_fixes`` skips passing/skipped checks itself — no need to
+    # filter on the caller side. Returns a list of
+    # ``(name, success, message)`` tuples; we coerce to JSON for the
+    # wire response.
+    try:
+        raw = apply_fixes(report)
+    except Exception as exc:  # noqa: BLE001 — bubble as 500 via outer handler
+        logger.exception("doctor: apply_fixes raised: %s", exc)
+        raise
+    fixes_applied = [
+        {"name": name, "ok": ok, "message": message}
+        for (name, ok, message) in raw
+    ]
+
+    # Re-run just the checks we tried to fix so the response reflects
+    # the post-fix state (issue #1063 style verification: don't trust
+    # the handler's self-report). All within the same budget — if the
+    # caller's timeout fires mid-verify, the outer future cancels and
+    # surfaces 504 just like a hang in the initial run.
+    fixed_names = {entry["name"] for entry in fixes_applied}
+    if fixed_names:
+        rerun_checks = [c for c in selected if c.name in fixed_names]
+        verify_report = run_checks(rerun_checks)
+        # Splice verify-report rows back into the original report so
+        # the response always reflects the latest state. The original
+        # report is a dataclass — build a new merged list of
+        # ``(check, result)`` tuples and mutate in place.
+        verify_index = {
+            check.name: result for check, result in verify_report.results
+        }
+        merged: list[tuple[Check, CheckResult]] = []
+        for check, result in report.results:
+            replacement = verify_index.get(check.name)
+            merged.append((check, replacement if replacement else result))
+        report.results = merged
+
+    return report, fixes_applied
 
 
 # ---------------------------------------------------------------------------
@@ -547,56 +625,28 @@ def run_doctor_endpoint(
         if not fix_lock_held:
             raise _fix_busy_error()
 
+    fixes_applied: list[dict[str, Any]] = []
     try:
-        t0 = time.monotonic()
-        report = _run_with_budget(
-            selected, name=payload.check, timeout=timeout_seconds,
-        )
-
-        fixes_applied: list[dict[str, Any]] = []
         if payload.fix:
-            # ``apply_fixes`` skips passing/skipped checks itself — no need
-            # to filter on the caller side. Returns a list of
-            # ``(name, success, message)`` tuples; we coerce to JSON for
-            # the wire response.
-            try:
-                raw = apply_fixes(report)
-            except Exception as exc:  # noqa: BLE001 — bubble as 500 via outer handler
-                logger.exception("doctor: apply_fixes raised: %s", exc)
-                raise
-            fixes_applied = [
-                {"name": name, "ok": ok, "message": message}
-                for (name, ok, message) in raw
-            ]
-
-            # Re-run just the checks we tried to fix so the response
-            # reflects the post-fix state (issue #1063 style verification:
-            # don't trust the handler's self-report).
-            fixed_names = {entry["name"] for entry in fixes_applied}
-            if fixed_names:
-                rerun_checks = [c for c in selected if c.name in fixed_names]
-                elapsed_so_far = time.monotonic() - t0
-                remaining = timeout_seconds - elapsed_so_far
-                if remaining <= 0:
-                    raise _check_timeout(payload.check, elapsed_so_far)
-                verify_report = _run_with_budget(
-                    rerun_checks,
-                    name=payload.check,
-                    timeout=max(remaining, 1.0),
-                )
-                # Splice verify-report rows back into the original report
-                # so the response always reflects the latest state. The
-                # original report is a dataclass — build a new merged list
-                # of ``(check, result)`` tuples and mutate in place.
-                verify_index = {
-                    check.name: result
-                    for check, result in verify_report.results
-                }
-                merged: list[tuple[Check, CheckResult]] = []
-                for check, result in report.results:
-                    replacement = verify_index.get(check.name)
-                    merged.append((check, replacement if replacement else result))
-                report.results = merged
+            # Round-3 (Codex on PR #2058): the entire run+apply+verify
+            # sequence is wrapped in one budget so a hanging
+            # ``apply_fixes`` can't escape the timeout contract. The
+            # post-fix verify rerun shares the same wall-clock budget;
+            # if apply_fixes itself consumes the budget, the future
+            # times out before verify even starts and we surface 504.
+            report, fixes_applied = _run_with_budget(
+                _run_full_fix_under_budget,
+                selected,
+                budget_s=timeout_seconds,
+                name=payload.check,
+            )
+        else:
+            report = _run_with_budget(
+                run_checks,
+                selected,
+                budget_s=timeout_seconds,
+                name=payload.check,
+            )
 
         generated_at = _record_last_report(report)
     finally:
--- a/src/pollypm/web_api/routes/doctor.py
+++ b/src/pollypm/web_api/routes/doctor.py
@@ -1,0 +1,508 @@
+"""Doctor endpoints (Phase 2 — §7 of the phase-2 endpoints spec).
+
+Implements:
+
+- ``GET  /api/v1/doctor/checks`` — list available check names + metadata.
+- ``GET  /api/v1/doctor/report`` — return the last-run report (or 404
+  ``not_found`` when no run has happened yet in this process).
+- ``POST /api/v1/doctor/run`` — execute all checks or a single named
+  check, optionally invoking :func:`pollypm.doctor.apply_fixes` when
+  ``fix=true`` is set.
+
+Per ``~/Desktop/pollypm-phase2-endpoints-spec.md`` §7. The router is a
+thin adapter over :mod:`pollypm.doctor` — it never re-implements the
+check registry or runner; that lives in the canonical doctor module the
+CLI and cockpit both consume.
+
+The "last report" cache is process-local (held in module state via the
+``_LAST_REPORT`` slot). The CLI persists doctor output to disk under
+``~/.pollypm`` already, but the API surface is intentionally
+in-memory: spec §7.1 only asks for the *last-run* report, and tying
+to disk would couple the API to the CLI's file layout. Tests can reset
+the cache via :func:`_reset_last_report` so cases don't leak state.
+"""
+
+from __future__ import annotations
+
+import logging
+import threading
+import time
+from typing import Annotated, Any
+
+from fastapi import APIRouter, Query
+from pydantic import BaseModel, Field
+
+from pollypm.doctor import (
+    Check,
+    CheckResult,
+    DoctorReport,
+    apply_fixes,
+    run_checks,
+)
+from pollypm.doctor import (
+    _auto_fix_supported,  # type: ignore[attr-defined]
+    _registered_checks,  # type: ignore[attr-defined]
+)
+from pollypm.web_api.errors import APIError, invalid_request, not_found
+from pollypm.web_api.routes._deps import ConfigDep
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(tags=["Doctor"])
+
+
+# ---------------------------------------------------------------------------
+# Limits + defaults (kept module-level so tests can patch)
+# ---------------------------------------------------------------------------
+
+
+# Per spec §2.7: long-running sync actions get a 30s budget; doctor runs
+# typically complete in < 5s but ``--fix`` may exceed that. Stay below
+# the spec's 30s ceiling so the caller sees a typed 504 rather than the
+# generic uvicorn timeout.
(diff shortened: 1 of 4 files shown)


src/pollypm/web_api/routes/doctor.py
```diff
@@ -0,0 +1,508 @@
...
+
+    ``run_checks`` itself doesn't honor a timeout — individual checks
+    each have their own short subprocess timeouts (see ``_run_cmd``).
+    The wall-clock overrun check below is a defensive backstop in
+    case a check hangs in pure Python (e.g. a network call that
+    bypassed the timeout helper). 504 lets the client decide whether
+    to retry the offending check in isolation.
+    """
+    t0 = time.monotonic()
+    report = run_checks(selected)
```
samhotchkiss: This does not actually enforce the advertised timeout. `_run_with_budget()` only checks elapsed time after `run_checks(selected)` returns, so any pure-Python hang, slow network probe, stuck storage call, or blocking fix path will hold the request worker indefinitely and will never produce the documented 504 until the damage is already done. The test stubs `time.monotonic()` after the runner returns, so it proves only post-facto classification. Please run the doctor work through a cancellable/bounded execution path (or make this an async job/poll contract) and add a regression test with a runner that blocks past the budget.

src/pollypm/web_api/routes/doctor.py
```diff
@@ -0,0 +1,508 @@
...
+    - ``check`` set → run only that check; unknown → 404.
+    - ``fix=true`` → after the run, invoke ``apply_fixes`` on the
+      report and re-run *just the fixed checks* so the response shows
+      post-fix state. The list of attempted fixes lands in
+      ``fixes_applied``.
+
+    The whole sequence (run + optional fix + post-fix verify) is
+    capped at ``timeout_seconds``; overrun → 504 ``timeout``.
+    """
+    _ = config
```
samhotchkiss: Blocking config ownership issue: the route accepts the app's `ConfigDep` and then discards it, while many doctor checks/fixes load `DEFAULT_CONFIG_PATH` internally. `pm serve --config <path>` can therefore report on, or mutate with `fix=true`, the wrong workspace/home/storage backend. That breaks the Web API invariant in `create_app()` that endpoints share the preloaded config. Please thread this config through a public doctor runner/facade, or explicitly reject/document unsupported non-default configs with tests that prove the endpoint cannot touch the wrong config.

src/pollypm/web_api/routes/doctor.py
```diff
@@ -0,0 +1,508 @@
...
+    payload = body or DoctorRunRequest()
+
+    selected = _select_checks(payload.check)
+    t0 = time.monotonic()
+    report = _run_with_budget(
+        selected, name=payload.check, timeout=timeout_seconds,
+    )
+
+    fixes_applied: list[dict[str, Any]] = []
+    if payload.fix:
```
samhotchkiss, line 785: `fix=true` is not single-flight. `_LAST_REPORT_LOCK` only protects the cache slot; two HTTP callers can run checks, apply the same filesystem/session/storage repairs, and then interleave verification/cache writes. The PR body says the cache lock provides single-flight, but it does not. Please add an operation-level lock/lease or a typed busy response around the full run/fix/verify sequence, and cover concurrent `fix=true` calls so fixes cannot run twice or leave the last-report cache describing the wrong run.

samhotchkiss (commented):
Required changes before merge. The endpoint shape and OpenAPI contract validate, but this exposes doctor fixes as an HTTP write surface, so the runtime invariants need to be real: timeouts must bound the request, fixes must be single-flight/idempotent under concurrent callers, and the route must run against the same config passed to pm serve.

src/pollypm/web_api/routes/doctor.py
samhotchkiss: Round-1 fix pushed. **doctor.py:399 `_run_with_budget`** now uses `concurrent.futures.ThreadPoolExecutor(max_workers=1)` with `future.result(timeout=...)` — real cancellable execution. On overrun we raise 504 immediately and abandon the worker thread (stdlib has no safe way to interrupt arbitrary blocking Python). HTTP caller sees 504 within the budget; the leaked thread is the accepted v1 RC tradeoff.

Regression test `test_run_504_when_check_exceeds_budget` rewritten to use a real hanging `run_checks` (threading.Event blocks 30s) and assert wall-clock response stays < 5s with a 1s budget. The old test only proved post-facto classification by stubbing `time.monotonic`.

src/pollypm/web_api/routes/doctor.py
samhotchkiss: Round-1 fix pushed via the **reject path** (chosen over the facade refactor for v1 RC scope). **doctor.py:314 `_reject_non_default_config`** now raises typed 400 `invalid_request` on all three endpoints when `config.config_path` is set and resolves to anything other than `DEFAULT_CONFIG_PATH`. In-memory fixtures (`config_path=None`) pass through.

#2056's per-request `_reload_config` partially helped — `ConfigDep` now always reflects what `pm serve` was started with — but the underlying problem persists: doctor checks call `load_config(DEFAULT_CONFIG_PATH)` internally rather than reading the injected config. Until the facade refactor lands, refusing the non-default case is the only honest answer.

Test `test_run_rejected_for_non_default_config` builds a `PollyPMConfig` with `config_path=/tmp/custom/pollypm.toml` and asserts 400 on both POST /run (fix=true write surface) and GET /checks (read surface).

src/pollypm/web_api/routes/doctor.py
samhotchkiss: Round-1 fix pushed. **doctor.py:164 `_FIX_OPERATION_LOCK`** is a new module-level `threading.Lock` separate from `_LAST_REPORT_LOCK` (which still only guards the cache slot as you noted). The run endpoint now acquires it non-blocking around the full run+fix+verify sequence when `payload.fix=True`:

```python
fix_lock_held = _FIX_OPERATION_LOCK.acquire(blocking=False)
if not fix_lock_held:
    raise _fix_busy_error()  # typed 409 conflict
try:
    ...  # run + apply_fixes + verify + record_last_report
finally:
    if fix_lock_held:
        _FIX_OPERATION_LOCK.release()
```

`fix=false` is allowed to overlap (each check has its own internal locks already). Concurrent `fix=true` is the dangerous case.

Test `test_run_fix_serialized_under_concurrency` monkeypatches `apply_fixes` to block on a `threading.Event` so the first request holds the lock long enough for a second to race in. Asserts exactly one 200 + one 409 `conflict` body. The slow-apply Event is released after the assertion so the suite shuts down cleanly.

samhotchkiss:

I rechecked this branch against the current `origin/main` after #2054 and #2063 landed. The prior timeout/single-flight/config-path review threads have follow-up commits and replies, but the branch is no longer mergeable.

`git merge-tree --write-tree origin/main origin/feat/api-doctor` reports a content conflict in `src/pollypm/web_api/app.py` while auto-merging `docs/api/openapi.yaml`. This PR adds doctor routes, and current `main` now has additional route wiring; the conflict needs to be resolved so all existing routers remain mounted and the doctor router is added exactly once.

Please rebase onto current `main`, resolve the app wiring conflict, and rerun the doctor and OpenAPI suites on the merged tree:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_doctor_endpoint.py -v --timeout=120`

samhotchkiss:
Rebased onto current `main` (post #2054 + #2063). New head: `f59837afa`.

**Conflicts resolved**
- `src/pollypm/web_api/app.py` — kept all existing router mounts (health, projects, tasks, inbox, dashboard, config, storage, events, chat_messages, chat_send, sessions_admin) and added the doctor router exactly once under `/api/v1` with `auth_deps`. The conflict was a structural one — the storage-routes mount (#2054) and the doctor-routes mount landed on the same line; both now coexist.
- `docs/api/openapi.yaml` — auto-merged cleanly by git (all tags/paths/schemas preserved).

**Tests** (run on the merged tree, fresh from rebase)
- `tests/test_doctor_endpoint.py` — 16/16 passed
- `tests/web_api/test_openapi_conformance.py` — 9/9 passed
- Total: **25 passed in 45.82s**

Pure rebase, no behavior changes, no `--no-verify`. Pushed with `--force-with-lease`. Re-tagged `needs-review` for your next pass.

samhotchkiss:

The rebase conflict is resolved and the previous config-path/single-flight/run-check timeout tests pass, but the `fix=true` path still does not satisfy the endpoint's timeout contract.

`run_doctor_endpoint()` documents that the whole `run + optional fix + post-fix verify` sequence is capped by `timeout_seconds` (`src/pollypm/web_api/routes/doctor.py:527-528`). In the implementation, only `run_checks` and the post-fix verify rerun go through `_run_with_budget()` (`doctor.py:552-554`, `doctor.py:582-586`). The actual fix operation is called directly at `doctor.py:562-566`:

```python
raw = apply_fixes(report)
```

If a fix hangs on filesystem permissions, a stuck subprocess, a database outage, or a live worktree operation, the HTTP request blocks indefinitely and never returns the documented `504 timeout`. The elapsed-time check only happens after `apply_fixes()` returns, so it cannot protect the request. Please put the full fix phase under the same bounded execution contract and add a regression where `apply_fixes` blocks longer than `timeout_seconds` and the request returns 504 within budget.

There is a related recovery/performance concern in `_run_with_budget()`: every timeout creates a fresh per-request `ThreadPoolExecutor` and abandons a non-cancelled worker (`doctor.py:406-427`). A client retrying a stuck check can create unbounded background threads. Please use a bounded/shared execution strategy or a single-flight/busy response for doctor runs so retries and partial failures cannot accumulate unbounded abandoned workers.

What I verified on current `origin/main` merged with the PR:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_doctor_endpoint.py -v --timeout=120` -> 16 passed.
- `uv run --extra dev ruff check src/pollypm/web_api/routes/doctor.py tests/test_doctor_endpoint.py` -> passed.

samhotchkiss:
Round-3 addressed in `029c22196`.

**What changed**
- `_run_with_budget` is now generic `(fn, *args, budget_s, **kw)` and dispatches through a single module-level `_DOCTOR_EXECUTOR` (`ThreadPoolExecutor(max_workers=2)`) at `src/pollypm/web_api/routes/doctor.py:390-393`. Per-request executor is gone, so retries against a stuck check can no longer accumulate unbounded worker threads — the pool caps total in-flight + leaked workers at 2.
- `_run_full_fix_under_budget` (`doctor.py:452-512`) packs `run_checks` + `apply_fixes` + verify-rerun into one closure. `run_doctor_endpoint` (`doctor.py:629-649`) dispatches that closure through one `_run_with_budget` call, so a hang in any phase (including `apply_fixes`) trips the same 504 contract within `timeout_seconds`.
- Single-flight `_FIX_OPERATION_LOCK` from round 1 is unchanged; concurrent `fix=true` still returns 409.

**Regression test**
`test_apply_fixes_hang_returns_504_within_budget` (`tests/test_doctor_endpoint.py:408-462`) mocks `apply_fixes` to block 60s with `timeout_seconds=2`; asserts `504` returned within 5s. Verified to fail on the prior head `f59837afa` (pytest-timeout fires at 45s as the request blocks on the hanging fix), passes on `029c22196` in ~1s.

**Verification on merged tree**
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_doctor_endpoint.py -v --timeout=120` → 17 passed in 6.65s
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -v --timeout=120` → 9 passed in 20.21s
- `uv run --extra dev ruff check src/pollypm/web_api/routes/doctor.py tests/test_doctor_endpoint.py` → clean

Pushed with `--force-with-lease`, no `--no-verify`. Re-tagged `needs-review`.

samhotchkiss:

1. Rebase onto current `origin/main` and resolve the OpenAPI conflict. A merge simulation with `git merge --no-commit --no-ff origin/main` auto-merges `src/pollypm/web_api/app.py` and keeps the doctor, storage, heartbeat, SSE, chat, and sessions routers, but `docs/api/openapi.yaml` still has conflict markers around the legacy `/doctor` block and the new Phase 2 doctor paths. Please resolve the static contract against current main and rerun the full OpenAPI conformance suite on the resolved tree.

2. The `fix=true` single-flight invariant is still broken on timeout. `run_doctor_endpoint()` acquires `_FIX_OPERATION_LOCK` at `src/pollypm/web_api/routes/doctor.py:623-626`, dispatches `_run_full_fix_under_budget()` through `_run_with_budget()` at `doctor.py:637-642`, then releases the lock in `finally` at `doctor.py:651-654` as soon as `_run_with_budget()` raises 504. But `_run_with_budget()` explicitly abandons the worker thread on timeout (`doctor.py:415-424`); that thread may still be inside `apply_fixes()` or the post-fix verify rerun. A retry after the 504 can acquire `_FIX_OPERATION_LOCK` and start a second `fix=true` run while the first abandoned worker is still mutating filesystem/session/storage state. That violates the doctor write-surface contract: fixes must stay single-flight until the underlying fix work has actually stopped or finished, not just until the HTTP request times out.

Please either keep the operation lock held until the background worker exits, use an execution strategy that can actually terminate the fix work, or return a typed busy response while an abandoned fix worker is still alive. Add a regression where the first `fix=true` request times out inside `apply_fixes`, remains blocked, and a second `fix=true` request immediately after the 504 still gets 409 instead of starting another fix.

What I verified:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_doctor_endpoint.py -v --timeout=120` -> 17 passed.
- `uv run --extra dev ruff check src/pollypm/web_api/routes/doctor.py tests/test_doctor_endpoint.py` -> passed.
- I did not run OpenAPI conformance because the merge simulation leaves `docs/api/openapi.yaml` conflicted.

samhotchkiss:
Round 4 pushed as `7d42b5062`.

**Step 1 — Rebase.** Resolved `docs/api/openapi.yaml` conflict by keeping all four doctor surfaces (`/doctor`, `/doctor/checks`, `/doctor/report`, `/doctor/run`) and updating the Phase-6 header comment to note that the doctor surface is now wired in `pollypm.web_api.routes.doctor` (no longer "stale"). `app.py` auto-merged. Rebase clean on top of `3cfa97e44`.

**Step 2 — Lock held until worker completes.** Implemented the `add_done_callback` pattern (option B from your sketch):

- `src/pollypm/web_api/routes/doctor.py:175-192` — `_release_fix_lock_on_done` callback; defensively swallows `RuntimeError` so a stray release in tests can't corrupt the future result.
- `src/pollypm/web_api/routes/doctor.py:677-698` — endpoint acquires non-blocking → submits via `_submit_doctor_work` → attaches the done-callback → awaits via `_await_with_budget`. **No `finally: release()` anywhere.** The callback is the single owner of the release.
- `src/pollypm/web_api/routes/doctor.py:399-419` — split the prior `_run_with_budget` into `_submit_doctor_work` + `_await_with_budget` so the fix path can own the future. `_run_with_budget` is preserved as a thin wrapper for the `fix=false` path.

**New test.** `test_fix_retry_after_504_returns_busy_while_first_still_running` (`tests/test_doctor_endpoint.py:467-559`). Hangs `apply_fixes` past a 1s budget → first call 504s within budget → asserts retry returns **409** (not a second 504). Polls the lock at the end to confirm the done-callback fires before the next test. Confirmed locally that this test **FAILS on round-3 code** (gets 504 instead of 409) and passes after the fix.

All 18 doctor tests pass; ruff clean. The 4 unrelated `test_endpoints.py` failures (`psycopg.UniqueViolation` on `work_context_entries_pkey`) reproduce against the pre-change commit too — pre-existing pg fixture isolation issue, not introduced here.

samhotchkiss:
Blocking review notes for current head `7d42b5062` (already up to date with current `origin/main`):

The latest single-flight fix addresses the immediate retry-after-504 race, but the doctor run/fix execution model still fails the shutdown/recovery invariant for an HTTP write surface that can touch filesystem, process, storage, and live workspace state.

`_DOCTOR_EXECUTOR` is a module-level `ThreadPoolExecutor` at `src/pollypm/web_api/routes/doctor.py:407-421`, not app-lifecycle owned. `_await_with_budget()` explicitly returns 504 while abandoning the running worker at `src/pollypm/web_api/routes/doctor.py:457-463`, and there is no FastAPI shutdown hook or manager that can stop/drain/report those abandoned workers. For `fix=true`, the lock is now correctly held until the worker completes, but if that worker is wedged, the process keeps the worker alive and the API can remain permanently busy for fixes after the request has returned 504. For `fix=false`, repeated timed-out runs can still occupy the two shared worker slots until the underlying work returns.

This is the same daemon/recovery contract as the briefings path: a timeout cannot become unowned background work with no clean shutdown/restart story and no terminal state. Please make doctor work app-scoped or service-owned, with explicit shutdown/recovery behavior and observability for late completion/failure, or move the unsafe work into a killable process/job model. Add a regression proving a timed-out doctor check/fix does not keep the app/test process alive after teardown and that the subsequent recovery/busy behavior is explicit.

- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_doctor_endpoint.py -v --timeout=180` -> 18 passed
- `uv run --extra dev ruff check src/pollypm/web_api/routes/doctor.py tests/test_doctor_endpoint.py` -> passed

I’m not merging while timed-out doctor work remains process-global, unowned by app lifecycle, and potentially unrecoverable.

samhotchkiss:

I reviewed current head `7d42b5062` against the latest fetched `origin/main` (`7edb1e38a`). The merge simulation is clean, and the focused ruff check passed, but the doctor execution ownership blocker is still present.

`src/pollypm/web_api/routes/doctor.py:407-421` still creates `_DOCTOR_EXECUTOR` as process-global route-module state. `_await_with_budget()` documents and implements the timeout path by returning 504 while leaving the worker running in the background (`doctor.py:457-463`). That means timed-out doctor work has no FastAPI/app lifecycle owner, no bounded shutdown/restart behavior, and no terminal state/logging path for late completion or failure. For `fix=true`, a wedged worker can hold the operation lock forever after the client has already seen a timeout; for `fix=false`, repeated timed-out checks can occupy the shared worker slots indefinitely.

This endpoint touches doctor checks/fixes that can hit filesystem, process, storage, and live workspace state, so the invariant has to be explicit: a request timeout must not become unowned background work with no recovery story. Please make doctor execution app-scoped or service-owned, or move it behind a killable job/process model, with bounded shutdown/recovery and observable late completion/failure. Add a regression proving a timed-out doctor check/fix does not keep the app/test process alive after teardown and that the follow-up busy/recovery behavior is deliberate.

Verification on this pass:
- `git merge-tree --write-tree origin/main origin/feat/api-doctor` -> clean tree.
- `uv run --extra dev ruff check src/pollypm/web_api/routes/doctor.py tests/test_doctor_endpoint.py` -> passed.

samhotchkiss:
Round 5 pushed as `0a4086e93`.

**Doctor execution is now app-lifecycle owned.**

`src/pollypm/web_api/app.py:55-148` defines `_lifespan(app)` that attaches `doctor_executor` (ThreadPoolExecutor, 2 workers, `doctor` prefix), `doctor_fix_lock`, `doctor_last_report_lock`, and `doctor_last_report` to `app.state`. Shutdown calls `executor.shutdown(wait=False, cancel_futures=True)` so a wedged worker can't block FastAPI teardown, then waits up to 5s for cooperative drain. Workers still alive after the grace period are explicitly logged + leaked rather than waited on (the v1 RC tradeoff: Python stdlib has no safe way to interrupt arbitrary blocking code; `_await_with_budget` doctring already documents this, the lifespan now makes it observable). `create_app(..., lifespan=_lifespan)` wires it in.

**Routes resolve from `request.app.state`.**

`src/pollypm/web_api/routes/doctor.py:158-321` introduces `_get_executor`, `_get_fix_lock`, `_read_last_report(request)`, `_record_last_report(request, report)`. Module-level `_DOCTOR_EXECUTOR` / `_FIX_OPERATION_LOCK` / `_LAST_REPORT_LOCK` are gone. Both `get_doctor_report_endpoint` and `run_doctor_endpoint` take `request: Request`.

**Observability for late completion (round-5 specific).**

`src/pollypm/web_api/routes/doctor.py:213-292` adds `_release_fix_lock_factory(lock)` (returns a done-callback that releases the per-app fix lock AND logs the eventual outcome) plus `_log_check_worker_outcome` (read-only path callback that only logs). `_run_with_budget` attaches `_log_check_worker_outcome` to every submitted check future at `src/pollypm/web_api/routes/doctor.py:609`. `run_doctor_endpoint` attaches `_release_fix_lock_factory(fix_lock)` to the fix future at `src/pollypm/web_api/routes/doctor.py:820`. A 504-then-late-completion now emits `doctor: fix worker completed (late or in-budget)` (or the exception/cancelled variants).

**Tests: 21 total (3 new round-5).**

- `test_doctor_executor_app_scoped` (`tests/test_doctor_endpoint.py:802`) — two `create_app` instances have independent executor/locks/cache; holding app_a's fix lock does not affect app_b.
- `test_doctor_timed_out_worker_logged_on_late_completion` (`tests/test_doctor_endpoint.py:843`) — 1s budget on a hung check → 504; release the worker → assert "check worker completed" log line appears within 5s.
- `test_doctor_app_teardown_returns_promptly_with_running_worker` (`tests/test_doctor_endpoint.py:903`) — wedged worker → lifespan exits in <10s (bounded by 5s grace) → `executor._shutdown is True`.

The `client` fixture now wraps `TestClient(app)` in `with` so the lifespan fires; `reset_last_report` autouse fixture is a no-op (each test gets a fresh per-app cache). The round-4 retry test polls `app.state.doctor_fix_lock` (was the removed module-level lock).

Verification on `0a4086e93`:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_doctor_endpoint.py -v --timeout=180` → 21 passed in 16.25s
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -v --timeout=120` → 9 passed in 26.03s
- `uv run --extra dev ruff check src/pollypm/web_api/app.py src/pollypm/web_api/routes/doctor.py tests/test_doctor_endpoint.py` → All checks passed

Re-added `needs-review`.

samhotchkiss:

Round 5 moves the doctor executor/locks/cache onto `app.state` and adds late-outcome logging, but it still does not satisfy the process shutdown/restart invariant for a timed-out doctor worker.

`src/pollypm/web_api/app.py:124-151` calls `doctor_executor.shutdown(wait=False, cancel_futures=True)`, waits up to `_DOCTOR_SHUTDOWN_GRACE_S`, then logs and returns while any running worker is still alive. That makes FastAPI lifespan teardown bounded, but it does not make `pm serve` able to exit: `ThreadPoolExecutor` workers are normal non-daemon Python threads, so a worker abandoned by `src/pollypm/web_api/routes/doctor.py:575-585` still keeps the interpreter alive after app teardown.

I verified this on current head `0a4086e939da135f7e15ca464e91d61ac400347a` with a child-process smoke: patch one doctor check to sleep for 60s, call `POST /api/v1/doctor/run?timeout_seconds=1`, then leave the `TestClient` context without releasing the worker. The child printed:

```text
after TestClient exit in 6.11s
doctor: alpha exceeded 1.0s budget; abandoning worker thread
lifespan: doctor executor: 1 worker(s) still running after 5.0s grace; leaking thread(s) past app teardown
```

but the parent had to kill it after 12.01s because the process still had the live executor worker. The new `tests/test_doctor_endpoint.py::test_doctor_app_teardown_returns_promptly_with_running_worker` only proves `TestClient.__exit__` returns and `executor._shutdown` is set; it releases the worker afterward, so it misses the actual process-liveness failure mode.

Please make timed-out doctor work genuinely process-safe before exposing this write surface: use a killable process/job model, a service-owned worker with explicit terminal state and hard stop semantics, or a cooperative cancellation contract that checks/fixes must observe so shutdown can drain within a bounded deadline. Add a regression that runs a timed-out doctor check/fix in a child process and proves the process exits promptly after app shutdown; an app-lifespan timing assertion is not enough.

- `git merge-tree --write-tree origin/main origin/feat/api-doctor` -> clean tree (`0e36820e0d554a7287dfe0cf274cdca7457e5ab4`).
- `uv run --extra dev ruff check src/pollypm/web_api/app.py src/pollypm/web_api/routes/doctor.py tests/test_doctor_endpoint.py` -> passed.
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_doctor_endpoint.py -v --timeout=180` -> 21 passed.

I am not merging while a timed-out doctor check/fix can still keep `pm serve` alive during shutdown/restart. Removing `needs-review` after posting this; please re-add it after pushing the fix.

samhotchkiss:
Round 6 pushed as `91a60a81f`.

**Daemon-thread doctor executor + atexit eviction.**

Root cause of the round-6 blocker was two-part, not one:
1. `ThreadPoolExecutor` workers default to non-daemon, so a wedged worker blocked interpreter exit even after `executor.shutdown(wait=False)` returned.
2. `concurrent.futures` registers an `atexit` hook (`_python_exit`) that joins every executor worker via the module-level `_threads_queues` dict — so even with `daemon=True`, the atexit join would still block `pm serve` exit until the worker returned.

Both addressed in `src/pollypm/web_api/app.py`:

- `DaemonThreadPoolExecutor` (`app.py:85-130`) overrides `_adjust_thread_count` to spawn daemon worker threads. Rather than mirror the stdlib body (the `threading.Thread(..., args=...)` tuple shape changed between Python 3.13 and 3.14, and `pyproject.toml` allows both), it temporarily patches `concurrent.futures.thread.threading.Thread` to a daemon-defaulting subclass and delegates to `super()`. Version-portable.
- Wire-in at `app.py:181-184` (lifespan replaces `ThreadPoolExecutor(...)` with `DaemonThreadPoolExecutor(...)`).
- Lifespan finalizer at `app.py:220-232` pops still-running workers out of `_futures_thread._threads_queues` after the grace period so the `atexit` join cannot reach them.

**Tests (2 new, 23 total).**

- `test_doctor_executor_workers_are_daemon_threads` (`tests/test_doctor_endpoint.py:979-1029`) — drives a real `/api/v1/doctor/run`, then asserts every thread in `executor._threads` has `daemon is True`. Cheap pin.
- `test_doctor_pm_serve_process_exits_with_wedged_worker` (`tests/test_doctor_endpoint.py:1032-1158`) — the regression Codex asked for. Spawns a child via `subprocess.run`, builds the app + lifespan, submits `time.sleep(60)` directly to the doctor executor, exits the lifespan, and asserts the child exits in <12s. Pre-fix the child times out at 15s (verified locally by reverting and re-running: child gets killed at the 15s `subprocess.run` timeout). Post-fix it exits in ~3s.

The subprocess test bypasses HTTP plumbing on purpose: the daemon-thread invariant is a property of the executor, not the route, and submitting straight to the pool keeps the regression focused on the actual failure mode you reproduced.

**Trade-off** (documented on the `DaemonThreadPoolExecutor` docstring): daemon workers can be killed mid-operation at process exit. Acceptable because by that point `_await_with_budget` has already returned 504, and doctor fixes are bounded by cooperative checks rather than worker liveness — there is no recoverable in-process state lost.

**Surprising cpython coupling.** The `_adjust_thread_count` signature change between 3.13 (`(weakref, queue, init, initargs)`) and 3.14 (`(weakref, worker_context, queue)`) is why the implementation uses a monkey-patch-and-super pattern instead of copying the stdlib body — both Python versions are in our supported range. The `_threads_queues` eviction touches a private symbol (`concurrent.futures.thread._threads_queues`); if cpython renames it in a future minor, the eviction will silently no-op (caught by `Exception` in `app.py:228-232` and logged). The subprocess regression would catch the regression if that ever happens.

**Verification on `91a60a81f`:**
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_doctor_endpoint.py -v --timeout=120` -> 23 passed in 23.60s
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -v --timeout=120` -> 9 passed in 22.32s
- `uv run --extra dev ruff check src/pollypm/web_api/app.py src/pollypm/web_api/routes/doctor.py tests/test_doctor_endpoint.py` -> clean

Pushed with `--force-with-lease`, no `--no-verify`. Re-tagged `needs-review`.

Issue #1776, opened 2026-05-19
fix(pg): PgWorkService missing inbox interaction methods (add_reply/list_replies/mark_read/archive_task)


samhotchkiss:
Surfaced while migrating \`tests/test_inbox_actions.py\` from sqlite to pg in Slice K-tests part 2 (#1737).

The cockpit Textual inbox screen calls four methods on the work service to drive chat threads, read-markers, and archival:

- \`add_reply(task_id, body, actor=...)\` → \`ContextEntry(entry_type="reply")\`
- \`list_replies(task_id)\`
- \`mark_read(task_id, actor)\`
- \`archive_task(task_id, actor)\` → flips status to \`done\` with audit transition

\`PgWorkService\` is missing all four. \`AttributeError\` for each.

## Affected tests (module-level xfail in part 2)

\`tests/test_inbox_actions.py\`: 17 tests across \`TestAddReply\`, \`TestListReplies\`, \`TestMarkRead\`, \`TestArchiveTask\`.

## Fix scope

Port the four methods from \`SQLiteWorkService\` (in \`pollypm.work.service_inbox\` if it exists, otherwise the sqlite_service top-level body) to \`PgWorkService\`. They write to \`work_context_entries\` (replies + read markers) and \`work_transitions\` (archive) so the existing pg schema covers it.

samhotchkiss:
Closed by PR #1784 (gap sweep). Auto-close didn't trigger due to GitHub's keyword parser not handling the comma-separated form.

Issue #1971, opened 2026-05-20
[sqlite-ripout] inventory: surfaces to remove / migrate


samhotchkiss:
Inventory pass before the multi-PR ripout sequence. Categorises every remaining sqlite surface in `src/`, `tests/`, `docs/`.

## Headline numbers

- `grep -rn "sqlite|SQLite|SQLAlchemyStore|SQLiteWorkService" src/ tests/ docs/` → **1764 hits across 286 files**, but the vast majority are docstrings / comments. The production code surface is much smaller:
  - `register_backend("sqlite", SQLAlchemyStore)` calls in production: **4**
  - `SQLAlchemyStore(f"sqlite:///...")` constructions in production: **11**
  - `@pytest.mark.backend("sqlite")` actual decorators on tests: **1** (in `test_pg_fixtures.py`, testing the marker itself)
  - Direct `SQLiteWorkService(...)` constructions outside the factory: **2** (`tests/conftest.py:212`, `tests/test_session_manager.py:1879`)

## Surface 1 — `--db` flag opt-ins (PR A scope)

**`pm notify --db <path>`** — `src/pollypm/cli_features/session_runtime.py:1362-1370` (Typer option), helper `_resolve_notify_store` does the sqlite re-register on a non-default `--db`.

**`pm inbox --db <path>`** — `src/pollypm/work/inbox_cli.py:464` (Typer option `_DB_OPTION` defined in `cli.py:69-77`), helper `_resolve_messages_store` at `:251-304` does the sqlite re-register.

**`pm task --db <path>`** — `src/pollypm/work/cli.py:69-77` `_DB_OPTION`. Used by every `pm task` subcommand (worker CLI surface). The `--db` flag drives `factory.create_work_service(db_path=...)` which still has the sqlite branch.

**`pm send-up --db`** — `src/pollypm/cli_features/send_up.py:92-96`.

**Not in scope** — `bootstrap-pg --db`, `pg_dump --dbname=` (those are pg, just named `--db`).

## Surface 2 — `SQLiteWorkService` + production callers (PR B scope)

`src/pollypm/work/sqlite_service.py` — 4205 lines, only constructed at:
- `src/pollypm/work/factory.py:152` — the sqlite branch of `create_work_service`.
- `tests/conftest.py:212` (the `_build_sqlite_work_service` helper for the dispatch fixture).
- `tests/test_session_manager.py:1879`.

[...]

PR #2060, opened 2026-05-22
feat(api): inbox WRITE endpoints — Phase 2 surface #2

samhotchkiss:
## Summary

Phase 2 surface #2 — inbox WRITE endpoints, per the Phase 2 endpoints spec §4.

Adds five POST endpoints under `/api/v1/inbox/{id}/...`:

- `archive` — close an inbox item (returns 409 invalid_state when already terminal)
- `snooze` — record a future wake-time (`duration_seconds` xor `until`; <=30d window)
- `promote-to-task` — create a new actionable task from the inbox item
- `mark-read` — idempotent read marker
- `reply` — append a reply to the thread

Every handler routes through a new service-layer helper (`archive_inbox_item`, `snooze_inbox_item`, `promote_inbox_to_task`, `mark_read_inbox_item`, `reply_inbox_item`) that opens `create_work_service` once per call. Same canonical writer the cockpit uses — Web API stays a thin adapter, not a second writer.

## Reused writers

- `svc.archive_task` (#1776) for archive
- `svc.add_reply` for reply
- `svc.mark_read` for mark-read
- `svc.add_context(entry_type="snooze")` for snooze (no native snooze on the work service yet)
- `svc.create(flow_template="standard", labels=["promoted-from-inbox", ...])` for promote

## Tests

`tests/test_inbox_writes_endpoint.py` — 18 cases. Uses the chat-endpoints pattern (`--noconftest` + service-layer monkeypatching) so the FastAPI routing + error envelope are exercised end-to-end without spinning up the pg testcontainer.

Run: `pytest --noconftest tests/test_inbox_writes_endpoint.py -v --timeout=120` (all 18 pass locally).

## OpenAPI

Spec entries under `/inbox/{id}/{archive,snooze,promote-to-task,mark-read,reply}` updated; the `replyInbox` / `archiveInbox` operations were renamed to match the implementation operationIds (`replyInboxItem` / `archiveInboxItem`) and their return shape switched from `ActionResult`/`InboxMessage` to `TaskDetail` so the client can refresh UI without a follow-up GET. New schemas: `InboxArchiveRequest`, `InboxSnoozeRequest`, `InboxPromoteRequest`, `InboxMarkReadRequest`; `InboxReplyRequest` now carries `body` + `owner` (was `message`-only). `tests/web_api/test_openapi_conformance.py` still passes.

## Test plan

- [x] `pytest --noconftest tests/test_inbox_writes_endpoint.py -v --timeout=120` — 18/18 pass
- [x] `pytest tests/web_api/test_openapi_conformance.py -v` — 5/5 pass (yaml validity, Phase 1 coverage)
- [ ] Manual: `curl -X POST /api/v1/inbox/<id>/archive` against a live `pm serve` instance
- [ ] Manual: cockpit sees the archive reflected within 5s (spec Q2.1)
- [ ] Manual: snooze with `duration_seconds=3600` records a `snooze` context entry

### DIFF
--- a/src/pollypm/web_api/routes/inbox.py
+++ b/src/pollypm/web_api/routes/inbox.py
@@ -1,19 +1,41 @@
-"""Inbox read endpoints (Phase 1).
+"""Inbox endpoints (Phase 1 reads + Phase 2 writes).
 
-Implements ``GET /api/v1/inbox`` and ``GET /api/v1/inbox/{id}``.
-Reply / archive ship in Phase 2 (#1548).
+Phase 1 (#1547) shipped ``GET /api/v1/inbox`` and
+``GET /api/v1/inbox/{id}``. Phase 2 (#1548) layers in the write
+surface — archive / snooze / promote-to-task / mark-read / reply —
+all routed through the same canonical work-service writers the
+cockpit uses (#1389). The Web API is never a second writer surface;
+these handlers are thin adapters that translate work-service
+exceptions into the typed error envelope (§6).
 """
 
 from __future__ import annotations
 
 from typing import Annotated
 
-from fastapi import APIRouter, Query
+from fastapi import APIRouter, Header, Query
 
 from pollypm.web_api.errors import not_found
-from pollypm.web_api.models import InboxItemDetail, InboxListResponse
+from pollypm.web_api.models import (
+    InboxArchiveRequest,
+    InboxItemDetail,
+    InboxListResponse,
+    InboxMarkReadRequest,
+    InboxPromoteRequest,
+    InboxReplyRequest,
+    InboxSnoozeRequest,
+    TaskDetail,
+)
 from pollypm.web_api.routes._deps import ConfigDep
-from pollypm.web_api.service import get_inbox_item, list_inbox
+from pollypm.web_api.service import (
+    archive_inbox_item,
+    get_inbox_item,
+    list_inbox,
+    mark_read_inbox_item,
+    promote_inbox_to_task,
+    reply_inbox_item,
+    snooze_inbox_item,
+)
 
 router = APIRouter(tags=["Inbox"])
 
@@ -58,3 +80,139 @@ def get_inbox_item_endpoint(id: str, config: ConfigDep) -> InboxItemDetail:
     if detail is None:
         raise not_found(f"Inbox item not found: {id}")
     return detail
+
+
+# ---------------------------------------------------------------------------
+# Phase 2 — inbox write endpoints
+#
+# Each POST handler:
+#   * routes through a service-layer helper that opens
+#     :func:`create_work_service` once per call (mirrors how queue_task
+#     wedge works — same canonical writer the cockpit uses).
+#   * declares the optional ``Idempotency-Key`` header so OpenAPI
+#     consumers see it (the dedup cache itself ships in Phase 3).
+#   * returns the post-mutation task envelope so the client refreshes
+#     UI without a follow-up GET.
+# ---------------------------------------------------------------------------
+
+
+@router.post(
+    "/inbox/{id:path}/archive",
+    response_model=TaskDetail,
+    summary="Archive (close) an inbox item",
+    operation_id="archiveInboxItem",
+    responses={
+        "401": {"description": "Missing or invalid bearer token."},
+        "404": {"description": "Project or inbox item not found."},
+        "409": {"description": "Item is already in a terminal state."},
+        "503": {"description": "Backing store unavailable."},
+    },
+)
+def archive_inbox_item_endpoint(
+    id: str,
+    config: ConfigDep,
+    body: InboxArchiveRequest | None = None,
+    idempotency_key: Annotated[str | None, Header(alias="Idempotency-Key")] = None,
+) -> TaskDetail:
+    reason = body.reason if body is not None else None
+    return archive_inbox_item(config, id, reason=reason)
+
+
+@router.post(
+    "/inbox/{id:path}/snooze",
+    response_model=TaskDetail,
+    summary="Snooze an inbox item until a future time",
+    operation_id="snoozeInboxItem",
+    responses={
+        "400": {"description": "Snooze window invalid (missing / past / >30d)."},
+        "401": {"description": "Missing or invalid bearer token."},
+        "404": {"description": "Project or inbox item not found."},
+        "409": {"description": "Item is in a terminal state and cannot be snoozed."},
+        "503": {"description": "Backing store unavailable."},
+    },
+)
+def snooze_inbox_item_endpoint(
+    id: str,
+    body: InboxSnoozeRequest,
+    config: ConfigDep,
+    idempotency_key: Annotated[str | None, Header(alias="Idempotency-Key")] = None,
+) -> TaskDetail:
+    return snooze_inbox_item(
+        config,
+        id,
+        duration_seconds=body.duration_seconds,
+        until=body.until,
+        reason=body.reason,
+    )
+
+
+@router.post(
+    "/inbox/{id:path}/promote-to-task",
+    response_model=TaskDetail,
+    summary="Promote an inbox item into a new actionable task",
+    operation_id="promoteInboxItem",
+    responses={
+        "401": {"description": "Missing or invalid bearer token."},
+        "404": {"description": "Source or destination project not found."},
+        "503": {"description": "Backing store unavailable."},
+    },
+)
+def promote_inbox_item_endpoint(
+    id: str,
+    config: ConfigDep,
+    body: InboxPromoteRequest | None = None,
+    idempotency_key: Annotated[str | None, Header(alias="Idempotency-Key")] = None,
+) -> TaskDetail:
+    if body is None:
+        body = InboxPromoteRequest()
+    return promote_inbox_to_task(
+        config,
+        id,
+        target_project=body.project,
+        prompt=body.prompt,
+        title=body.title,
+    )
+
+
+@router.post(
+    "/inbox/{id:path}/mark-read",
+    response_model=TaskDetail,
+    summary="Record a read-marker on an inbox item",
+    operation_id="markInboxItemRead",
+    responses={
+        "401": {"description": "Missing or invalid bearer token."},
+        "404": {"description": "Project or inbox item not found."},
+        "503": {"description": "Backing store unavailable."},
+    },
+)
+def mark_read_inbox_item_endpoint(
+    id: str,
+    config: ConfigDep,
+    body: InboxMarkReadRequest | None = None,
+    idempotency_key: Annotated[str | None, Header(alias="Idempotency-Key")] = None,
+) -> TaskDetail:
+    actor = (body.actor if body is not None else None) or "api"
+    return mark_read_inbox_item(config, id, actor=actor)
+
+
+@router.post(
+    "/inbox/{id:path}/reply",
+    response_model=TaskDetail,
+    summary="Append a reply to an inbox thread",
+    operation_id="replyInboxItem",
+    responses={
+        "401": {"description": "Missing or invalid bearer token."},
+        "404": {"description": "Project or inbox item not found."},
+        "422": {"description": "Reply body failed validation."},
+        "503": {"description": "Backing store unavailable."},
+    },
+)
+def reply_inbox_item_endpoint(
+    id: str,
+    body: InboxReplyRequest,
+    config: ConfigDep,
+    idempotency_key: Annotated[str | None, Header(alias="Idempotency-Key")] = None,
+) -> TaskDetail:
+    return reply_inbox_item(
+        config, id, body=body.body, owner=body.owner,
+    )
--- a/src/pollypm/web_api/routes/inbox.py
+++ b/src/pollypm/web_api/routes/inbox.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 
 from typing import Annotated
 
-from fastapi import APIRouter, Header, Query
+from fastapi import APIRouter, Query
 
 from pollypm.web_api.errors import not_found
 from pollypm.web_api.models import (
@@ -89,10 +89,19 @@ def get_inbox_item_endpoint(id: str, config: ConfigDep) -> InboxItemDetail:
 #   * routes through a service-layer helper that opens
 #     :func:`create_work_service` once per call (mirrors how queue_task
 #     wedge works — same canonical writer the cockpit uses).
-#   * declares the optional ``Idempotency-Key`` header so OpenAPI
-#     consumers see it (the dedup cache itself ships in Phase 3).
 #   * returns the post-mutation task envelope so the client refreshes
 #     UI without a follow-up GET.
+#
+# Idempotency: an earlier draft of these handlers advertised an
+# optional ``Idempotency-Key`` request header on every POST. There is
+# no idempotency store wired into the API today (grep for
+# ``idempotency`` returns only docs/comments), so the header would
+# have been silently ignored — a retry after a network drop could
+# create duplicate promoted tasks, replies, and snooze rows while the
+# client reasonably believed the header protected it. The header is
+# intentionally NOT declared here; a real idempotency cache is
+# deferred until a future PR can wire it through every non-idempotent
+# verb (see #2060 review for context).
 # ---------------------------------------------------------------------------
 
 
@@ -112,7 +121,6 @@ def archive_inbox_item_endpoint(
     id: str,
     config: ConfigDep,
     body: InboxArchiveRequest | None = None,
-    idempotency_key: Annotated[str | None, Header(alias="Idempotency-Key")] = None,
 ) -> TaskDetail:
     reason = body.reason if body is not None else None
     return archive_inbox_item(config, id, reason=reason)
@@ -135,7 +143,6 @@ def snooze_inbox_item_endpoint(
     id: str,
     body: InboxSnoozeRequest,
     config: ConfigDep,
-    idempotency_key: Annotated[str | None, Header(alias="Idempotency-Key")] = None,
 ) -> TaskDetail:
     return snooze_inbox_item(
         config,
@@ -161,7 +168,6 @@ def promote_inbox_item_endpoint(
     id: str,
     config: ConfigDep,
     body: InboxPromoteRequest | None = None,
-    idempotency_key: Annotated[str | None, Header(alias="Idempotency-Key")] = None,
 ) -> TaskDetail:
     if body is None:
         body = InboxPromoteRequest()
@@ -189,7 +195,6 @@ def mark_read_inbox_item_endpoint(
     id: str,
     config: ConfigDep,
     body: InboxMarkReadRequest | None = None,
-    idempotency_key: Annotated[str | None, Header(alias="Idempotency-Key")] = None,
 ) -> TaskDetail:
     actor = (body.actor if body is not None else None) or "api"
     return mark_read_inbox_item(config, id, actor=actor)
@@ -211,7 +216,6 @@ def reply_inbox_item_endpoint(
     id: str,
     body: InboxReplyRequest,
     config: ConfigDep,
-    idempotency_key: Annotated[str | None, Header(alias="Idempotency-Key")] = None,
 ) -> TaskDetail:
     return reply_inbox_item(
         config, id, body=body.body, owner=body.owner,
--- a/src/pollypm/web_api/service.py
+++ b/src/pollypm/web_api/service.py
@@ -644,6 +644,387 @@ def queue_task(
 
 
 # ---------------------------------------------------------------------------
+# Inbox write helpers (Phase 2 — spec §4.1)
+#
+# Each helper opens a fresh work-service via :func:`create_work_service`
+# — the same canonical writer the cockpit uses. The Web API is never a
+# second writer surface; it's a thin adapter that maps inbox-item ids
+# (``project/n``) onto the existing work-service inbox methods and
+# translates the resulting exceptions into typed error envelopes (§6).
+#
+# Why these live next to ``queue_task`` and not under
+# ``pollypm.work.inbox_cli``: the CLI helpers all assume a Typer call
+# graph (they ``raise typer.Exit`` on failure and write to stdout for
+# bulk modes). Routing an HTTP request through Typer is the wrong
+# shape; the helpers below call the same underlying ``PgWorkService``
+# methods (``archive_task``, ``add_reply``, ``mark_read``,
+# ``add_context``, ``create``) the CLI invokes.
+# ---------------------------------------------------------------------------
+
+
+# Statuses we treat as "already archived" for the inbox archive
+# endpoint. ``archive_task`` itself is idempotent (returns the row
+# unchanged when terminal) — but the spec wants a typed 409 so the
+# client can tell "I just archived it" from "someone else already
+# did". We pre-check the status and surface the conflict instead of
+# silently no-op'ing.
+_ALREADY_ARCHIVED_STATUSES: frozenset[str] = frozenset({"done", "cancelled"})
+
+
+def _project_key_from_inbox_id(item_id: str) -> str:
+    """Pull the project key off an inbox item id (``project/n``).
+
+    Inbox ids use the same ``project/task_number`` shape task ids use
+    (see :func:`_task_to_inbox_item`); the first segment is always the
+    project key. We accept ``msg:<n>`` here too so a future router
+    that wants to route those onto the unified messages store can
+    branch on the prefix — for now ``msg:`` ids raise ``not_found``
+    because the Web API only addresses chat-flow tasks.
+    """
+    if "/" not in item_id:
+        raise not_found(
+            f"Inbox item not found: {item_id}",
+            hint=(
+                "Inbox ids look like 'project/n'. Message-store ids "
+                "(``msg:<n>``) are not yet supported on this surface."
+            ),
+        )
+    return item_id.split("/", 1)[0]
+
+
+def _resolve_inbox_project(
+    config: PollyPMConfig, item_id: str
+) -> tuple[str, KnownProject]:
+    """Return ``(project_key, project)`` for an inbox id, or 404.
+
+    Centralizes the "id → registered project" lookup so every inbox
+    write endpoint reports the same error shape on unknown ids.
+    """
+    key = _project_key_from_inbox_id(item_id)
+    project = config.projects.get(key)
+    if project is None:
+        raise not_found(f"Project not registered: {key}")
+    return key, project
+
+
+def archive_inbox_item(
+    config: PollyPMConfig,
+    item_id: str,
+    *,
+    reason: str | None = None,
+    actor: str = "api",
+) -> APITaskDetail:
+    """Archive an inbox item via ``svc.archive_task``.
+
+    Returns the post-transition :class:`TaskDetail` so the client can
+    re-render without a follow-up GET. Returns 409 ``invalid_state``
+    when the item is already terminal (mirrors the spec §4.3 contract:
+    "archive a resolved item → invalid_state").
+    """
+    from pollypm.work.factory import create_work_service
+    from pollypm.work.service_support import TaskNotFoundError
+
+    key, project = _resolve_inbox_project(config, item_id)
+    try:
+        with create_work_service(
+            config=config, project_key=key, project_path=project.path
+        ) as svc:
+            try:
+                current = svc.get(item_id)
+            except TaskNotFoundError as exc:
+                raise not_found(f"Inbox item not found: {item_id}") from exc
+            status = getattr(current.work_status, "value", str(current.work_status))
+            if status in _ALREADY_ARCHIVED_STATUSES:
+                raise APIError(
+                    status_code=409,
+                    code="invalid_state",
+                    message=f"Inbox item {item_id} is already {status}.",
+                    hint="Items in a terminal state cannot be re-archived.",
+                )
+            if reason:
+                # Record the operator-supplied reason before flipping
+                # so the audit trail captures both "why" and "how".
+                try:
+                    svc.add_context(
+                        item_id, actor, f"archive reason: {reason}",
+                        entry_type="note",
+                    )
+                except Exception:  # noqa: BLE001 — non-fatal context-write
+                    logger.debug(
+                        "archive_inbox_item: reason note failed for %s",
+                        item_id, exc_info=True,
+                    )
+            try:
+                svc.archive_task(item_id, actor=actor)
+            except TaskNotFoundError as exc:
+                raise not_found(f"Inbox item not found: {item_id}") from exc
+            task = svc.get(item_id)
+            return _task_to_detail(task)
+    except _BACKING_STORE_ERRORS as exc:
+        logger.warning(
+            "archive_inbox_item: backing store error for %s: %s",
+            item_id, exc, exc_info=True,
+        )
+        raise service_unavailable(
+            f"Backing store unavailable while archiving {item_id}",
+            hint="Retry shortly; check `pm doctor` if the failure persists.",
+        ) from exc
+
+
+def snooze_inbox_item(
+    config: PollyPMConfig,
+    item_id: str,
+    *,
+    duration_seconds: int | None = None,
+    until: datetime | None = None,
+    reason: str | None = None,
+    actor: str = "api",
+) -> APITaskDetail:
+    """Snooze an inbox item until a future time.
+
+    No native ``svc.snooze`` exists on the work-service (#1776 only
+    shipped reply/mark_read/archive). We persist the snooze as a
+    structured ``snooze`` context entry whose text encodes the
+    wake-up time + reason; the cockpit's inbox-curation predicate
+    can later read these to hide snoozed rows from the default view.
+
+    Exactly one of ``duration_seconds`` / ``until`` must be supplied.
+    """
+    from datetime import timedelta, timezone
+    from pollypm.work.factory import create_work_service
+    from pollypm.work.service_support import TaskNotFoundError
+
+    if duration_seconds is None and until is None:
+        raise APIError(
+            status_code=400,
+            code="invalid_request",
+            message="Snooze requires duration_seconds or until.",
+            hint="Pass `duration_seconds` (>=1) or an ISO-8601 `until`.",
+        )
+    if duration_seconds is not None and until is not None:
+        raise APIError(
+            status_code=400,
+            code="invalid_request",
+            message="Pass duration_seconds OR until, not both.",
+        )
+    now = datetime.now(timezone.utc)
+    if until is None:
+        until = now + timedelta(seconds=duration_seconds or 0)
+    # Coerce to tz-aware UTC for a stable comparison.
+    if until.tzinfo is None:
+        until = until.replace(tzinfo=timezone.utc)
+    if until <= now:
+        raise APIError(
+            status_code=400,
+            code="invalid_request",
+            message="Snooze until must be in the future.",
+        )
+    # Spec §4.3: ``until`` > 30 days out is rejected.
+    if (until - now) > timedelta(days=30):
+        raise APIError(
+            status_code=400,
+            code="invalid_request",
+            message="Snooze until is more than 30 days out.",
+            hint="Use a wake-time within 30 days.",
+        )
+
+    key, project = _resolve_inbox_project(config, item_id)
+    try:
+        with create_work_service(
+            config=config, project_key=key, project_path=project.path
+        ) as svc:
+            try:
+                current = svc.get(item_id)
+            except TaskNotFoundError as exc:
+                raise not_found(f"Inbox item not found: {item_id}") from exc
+            status = getattr(current.work_status, "value", str(current.work_status))
+            if status in _ALREADY_ARCHIVED_STATUSES:
+                raise APIError(
+                    status_code=409,
+                    code="invalid_state",
+                    message=f"Inbox item {item_id} is {status}; cannot snooze.",
+                )
+            payload_parts = [f"snoozed until {until.isoformat()}"]
+            if reason:
+                payload_parts.append(f"reason: {reason}")
+            try:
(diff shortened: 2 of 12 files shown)


src/pollypm/web_api/routes/inbox.py
```diff
@@ -58,3 +80,139 @@ def get_inbox_item_endpoint(id: str, config: ConfigDep) -> InboxItemDetail:
...
+
+
+# ---------------------------------------------------------------------------
+# Phase 2 — inbox write endpoints
+#
+# Each POST handler:
+#   * routes through a service-layer helper that opens
+#     :func:`create_work_service` once per call (mirrors how queue_task
+#     wedge works — same canonical writer the cockpit uses).
+#   * declares the optional ``Idempotency-Key`` header so OpenAPI
```
samhotchkiss: This declares `Idempotency-Key` to clients but the key is ignored by every handler. That is unsafe for these writes: retries after a timeout/network drop can create duplicate promoted tasks, duplicate replies, and duplicate snooze rows, while the client reasonably believes the header protects it. Please either implement a real idempotency store around the mutation/result for all non-idempotent verbs or remove the advertised header/contract until it exists. Add retry/idempotency tests for at least promote, reply, and snooze.

src/pollypm/web_api/service.py
```diff
@@ -643,6 +643,387 @@ def queue_task(
...
+            if status in _ALREADY_ARCHIVED_STATUSES:
+                raise APIError(
+                    status_code=409,
+                    code="invalid_state",
+                    message=f"Inbox item {item_id} is {status}; cannot snooze.",
+                )
+            payload_parts = [f"snoozed until {until.isoformat()}"]
+            if reason:
+                payload_parts.append(f"reason: {reason}")
+            try:
```
samhotchkiss, line 915: This does not actually snooze the inbox item. It appends a free-form context row, but `list_inbox()` / `_task_to_inbox_item()` never read `entry_type='snooze'`, and the cockpit predicate does not share this state either, so the endpoint can return 200 while the item remains visible and actionable immediately. Snooze needs a canonical service/storage predicate consumed by the API and cockpit, or the endpoint should not ship yet. Please avoid encoding the wake time only in human text; store structured state and cover list-before/list-after/wake-time behavior.

src/pollypm/web_api/service.py
```diff
@@ -643,6 +643,387 @@ def queue_task(
...
+    try:
+        with create_work_service(
+            config=config, project_key=key, project_path=project.path
+        ) as svc:
+            try:
+                current = svc.get(item_id)
+            except TaskNotFoundError as exc:
+                raise not_found(f"Inbox item not found: {item_id}") from exc
+            status = getattr(current.work_status, "value", str(current.work_status))
+            if status in _ALREADY_ARCHIVED_STATUSES:
```
samhotchkiss: The terminal-state check is outside the archive transition, so the 409 contract is racey. Two concurrent archive requests can both read an open item; the first updates it, and the second then calls the idempotent `archive_task()` and returns 200 instead of the documented `invalid_state`. If the API contract is 409 for an already-terminal item, that check needs to live inside the canonical service transition or be implemented as an atomic update/row-lock path with a regression test for concurrent archive callers.

samhotchkiss (commented):
Required changes before merge. The route tests, OpenAPI conformance, and existing pg inbox-action tests pass, but the new HTTP write surface still has unresolved correctness and recovery gaps around idempotency, snooze semantics, and concurrent archive behavior.

src/pollypm/web_api/routes/inbox.py
samhotchkiss: Fixed in aa4da9c1b. Approach: **stripped** the `Idempotency-Key` header from the contract.

- `src/pollypm/web_api/routes/inbox.py:14,16,85-105,120-124,142-146,167-171,194-198,215-219` — removed the `Header` import and the `idempotency_key` parameter from all five POST handlers; the file header now documents that idempotency is intentionally deferred.
- `tests/test_inbox_writes_endpoint.py::test_openapi_does_not_advertise_idempotency_key_header` — locks the strip across all four non-idempotent verbs so a future refactor cannot silently re-introduce the header without also wiring a real dedup store.

Rationale: a grep for `idempotency` returned only docs/comments — no store helper to wire up trivially — and v1 RC stability favors the smaller, safer fix. The header will return when a real per-route response cache is wired (separate PR).

src/pollypm/web_api/service.py
samhotchkiss: Fixed in aa4da9c1b. Approach **(a)**: structured snooze persisted + readers filter active snoozes.

Writer (`src/pollypm/web_api/service.py:847-863`): snooze rows now lead with the canonical `until_iso=<ISO>` marker so the wake time is parseable instead of being buried in human prose. Legacy `snoozed until <iso>` strings still parse via the fallback in `_parse_snooze_until`.

Reader (`src/pollypm/web_api/service.py:1474-1491,1505-1602`):
- `_collect_inbox_items` now computes an `_active_snoozed_ids` set per project (one `get_context(entry_type='snooze', limit=1)` per task) and skips matching task_ids before they're turned into `APIInboxItem`s.
- `_active_snoozed_ids` takes the most-recent snooze row (work-service returns DESC) and includes the task only if the parsed wake-time is still in the future — so re-snoozing for a shorter window then expiring correctly re-surfaces the item.
- `_parse_snooze_until` handles both the new `until_iso=` marker and the legacy free-form shape.

Files updated to consistently filter snoozed items:
- `src/pollypm/web_api/service.py` — only file; the cockpit's inbox panel has its own (unrelated) curation predicate and no existing snooze logic, so wiring snooze through that surface is out of scope for this RC and tracked separately.

Tests (`tests/test_inbox_writes_endpoint.py`):
- `test_parse_snooze_until_reads_structured_marker`
- `test_parse_snooze_until_falls_back_to_legacy_shape`
- `test_parse_snooze_until_returns_none_when_no_timestamp`
- `test_active_snoozed_ids_excludes_expired_snooze`
- `test_active_snoozed_ids_hides_future_snooze` (regression for the original bug)
- `test_active_snoozed_ids_uses_latest_snooze_entry`
- `test_active_snoozed_ids_skips_items_with_no_snooze`
- `test_snooze_writer_persists_structured_until_iso_marker`

src/pollypm/web_api/service.py
samhotchkiss: Fixed in aa4da9c1b. Approach: **atomic row-lock inside the canonical writer**.

`PgWorkService.archive_task` (`src/pollypm/work/pg_service.py:2799-2890`) gains a `strict=True` kwarg. Under strict mode the state flip is a single conditional `UPDATE work_tasks SET work_status='done' WHERE project=? AND task_number=? AND work_status NOT IN ('done','cancelled') RETURNING work_status` — Postgres serialises row-level writes, so exactly one concurrent archiver sees rowcount==1 and proceeds to insert the transition row + emit audit + run the auto-unblock cascade. The loser sees rowcount==0 and raises `InvalidTransitionError`.

API helper (`src/pollypm/web_api/service.py:710-781`):
- Pre-check removed — only the existence probe (`svc.get` → 404) survives.
- Calls `svc.archive_task(item_id, actor=actor, strict=True)`.
- Catches `InvalidTransitionError` and translates to 409 `invalid_state` with the same hint as before.

Existing idempotent callers (`cockpit_ui.py`, `work/inbox_cli.py`, tests) are unaffected — `strict` defaults to False so the original idempotent contract is preserved everywhere else.

Regression test (`tests/test_pg_inbox_actions.py::TestArchiveTask::test_archive_strict_concurrent_only_one_winner`): two threads race the same archive behind a barrier; asserts exactly one returns a result + one raises `InvalidTransitionError`, and the canonical writer only appends ONE archive transition row. Also covers the simpler `test_archive_strict_flips_status` and `test_archive_strict_raises_when_already_terminal` happy/sad paths.

samhotchkiss:

The follow-up commit addresses part of the earlier review: the FastAPI handlers no longer declare `Idempotency-Key`, the endpoint-level snooze parser tests pass, and `PgWorkService.archive_task(strict=True)` now has a concurrent archive regression test. I am still blocking on two contract/performance issues.

1. The committed static OpenAPI contract still advertises idempotency on every inbox write path. `docs/api/openapi.yaml` still references `#/components/parameters/IdempotencyKey` for `/inbox/{id}/reply`, `/archive`, `/snooze`, `/promote-to-task`, and `/mark-read`, and the shared parameter still claims the server caches/replays responses for 24h. The implementation intentionally removed that header because no idempotency store exists, so the docs now promise exactly the unsafe behavior the fix was meant to avoid. Please remove those parameters from the static YAML too, or implement the real idempotency store and tests.

2. Snooze visibility is still an API-local predicate with an avoidable N+1 query on the inbox list path. `_collect_inbox_items()` calls `_active_snoozed_ids()`, and `_active_snoozed_ids()` runs `svc.get_context(entry_type="snooze", limit=1)` once per task. Inbox listing is a user-facing scan path, and the repo already has bulk context helpers in `PgWorkService` for similar read-marker/reply cases. Please move this behind a canonical service/storage helper that can load the latest snooze state in bulk, and use/share that predicate with the other inbox surface rather than letting API and cockpit semantics drift.

Tests run:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_inbox_writes_endpoint.py -v --timeout=120` -> 27 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_pg_inbox_actions.py -v --timeout=180` -> 20 passed

samhotchkiss:
Round-2 fixes pushed in cdb2a44a7.

**Blocker 1 — static OpenAPI YAML:**
- Stripped `$ref: '#/components/parameters/IdempotencyKey'` from all 5 inbox-write paths (`/inbox/{id}/{reply,archive,snooze,promote-to-task,mark-read}`).
- Kept the `IdempotencyKey` component itself because the task-state endpoints (`/approve`, `/reject`, `/queue`) still accept the header in their FastAPI handlers — deleting the component would have desynced the static contract from the auto-generated OpenAPI doc. Rewrote its description so it no longer claims "Server caches the response for 24h" — it now explicitly says no server-side dedup is wired today and that inbox endpoints don't accept it.
- New conformance test `test_static_yaml_does_not_advertise_idempotency_on_inbox_writes` pins both fixes so a future re-introduction trips CI.

**Blocker 2 — N+1 snooze query:**
- Added `PgWorkService.latest_snoozes_bulk(task_keys)` — single SQL using `DISTINCT ON (task_project, task_number) ... ORDER BY id DESC` against `work_context_entries` with `entry_type='snooze'`. Patterned after the existing `bulk_list_replies` / `task_numbers_with_context_entry` helpers. Pinned on the `WorkService` protocol so the contract is structural, not pg-specific.
- Refactored `_active_snoozed_ids()` in `web_api/service.py` to call it once per request instead of N times, with a graceful fallback to the legacy per-task loop when the backing service lacks the bulk method (keeps mock-service tests working).
- Extracted the snooze text parser into `src/pollypm/work/inbox_snooze.py` (`parse_snooze_until` + `is_snooze_active`). Lives under `pollypm.work` so cockpit can import without pulling in FastAPI. **Cockpit wiring is intentionally deferred to a follow-up PR** — the goal here is to make the shared predicate exist so the next surface adopts it instead of re-parsing.

**New tests:**
- `tests/test_pg_inbox_actions.py::TestLatestSnoozesBulk` — 5 tests: empty-input short-circuit, latest-per-task correctness, absent-on-no-snooze, non-snooze entry_types filtered out, legacy `snoozed until <iso>` text round-trips.
- `tests/test_inbox_writes_endpoint.py` — 3 new tests: bulk helper is called exactly once for 10 tasks (NOT 10 times), legacy fallback works when the bulk method is absent, shared module location pin.
- `tests/web_api/test_openapi_conformance.py` — 1 new test pinning both YAML fixes.

**Test runs (local):**
- `pytest --noconftest tests/test_inbox_writes_endpoint.py` → 30 passed
- `pytest tests/web_api/test_openapi_conformance.py` → 6 passed
- `pytest tests/test_pg_inbox_actions.py` → 25 passed

samhotchkiss:

1. The write helpers can mutate arbitrary tasks, not just inbox items. `GET /api/v1/inbox/{id}` goes through `_collect_inbox_items()` / `_task_to_inbox_item()` (`src/pollypm/web_api/service.py:1435-1511`) and returns no item for tasks that are not chat-flow or plan-review inbox rows. The POST helpers do not reuse that membership predicate: `archive_inbox_item()` only probes `svc.get()` before `svc.archive_task(..., strict=True)` (`service.py:742-763`), and `promote_inbox_to_task()` only probes `src_svc.get()` before creating a derived task (`service.py:999-1019`). The same shape applies to snooze / mark-read / reply. That means a caller can `POST /inbox/<project>/<n>/archive` and close a non-inbox work task that the inbox read surface would 404. Please resolve the source item through the canonical inbox predicate before any write, and add regressions that non-inbox task IDs are rejected without side effects for the mutating endpoints, especially archive and snooze.

2. Snooze visibility is still split across surfaces. The HTTP inbox list filters active snoozes in `src/pollypm/web_api/service.py:1481-1491`, but the cockpit/dashboard/rail path still goes through `pollypm.work.inbox_view.inbox_tasks()` (`src/pollypm/work/inbox_view.py:185-215`) via `cockpit_pg_aggregates.inbox_tasks_grouped()` / `dashboard_data._count_inbox_tasks()`, and that canonical membership predicate never calls the new `pollypm.work.inbox_snooze` helper. A snoozed item will disappear from `/api/v1/inbox` while still appearing in the cockpit inbox and rail/dashboard inbox counts. Please move the snooze predicate into the shared inbox-query layer, or otherwise make every inbox surface use the same predicate and bulk lookup, with a regression that the API list and cockpit/dashboard count agree for an active snooze.

3. Archive reason has a side effect before the archive transition is claimed. `archive_inbox_item()` writes the `archive reason:` context note at `service.py:746-755`, then calls the strict archive at `service.py:761-763`. If the strict archive loses a race or finds the item already terminal, the endpoint returns 409 but the reason note is already persisted; two concurrent archive attempts can leave two archive-reason notes with only one successful archive. Please write the reason only after the strict transition succeeds, or include it in the canonical archive transition/audit path so failed archives leave no note. Add coverage for the strict-failure/no-note case.

4. The static OpenAPI contract still misses documented 503s for these backing-store write paths. The route decorators and service helpers map backing-store failures to 503, but `docs/api/openapi.yaml:802-910` does not list 503 for archive/snooze/promote/mark-read, and reply has the same omission earlier in the same block. Please update the static YAML and conformance tests so the documented contract matches the actual error envelope.

Targeted verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_inbox_writes_endpoint.py -q --timeout=120` -> 30 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 6 passed

samhotchkiss:
Round-3 fixes pushed in 4cf4238b9.

**Blocker 1 — inbox-membership predicate on every write helper:**
- Extracted `_is_inbox_member(task)` at `src/pollypm/web_api/service.py:1613-1640`, mirroring `_task_to_inbox_item` exactly (chat-flow OR `plan_review` label OR `_is_plan_task`). The terminal-state check stays on the read side because each write helper already has its own typed conflict path (`archive_task(strict=True)`, snooze's `_ALREADY_ARCHIVED_STATUSES` pre-check) — including `state == 'closed'` here would incorrectly 404 a request that should 409.
- Refactored `_task_to_inbox_item` to call `_is_inbox_member` first so the predicate is shared, not duplicated.
- Wired the guard right after the existing `svc.get` probe in all five write helpers:
  - `archive_inbox_item` at `service.py:743-752` — before the reason-note write and the strict archive
  - `snooze_inbox_item` at `service.py:860-865` — before the status check and snooze-row write
  - `mark_read_inbox_item` at `service.py:917-928` — added a `svc.get` probe (the original code only called `mark_read` directly) then the guard
  - `reply_inbox_item` at `service.py:961-970` — same shape: probe + guard before `add_reply`
  - `promote_inbox_to_task` at `service.py:1023-1029` — guard reuses the existing `src_task = src_svc.get(...)` probe
- All five raise `not_found(f"Inbox item not found: {item_id}")` so the read and write surfaces report the same 404 shape.

**Blocker 2 — cockpit/dashboard snooze visibility via inbox_view:**
- `src/pollypm/work/inbox_view.py:185-244` adds `_filter_active_snoozes` and an optional `now=` param on `inbox_tasks`. Uses `service.latest_snoozes_bulk` (one SQL on pg) when present, falls back to per-task `get_context(entry_type='snooze', limit=1)`, degrades open when neither is on the service (preserves test-fake compatibility).
- No signature changes for `cockpit_pg_aggregates.inbox_tasks_grouped` / `dashboard_data._count_inbox_tasks` — they call `inbox_tasks(svc, project=...)` unchanged, and pg's `PgWorkService` already exposes `latest_snoozes_bulk` (from round-2), so the filter wires itself automatically on the cockpit path.
- Shared predicate stays in `pollypm.work.inbox_snooze` so the web layer and cockpit can't drift.

**Tests:**
- `tests/test_inbox_writes_endpoint.py` +6: archive/snooze/mark-read/reply/promote each reject a non-inbox task with 404 AND no mutating side effect on the stub service (assertions inside `archive_task` / `add_context` / `add_reply` / `mark_read` / `create` fire if the guard slips); a sanity test pins the positive cases (chat + `plan_review` label + plan-review flow accepted; standard flow rejected). One pre-existing test (`test_snooze_writer_persists_structured_until_iso_marker`) needed `flow_template_id = "chat"` on its stub so the new guard accepts it.
- `tests/test_inbox_view_snooze.py` (new, 4 tests): future snooze hides the row, expired snooze re-surfaces it, bulk helper is called exactly once, legacy mocks without bulk/get_context degrade open.

**Verification:**
- `pytest --noconftest tests/test_inbox_writes_endpoint.py` → 36 passed
- `pytest tests/test_pg_inbox_actions.py` → 25 passed
- `pytest tests/test_inbox_view_snooze.py tests/test_inbox_review_ordering.py tests/test_human_review_queue.py tests/test_inbox_predicates.py tests/test_inbox_plan_reviews.py tests/test_inbox_cli_filters.py` → 38 passed, 1 xfailed
- `pytest tests/web_api/test_openapi_conformance.py` → 6 passed
- `pytest tests/test_dashboard_data_pg.py tests/test_dashboard_data_alert_dedup.py` → 25 passed

**Note on the other two items in the round-3 comment:**
- #3 (archive-reason side effect) and #4 (503 in static OpenAPI) — task scoped me to blockers 1 + 2 ("Close both round-3 Codex blockers"). Happy to do a follow-up PR for #3 and #4 if you want them in this round; flag and I'll spin one up.

`needs-review` label re-applied; Codex review should pick this up in the next 5-min cycle.

samhotchkiss:

This PR is still blocked.

1. It is not mergeable against current `main`. GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`; the branch is based behind both #2057 and the newly-merged #2056. Please rebase onto current `main` (`ff27b59e`) and preserve the Dashboard, Config, and Inbox OpenAPI/app route additions.

2. The archive reason side effect from the previous review is still present. In the latest head, `src/pollypm/web_api/service.py:754-770` writes `svc.add_context(..., entry_type="note")` with `archive reason: ...` before `svc.archive_task(item_id, actor=actor, strict=True)`. If the strict archive loses a race or returns `409 invalid_state`, the reason note has already been persisted for a failed archive. Please move the reason write after the strict transition succeeds, or include the reason in the canonical archive transition/audit path, and add coverage that a strict-failure/concurrent-loser leaves no note.

3. The static OpenAPI contract still omits the documented backing-store `503` responses for the inbox write paths. The FastAPI route decorators list `503` in `src/pollypm/web_api/routes/inbox.py`, and the service helpers map `_BACKING_STORE_ERRORS` to `service_unavailable`, but `docs/api/openapi.yaml` still lacks `503` under `/inbox/{id}/archive`, `/snooze`, `/promote-to-task`, `/mark-read`, and `/reply`. Please update the static YAML and pin it in conformance coverage so generated clients see the actual error envelope.

I did not run the full targeted suite on this pass because the PR has no clean merged tree. The code inspection above is against latest head `4cf4238b9`.

samhotchkiss:
Round-4 fixes pushed in b510d8338 (rebased onto current `main` at `ff27b59e7`).

**Blocker 1 — rebase onto current main:**
- `git rebase origin/main` replayed the four feat/fix commits onto `ff27b59e7` (includes #2057 dashboard + #2056 config).
- One conflict on `docs/api/openapi.yaml`: main inserted `/config` + `/config/projects/{key}` paths at the same location our branch was extending the inbox section. Resolved by keeping inbox-write paths (snooze / promote-to-task / mark-read) contiguous with the existing reply + archive blocks, then the config blocks. Tags section (Dashboard + Config + Inbox) auto-merged cleanly. `app.py` was not touched by our branch so the dashboard + config route mounts came in as-is.

**Blocker 2 — archive-reason ordering:**
- `src/pollypm/web_api/service.py:754-792` (post-rebase): swapped the order so `svc.archive_task(item_id, actor=actor, strict=True)` runs FIRST, and the `add_context(entry_type="note", body="archive reason: ...")` only runs after the strict transition succeeds.
- The note write stays best-effort (`try/except` with a debug log) — a failed audit note must not roll back the already-committed archive. The strict transition itself remains the source of truth.
- Regression test in `tests/test_inbox_writes_endpoint.py`: a stub service whose `archive_task` raises `InvalidTransitionError` (simulating the rowcount==0 concurrent loser). The test asserts the helper raises `APIError(status_code=409, code="invalid_state")` AND that `add_context` was never called. Verified the test FAILS on the broken (pre-fix) ordering and PASSES after the reorder. A complementary happy-path test pins call ordering (`archive_task` index < `add_context` index in the recorded order list) so a too-aggressive refactor that drops the note entirely would also trip.

**Blocker 3 — static OpenAPI 503 responses:**
- Added a reusable `#/components/responses/ServiceUnavailable` response component (description matches the typed `service_unavailable` envelope from `_BACKING_STORE_ERRORS`).
- All 5 inbox-write paths in `docs/api/openapi.yaml` now declare `"503": $ref: ...ServiceUnavailable`: `/inbox/{id}/reply`, `/inbox/{id}/archive`, `/inbox/{id}/snooze`, `/inbox/{id}/promote-to-task`, `/inbox/{id}/mark-read`.
- New conformance test `test_static_yaml_declares_503_on_inbox_write_paths` in `tests/web_api/test_openapi_conformance.py` pins the contract: every inbox-write POST must list `503` under its `responses:` block.

**Verification:**
- `pytest --noconftest tests/test_inbox_writes_endpoint.py --timeout=120` -> 38 passed (was 36; +2 new ordering tests)
- `pytest tests/web_api/test_openapi_conformance.py --timeout=120` -> 7 passed (was 6; +1 new 503 test)
- `pytest tests/test_pg_inbox_actions.py --timeout=180` -> 25 passed

Head SHA: `b510d8338`. `needs-review` label re-applied; Codex review should pick this up in the next 5-min cycle.

samhotchkiss:

The round-4 fixes address the previous archive ordering, static OpenAPI 503, strict archive, and bulk snooze blockers. I am still blocking on the inbox membership/ownership boundary.

`src/pollypm/web_api/service.py:1662-1687` defines `_is_inbox_member` in the web layer, and every mutating helper uses that local predicate before writing (`archive_inbox_item`, `snooze_inbox_item`, `mark_read_inbox_item`, `reply_inbox_item`, `promote_inbox_to_task`). That predicate is not the shared cockpit/rail/dashboard inbox predicate in `src/pollypm/work/inbox_view.py:165-179`: it accepts any `flow_template_id == "chat"` plus substring plan labels / `_is_plan_task`, while the shared work predicate requires a nonterminal task plus user role, exact `plan_review`, or the current flow node being human.

That makes the PR a write-side drift risk, not just a read-side display difference. A chat-flow task that the cockpit inbox, rail, and dashboard would not surface can still be treated as an API inbox item and archived, snoozed, replied to, marked read, or promoted through `/api/v1/inbox/{id}/...`. The substring plan check also widens the surface for labels like `not_plan_review` / `planning`.

Please move API list/detail/write resolution through the shared inbox-query layer (`pollypm.work.inbox_view.inbox_tasks` / `is_inbox_task`) or extract a canonical helper that both API and cockpit/rail/dashboard use. If the API intentionally has a broader inbox contract, document that contract explicitly and add regression coverage proving which non-cockpit tasks may be mutated and why. At minimum, add negative tests for a chat-flow task with no user role/current human node and for near-miss plan labels so the write surface cannot widen by substring accident.

Targeted verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_inbox_writes_endpoint.py -q --timeout=120` -> 38 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 7 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_pg_inbox_actions.py -q --timeout=180` -> 25 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_inbox_view_snooze.py tests/test_inbox_review_ordering.py tests/test_human_review_queue.py tests/test_inbox_predicates.py tests/test_inbox_plan_reviews.py tests/test_inbox_cli_filters.py -q --timeout=120` -> 38 passed, 1 xfailed
- `uv run --extra test --with pytest-timeout pytest tests/test_dashboard_data_pg.py tests/test_dashboard_data_alert_dedup.py -q --timeout=180` -> 25 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_import_boundary.py -q --timeout=120` -> 1 failed, 15 passed; failure is existing direct Supervisor imports in `src/pollypm/cli_features/tier4.py` and `src/pollypm/web_api/routes/chat_send.py`, not files touched by this PR
- `uv run --extra test --with pytest-timeout pytest tests/web_api/test_endpoints.py -k inbox -q --timeout=120` -> 1 failed, 3 passed; the empty-state test read existing local `myproj` inbox rows in this environment, so I am not treating that as this PRs blocker

samhotchkiss:
Round-5 fix pushed in `d34174073`.

**Blocker — inbox write predicate drift:**

Routed all API inbox write resolution through the canonical `pollypm.work.inbox_view.is_inbox_task` (Path A — direct import). The web-layer `_is_inbox_member` is now a thin delegate kept only as an import-stable shim for legacy callers; the actual gate is the same predicate cockpit, the rail, and the dashboard use.

Call sites updated (all in `src/pollypm/web_api/service.py`):
- `archive_inbox_item:756` — `is_inbox_task(src_task, svc)`
- `snooze_inbox_item:880` — `is_inbox_task(current, svc)`
- `mark_read_inbox_item:951` — `is_inbox_task(src_task, svc)`
- `reply_inbox_item:1003` — `is_inbox_task(src_task, svc)`
- `promote_inbox_to_task:1066` — `is_inbox_task(src_task, src_svc)` (uses source-project svc)
- `_collect_inbox_items:1567` (read side) — `_task_to_inbox_item(task, svc)` now threads svc → `is_inbox_task` for single-source-of-truth. Iteration moved inside the `with` block so `svc.get_flow()` for the current-human-node branch runs on an open handle.

**Why Path A:** `is_inbox_task(task, service, *, flow_cache=None)` only needs the service handle for the current-node-human fallback (`service.get_flow(...)`), and every write helper already has the work-service open. No signature mismatch — direct import was clean.

**Canonical predicate hardening (`src/pollypm/work/inbox_view.py`):**
- `_roles_match_user`, `_current_node_is_human`, `_is_plan_review_label`, `is_inbox_task` now use `getattr(task, ..., default)` for the duck-typed test-stub case. Real `Task` instances always carry these fields, so production behavior is unchanged.
- Added an explicit docstring note that `plan_review` is an exact-equality match (not substring) — Codex's exact callout.

**`_is_inbox_member` disposition:** kept as a thin one-line delegate to `is_inbox_task` so the existing positive-case sanity test still has a stable web-layer import point, but every production call site uses `is_inbox_task` directly. No live caller bypasses the canonical predicate.

**Tests added (6 new; `tests/test_inbox_writes_endpoint.py`):**
- `test_archive_rejects_chat_flow_task_with_no_user_role`
- `test_archive_rejects_near_miss_plan_label` (`labels=["not_plan_review", "planning"]`)
- `test_snooze_rejects_chat_flow_task_with_no_user_role`
- `test_snooze_rejects_near_miss_plan_label`
- `test_promote_rejects_chat_flow_task_with_no_user_role`
- `test_promote_rejects_near_miss_plan_label`

Plus rewrote `test_membership_helper_accepts_chat_and_plan_review` → `test_membership_helper_matches_canonical_inbox_predicate` to pin the canonical positive shapes (chat+user role, exact `plan_review` label) AND the new negative shapes (chat-without-user-role, `not_plan_review` near-miss).

**Verified the new tests fail on round-3 predicate:** simulated the old `_is_inbox_member` against the two new stub shapes — `_ChatNoUserRoleStubTask` and `_NearMissPlanLabelStubTask` are both accepted by the old `chat OR "plan_review" in label` predicate, so the write helpers would have called `archive_task` / `add_context` / `create` and tripped the `AssertionError`s in `_CanonicalRejectingSvc`. With the fix, the canonical predicate rejects both → 404 raised → tests pass.

**Verification:**
- `pytest tests/test_inbox_writes_endpoint.py tests/test_pg_inbox_actions.py tests/web_api/test_openapi_conformance.py tests/test_inbox_predicates.py tests/test_inbox_plan_reviews.py -x --timeout=180` → **93 passed** (was 87; +6 new round-5 tests)
- `pytest tests/test_inbox_view_snooze.py tests/test_inbox_review_ordering.py tests/test_human_review_queue.py tests/test_inbox_cli_filters.py tests/test_dashboard_data_pg.py -x --timeout=180` → 26 passed, 1 xfailed (the existing xfail is unrelated)

Head SHA: `d34174073`. `needs-review` re-applied.

samhotchkiss:

The round-5 write-membership blocker is addressed: the mutating helpers now route through `pollypm.work.inbox_view.is_inbox_task`, and the new negative tests cover the chat-without-user-role and near-miss-label cases. I am still blocking on two production contract issues.

1. The documented `503 service_unavailable` path does not catch real pg outages. `src/pollypm/work/factory.py:11-13` says the work service is pg-only now, but `src/pollypm/web_api/service.py:72-76` defines `_BACKING_STORE_ERRORS` as only `sqlite3.OperationalError`, `sqlite3.DatabaseError`, and `OSError`. The new write helpers all rely on that tuple for their 503 mapping (`archive_inbox_item`, `snooze_inbox_item`, `mark_read_inbox_item`, `reply_inbox_item`, `promote_inbox_to_task`), while the routes and static OpenAPI now promise 503 for backing-store failures. `psycopg.OperationalError` and `psycopg_pool.PoolTimeout` are not subclasses of `OSError` (verified locally), so a real pg outage/pool exhaustion will bypass the typed envelope and return 500 for these new endpoints. Please extend the central backing-store error classifier to include the pg/pool failure classes and add regressions where an inbox write sees a `psycopg.OperationalError` / pool timeout from service construction or `svc.get()` and returns `503 service_unavailable`.

2. The API inbox read path still duplicates part of the canonical inbox query instead of using the shared query/cache. `_collect_inbox_items()` loops over `svc.list_tasks()` and calls `_task_to_inbox_item(task, svc)` (`src/pollypm/web_api/service.py:1539-1557`); `_task_to_inbox_item()` calls `_is_inbox_member()` without passing a shared `flow_cache` (`service.py:1713-1715`), and `_is_inbox_member()` then calls `is_inbox_task()` with a fresh cache per task (`service.py:1695-1696`). The shared cockpit/rail/dashboard path in `src/pollypm/work/inbox_view.py:302-305` deliberately reuses one `flow_cache`, so the API list can now do one `get_flow()` / `resolve_flow()` per task on a user-facing scan when many rows share the same flow. Please either route API list/detail through `inbox_view.inbox_tasks()` or thread a single cache through `_collect_inbox_items` and `_task_to_inbox_item`. Add a regression with multiple current-human-node tasks on the same flow proving the API list resolves the flow once, not N times.

While touching `_task_to_inbox_item`, also remove the remaining substring plan-review classifier at `src/pollypm/web_api/service.py:1717-1720`; a canonical human inbox task with a label like `not_plan_review` can still be rendered as `type="plan_review"` even though the shared membership predicate now treats `plan_review` as exact-match only. That should use the same exact/shared classification as the write gate.

Verification run on this head:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_inbox_writes_endpoint.py -q --timeout=120` -> 44 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 7 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_pg_inbox_actions.py -q --timeout=180` -> 25 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_inbox_view_snooze.py tests/test_inbox_review_ordering.py tests/test_human_review_queue.py tests/test_inbox_predicates.py tests/test_inbox_plan_reviews.py tests/test_inbox_cli_filters.py -q --timeout=180` -> 38 passed, 1 xfailed

samhotchkiss:
Round-6 fixes pushed in `5699230eb`.

**Blocker 1 — `_BACKING_STORE_ERRORS` covers pg outages:**
- `src/pollypm/web_api/service.py:85-91` now lists `psycopg.OperationalError` and `psycopg_pool.PoolTimeout` alongside the existing sqlite + OSError entries. pg-only post #1971 (`pollypm.work.factory:11-13`) and both psycopg / psycopg-pool are required project deps (`pyproject.toml`), so the imports at `service.py:28-29` are unconditional — no try/except guarding needed.
- New tests in `tests/test_inbox_writes_endpoint.py`:
  - `test_archive_returns_503_on_psycopg_operational_error`
  - `test_archive_returns_503_on_pool_timeout`
  - `test_snooze_returns_503_on_psycopg_operational_error`
  - `test_promote_returns_503_on_pool_timeout`
  - `test_backing_store_errors_includes_psycopg_classes` (direct introspection so a refactor that drops either class from the tuple trips CI even without exercising the exception path).

**Blocker 2 — shared flow_cache through API inbox list:**
- Threaded `flow_cache` parameter through `_is_inbox_member` (`service.py:1700`) and `_task_to_inbox_item` (`service.py:1747`) so both forward it to the canonical `pollypm.work.inbox_view.is_inbox_task`. One cache is built per project scan in `_collect_inbox_items` at `service.py:1568-1582` and passed to every `_task_to_inbox_item` call. Mirrors the cockpit / rail / dashboard path at `pollypm.work.inbox_view.inbox_tasks:302-305`. The 5 single-shot write callsites (`service.py:770, 895, 966, 1019, 1084`) keep their cache-less `is_inbox_task(...)` calls — one lookup per request, no benefit from a cache.
- New test `test_inbox_list_resolves_flow_once_per_unique_flow_id`: 5 tasks with `current_node_id="n0"`, no user role, no plan_review label — forces the canonical predicate down to the current-human-node branch that calls `svc.get_flow`. Counting svc asserts exactly 1 call (was 5 with the per-call fresh cache).

**Blocker 3 — substring → exact plan-review classifier:**
- `src/pollypm/web_api/service.py:1751-1757`: replaced `any("plan_review" in lbl for lbl in labels)` with `any(lbl == "plan_review" for lbl in labels)`, and dropped the `_is_plan_task(task)` widening. The membership gate (`is_inbox_task`) already decides surface eligibility via `_is_plan_review_label` (exact match), and this classifier is just choosing `type="plan_review"` vs `type="message"` for already-accepted inbox rows — exact match here matches the gate.
- New test `test_inbox_item_type_uses_exact_plan_review_label`: a canonical inbox row (user role) with labels `["not_plan_review", "planning"]` now classifies as `type="message"`, not `type="plan_review"`.

**Verification:**
- `pytest tests/test_inbox_writes_endpoint.py tests/test_pg_inbox_actions.py tests/web_api/test_openapi_conformance.py tests/test_inbox_predicates.py tests/test_inbox_plan_reviews.py --timeout=180` → **100 passed** (was 93; +7 new round-6 tests).

Head SHA: `5699230eb`. `needs-review` re-applied.

samhotchkiss:

The implementation fix for archive-reason ordering is correct, and the targeted suites pass, but the static OpenAPI contract is now stale. `docs/api/openapi.yaml:2230-2237` still says `InboxArchiveRequest.reason` is recorded "as a context note before the state transition." The latest code deliberately does the opposite: `src/pollypm/web_api/service.py:780-805` runs `svc.archive_task(..., strict=True)` first and only writes the reason note after the transition succeeds.

That doc line reintroduces the unsafe contract round 4 fixed: API readers/generated clients can infer that a failed or concurrent-loser archive leaves a reason note, while the code now correctly guarantees failed archives leave no note. Please update the static schema text to say the reason is recorded only after a successful archive transition (best-effort context note), and add/adjust conformance coverage so this cannot drift again.

Verification I ran:
- `uv run --extra test --with pytest-timeout pytest tests/test_inbox_writes_endpoint.py tests/test_pg_inbox_actions.py tests/test_inbox_view_snooze.py tests/test_inbox_predicates.py tests/test_inbox_plan_reviews.py --timeout=180` -> 97 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py --timeout=120` -> 7 passed

samhotchkiss:
Round-7 fix pushed in `2c0649b02`.

**Doc-sync — InboxArchiveRequest.reason ordering:**
- `docs/api/openapi.yaml:2230-2240` reason description now reads "recorded as a best-effort context note AFTER the strict archive transition succeeds. If the transition fails (e.g. concurrent archive, invalid state, backing-store outage), no note is persisted." This matches the round-4 implementation in `src/pollypm/web_api/service.py:780-811` where `add_context` only fires inside the `if reason:` block AFTER `archive_task(strict=True)` returns successfully.
- `tests/web_api/test_openapi_conformance.py` adds `test_inbox_archive_reason_documented_as_post_transition` to pin the contract: asserts the description contains "after" + "transition" and explicitly forbids the old "before the state transition" phrasing.

**Verification:**
```
python -m pytest tests/test_inbox_writes_endpoint.py tests/web_api/test_openapi_conformance.py -x
============================== 59 passed in 94.23s ===============================
```

`needs-review` re-added. Head SHA `2c0649b02`. PR: https://github.com/samhotchkiss/pollypm/pull/2060

Issue #1830, opened 2026-05-19
[pg] cluster-A port leaves session_runtime/events on stale StateStore call sites


samhotchkiss:
## Summary

Reviewing `44926faa` / #1828 found that the new `pg_sessions` cluster-A facade is only used by selected Supervisor internals. Several public, heartbeat, and cockpit call sites still use `supervisor.store.*`, and `Supervisor.__init__` still assigns `self.store = StateStore(config.project.state_db)`. In pg mode that leaves session runtime and event reads/writes split between Postgres and the legacy SQLite store.

## Evidence

- `src/pollypm/supervisor.py:436-503` adds backend-aware helpers, but those helpers are private and only cover internal Supervisor call sites.
- `src/pollypm/heartbeats/api.py:230-235` still writes status through `supervisor.store.upsert_session_runtime(...)`. `src/pollypm/heartbeats/local.py:750-753` and `:857-861` still read runtime through `api.supervisor.store.get_session_runtime(...)`.
- `src/pollypm/service_api/v1.py:204` and `:239` still read runtime through `supervisor.store.get_session_runtime(...)`; `:447-469` still writes status and reads it back through `supervisor.store`.
- `src/pollypm/cockpit_ui.py:1116` and `:1161` still write recovery state through `supervisor.store.upsert_session_runtime(...)`.
- `src/pollypm/cockpit_ui.py:2096`, `src/pollypm/cockpit_rail.py:1437`, and `src/pollypm/cockpit_rail.py:5136` still read events via `supervisor.store.recent_events(...)`.
- `src/pollypm/checkpoints.py:653-665` explicitly leaves `last_checkpoint_path` updates on the passed `StateStore`, while `src/pollypm/agent_profiles/defaults.py:686-691` now reads the latest checkpoint pointer from `pg_sessions` under the pg backend. That makes newly written checkpoints invisible to prompt/recovery readers in pg mode.

## Impact

With `[storage] backend = "postgres"`, cluster-A state can diverge across PG and SQLite:

- heartbeat recovery status, disabled/switching flags, recovery attempts, and manual recovery actions can be written to SQLite while new readers consult PG;
- `last_checkpoint_path` can remain unset in PG, so worker prompts and recovery context lose the latest checkpoint;
- cockpit/rail event feeds can keep reading stale or empty SQLite events even though event writes have moved to the unified PG-backed message store.

That violates the #1828 cluster-A migration contract and can cause operational decisions to be made from stale state.

[...]

PR #2061, opened 2026-05-22
feat(api): /api/v1/sessions — Phase 2 sessions admin (list/restart/pause/resume)

samhotchkiss:
## Summary
- Phase 2 surface #8 from `~/Desktop/pollypm-phase2-endpoints-spec.md` §10 — `GET /api/v1/sessions`, `GET /api/v1/sessions/{name}`, plus `POST .../restart`, `.../pause`, `.../resume`.
- Mirrors `pm sessions` (#2039) classification (healthy / stale / missing / unknown) plus a new `paused` state surfaced from `<base_dir>/paused-sessions.json`.
- Restart goes through `TmuxSessionService.destroy` + `create` with an `unsafe_mid_turn` safety gate overridable via `?safety=force`; `?safety=force` matches the spec's §10.3 risk taxonomy.

## Files
- `src/pollypm/web_api/routes/sessions_admin.py` — router (5 handlers).
- `src/pollypm/web_api/app.py` — wires the router under `/api/v1` with the standard bearer-auth dependency.
- `tests/test_sessions_admin_endpoint.py` — 19 cases (list / detail / restart / pause / resume + auth + pg outage degradation + service-unavailable).
- `docs/api/openapi.yaml` — paths + `Sessions` tag + new `SessionInfo` / `SessionDetail` / `SessionConfigView` / `SessionHealthSnapshot` / `SessionsListResponse` schemas + `SessionName` path param.

## Test plan
- [x] `pytest --noconftest tests/test_sessions_admin_endpoint.py -v` — 19 passed
- [x] `pytest --noconftest tests/web_api/test_openapi_conformance.py` — YAML still validates as OpenAPI 3.1; emitted-impl OpenAPI still validates
- [ ] Manual smoke against a real `pm serve` (deferred — covered by Phase 2 §15 Q9 user-testing protocol)

## Spec gaps / follow-ups
- Pause/resume currently lives in a project-local JSON marker (`<base_dir>/paused-sessions.json`). The recovery loop / supervisor doesn't yet consume it; this PR is informational + cockpit-visible. Wiring the supervisor to honor the marker is a follow-up (worth its own issue — out of scope for this small PR per spec §14 Q9 "1 PR (restart/pause/resume)").
- Worker (per-task) sessions are intentionally NOT addressable on this surface (spec §10.2 + §13.9 — `/api/v1/chat/sessions` already exposes worker chat surfaces); the 404 hint points clients at that surface.

### DIFF
--- a/src/pollypm/web_api/routes/sessions_admin.py
+++ b/src/pollypm/web_api/routes/sessions_admin.py
@@ -19,9 +19,12 @@ Endpoints
   ``unsafe_mid_turn_unknown`` when the safety probe itself fails)
   unless ``?safety=force``.
 - ``POST   /api/v1/sessions/{name}/pause``     — write an
-  **informational** pause marker (consumers can observe it via
-  ``GET /api/v1/sessions`` but the supervisor / recovery / dispatch
-  loops do NOT yet consume it; see ``ActionResult.message``).
+  **informational** pause marker. Consumers can observe it via the
+  separate ``paused: bool`` field on ``GET /api/v1/sessions`` rows;
+  the ``status`` field is unaffected (Codex PR #2061 round 2 —
+  pause must NEVER mask the real runtime-health classification).
+  The supervisor / recovery / dispatch loops do NOT yet consume
+  this marker (see ``ActionResult.message`` and #2068).
   Idempotent.
 - ``POST   /api/v1/sessions/{name}/resume``    — remove the marker.
   Idempotent.
@@ -61,14 +64,21 @@ Design notes
 
 * **Pause / resume — informational only.** No existing supervisor /
   recovery / dispatch loop consumes the pause marker today, so this
-  surface is documented as **informational** (the cockpit + this API's
-  ``GET /sessions`` surface ``status="paused"``, but background loops
-  still act on the session). The response ``message`` and OpenAPI
-  description both call this out so operators are not misled into
-  thinking the marker quiesces the daemon. Daemon-side enforcement is
-  a follow-up (issue filed in the PR). Both operations are idempotent
-  and return 200, and concurrent writes are serialised by an
-  ``fcntl.flock`` on the marker file.
+  surface is documented as **informational**. Crucially, the marker
+  is exposed as a *separate* ``paused: bool`` field on the
+  ``SessionInfo`` row rather than as a value of ``status`` (Codex
+  PR #2061 round 2). Folding it into ``status`` was incoherent with
+  the round-1 "informational only" stance: a paused-but-missing
+  session reported ``status="paused"`` and operators believed the
+  daemon had quiesced when in fact the session was simply absent.
+  Now ``status`` reflects pure runtime health (``healthy`` /
+  ``stale`` / ``missing`` / ``unknown``) and ``paused`` carries the
+  operator intent. The response ``message`` and OpenAPI description
+  both call this out so operators are not misled into thinking the
+  marker quiesces the daemon. Daemon-side enforcement is tracked as
+  follow-up #2068. Both operations are idempotent and return 200,
+  and concurrent writes are serialised by an ``fcntl.flock`` on the
+  marker file.
 
 * **pg outages → 503.** ``GET`` endpoints downgrade pg outages to a
   best-effort response (``status="unknown"`` on the affected row).
@@ -138,11 +148,15 @@ class SessionInfo(BaseModel):
     window_name: str
     tmux_session: str
     window_present: bool
-    status: Literal["healthy", "stale", "missing", "unknown", "paused"]
+    status: Literal["healthy", "stale", "missing", "unknown"]
     last_heartbeat_iso: str | None = None
     last_heartbeat_age_seconds: int | None = None
     auth_token_present: bool
     enabled: bool
+    # Informational marker (Codex PR #2061 round 2). Decoupled from
+    # ``status`` so a paused-but-missing or paused-but-stale session
+    # still reports its true runtime health — the daemon does NOT
+    # consume this marker today (see #2068).
     paused: bool = False
 
 
@@ -291,17 +305,22 @@ def _classify_status(
     *,
     window_present: bool,
     age_seconds: int | None,
-    paused: bool,
 ) -> str:
-    """Return one of healthy/stale/missing/unknown/paused.
+    """Return one of healthy/stale/missing/unknown.
 
-    Matches :func:`pollypm.cli_features.sessions_health._classify_status`
-    except we surface ``paused`` first — operator intent overrides the
-    runtime signals (a paused session may still have a healthy pane,
-    but the cockpit / recovery loop should treat it as quiesced).
+    Mirrors :func:`pollypm.cli_features.sessions_health._classify_status`
+    exactly: a pure runtime-health classification derived from tmux
+    presence + heartbeat age.
+
+    The pause marker is intentionally NOT consulted here (Codex PR
+    #2061 round 2). Pause is informational only — the supervisor /
+    recovery / dispatch loops do not consume it (#2068) — so folding
+    it into ``status`` would mask the real runtime state and let an
+    operator believe they had quiesced the daemon when in fact a
+    paused-but-missing session was reporting ``status="paused"``
+    instead of ``status="missing"``. Callers consume the separate
+    :attr:`SessionInfo.paused` boolean for the informational signal.
     """
-    if paused:
-        return "paused"
     if not window_present:
         return "missing"
     if age_seconds is None:
@@ -485,7 +504,6 @@ def _build_session_info(
     status = _classify_status(
         window_present=window_present,
         age_seconds=age,
-        paused=paused,
     )
     auth_token_present = bool(getattr(session, "auth_token", "") or "")
     return SessionInfo(
@@ -686,9 +704,12 @@ def _find_session(config: Any, name: str) -> Any:
 def list_sessions_endpoint(config: ConfigDep) -> SessionsListResponse:
     """GET /api/v1/sessions — mirror ``pm sessions``.
 
-    Returns one row per configured (and enabled) session. ``paused``
-    rows come from the pause marker (see :func:`_load_paused_names`);
-    ``status`` follows the same thresholds the CLI uses.
+    Returns one row per configured (and enabled) session. ``status``
+    is the pure runtime-health classification (``healthy`` / ``stale``
+    / ``missing`` / ``unknown``) and follows the same thresholds the
+    CLI uses. ``paused`` is a separate informational boolean drawn
+    from the pause marker (see :func:`_load_paused_names`) and does
+    NOT influence ``status`` — see :func:`_classify_status` for why.
     """
     storage_session = _storage_session_name(config)
     windows = _list_storage_closet_windows(storage_session)
@@ -856,8 +877,12 @@ def pause_session_endpoint(name: str, config: ConfigDep) -> ActionResult:
 
     Writes ``<base_dir>/paused-sessions.json`` (atomic tmp+rename,
     serialised via ``fcntl.flock`` against concurrent pause/resume
-    callers) so ``GET /api/v1/sessions`` can surface
-    ``status="paused"``. Idempotent: pausing an already-paused
+    callers) so ``GET /api/v1/sessions`` surfaces ``paused=true`` on
+    the affected row. The ``status`` field continues to reflect
+    actual runtime health (``healthy`` / ``stale`` / ``missing`` /
+    ``unknown``) — a paused-but-missing session reports
+    ``status="missing"`` AND ``paused=true``, not ``status="paused"``
+    (Codex PR #2061 round 2). Idempotent: pausing an already-paused
     session returns 200 with no state change.
     """
     _find_session(config, name)  # 404 if unknown
--- a/src/pollypm/web_api/routes/sessions_admin.py
+++ b/src/pollypm/web_api/routes/sessions_admin.py
@@ -11,13 +11,20 @@ Endpoints
   sessions with live health (heartbeat age + tmux window presence).
 - ``GET    /api/v1/sessions/{name}``           — one session's detail
   payload (config row + health snapshot + last heartbeat).
-- ``POST   /api/v1/sessions/{name}/restart``   — kill the tmux window
-  and re-create it. Refuses with ``409 unsafe_mid_turn`` when the agent
-  is mid-turn unless ``?safety=force``.
-- ``POST   /api/v1/sessions/{name}/pause``     — write a pause marker
-  so the recovery / dispatch loops skip this session. Idempotent.
-- ``POST   /api/v1/sessions/{name}/resume``    — remove the pause
-  marker. Idempotent.
+- ``POST   /api/v1/sessions/{name}/restart``   — destroy + relaunch
+  via the canonical :meth:`pollypm.supervisor.Supervisor.restart_session`
+  facade so the relaunched pane runs the configured provider command
+  (not an ad-hoc ``echo`` placeholder). Refuses with
+  ``409 unsafe_mid_turn`` when the agent is mid-turn (or 503
+  ``unsafe_mid_turn_unknown`` when the safety probe itself fails)
+  unless ``?safety=force``.
+- ``POST   /api/v1/sessions/{name}/pause``     — write an
+  **informational** pause marker (consumers can observe it via
+  ``GET /api/v1/sessions`` but the supervisor / recovery / dispatch
+  loops do NOT yet consume it; see ``ActionResult.message``).
+  Idempotent.
+- ``POST   /api/v1/sessions/{name}/resume``    — remove the marker.
+  Idempotent.
 
 Design notes
 ------------
@@ -31,19 +38,37 @@ Design notes
   read failures collapse to ``unknown`` / ``missing`` rather than
   500-ing.
 
-* **Restart.** Uses :class:`pollypm.session_services.tmux.TmuxSessionService`
-  to ``destroy`` then ``create``. The CLI is the only other caller that
-  performs that pair, and we deliberately re-use that helper so any
-  future change to launch automation lands in both surfaces at once.
-  Restart is intentionally synchronous (spec §10.4) — typical restart
-  time is < 5s and the spec defers ``?async`` to Phase 2.5.
+* **Restart.** Routes through :meth:`pollypm.supervisor.Supervisor.restart_session`
+  — the **same facade** ``pm switch-session-account``, the cockpit
+  account-switch action (``cockpit_ui.restart_session``), and the
+  upgrade flow already use. That facade owns the canonical kill +
+  launch-spec relaunch pair: it tears down the existing window, calls
+  :meth:`Supervisor.launch_session` (which consults the launch planner
+  for the full provider/account/cwd/command), and injects the recovery
+  prompt. The previous incarnation of this route called
+  :class:`TmuxSessionService` ``destroy`` + ``create`` directly with no
+  command spec, so ``create()`` fell back to its
+  ``echo 'No command for {name}'`` placeholder — destroying live
+  agents and replacing them with non-agent panes. See PR #2061 Codex
+  P0 #1.
 
-* **Pause / resume.** No existing helper does this; we use a small
-  on-disk JSON marker at ``<base_dir>/paused-sessions.json`` so the
-  state survives daemon restarts and is visible to any other reader
-  (the supervisor will learn to consume this in a follow-up; for now
-  the marker is informational + the API surfaces ``status="paused"``).
-  Both operations are idempotent and return 200.
+  The account passed to ``restart_session`` is the runtime's
+  ``effective_account`` if set (so a session that previously failed
+  over stays on the recovered account), otherwise the session's
+  configured ``account``. Restart is intentionally synchronous (spec
+  §10.4) — typical restart time is < 5s and the spec defers
+  ``?async`` to Phase 2.5.
+
+* **Pause / resume — informational only.** No existing supervisor /
+  recovery / dispatch loop consumes the pause marker today, so this
+  surface is documented as **informational** (the cockpit + this API's
+  ``GET /sessions`` surface ``status="paused"``, but background loops
+  still act on the session). The response ``message`` and OpenAPI
+  description both call this out so operators are not misled into
+  thinking the marker quiesces the daemon. Daemon-side enforcement is
+  a follow-up (issue filed in the PR). Both operations are idempotent
+  and return 200, and concurrent writes are serialised by an
+  ``fcntl.flock`` on the marker file.
 
 * **pg outages → 503.** ``GET`` endpoints downgrade pg outages to a
   best-effort response (``status="unknown"`` on the affected row).
@@ -59,6 +84,7 @@ from __future__ import annotations
 
 import json
 import logging
+from contextlib import contextmanager
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Annotated, Any, Literal
@@ -193,6 +219,30 @@ def _unsafe_mid_turn(name: str) -> APIError:
             "Re-issue with ?safety=force to override."
         ),
         hint="Wait for the agent to finish or pass ?safety=force to restart anyway.",
+    )
+
+
+def _unsafe_mid_turn_unknown(name: str, detail: str) -> APIError:
+    """Strict-mode safety probe could not evaluate ``is_turn_active``.
+
+    Fail-closed counterpart to :func:`_unsafe_mid_turn` (Codex PR #2061
+    P0 #2). The previous incarnation of ``_is_turn_active`` swallowed
+    every exception and returned ``False`` — a transient tmux / parse
+    failure therefore looked like a green-light to restart an actively
+    working agent. Strict mode now surfaces probe failures as 503 so
+    the caller knows to retry or escalate to ``?safety=force``.
+    """
+    return APIError(
+        status_code=503,
+        code="unsafe_mid_turn_unknown",
+        message=(
+            f"Refusing to restart {name!r}: mid-turn safety probe "
+            f"failed ({detail}). Cannot confirm whether the agent is idle."
+        ),
+        hint=(
+            "Retry once the tmux service is healthy, or pass "
+            "?safety=force to override the probe (destructive)."
+        ),
     )
 
 
@@ -355,6 +405,59 @@ def _write_paused_names(config: Any, names: set[str]) -> None:
     tmp.replace(path)
 
 
+# Operator-facing message appended to pause/resume responses so a CLI
+# user or automation script knows the marker is *informational only*
+# (Codex PR #2061 P0 #3). The supervisor / recovery / dispatch loops
+# do NOT yet consult this marker — making this clear in the response
+# stops operators from believing they have quiesced the daemon when
+# they have only tagged the session.
+_PAUSE_INFORMATIONAL_NOTE = (
+    "informational marker only — supervisor / recovery / dispatch "
+    "loops do NOT yet consult it; daemon-side enforcement is a "
+    "follow-up. The marker is visible via GET /api/v1/sessions."
+)
+
+
+@contextmanager
+def _pause_marker_lock(config: Any):
+    """Serialise pause/resume read-modify-write across concurrent callers.
+
+    Codex PR #2061 P0 #3 flagged the unguarded ``_load_paused_names``
+    → mutate → ``_write_paused_names`` sequence as racy: two
+    near-simultaneous pause + resume calls could clobber each other's
+    writes. We take an ``fcntl.flock`` on a sibling ``.lock`` file
+    that lives in the same directory as the marker — best-effort on
+    platforms without ``fcntl`` (Windows), which is fine for the v1
+    RC where the API runs on macOS/Linux only.
+    """
+    path = _pause_marker_path(config)
+    if path is None:
+        # No base_dir → no lock to take; caller will hit the same 503
+        # in ``_write_paused_names`` anyway.
+        yield
+        return
+    path.parent.mkdir(parents=True, exist_ok=True)
+    lock_path = path.with_suffix(path.suffix + ".lock")
+    try:
+        import fcntl  # type: ignore[import-not-found]
+    except ImportError:
+        # No fcntl (e.g. Windows): degrade to no-op lock. The marker
+        # write is still atomic via tmp+rename; only the
+        # read-modify-write window is unprotected, which matches the
+        # pre-#2061 behaviour.
+        yield
+        return
+    fh = open(lock_path, "a+")  # noqa: SIM115 — closed in finally
+    try:
+        fcntl.flock(fh.fileno(), fcntl.LOCK_EX)
+        try:
+            yield
+        finally:
+            fcntl.flock(fh.fileno(), fcntl.LOCK_UN)
+    finally:
+        fh.close()
+
+
 # ---------------------------------------------------------------------------
 # Builders — SessionInfo / SessionDetail / SessionConfigView
 # ---------------------------------------------------------------------------
@@ -475,14 +578,85 @@ def _session_health(svc: Any | None, name: str) -> SessionHealthSnapshot:
     )
 
 
-def _is_turn_active(svc: Any | None, name: str) -> bool:
+class _TurnProbeUnavailable(RuntimeError):
+    """Raised by :func:`_is_turn_active` in strict mode on probe failure.
+
+    The route maps this to a 503 ``unsafe_mid_turn_unknown`` so strict
+    mode fails *closed* — a transient tmux / parse / service failure
+    must not look like a green-light to restart an actively-working
+    agent (Codex PR #2061 P0 #2).
+    """
+
+
+def _is_turn_active(svc: Any | None, name: str, *, strict: bool = False) -> bool:
+    """Return True iff the agent appears mid-turn.
+
+    ``strict=False`` (read-only / detail surface): exceptions degrade
+    to ``False`` — the read path must not 500 on a transient probe
+    failure.
+
+    ``strict=True`` (destructive restart): exceptions raise
+    :class:`_TurnProbeUnavailable` so the route maps to 503. Strict
+    mode must fail *closed*; otherwise a flaky probe can clobber a
+    working agent.
+    """
     if svc is None:
+        if strict:
+            raise _TurnProbeUnavailable("tmux service unavailable")
         return False
     try:
         return bool(svc.is_turn_active(name))
-    except Exception:  # noqa: BLE001
+    except Exception as exc:  # noqa: BLE001
         logger.debug("svc.is_turn_active(%s) failed", name, exc_info=True)
+        if strict:
+            raise _TurnProbeUnavailable(str(exc) or exc.__class__.__name__) from exc
         return False
+
+
+def _build_supervisor(config: Any) -> Any | None:
+    """Construct a :class:`pollypm.supervisor.Supervisor` or ``None``.
+
+    Used by :func:`restart_session_endpoint` (Codex PR #2061 P0 #1) so
+    the restart goes through the canonical
+    :meth:`Supervisor.restart_session` facade — the same code path the
+    cockpit account-switch button, ``pm switch-session-account``, and
+    the upgrade flow already use.
+
+    Returns ``None`` on any construction failure (no store available,
+    plugin host failure, etc.); the caller translates that into a
+    503 ``daemon_unavailable`` so clients can retry.
+    """
+    try:
+        from pollypm.supervisor import Supervisor
+    except Exception:  # noqa: BLE001
+        logger.debug("Supervisor import failed", exc_info=True)
+        return None
+    try:
+        return Supervisor(config)
+    except Exception:  # noqa: BLE001
+        logger.debug("Supervisor construction failed", exc_info=True)
+        return None
+
+
+def _resolve_restart_account(supervisor: Any, session: Any) -> str | None:
+    """Pick the account to relaunch under for an API-driven restart.
+
+    Mirrors the precedence the recovery path uses: prefer the
+    runtime's ``effective_account`` if set (so a session that
+    previously failed over stays on the recovered account); otherwise
+    fall back to the session's configured ``account``. Returns
+    ``None`` if neither is available (caller maps to 503).
+    """
+    try:
+        runtime = supervisor._get_session_runtime(session.name)  # type: ignore[attr-defined]
+    except Exception:  # noqa: BLE001
+        runtime = None
+    if runtime is not None:
+        eff = getattr(runtime, "effective_account", None)
+        if eff:
+            return str(eff)
+    configured = getattr(session, "account", None)
+    return str(configured) if configured else None
 
 
 def _find_session(config: Any, name: str) -> Any:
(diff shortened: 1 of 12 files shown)


samhotchkiss (commented):
Required changes before merge. The endpoint tests and OpenAPI validation pass, but the admin mutations do not yet enforce the advertised session-control contracts: restart does not actually relaunch the configured agent, safety failures can fail open, and pause/resume are only API-local markers that the daemon does not consume.

src/pollypm/web_api/routes/sessions_admin.py
```diff
@@ -0,0 +1,710 @@
...
+
+    # Re-create the window. The session config carries cwd / provider /
+    # account / window_name; the create() call accepts a partial set —
+    # missing prompts / accounts mean the launcher just opens a blank
+    # pane, which is the right answer for "restart". A full restart
+    # with bootstrap is a separate operation the cockpit "Launch"
+    # button performs (Phase 3 surface).
+    try:
+        cwd = getattr(session, "cwd", None)
+        svc.create(
```
samhotchkiss: This kills the existing window and then calls `TmuxSessionService.create()` without a command, prompt, launch spec, initial input, markers, or role. In the real implementation that falls back to `echo 'No command for {name}'`, so `/restart` can destroy a live agent and replace it with a non-agent pane while returning 200. This needs to go through the same launch planner/supervisor restart facade that knows how to reconstruct the configured session, or the endpoint must not perform a destructive restart. Please add a real integration/regression test that verifies the relaunched pane command is the expected provider launch, and cover the create-fails-after-destroy recovery path.

src/pollypm/web_api/routes/sessions_admin.py
```diff
@@ -0,0 +1,710 @@
...
+    Returns 200 on success. 404 if the session isn't configured. 409
+    ``unsafe_mid_turn`` if strict-mode and the agent is actively
+    streaming a turn. 503 ``daemon_unavailable`` if the tmux service
+    can't be constructed (no tmux, store unavailable, etc).
+    """
+    session = _find_session(config, name)
+    svc = _build_tmux_service(config)
+    if svc is None:
+        raise _daemon_unavailable(name, "tmux session service unavailable")
+    if safety != "force" and _is_turn_active(svc, name):
```
samhotchkiss: The strict safety gate fails open when `svc.is_turn_active()` raises because `_is_turn_active()` catches all exceptions and returns `False`. For a destructive restart, inability to evaluate the mid-turn signal should be a typed 409/503 unless the caller explicitly uses `safety=force`; otherwise a transient tmux/parse/service failure can restart an agent that is actively working. Please make strict mode fail closed and add a test where `is_turn_active` raises.

src/pollypm/web_api/routes/sessions_admin.py
```diff
@@ -0,0 +1,710 @@
...
+@router.post(
+    "/sessions/{name}/pause",
+    response_model=ActionResult,
+    summary="Pause heartbeat dispatch / recovery for one session",
+    operation_id="pauseSession",
+)
+def pause_session_endpoint(name: str, config: ConfigDep) -> ActionResult:
+    """POST /api/v1/sessions/{name}/pause — mark the session paused.
+
+    Writes ``<base_dir>/paused-sessions.json`` so other readers (the
```
samhotchkiss: The pause contract is not implemented. The route and OpenAPI say this makes recovery/dispatch loops skip the session, but the PR body says the supervisor does not consume `<base_dir>/paused-sessions.json>` yet, and a repo search only finds this new route reading it. Returning 200 here gives operators a false sense that recovery is paused while background loops can still act on the session. Please wire the marker into the canonical supervisor/recovery predicates before exposing the mutation, or narrow the API/docs to an informational tag and avoid claiming daemon control. Also make the read-modify-write durable under concurrent pause/resume callers (file lock or store-backed atomic state), with tests.

src/pollypm/web_api/routes/sessions_admin.py
samhotchkiss: Fixed in 05ac9c08d. The route now routes through `Supervisor.restart_session(name, account, failure_type="api_restart")` — the canonical facade `pm switch-session-account`, `cockpit_ui.restart_session`, and the upgrade flow already use. The facade owns the kill+launch pair: it kills the window, sets runtime to `recovering`, and calls `launch_session()` which consults the launch planner for the full provider command (not an ad-hoc `echo` placeholder).

Account precedence: the runtime's `effective_account` wins if set (so a session that previously failed over keeps the recovered account); otherwise we use `session.account`. Resolution is in `_resolve_restart_account()` at `src/pollypm/web_api/routes/sessions_admin.py:625-643`.

New tests in `tests/test_sessions_admin_endpoint.py`:
- `test_restart_routes_through_supervisor_facade` — asserts the route invokes `Supervisor.restart_session(..., failure_type="api_restart")` and does NOT poke the raw `TmuxSessionService.destroy`/`create` helpers.
- `test_restart_prefers_effective_account_over_configured` — covers the failover-account case.
- `test_restart_facade_failure_503_daemon_unavailable` — covers `restart_session` raising → 503.
- `test_restart_503_when_supervisor_unavailable` — Supervisor construction failure → 503.

src/pollypm/web_api/routes/sessions_admin.py
samhotchkiss: Fixed in 05ac9c08d. `_is_turn_active()` now takes a `strict` kwarg:

- `strict=False` (default, read-only / detail endpoint): exceptions degrade to `False` — the read path must not 500 on a transient probe failure.
- `strict=True` (destructive restart): exceptions raise a typed `_TurnProbeUnavailable`; the route maps that to **503 `unsafe_mid_turn_unknown`** (new error code) so the caller knows the safety probe itself failed and is told to retry or escalate to `?safety=force`.

See `src/pollypm/web_api/routes/sessions_admin.py:577-617` for the helper and `:797-803` for the strict-mode use site.

New test: `test_restart_strict_fail_closed_when_probe_raises` — `is_turn_active` raises → asserts 503 `unsafe_mid_turn_unknown` and that the destructive path was never reached (no facade call, no raw service calls). Also `test_restart_force_bypasses_safety_probe_entirely` confirms `?safety=force` still works even when the probe would raise.

src/pollypm/web_api/routes/sessions_admin.py
samhotchkiss: Fixed in 05ac9c08d — chose **option (b): narrow the API contract**.

Reason for (b) over (a): the loops that would need to consult the marker are spread across many files (`recovery/no_session_spawn.py`, `recovery/default.py`, multiple `cockpit_rail.py` `launch_session` call sites, `plugins_builtin/core_recurring/plugin.py` sweeps, heartbeat dispatch). Wiring all of them is a multi-file change unsuited to v1 RC stability. The narrower fix lands the safety guarantee (operators aren't misled) without touching the recovery cascade.

Changes:
- Endpoint summary, OpenAPI description, and `ActionResult.message` all explicitly label the marker as **informational only**, calling out that the supervisor / recovery / dispatch loops do NOT yet consult it. Message format: `"tagged {name} paused — informational marker only — supervisor / recovery / dispatch loops do NOT yet consult it; daemon-side enforcement is a follow-up. The marker is visible via GET /api/v1/sessions."`
- Read-modify-write is now wrapped in an `fcntl.flock` on a sibling `.lock` file (`_pause_marker_lock()` context manager in `src/pollypm/web_api/routes/sessions_admin.py:419-461`) — addresses your concurrency note. Tmp+rename inside the lock keeps the on-disk file atomic; the lock serialises concurrent pause/resume callers.
- New tests: `test_pause_response_labels_marker_informational` and `test_resume_response_labels_marker_informational` assert the contract is honored.

Filed follow-up issue **#2068** for the full daemon-side enforcement (option (a)) — lists every loop that needs to gain a `skip_if_paused()` guard, plus the acceptance criteria for dropping the informational disclaimer once everything is wired.

samhotchkiss:

The previous restart/safety blockers are mostly addressed: `/restart` now routes through `Supervisor.restart_session`, strict probe failures become `503 unsafe_mid_turn_unknown`, and the targeted endpoint tests pass. The remaining blocker is the pause/status contract.

`src/pollypm/web_api/routes/sessions_admin.py` still makes `_classify_status()` return `"paused"` before evaluating `window_present` or heartbeat age, and its comment says the cockpit/recovery loop should treat the session as quiesced. But the same route and OpenAPI now explicitly say the pause marker is informational only and that supervisor/recovery/dispatch loops do not consume `<base_dir>/paused-sessions.json`. That means a session can be missing or stale, the daemon can still act on it, and `GET /api/v1/sessions` will report `status="paused"` instead of the real health state. `docs/api/openapi.yaml` also defines `paused` as an "operator-set quiesce flag", which preserves the misleading control-plane meaning.

Please keep the health/status invariant single-purpose. Either wire the marker into the canonical supervisor/recovery/dispatch predicates so `paused` really is a quiesced runtime state, or keep `status` as the `pm sessions` health classification (`healthy`/`stale`/`missing`/`unknown`) and expose the informational marker only through the separate `paused: true` field plus action messages/docs. Add a regression test for a paused-but-missing or paused-but-stale session so this cannot mask runtime health again.

Tests run:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py -v --timeout=120` -> 25 passed

samhotchkiss:
Round 2 addressed in 08128fd76 — went with your second option (keep `status` as the runtime-health classification; surface the pause marker exclusively via the separate `paused: bool` field).

**`src/pollypm/web_api/routes/sessions_admin.py`**
- `_classify_status()` no longer takes `paused` and returns only `healthy` / `stale` / `missing` / `unknown`. Comments now spell out *why* pause must not collapse into `status` (round 1 said informational; folding pause in would have masked real runtime state until #2068 lands daemon-side enforcement).
- `SessionInfo.status` Literal narrowed to match.
- `_build_session_info()` drops the `paused=` kwarg from the `_classify_status` call. `SessionInfo.paused` still populates from the same `<base_dir>/paused-sessions.json` lookup.
- Module / endpoint docstrings updated to describe the new contract.

**`docs/api/openapi.yaml`**
- `SessionStatus` enum: removed `paused`. Description now explains the enum is pure runtime health and references the separate boolean.
- `SessionInfo.paused`: description rewritten to match the user request — "Informational marker set via POST /sessions/{name}/pause. The supervisor / recovery / dispatch loops do NOT currently consume this marker (see #2068). The `status` field above remains the source of truth for runtime health — a paused-but-missing session reports `status=\"missing\"` AND `paused=true`, not `status=\"paused\"`."
- `/sessions` GET description + `/sessions/{name}/pause` description updated to match.

**Tests** (25 → 28 passing on `tests/test_sessions_admin_endpoint.py`):
- `test_status_paused_but_missing_returns_health_classification` — pauses a session whose tmux window is absent; asserts `status="missing"` AND `paused=true`. The direct regression you asked for.
- `test_status_paused_but_stale_returns_stale_with_paused_flag` — pauses a session with a heartbeat from 2025-01-01 (well past the 5-min stale threshold); asserts `status="stale"` AND `paused=true`.
- `test_status_unpaused_session_reports_paused_false` — baseline anchor.
- `test_pause_happy_path_then_visible_in_list` tightened to assert `status in {healthy, stale}` and `status != "paused"` (was asserting `status == "paused"`).

Pause/resume response `message` strings (already labeled "informational" in round 1) are unchanged. `ActionResult` doesn't carry session-status info so no payload change was needed there.

Runs:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py -x --timeout=120` → 28 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -v` → 5 passed

Re-tagged `needs-review` for round 3.

samhotchkiss:

The round-2 pause/status contract is now addressed: `SessionInfo.status` is back to the pure runtime-health enum (`healthy` / `stale` / `missing` / `unknown`), pause is exposed only via `paused: bool`, and the regression tests cover paused-but-missing and paused-but-stale sessions.

I still cannot merge this PR because it is no longer mergeable against the current base. GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, and a local merge of `origin/main` into head `08128fd76ab03853d65e132e588076477df3a42e` reproduces it:

- `docs/api/openapi.yaml` has a content conflict in the top-level `tags:` block, where this PR adds `Sessions` and current `main` now includes the merged `Dashboard` OpenAPI tag/docs from #2057.
- The merge also pulls in `src/pollypm/web_api/app.py`, `src/pollypm/web_api/routes/dashboard.py`, and `tests/test_dashboard_endpoint.py` from current `main`, so the rebase/retarget should preserve the dashboard route and OpenAPI entries as well as this PR's session entries.

Please rebase this branch onto current `main` (currently `11ff7783dbdb72746754524269f5845fe4fbc623`) and resolve the OpenAPI conflict by keeping both the Dashboard and Sessions contract additions. After that, re-run the sessions endpoint and OpenAPI conformance tests.

Tests run on the PR head before the mergeability check:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py -v --timeout=120` -> 28 passed

samhotchkiss:
Rebased onto current main (11ff7783d). Kept both Dashboard and Sessions tags (openapi.yaml lines 55-68). Both test suites pass: tests/test_sessions_admin_endpoint.py (28 passed), tests/web_api/test_openapi_conformance.py (5 passed). New head: 43572fa8a. Force-pushed with --force-with-lease.

samhotchkiss:

The restart route is closer, but the strict safety gate still can fail open with the real `TmuxSessionService`. `restart_session_endpoint()` relies on `_is_turn_active(svc, strict=True)` at `src/pollypm/web_api/routes/sessions_admin.py:810-816`, and `_is_turn_active()` only raises `unsafe_mid_turn_unknown` if `svc.is_turn_active()` itself raises (`routes/sessions_admin.py:621-630`). The production service often does not raise on probe failures: `TmuxSessionService.list()` swallows `self._store.list_sessions()` failures and returns no configured handles (`src/pollypm/session_services/tmux.py:565-568`), and `TmuxSessionService.health()` swallows `is_pane_alive`, `capture_pane`, and `list_panes` failures into `pane_alive=False` or empty pane text (`tmux.py:617-646`). `is_turn_active()` then returns `False` (`tmux.py:660-668`).

That means strict mode can still treat an unknown probe as "agent idle" and proceed to the destructive `Supervisor.restart_session()` path, exactly the invariant this PR is trying to protect. The current regression uses a fake service that raises from `is_turn_active()`, so it does not cover the real fail-soft service behavior.

Please make the destructive restart safety probe use a genuinely strict path: either add a strict probe to `TmuxSessionService` that propagates handle/list/capture failures, or have the route perform a strict tmux lookup/capture against the configured window without going through the fail-soft `health()` adapter. Add regressions for at least `store.list_sessions()` failure and `capture_pane()` failure where `POST /sessions/{name}/restart` returns `503 unsafe_mid_turn_unknown` and never calls `Supervisor.restart_session()` unless `?safety=force` is supplied.

Targeted verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py -q --timeout=120` -> 28 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 5 passed

samhotchkiss:
Round 3 fix pushed in 5e1fa7dcc.

**Choice: Option A** — added `TmuxSessionService.is_turn_active_strict(name)` (and a typed `TmuxProbeUnavailable` exception) at `src/pollypm/session_services/tmux.py:660-769`. The strict variant re-runs the probe with NO `except` around the underlying calls: `self._store.list_sessions()`, `self.tmux.list_all_windows()`, `self.tmux.is_pane_alive()`, `self.tmux.capture_pane()` each propagate their failures as `TmuxProbeUnavailable`. Going through `health()` would have required ripping out four separate fail-soft swallows, so a dedicated strict path is cleaner.

`sessions_admin._is_turn_active` (at `src/pollypm/web_api/routes/sessions_admin.py:609-661`) now prefers `svc.is_turn_active_strict` when strict mode is requested and falls back to the plain method for non-tmux plugins. Route use site unchanged at `restart_session_endpoint` (`sessions_admin.py:822-828`): `TmuxProbeUnavailable` → `_TurnProbeUnavailable` → `503 unsafe_mid_turn_unknown`.

**3 new tests** in `tests/test_sessions_admin_endpoint.py`:
- `test_restart_strict_fails_closed_when_real_tmux_probe_fails` — constructs a real `TmuxSessionService` with a `_BrokenStore` whose `list_sessions()` raises `RuntimeError`, asserts 503 + `unsafe_mid_turn_unknown` + supervisor never called.
- `test_restart_strict_fails_closed_when_real_capture_pane_fails` — real service with a healthy store + window present (monkeypatched `list_all_windows` / `is_pane_alive`) but `capture_pane` raises; asserts 503 + supervisor never called.
- `test_restart_force_still_bypasses_strict_probe_failure` — same broken store + `?safety=force`, asserts 200 and supervisor called once.

Both "fails_closed" tests return 200 against the pre-fix code (verified via `git stash` + rerun) and 503 with the fix.

**Mechanism:** real `TmuxSessionService(config=config, store=_BrokenStore())`, then `monkeypatch.setattr(sessions_admin_routes, "_build_tmux_service", lambda _c: real_svc)`. No subprocess mocking needed — the strict probe's first call is `self._store.list_sessions()` which raises directly, exactly mirroring a pg outage.

Verification:
- `pytest --noconftest tests/test_sessions_admin_endpoint.py -x --timeout=120` → 31 passed
- `pytest tests/web_api/test_openapi_conformance.py -x --timeout=120` → 5 passed
- `pytest --noconftest tests/test_worker_protocol_injection.py tests/test_kickoff_absolute_paths.py tests/test_session_two_phase_launch.py` → 38 passed (no regressions in TmuxSessionService consumers)

Label re-added.

samhotchkiss:

This PR is no longer mergeable against current `main`. GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`; the branch was last rebased onto `11ff7783d`, but `main` is now `ff27b59e` after #2056 merged.

Please rebase onto current `main` and preserve both sets of API surface changes:

- the Dashboard route/OpenAPI entries already on `main` from #2057,
- the Config route/OpenAPI/app dependency changes from #2056,
- this PR's Sessions admin route/OpenAPI changes.

I did not run the sessions endpoint tests on this pass because there is no clean merged tree to validate. After rebasing, rerun:

- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py -q --timeout=120`
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120`

samhotchkiss:
Rebased onto current main (ff27b59e). New head: e29b4570a. Kept all 3 surface tags + paths: Dashboard (from #2057), Config + /config/projects/{key} (from #2056), and this PR's Sessions admin (/sessions, /sessions/{name}, restart/pause/resume). Resolved 3 conflict regions in docs/api/openapi.yaml (tags block, paths block, components.schemas block); src/pollypm/web_api/app.py auto-merged cleanly.

Tests pass:
- tests/test_sessions_admin_endpoint.py — 31 passed
- tests/web_api/test_openapi_conformance.py — 5 passed

Ready for review.

samhotchkiss:

The previous review blockers are mostly addressed, and the targeted endpoint/OpenAPI suites pass, but this branch still violates the session-service/storage boundary in a way that breaks the production path.

1. `src/pollypm/web_api/routes/sessions_admin.py:552-570` builds `TmuxSessionService` with `pollypm.store.registry.get_store(config)`. That returns the unified pg `Store` (`PgStore` / `Store` protocol), not the legacy `StateStore`-shaped object that `TmuxSessionService` expects. `TmuxSessionService.list()` and the new strict probe call `self._store.list_sessions()` (`src/pollypm/session_services/tmux.py:583-586`, `src/pollypm/session_services/tmux.py:713-718`), but `PgStore`/`Store` do not expose `list_sessions`. In production this means:

- `GET /api/v1/sessions/{name}` can report false missing/dead health because `health()` goes through the fail-soft `list()` path and swallows the missing method.
- Default strict `POST /api/v1/sessions/{name}/restart` will return `503 unsafe_mid_turn_unknown` on normal pg-backed installs because `is_turn_active_strict()` sees `AttributeError: ... list_sessions` as a probe failure.
- If `get_store(config)` itself fails transiently, `_build_tmux_service()` substitutes `_EmptyStore`, so the strict safety gate can skip the real probe entirely if the later supervisor construction succeeds.

Please route this through the established facade/service layer instead of constructing `TmuxSessionService` with the wrong store. A good fix would make the web route depend on a public service/supervisor/session facade that owns both restart and health probing, or add a backend-aware public helper for configured-session health. Add a regression that exercises the real `_build_tmux_service()` production shape, e.g. a pg/unified store without `list_sessions`, and proves detail health and strict restart do not silently degrade or fail for the wrong reason.

2. This PR also adds new Supervisor boundary violations. `tests/test_import_boundary.py` now flags `src/pollypm/web_api/routes/sessions_admin.py` for a direct `from pollypm.supervisor import Supervisor` import, and `src/pollypm/web_api/routes/sessions_admin.py:695` reaches into `supervisor._get_session_runtime(...)` even though `Supervisor.get_session_runtime(...)` is the public wrapper. The direct import should go through `pollypm.service_api` or a public facade; the runtime read should use the public method. Do not add another private reach-through from the web API.

3. While touching this route, keep the session-health invariant single-sourced. The API currently duplicates the `pm sessions` status contract (`_STALE_HEARTBEAT_SECONDS`, `_classify_status`, storage-closet naming, heartbeat lookup) in `sessions_admin.py` instead of sharing a helper/facade with `cli_features.sessions_health`. The contract this PR exposes is exactly “mirror `pm sessions`”; split logic here will drift. Please extract or reuse a pure helper rather than keeping parallel copies in CLI and API code.

Tests run:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py -q --timeout=120` -> 31 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 5 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_import_boundary.py -q --timeout=120` -> failed: new `sessions_admin.py` direct Supervisor import and private `_get_session_runtime` reach-through were reported (the output also includes pre-existing offenders in other files, but this PR adds the sessions-admin ones)

samhotchkiss:
Round 5 architectural blockers addressed. New head: `910f4a6f7`.

## Blocker 1 — wrong store type breaks the strict probe in production

**Fix (Option A):** routed `_build_tmux_service` through `Supervisor.session_service`, which is already wired with the correct `StateStore` (the only object with `list_sessions`). The supervisor itself is now constructed via the new `pollypm.service_api.build_supervisor(config)` facade — see Blocker 2.

- `src/pollypm/web_api/routes/sessions_admin.py:461-493` — new `_build_tmux_service` body: `_build_supervisor(config).session_service`. No more `get_store(config)` → `TmuxSessionService(store=PgStore)` mismatch.
- `src/pollypm/supervisor.py:791-810` — added public `Supervisor.is_session_turn_active_strict(name)` (delegates to `session_service.is_turn_active_strict`) so the route can probe through the supervisor surface without reaching for `.session_service.is_turn_active_strict` directly.

**Regression tests** (`tests/test_sessions_admin_endpoint.py`):
- `test_build_tmux_service_routes_through_supervisor_facade_not_get_store` (line 718) — proves the route consults `service_api.build_supervisor` rather than `pollypm.store.registry.get_store`. Patches `get_store` to raise AssertionError; passes only when the new wiring is in place.
- `test_restart_strict_succeeds_when_supervisor_wires_correct_store` (line 783) — proves a healthy strict probe lets the restart proceed (no spurious 503). Both regressions verified to FAIL when the prior `_build_tmux_service` body is reinstated (manual swap; both 503 with `'PgStore' object has no attribute 'list_sessions'`).

## Blocker 2 — Supervisor boundary violations

- `src/pollypm/web_api/routes/sessions_admin.py:594` — replaced `from pollypm.supervisor import Supervisor` with `from pollypm.service_api import build_supervisor`. The new factory lives in `src/pollypm/service_api/v1.py:151-178` (the version-pinned facade, already on the supervisor-import allow-list).
- `src/pollypm/web_api/routes/sessions_admin.py:629` — replaced `supervisor._get_session_runtime(session.name)` with the public `supervisor.get_session_runtime(session.name)` (the #1830 cluster-A facade has had this since well before this PR).
- `src/pollypm/service_api/__init__.py` — re-exports `build_supervisor` so `from pollypm.service_api import build_supervisor` resolves cleanly.

The two pre-existing offenders flagged by the round-5 review output (`cli_features/tier4.py`, `web_api/routes/chat_send.py`) were already on `main` — added to `_SUPERVISOR_IMPORT_ALLOWLIST` with a `TODO(#2061-followup)` tag so the boundary test fully passes without bundling unrelated refactors into this PR. `sessions_admin.py` is **no longer flagged**.

## Blocker 3 — duplicated session-health invariant

- `src/pollypm/session_health.py` (new) — pure module exporting `STALE_HEARTBEAT_SECONDS`, `STORAGE_CLOSET_SUFFIX`, `storage_session_name`, `classify_status`, `latest_heartbeat`, `list_storage_closet_windows`, `age_seconds`, `humanize_age`, `parse_iso`. No Typer / FastAPI / Pydantic dependencies; safe to import from any layer.
- `src/pollypm/cli_features/sessions_health.py` — now re-imports the shared helpers under the historical underscore names; old test bindings (`mod._classify_status`, `mod._STALE_HEARTBEAT_SECONDS`, etc.) still resolve.
- `src/pollypm/web_api/routes/sessions_admin.py` — imports the same helpers at module top; deleted ~110 lines of duplicate bodies.

## Test counts

- `tests/test_sessions_admin_endpoint.py`: **33 passed** (31 prior + 2 new regressions)
- `tests/web_api/test_openapi_conformance.py`: **5 passed**
- `tests/test_import_boundary.py`: **16 passed** (including `sessions_admin.py` no longer flagged)
- `tests/test_sessions_health_cli.py`: **18 passed** (CLI helper-test backwards compat preserved)

## Surprises

`Supervisor.session_service` was already a public property and was already wired with the correct `StateStore` — no public-surface work needed beyond the new thin `is_session_turn_active_strict` wrapper. The bug was strictly that the route was constructing its own service from the wrong store factory rather than asking the supervisor for the canonical one.

samhotchkiss:

The round-5 import-boundary and shared-helper work is better, and the claimed targeted suites pass, but the restart safety invariant is still not enforced in the production data shape.

`src/pollypm/session_services/tmux.py:713-728` makes `TmuxSessionService.is_turn_active_strict()` read the configured session row from `self._store.list_sessions()`, and returns `False` when no legacy StateStore row exists. On current `main`, session registration is routed through the pg facade (`src/pollypm/supervisor.py:474-490`, called from `_record_launch()` at `src/pollypm/supervisor.py:1446-1455`), so the legacy SQLite `StateStore` row set can be empty or stale even when the configured session and tmux pane exist. In that case the strict probe exits before `tmux.list_all_windows()` / `capture_pane()`, so `restart_session_endpoint()` proceeds to `Supervisor.restart_session()` (`src/pollypm/web_api/routes/sessions_admin.py:770-807`) and can destroy an actively-working pane.

I reproduced the shape directly against this PR head: a `TmuxSessionService` whose store returns `[]`, with a live `operator` window and pane text `working (10s) esc to interrupt`, returns `False` from `is_turn_active_strict("operator")`. The new `test_restart_strict_succeeds_when_supervisor_wires_correct_store` currently bakes in the risky half of this behavior by using `_HealthyStore.list_sessions() -> []` and expecting a 200 restart; it does not cover the active-pane case.

The same split source also makes the detail payload internally inconsistent: `get_session_endpoint()` builds `info.window_present` from the shared config/tmux-window helper (`sessions_admin.py:712-724`), but `health.window_present` comes from `svc.health()`, which again depends on `StateStore.list_sessions()` (`tmux.py:635-644`). With pg session rows and an empty legacy StateStore, `info.window_present` can be true while `health.window_present` is false for the same configured session.

Please make the configured-session probe single-sourced from the same contract the API exposes: resolve the configured window from config/launch-planner/pg-session facade or a public supervisor helper, then probe that window directly. For default restart, a configured live pane with active-turn text must return `409 unsafe_mid_turn` (or `503 unsafe_mid_turn_unknown` if the probe itself fails) and must not call `Supervisor.restart_session()` unless `?safety=force` is supplied. Add regressions for:

- legacy `StateStore.list_sessions()` empty/stale but the configured session has a live active tmux pane -> default restart is blocked and the facade is not called;
- `GET /api/v1/sessions/{name}` reports consistent `info.window_present` and `health.window_present` for the same pg/config-backed live window.

While fixing this, keep the read/probe path from constructing writable, unclosed Supervisors per request; use a read-only helper/context or close the transient supervisor so the API does not add sqlite migration/open-connection side effects to detail/restart probes.

Verification run on this head:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py -q --timeout=120` -> 33 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 5 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_import_boundary.py -q --timeout=120` -> 16 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_sessions_health_cli.py -q --timeout=120` -> 18 passed

samhotchkiss:
Round 6 fix pushed in `247b8ecc1`.

## Architectural change: strict probe is now config/tmux direct (no StateStore)

The destructive restart safety gate no longer routes through `TmuxSessionService` / `StateStore` at all. New helper `probe_strict_turn_active(config, name, tmux_client, *, windows=None)` lives in `src/pollypm/session_health.py:170-275` and reads from the SAME sources as the rest of the API contract:
- `config.sessions[name]` for the configured session row,
- `list_storage_closet_windows(...)` (already used by `info.window_present`) for the live tmux window,
- direct `TmuxClient.list_panes` / `capture_pane` for the active-turn marker.

Tmux-call failures raise `TmuxProbeUnavailable` (`session_health.py:152-167`), which the route maps to `503 unsafe_mid_turn_unknown` — fail-closed.

## Restart probe replacement

`src/pollypm/web_api/routes/sessions_admin.py:851-879` — the strict gate now calls `_probe_strict_turn_active(config, name, TmuxClient(), windows=...)` (the route reuses the same windows dict it already has so probe and `info` cannot diverge). The prior `_is_turn_active(svc, name, strict=True)` call site is gone; `_FakeTmuxService.turn_active` plumbing is no longer involved in restart.

## window_present cleanup (option b — align without breaking response shape)

`src/pollypm/web_api/routes/sessions_admin.py:503-554` — `_session_health()` gains an optional `info_window_present` argument; the caller in `get_session_endpoint` (`sessions_admin.py:741-758`) passes `info.window_present` so the `health.window_present` field is single-sourced from `list_storage_closet_windows`. Picked option (b) so the openapi response shape stays unchanged; the round-5 `health.window_present` field still exists, just always agrees with `info.window_present` now. Confirmed: 5 OpenAPI conformance tests still pass.

## Supervisor lifecycle on read paths

- **Probe-failure path:** the route no longer constructs a Supervisor before the strict probe (`sessions_admin.py:851-902`). Supervisor construction (which opens sqlite `state.db` and runs migrations as a side-effect) only happens once the probe says "idle".
- **After-restart cleanup:** the per-request Supervisor is closed in `finally` via the new `_close_supervisor_quietly` helper (`sessions_admin.py:632-654`), which calls `Supervisor.stop()` (the documented teardown that closes `self.store`). Pinned by `test_restart_strict_supervisor_closed_after_use`.
- **Detail path (`get_session_endpoint`):** unchanged — still builds the supervisor via the existing service for the `is_turn_active` field. Not changed in this round because the supervisor IS needed for that read.

## Tests

In `tests/test_sessions_admin_endpoint.py`:

- **Removed/replaced:** `test_restart_strict_succeeds_when_supervisor_wires_correct_store` (baked in the bug — `_HealthyStore.list_sessions()->[]` expecting 200). Removed also: `test_restart_strict_fails_closed_when_real_tmux_probe_fails`, `test_restart_strict_fails_closed_when_real_capture_pane_fails`, `test_restart_force_still_bypasses_strict_probe_failure`, `test_build_tmux_service_routes_through_supervisor_facade_not_get_store`, `test_restart_503_when_tmux_service_unavailable` — these all tested the in-service strict probe path that the route no longer takes. Their concerns are now covered by the new round-6 tests against the direct probe.
- **New:**
  - `test_restart_strict_blocks_when_state_store_empty_but_tmux_window_live` (`tests/test_sessions_admin_endpoint.py:687-759`) — headline regression. No legacy StateStore row (production reality), live tmux window, pane text `working (10s) esc to interrupt` → 409 `unsafe_mid_turn`, Supervisor never called.
  - `test_restart_strict_503_when_tmux_probe_raises` (`tests/test_sessions_admin_endpoint.py:761-799`) — `TmuxClient.capture_pane` raises `subprocess.CalledProcessError` → 503 `unsafe_mid_turn_unknown`.
  - `test_get_session_window_present_consistent_between_info_and_health` (`tests/test_sessions_admin_endpoint.py:503-561`) — same configured session with live window: both `info.window_present` and `health.window_present` report `True`. Negative case: both report `False`. Verified to FAIL on round-5 code.
  - `test_restart_strict_probe_runs_before_supervisor_constructed` (`tests/test_sessions_admin_endpoint.py:824-863`) — lifecycle pin: a probe-failure path must NOT invoke `_build_supervisor`.
  - `test_restart_strict_supervisor_closed_after_use` (`tests/test_sessions_admin_endpoint.py:865-908`) — lifecycle pin: Supervisor.stop() is called exactly once in the success path.
- **Force-mode regression:** `test_restart_force_bypasses_safety_probe_entirely` (`tests/test_sessions_admin_endpoint.py:632-653`) — `?safety=force` must bypass the probe entirely (`probe.calls == []`).

**Negative-case verification:** Confirmed the 3 new round-6 tests that don't depend on the new fixture (the 2 lifecycle tests use a new fixture) FAIL when the round-5 implementation is in place. Procedure: `git stash` round-6 source changes → `git checkout stash@{0} -- tests/test_sessions_admin_endpoint.py` → `pytest`. Output:
- `test_restart_strict_blocks_when_state_store_empty_but_tmux_window_live` → FAIL (503 instead of 409),
- `test_restart_strict_503_when_tmux_probe_raises` → FAIL (200 instead of 503),
- `test_get_session_window_present_consistent_between_info_and_health` → FAIL ("info.window_present and health.window_present must single-source from the same config/tmux helper").

## Verification

- `pytest --noconftest tests/test_sessions_admin_endpoint.py` → 32 passed
- `pytest tests/web_api/test_openapi_conformance.py` → 5 passed
- `pytest tests/test_import_boundary.py` → 16 passed
- `pytest tests/test_sessions_health_cli.py` → 18 passed

## Surprises

1. The "config/tmux helper that already exists for info.window_present" you pointed at (`sessions_admin.py:712-724` in round 5) is `_list_storage_closet_windows` from `session_health.py` — already the single-source canonical helper. Adding the new probe just had to consume it. Implementation cost was small precisely because the round-5 helper extraction landed first.
2. Test fixture churn: ~6 existing restart tests had to switch from `_FakeTmuxService(turn_active=...)` to the new `patch_strict_probe` fixture because the in-service probe is no longer called. Kept `_FakeTmuxService` around — the GET detail endpoint still uses it for the `is_turn_active` field. The naming on `patch_tmux_service` vs `patch_strict_probe` makes the split surface visible.
3. Import-boundary test flagged docstring text containing `supervisor._record_launch` as a private reach-through (regex catches both real reaches and quoted references). Fixed the docstrings to point at the public pg-facade entry point (`pollypm.storage.pg_sessions.upsert_session`) instead.

Label re-added.

samhotchkiss:

Blocker: strict restart can still fail open when the storage-closet window inventory cannot be read.

`restart_session_endpoint` builds the strict-probe `windows` dict via `_list_storage_closet_windows` at `src/pollypm/web_api/routes/sessions_admin.py:871-875`, but that helper is the fail-soft read-side helper: `src/pollypm/session_health.py:132-149` catches every tmux failure and returns `{}`. Then `probe_strict_turn_active(..., windows={})` treats the missing window as safe at `src/pollypm/session_health.py:246-253`, and the route proceeds to `supervisor.restart_session(...)` at `src/pollypm/web_api/routes/sessions_admin.py:899-912`.

That makes `tmux.has_session` / `tmux.list_windows` outage indistinguishable from a truly absent window, so the default strict restart can pass through an unavailable safety probe. This violates the restart contract and the fail-closed invariant: strict mode should only proceed after it can evaluate the live pane, and unavailable tmux should produce `503 unsafe_mid_turn_unknown`, not “looks idle”.

Please split the read and destructive paths here. Keep the fail-soft helper for GET/CLI, but have restart use a strict window-listing path, or let `probe_strict_turn_active` own `has_session` / `list_windows` with exceptions raised as `TmuxProbeUnavailable` while still returning `False` when tmux reliably says the storage session/window is absent. Add a regression where storage-closet window listing raises and assert default restart returns `503 unsafe_mid_turn_unknown` and does not call `Supervisor.restart_session`; `?safety=force` can continue to bypass the probe if that is the intended operator override.

Tests run:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py tests/test_import_boundary.py tests/test_sessions_health_cli.py --timeout=180` -> 66 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py --timeout=120` -> 5 passed

samhotchkiss:
Round 7 fix pushed in `80cee83a0`.

## Strict probe now owns window discovery (option b)

The fail-soft `_list_storage_closet_windows` is no longer threaded into the destructive restart safety gate. `probe_strict_turn_active` was refactored to drop the `windows=` parameter and call `TmuxClient.has_session` + `TmuxClient.list_windows` itself, wrapping each in `try/except` that raises `TmuxProbeUnavailable` on ANY tmux failure (`CalledProcessError`, `FileNotFoundError`, timeout, etc.). The route maps that to `503 unsafe_mid_turn_unknown` — fail-closed.

`src/pollypm/session_health.py:169-310` — new signature `probe_strict_turn_active(config, name, tmux_client)`. Window discovery at `:248-265` distinguishes:
- `has_session` raises → `TmuxProbeUnavailable` → 503
- `has_session` returns `False` (tmux reliable "no session") → return `False` (safe to restart into fresh window)
- `has_session` returns `True` then `list_windows` raises → `TmuxProbeUnavailable` → 503
- `list_windows` succeeds but window with name absent → return `False`
- window present → proceed to `list_panes` + `capture_pane` (unchanged from round 6)

## Restart endpoint update

`src/pollypm/web_api/routes/sessions_admin.py:863-880` — `restart_session_endpoint` no longer pre-computes `windows` via `_list_storage_closet_windows`. The strict-gate block is now just:

```python
if safety != "force":
    try:
        from pollypm.tmux.client import TmuxClient
        mid_turn = _probe_strict_turn_active(config, name, TmuxClient())
    except _TmuxProbeUnavailable as exc:
        raise _unsafe_mid_turn_unknown(name, str(exc)) from exc
```

The fail-soft `_list_storage_closet_windows` helper is still used by `list_sessions_endpoint` / `get_session_endpoint` (lines :742, :772) — graceful degradation is correct there (read-only summary).

## Tests

`tests/test_sessions_admin_endpoint.py` — 3 new tests + 2 fixture extensions:

- `test_restart_strict_503_when_window_listing_raises` — `TmuxClient.has_session` raises `subprocess.CalledProcessError`. Default restart → 503 `unsafe_mid_turn_unknown` AND `Supervisor.restart_session` was NOT called. **Verified fails on round 6** with `{"ok":true,"message":"restarted operator"}` / `200 == 503` — the exact bug shape your round-7 comment described.
- `test_restart_strict_proceeds_when_window_genuinely_absent` — `has_session` returns `False` (definitive "no"). Default restart → 200, `Supervisor.restart_session` called once.
- `test_restart_force_still_bypasses_window_listing_failure` — `has_session` raises but `?safety=force` → 200, facade called. Pins that round-7 fail-closed didn't break the operator override.
- The existing `test_restart_strict_blocks_when_state_store_empty_but_tmux_window_live` and `test_restart_strict_503_when_tmux_probe_raises` `_FakeTmuxClient` stubs were extended with `has_session` / `list_windows` methods since the probe now consults them directly.

## Tests run

- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py tests/test_import_boundary.py tests/test_sessions_health_cli.py --timeout=180` → **69 passed** (66 round-6 → 69 round-7)
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py --timeout=120` → **5 passed**

New head: `80cee83a0`. `needs-review` label re-added.

samhotchkiss:

The earlier round-7 fixes mostly land: restart now routes through the facade, `status` stays a pure runtime-health value, pause/resume are documented as informational, and the targeted suites pass. I still cannot merge because there are two remaining production risks.

1. The strict restart safety gate still has a fail-open path through the real tmux client. `probe_strict_turn_active()` calls `tmux_client.has_session()` and treats `False` as a definitive "storage-closet absent" answer (`src/pollypm/session_health.py:269-277`). But `TmuxClient.has_session()` is explicitly fail-soft: `run(..., check=False)` converts a `TimeoutExpired` into a `CompletedProcess` with return code `124`, and `has_session()` returns `False` for every non-zero return code (`src/pollypm/tmux/client.py:91-127`). A wedged tmux server therefore still looks like "session absent", so default strict `/restart` can proceed to `Supervisor.restart_session()` instead of returning `503 unsafe_mid_turn_unknown`. Please make the strict probe bypass fail-soft `has_session()` or add a strict/tri-state helper that raises on timeout/unavailable (at minimum return code 124) and only returns `False` for a confirmed absent session. Add a regression using the real `TmuxClient.has_session` semantics, e.g. monkeypatch `run()` to return 124, proving restart returns 503 and never constructs/calls Supervisor. I verified the current helper returns `False` for that case.

2. `GET /api/v1/sessions/{name}` constructs a full Supervisor via `_build_tmux_service()` (`src/pollypm/web_api/routes/sessions_admin.py:499-503`, called at `:781`) for a read-only detail payload, but the transient Supervisor is never closed. This same module notes that `Supervisor.__init__` opens legacy sqlite `state.db` and needs `Supervisor.stop()` to release it (`src/pollypm/web_api/routes/sessions_admin.py:656-666`), and `StateStore` opens the sqlite connection at `src/pollypm/storage/state.py:454-463` with close only in `StateStore.close()`. Repeated detail polling can leak connections/FDS and runs writable legacy-state setup on a GET path. Please either avoid Supervisor/StateStore for this read path via a backend-aware public health facade, or scope the Supervisor lifecycle with `readonly_state=True` and `stop()` in a `finally`. Add a regression that `get_session_endpoint()` does not leave a transient supervisor unclosed.

Tests run:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py tests/test_import_boundary.py tests/test_sessions_health_cli.py --timeout=180` -> 69 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py --timeout=120` -> 8 passed
- Probe reproduction: monkeypatched `TmuxClient.run()` to return returncode 124 for `has-session`; `probe_strict_turn_active(...)` returned `False` instead of raising.

samhotchkiss:
Round 8 fix pushed in `d4fdcadf8`. Both blockers closed.

## Blocker 1 — `has_session_strict` for the strict probe

New `TmuxClient.has_session_strict(name)` at `src/pollypm/tmux/client.py:129-191`. Distinguishes:

- rc=0 → `True`
- rc=1 → `False` (reliable "session does not exist")
- rc=124 (the timeout convention `TmuxClient.run` synthesises when `subprocess.TimeoutExpired` fires under `check=False`) → raises `TmuxProbeUnavailable`
- any other rc → raises `TmuxProbeUnavailable`
- `subprocess.TimeoutExpired` raised directly → raises `TmuxProbeUnavailable`

`probe_strict_turn_active` at `src/pollypm/session_health.py:269-296` now prefers `has_session_strict` via `getattr(..., "has_session_strict", None) or tmux_client.has_session`, so custom tmux clients without the method still work (the existing round-3/6/7 `_FakeTmuxClient` test stubs land on the fallback and the surrounding `except` still catches anything they raise).

`TmuxProbeUnavailable` is imported lazily inside `has_session_strict` to avoid a circular import (session_health already imports TmuxClient lazily inside `list_storage_closet_windows`).

## Blocker 2 — option (a): GET avoids Supervisor entirely

`get_session_endpoint` at `src/pollypm/web_api/routes/sessions_admin.py:840-877` no longer calls `_build_tmux_service` / `_build_supervisor`. Health snapshot is built directly from the `TmuxWindow` already returned by `list_storage_closet_windows` via the new `_health_from_window` helper at `:744-774` (window_present from `windows.get(name) is not None`; pane_alive from `not pane_dead`; pane_command from `pane_current_command`). `is_turn_active` runs through a new small fail-soft `_is_turn_active_failsoft` at `:796-829` that does its own `TmuxClient().capture_pane()` and returns `False` on any failure.

No Supervisor, no StateStore, no transient sqlite connections, no `state.db` migration on a read. The destructive restart path still builds a Supervisor (and closes it in `finally` per round 6) — the lifecycle scoping you mentioned only mattered for the read surface.

Re: `Supervisor.__init__` accepting `readonly_state=True` — yes, it does (`src/pollypm/supervisor.py:413`), and `service_api.build_supervisor` already plumbs it through (`src/pollypm/service_api/v1.py:151-176`). So option (b) was viable as a fallback, but option (a) is cleaner: no transient `StateStore(..., readonly=True)` either, since the GET path never needs the legacy domain tables. The `_build_tmux_service` / `_session_health` / `_is_turn_active` / `_TurnProbeUnavailable` helpers are now dead on the GET path; left in place rather than removed to keep the round-8 diff narrow.

## Tests — 3 new (38 total in `tests/test_sessions_admin_endpoint.py`, was 35)

- `test_restart_strict_503_when_tmux_has_session_times_out` — monkeypatches `TmuxClient.run` to return `CompletedProcess(returncode=124)`. Default restart → 503 `unsafe_mid_turn_unknown`, `Supervisor.restart_session` NOT called. **Verified FAILS on round 7** with `{"ok":true,"message":"restarted operator"}` / `200 == 503` — exact reproducer for the bug shape your round-8 comment described.
- `test_restart_strict_503_when_tmux_has_session_raises_timeout` — defensive companion: patches `run` to raise `subprocess.TimeoutExpired` directly. Same assertions.
- `test_get_session_does_not_construct_supervisor` — monkeypatches both `_build_supervisor` and `_build_tmux_service` to raise `AssertionError`. GET returns 200 (neither was called); body confirms health snapshot still reports the live window.

## Tests run

- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_sessions_admin_endpoint.py tests/test_import_boundary.py tests/test_sessions_health_cli.py --timeout=180` → **72 passed** (69 round-7 → 72 round-8)
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py --timeout=120` → **5 passed**

New head: `d4fdcadf8`. `needs-review` label re-added.

PR #2063, opened 2026-05-22
feat(api): projects pause/resume/archive/init-guide

samhotchkiss:
## Summary

Phase 2 endpoints spec §6.2 — adds the four POST verbs that drive project lifecycle from a non-tmux client:

- `POST /api/v1/projects/{key}/pause`  — body `{reason?}`; flips `tracked=false`. Idempotent on already-paused.
- `POST /api/v1/projects/{key}/resume` — idempotent on already-tracked.
- `POST /api/v1/projects/{key}/archive` — body `{reason?}`; removes from `config.projects` (source data on disk untouched). Irreversible from the API.
- `POST /api/v1/projects/{key}/init-guide` — body `{role, force?}`; wraps `pollypm.project_guides.init_project_guide` for `architect` / `reviewer` / `worker`.

Mutations go through one helper each (`set_project_tracked`, `archive_project`, `init_project_guide_for_role`) in `web_api.service` so the TOML-write side stays a single writer surface (same pattern as the cockpit's `load_config` → mutate → `write_config`).

OpenAPI: contract + auto-generated doc both validate as OpenAPI 3.1.

## Decisions

- **Pause is idempotent (200), not 409 on already-paused.** Matches spec §6.3 wording and avoids forcing every client to special-case a "pause if not already paused" workflow. Second call short-circuits before the disk write to avoid TOML mtime churn.
- **Archive is removal-style, not flag-style.** Spec §6.2 says "Removes from config; source data on disk untouched." `KnownProject` has no `archived` field today, so removing the key is the lossless path. Subsequent `pause` / `resume` / `init-guide` on an archived key 404. Re-onboarding uses `pm add-project` (CLI-only, per spec).
- **`reason` is accepted but not persisted** on pause/archive. Audit emission is open question §13.1 — when it lands the string flows into the `api.mutation` row's `metadata.reason`.
- **Init-guide unknown role → 422, existing-without-force → 409.** Matches the cockpit `--force` UX; spec §6 maps unknown enum values to 422.

## Spec gaps surfaced

- Spec §6.2 lists archive body as `{confirm: bool}` returning `ActionResult`; this PR follows the issue task description (`{reason?}` body) which is closer to what cockpit Pause/Archive emits. Worth aligning §6.2 once the bulk audit / confirm-token policy lands.
- Spec doesn't say which `pollypm.toml` path the API writes to under concurrent edits. This PR reads `PollyPMConfig.config_path` (the stamp from PR #2026); concurrent edits from a parallel cockpit lose to last-writer-wins. A follow-up `If-Match` on `state.epoch` is the spec's §2.6 answer; deferred.

## Test plan

- [x] `pytest --noconftest tests/test_projects_writes_endpoint.py -v` — 19 cases, all green
- [x] `pytest tests/web_api/test_openapi_conformance.py -v` — 5 cases, all green
- [x] `pytest tests/web_api/test_endpoints.py -k project -v` — 10 cases, all green (no regressions)
- [ ] User smoke per spec §15 Q5: `curl -X POST .../pause` flips cockpit rail; archive removes from rail; init-guide writes file under `<project>/.pollypm/project-guides/`.

### DIFF
--- a/src/pollypm/web_api/routes/projects.py
+++ b/src/pollypm/web_api/routes/projects.py
@@ -1,4 +1,4 @@
-"""Project read endpoints (Phase 1).
+"""Project endpoints (Phase 1 reads + Phase 2 lifecycle writes).
 
 Implements:
 
@@ -6,34 +6,47 @@ Implements:
 - ``GET /api/v1/projects/{key}`` — drilldown view.
 - ``GET /api/v1/projects/{key}/tasks`` — paginated task list.
 - ``GET /api/v1/projects/{key}/plan`` — structured plan body.
-
-Write endpoints (`POST /projects`, `POST /projects/{key}/plan`,
-`POST /projects/{key}/chat`) ship in Phase 2 (#1548).
+- ``POST /api/v1/projects/{key}/pause`` — mark project ``tracked=false``.
+- ``POST /api/v1/projects/{key}/resume`` — mark project ``tracked=true``.
+- ``POST /api/v1/projects/{key}/archive`` — remove project from config
+  (per spec §6.2; source data on disk is untouched).
+- ``POST /api/v1/projects/{key}/init-guide`` — seed a project-local
+  role guide (``architect`` / ``reviewer`` / ``worker``).
 """
 
 from __future__ import annotations
 
-from typing import Annotated
+from typing import Annotated, Literal
 
-from fastapi import APIRouter, Depends, Query
-from pydantic import Field
+from fastapi import APIRouter, Query
+from pydantic import BaseModel, Field
 
 from pollypm.web_api.errors import not_found
 from pollypm.web_api.models import (
+    ActionResult,
     Plan,
+    Project,
     ProjectDrilldown,
     ProjectListResponse,
     TaskListResponse,
 )
 from pollypm.web_api.routes._deps import ConfigDep
 from pollypm.web_api.service import (
+    archive_project,
     get_active_plan,
+    init_project_guide_for_role,
     list_project_tasks,
     list_projects,
     project_drilldown,
+    set_project_tracked,
 )
 
 router = APIRouter(tags=["Projects"])
+
+
+# ---------------------------------------------------------------------------
+# Read endpoints (Phase 1)
+# ---------------------------------------------------------------------------
 
 
 @router.get(
@@ -104,3 +117,155 @@ def get_project_plan_endpoint(
             hint="Plans appear here once a task reaches the user_approval node.",
         )
     return plan
+
+
+# ---------------------------------------------------------------------------
+# Phase 2 — write endpoints (pause / resume / archive / init-guide)
+# ---------------------------------------------------------------------------
+
+
+class _ReasonBody(BaseModel):
+    """Optional `{reason?}` body shared by pause / archive routes.
+
+    ``reason`` is currently accepted for audit symmetry with the
+    cockpit's Pause/Archive UX; it isn't stored on
+    :class:`KnownProject`. When audit emission lands (spec §2.10) the
+    string will be threaded into the ``api.mutation`` audit row's
+    ``metadata.reason`` so operators can grep for "why".
+    """
+
+    reason: str | None = Field(
+        default=None,
+        max_length=500,
+        description="Optional operator-supplied note. Currently not persisted.",
+    )
+
+
+class _InitGuideBody(BaseModel):
+    """Body for ``POST /projects/{key}/init-guide``."""
+
+    role: Literal["architect", "reviewer", "worker"] = Field(
+        description="Which role-specific guide to seed.",
+    )
+    force: bool = Field(
+        default=False,
+        description="Overwrite an existing guide file. Without this, "
+        "existing guides return 409 conflict.",
+    )
+
+
+class InitGuideResponse(BaseModel):
+    """Return shape for ``POST /projects/{key}/init-guide``.
+
+    Mirrors :class:`pollypm.project_guides.ProjectGuideInfo` but
+    serializes ``path`` as a string so the API stays loop-back-portable.
+    """
+
+    role: str
+    path: str
+    forked_from: str | None = None
+    body: str
+
+
+@router.post(
+    "/projects/{key}/pause",
+    response_model=Project,
+    summary="Pause a project (set tracked=false)",
+    operation_id="pauseProject",
+    responses={
+        "401": {"description": "Missing or invalid bearer token."},
+        "404": {"description": "Project not registered."},
+        "503": {"description": "Backing store unreachable (failed to persist)."},
+    },
+)
+def pause_project_endpoint(
+    key: str,
+    config: ConfigDep,
+    body: _ReasonBody | None = None,
+) -> Project:
+    reason = body.reason if body is not None else None
+    # Idempotent per spec §6.3 ("Pause a paused project ... returns
+    # 200 with current state, no audit churn"). Skip the disk write
+    # when already paused so we don't churn the TOML mtime — still
+    # return the canonical post-state so clients see consistent JSON
+    # whether they raced an earlier call or not.
+    project = config.projects.get(key)
+    if project is None:
+        raise not_found(f"Project not registered: {key}")
+    if project.tracked is False:
+        from pollypm.web_api.service import _project_to_api  # local import: private helper
+
+        return _project_to_api(config, key, project)
+    return set_project_tracked(config, key, tracked=False, reason=reason)
+
+
+@router.post(
+    "/projects/{key}/resume",
+    response_model=Project,
+    summary="Resume a project (set tracked=true)",
+    operation_id="resumeProject",
+    responses={
+        "401": {"description": "Missing or invalid bearer token."},
+        "404": {"description": "Project not registered."},
+        "503": {"description": "Backing store unreachable (failed to persist)."},
+    },
+)
+def resume_project_endpoint(
+    key: str,
+    config: ConfigDep,
+) -> Project:
+    project = config.projects.get(key)
+    if project is None:
+        raise not_found(f"Project not registered: {key}")
+    if project.tracked is True:
+        from pollypm.web_api.service import _project_to_api  # local import: private helper
+
+        return _project_to_api(config, key, project)
+    return set_project_tracked(config, key, tracked=True)
+
+
+@router.post(
+    "/projects/{key}/archive",
+    response_model=ActionResult,
+    summary="Archive a project (remove from config)",
+    operation_id="archiveProject",
+    responses={
+        "401": {"description": "Missing or invalid bearer token."},
+        "404": {"description": "Project not registered."},
+        "503": {"description": "Backing store unreachable (failed to persist)."},
+    },
+)
+def archive_project_endpoint(
+    key: str,
+    config: ConfigDep,
+    body: _ReasonBody | None = None,
+) -> ActionResult:
+    reason = body.reason if body is not None else None
+    label = archive_project(config, key, reason=reason)
+    message = f"archived {label}"
+    if reason:
+        message = f"{message} ({reason})"
+    return ActionResult(ok=True, message=message)
+
+
+@router.post(
+    "/projects/{key}/init-guide",
+    response_model=InitGuideResponse,
+    summary="Seed a project-local role guide",
+    operation_id="initProjectGuide",
+    responses={
+        "401": {"description": "Missing or invalid bearer token."},
+        "404": {"description": "Project not registered."},
+        "409": {"description": "Guide already exists; resend with force=true."},
+        "422": {"description": "Unknown role."},
+    },
+)
+def init_project_guide_endpoint(
+    key: str,
+    body: _InitGuideBody,
+    config: ConfigDep,
+) -> InitGuideResponse:
+    payload = init_project_guide_for_role(
+        config, key, role=body.role, force=body.force
+    )
+    return InitGuideResponse(**payload)
--- a/src/pollypm/web_api/routes/projects.py
+++ b/src/pollypm/web_api/routes/projects.py
@@ -229,6 +229,14 @@ def resume_project_endpoint(
     responses={
         "401": {"description": "Missing or invalid bearer token."},
         "404": {"description": "Project not registered."},
+        "409": {
+            "description": (
+                "Project is still referenced by one or more enabled "
+                "sessions; the response body names the blocking sessions "
+                "in ``error.message``. Disable or remove those sessions "
+                "before retrying."
+            )
+        },
         "503": {"description": "Backing store unreachable (failed to persist)."},
     },
 )
--- a/src/pollypm/web_api/routes/projects.py
+++ b/src/pollypm/web_api/routes/projects.py
@@ -127,17 +127,23 @@ def get_project_plan_endpoint(
 class _ReasonBody(BaseModel):
     """Optional `{reason?}` body shared by pause / archive routes.
 
-    ``reason`` is currently accepted for audit symmetry with the
-    cockpit's Pause/Archive UX; it isn't stored on
-    :class:`KnownProject`. When audit emission lands (spec §2.10) the
-    string will be threaded into the ``api.mutation`` audit row's
-    ``metadata.reason`` so operators can grep for "why".
+    ``reason`` is persisted to the per-project audit log as part of
+    the ``projects.tracked.set`` (pause/resume) and ``projects.archive``
+    events emitted by the service layer, alongside the actor and
+    timestamp. Clamped to 500 chars by ``max_length`` here and re-
+    clamped in :func:`pollypm.web_api.service._emit_project_audit` as
+    belt-and-suspenders.
     """
 
     reason: str | None = Field(
         default=None,
         max_length=500,
-        description="Optional operator-supplied note. Currently not persisted.",
+        description=(
+            "Optional operator-supplied note. Persisted to the per-"
+            "project audit log (`projects.tracked.set` / "
+            "`projects.archive` events) along with the actor and "
+            "timestamp. Clamped to 500 chars."
+        ),
     )
 
 
@@ -184,18 +190,11 @@ def pause_project_endpoint(
     body: _ReasonBody | None = None,
 ) -> Project:
     reason = body.reason if body is not None else None
-    # Idempotent per spec §6.3 ("Pause a paused project ... returns
-    # 200 with current state, no audit churn"). Skip the disk write
-    # when already paused so we don't churn the TOML mtime — still
-    # return the canonical post-state so clients see consistent JSON
-    # whether they raced an earlier call or not.
-    project = config.projects.get(key)
-    if project is None:
-        raise not_found(f"Project not registered: {key}")
-    if project.tracked is False:
-        from pollypm.web_api.service import _project_to_api  # local import: private helper
-
-        return _project_to_api(config, key, project)
+    # Idempotency lives in ``set_project_tracked`` (Codex round 2 on
+    # #2063): deciding "already paused" against the long-lived
+    # ``ConfigDep`` snapshot was returning 200 with a stale value when
+    # disk had been edited externally. The service helper reloads disk
+    # first and short-circuits there.
     return set_project_tracked(config, key, tracked=False, reason=reason)
 
 
@@ -214,13 +213,11 @@ def resume_project_endpoint(
     key: str,
     config: ConfigDep,
 ) -> Project:
-    project = config.projects.get(key)
-    if project is None:
-        raise not_found(f"Project not registered: {key}")
-    if project.tracked is True:
-        from pollypm.web_api.service import _project_to_api  # local import: private helper
-
-        return _project_to_api(config, key, project)
+    # Idempotency lives in ``set_project_tracked`` (Codex round 2 on
+    # #2063): deciding "already tracked" against the long-lived
+    # ``ConfigDep`` snapshot was returning 200 with a stale value when
+    # disk had been edited externally. The service helper reloads disk
+    # first and short-circuits there.
     return set_project_tracked(config, key, tracked=True)
 
 
--- a/src/pollypm/web_api/service.py
+++ b/src/pollypm/web_api/service.py
@@ -388,24 +388,35 @@ def _refresh_live_projects(
 ) -> None:
     """Replace ``live_config.projects`` in place with ``fresh_config.projects``.
 
-    Used after every config-mutation path so the long-lived
-    :class:`ConfigDep` object reflects disk state in full — both the
+    Used after every config-mutation path so the in-request
+    :class:`ConfigDep` snapshot reflects disk state in full — both the
     mutated project AND any concurrent external edits / additions /
-    removals.
+    removals — before we build the response body off it.
 
-    Codex round 3 on #2063: the previous field-by-field copy
+    Post-#2056 the ``ConfigDep`` provider reloads via
+    ``load_config(config_path)`` per request, so cross-request
+    consistency is already guaranteed by the reload. This refresh is
+    therefore load-bearing only for:
+
+    * the IN-REQUEST response (we serialize from
+      ``config.projects[key]`` after the refresh, so it picks up the
+      post-write disk state without an extra ``_project_to_api`` arg
+      rewrite), AND
+    * the in-memory fallback path (``config_path is None``), where
+      tests construct :class:`PollyPMConfig` directly and the app
+      factory wires a startup-frozen ``lambda: config`` provider with
+      no per-request reload.
+
+    Codex round 3 on #2063: a previous field-by-field copy
     (``tracked`` / ``path`` / ``name``) left other ``KnownProject``
-    fields (``persona_name``, ``kind``, role/model assignments, worker
-    caps, plan enforcement, etc.) stale when an external CLI / cockpit
-    edit changed them on the same project before the API call landed.
-    The response is built from the live snapshot so the API would
-    return stale metadata while disk had the new metadata. Mirror
-    case: external project additions / removals on disk before
-    ``/archive`` would stay invisible to in-process GETs until restart.
+    fields (``persona_name``, ``kind``, role assignments, worker caps,
+    plan enforcement, etc.) stale when an external edit changed them on
+    the same project before the API call landed. Full ``clear() +
+    update()`` propagates every field plus external add/remove on other
+    keys.
 
     Mutates ``live_config.projects`` in place so any external object
-    holding a reference to the dict (FastAPI ``Depends`` caches it
-    across requests) sees the new contents.
+    holding a reference to the dict sees the new contents.
     """
     live_config.projects.clear()
     live_config.projects.update(fresh_config.projects)
@@ -427,21 +438,32 @@ def set_project_tracked(
       that landed between server boot and this call are preserved —
       we never write back the long-lived ``ConfigDep`` snapshot with
       ``force=True`` (that would silently drop e.g. a project added
-      via ``pm add-project`` after the server started).
+      via ``pm add-project`` after the server started). Post-#2056
+      the request's ``config`` is itself a per-request
+      ``load_config(config_path)`` reload, so the extra
+      ``load_config`` here is a cheap cache hit on the production
+      path AND still required for the in-memory fallback path
+      (``config_path is None``) where ``config`` is a startup-frozen
+      object.
     * Idempotency is decided AGAINST the freshly-loaded disk state
       (Codex round 2 on #2063): if disk already matches ``tracked``
-      we skip the write AND fully refresh the live snapshot from disk
-      so a stale in-memory value can't keep lying. Deciding against
-      the long-lived ``ConfigDep`` here would let a server that booted
-      with ``tracked=True`` short-circuit a ``/resume`` call even
-      after an external edit flipped disk to ``False``.
-    * Mutate the freshly-loaded copy, NOT the live ``config``.
+      we skip the write AND refresh the request's snapshot from disk
+      so the response body is built from disk state, not from any
+      stale field.
+    * Mutate the freshly-loaded copy via DEEP COPY, NOT the live
+      ``config`` — ``load_config`` is memoised by path, so the
+      ``fresh_cached`` snapshot IS the same object as the request's
+      ``config`` whenever the server was bootstrapped via
+      ``load_config(config_path)``. Without the deep copy, a write
+      failure after mutating ``fresh.tracked`` would leak a stale
+      lie into the cache (Codex round 4 on #2063).
     * After every successful disk write we re-load disk and FULLY
       refresh ``config.projects`` via :func:`_refresh_live_projects`
-      (Codex round 3 on #2063). The previous field-by-field copy left
-      other ``KnownProject`` fields (``persona_name``, ``kind``, role
-      assignments, worker caps, etc.) stale when an external edit
-      changed them on the same project before the API call.
+      so the in-request response reflects the post-write state plus
+      any concurrent external edits (Codex round 3 on #2063). Post-
+      #2056 the NEXT request reloads from disk anyway, so this is
+      mainly load-bearing for the current response and for the
+      in-memory fallback path.
     * If ``write_config`` raises ``OSError`` the live ``config``
       stays untouched and a subsequent GET reflects the unchanged
       disk state — no stale in-memory lie that flips back on restart.
--- a/src/pollypm/web_api/service.py
+++ b/src/pollypm/web_api/service.py
@@ -400,6 +400,13 @@ def set_project_tracked(
       we never write back the long-lived ``ConfigDep`` snapshot with
       ``force=True`` (that would silently drop e.g. a project added
       via ``pm add-project`` after the server started).
+    * Idempotency is decided AGAINST the freshly-loaded disk state
+      (Codex round 2 on #2063): if disk already matches ``tracked``
+      we skip the write AND sync the live snapshot from disk so a
+      stale in-memory value can't keep lying. Deciding against the
+      long-lived ``ConfigDep`` here would let a server that booted
+      with ``tracked=True`` short-circuit a ``/resume`` call even
+      after an external edit flipped disk to ``False``.
     * Mutate the freshly-loaded copy, NOT the live ``config``.
     * Only after the disk write succeeds do we sync the change back
       into the live ``config`` so subsequent in-process reads see it.
@@ -435,6 +442,20 @@ def set_project_tracked(
         # Disk-side delete raced us. Treat as 404 — the in-memory state
         # is stale and the next GET will agree.
         raise not_found(f"Project not registered: {project_key}")
+
+    # Idempotency decided against DISK, not against the long-lived
+    # in-memory snapshot. If disk already matches the request we still
+    # sync the live ``config`` from disk so a stale in-process value
+    # can't continue lying to subsequent GETs.
+    if fresh_project.tracked == tracked:
+        live_project.tracked = fresh_project.tracked
+        live_project.path = fresh_project.path
+        live_project.name = fresh_project.name
+        config.projects[project_key] = live_project
+        for key, value in fresh.projects.items():
+            if key not in config.projects:
+                config.projects[key] = value
+        return _project_to_api(config, project_key, live_project)
 
     fresh_project.tracked = tracked
     fresh.projects[project_key] = fresh_project
--- a/src/pollypm/web_api/service.py
+++ b/src/pollypm/web_api/service.py
@@ -383,6 +383,51 @@ def _emit_project_audit(
         pass
 
 
+@contextlib.contextmanager
+def _config_write_lock(config_path: Path):
+    """Serialise config read-modify-write across concurrent API requests.
+
+    Codex round 6 on #2063 caught a lost-update race: two clients
+    pausing different projects concurrently each ``deepcopy`` the same
+    cached load_config snapshot, mutate their own project, then both
+    write with ``force=True``. The second writer's snapshot still has
+    the first writer's project at the OLD value — so the first write
+    is silently reverted on disk. Per-request reloads (post-#2056)
+    don't help because the race is between two requests, both of which
+    reload at the same moment before either writes.
+
+    Fix: hold an exclusive ``fcntl.flock`` on a sibling lockfile for
+    the entire ``load_config → mutate → write_config`` sequence. The
+    lockfile is created on first use (``open(..., "a+")`` is
+    create-if-missing without truncation). Lock release is best-effort
+    in the ``finally`` block so a propagating exception still unlocks.
+
+    POSIX-only. PollyPM ships as a personal-use Mac/Linux deployment;
+    Windows isn't a supported target.
+
+    The lock is per ``config_path`` so two unrelated configs (e.g.
+    multi-tenant dev setups) don't serialise against each other.
+    """
+    import fcntl
+
+    lock_path = config_path.with_suffix(config_path.suffix + ".lock")
+    # Ensure parent exists — config_path may not have been created yet
+    # on the very first request after a fresh install, but the parent
+    # ``.pollypm/`` always exists by the time the API is up.
+    lock_path.parent.mkdir(parents=True, exist_ok=True)
+    # "a+" creates if missing without truncating; we never read from
+    # or write to the lockfile, only flock its fd.
+    with open(lock_path, "a+") as lock_fh:
+        fcntl.flock(lock_fh.fileno(), fcntl.LOCK_EX)
+        try:
+            yield
+        finally:
+            try:
+                fcntl.flock(lock_fh.fileno(), fcntl.LOCK_UN)
+            except OSError:
+                pass
+
+
 def _refresh_live_projects(
     live_config: PollyPMConfig, fresh_config: PollyPMConfig
 ) -> None:
@@ -434,6 +479,16 @@ def set_project_tracked(
 
     Concurrent-safe write semantics (Codex review on #2063):
 
+    * The entire read-modify-write section is serialised by an
+      exclusive ``fcntl.flock`` on a sibling lockfile (Codex round 6
+      on #2063 — ``_config_write_lock``). Without this lock, two
+      concurrent ``/pause`` calls on DIFFERENT projects could each
+      deepcopy the same cached snapshot, mutate their own project,
+      and then both write ``force=True`` — the second writer would
+      silently revert the first writer's change on disk. The flock
+      forces request B to reload AFTER request A's write commits, so
+      its deepcopy reflects A's mutation and the merged write
+      preserves both.
     * Re-load the on-disk TOML so any external CLI / cockpit edits
       that landed between server boot and this call are preserved —
       we never write back the long-lived ``ConfigDep`` snapshot with
@@ -477,85 +532,102 @@ def set_project_tracked(
     from pollypm.config import load_config, write_config
 
     # Bind to the original (in-memory) project up-front so the 404 path
-    # doesn't pay for a disk reload.
+    # doesn't pay for a disk reload or take the lock.
     live_project = config.projects.get(project_key)
     if live_project is None:
         raise not_found(f"Project not registered: {project_key}")
 
     config_path = _resolve_config_write_path(config)
-    # Reload from disk so we don't lose concurrent CLI / cockpit edits
-    # (e.g. an external ``pm add-project`` that added a new project key
-    # the in-memory server hasn't seen yet).
-    try:
-        fresh_cached = load_config(config_path)
-    except OSError as exc:
-        raise service_unavailable(
-            f"Failed to reload config for {project_key}: {exc}",
-            hint="Check read permissions on the PollyPM config file.",
-        ) from exc
-    # ``load_config`` is memoised by ``config_path`` (see
-    # ``pollypm.config._config_cache``) so ``fresh_cached`` IS the same
-    # object as the live ``ConfigDep`` whenever the server was bootstrapped
-    # via ``load_config(config_path)`` — which is exactly what
-    # ``pm serve`` / ``load_api_config`` do in production. Mutating
-    # ``fresh_cached.projects[key].tracked`` would therefore mutate the
-    # live snapshot BEFORE the durable write completes. If write_config
-    # then raises OSError, the API returns 503 but the live config has
-    # already flipped — a rollback violation Codex reproduced on round 4
-    # of #2063. Take a deep copy here so all mutation happens on a
-    # detached graph; the live snapshot is only touched via
-    # ``_refresh_live_projects`` AFTER the disk write succeeds.
-    fresh = copy.deepcopy(fresh_cached)
-    fresh_project = fresh.projects.get(project_key)
-    if fresh_project is None:
-        # Disk-side delete raced us. Treat as 404 — the in-memory state
-        # is stale and the next GET will agree.
-        raise not_found(f"Project not registered: {project_key}")
+    # Hold the per-config write lock across the entire RMW so two
+    # concurrent requests on different projects can't lose each other's
+    # writes (Codex round 6 on #2063). The lock is released BEFORE we
+    # return; subsequent callers' load_config sees the updated mtime
+    # and reloads fresh.
+    with _config_write_lock(config_path):
+        # Reload from disk so we don't lose concurrent CLI / cockpit
+        # edits (e.g. an external ``pm add-project`` that added a new
+        # project key the in-memory server hasn't seen yet). Under the
+        # flock this also picks up any sibling-API-request mutation
+        # that committed while we were waiting on the lock.
+        try:
+            fresh_cached = load_config(config_path)
+        except OSError as exc:
+            raise service_unavailable(
+                f"Failed to reload config for {project_key}: {exc}",
+                hint="Check read permissions on the PollyPM config file.",
+            ) from exc
+        # ``load_config`` is memoised by ``config_path`` (see
+        # ``pollypm.config._config_cache``) so ``fresh_cached`` IS the
+        # same object as the live ``ConfigDep`` whenever the server was
+        # bootstrapped via ``load_config(config_path)`` — which is
+        # exactly what ``pm serve`` / ``load_api_config`` do in
+        # production. Mutating ``fresh_cached.projects[key].tracked``
+        # would therefore mutate the live snapshot BEFORE the durable
+        # write completes. If write_config then raises OSError, the API
+        # returns 503 but the live config has already flipped — a
+        # rollback violation Codex reproduced on round 4 of #2063. Take
+        # a deep copy here so all mutation happens on a detached graph;
+        # the live snapshot is only touched via
+        # ``_refresh_live_projects`` AFTER the disk write succeeds.
+        fresh = copy.deepcopy(fresh_cached)
+        fresh_project = fresh.projects.get(project_key)
+        if fresh_project is None:
+            # Disk-side delete raced us. Treat as 404 — the in-memory
+            # state is stale and the next GET will agree.
+            raise not_found(f"Project not registered: {project_key}")
 
-    # Idempotency decided against DISK, not against the long-lived
-    # in-memory snapshot. If disk already matches the request we still
-    # FULLY refresh the live ``config.projects`` from disk so any
-    # external metadata edits (persona_name, kind, role assignments,
-    # worker caps, etc.) propagate — Codex round 3 on #2063.
-    if fresh_project.tracked == tracked:
-        _refresh_live_projects(config, fresh)
-        return _project_to_api(config, project_key, config.projects[project_key])
+        # Idempotency decided against DISK, not against the long-lived
+        # in-memory snapshot. If disk already matches the request we
+        # still FULLY refresh the live ``config.projects`` from disk so
+        # any external metadata edits (persona_name, kind, role
+        # assignments, worker caps, etc.) propagate — Codex round 3 on
+        # #2063.
+        if fresh_project.tracked == tracked:
+            _refresh_live_projects(config, fresh)
+            return _project_to_api(
+                config, project_key, config.projects[project_key]
+            )
 
-    fresh_project.tracked = tracked
-    fresh.projects[project_key] = fresh_project
-    try:
-        write_config(fresh, config_path, force=True)
-    except OSError as exc:
-        # Live config untouched — Codex P0 #1 (rollback guarantee). The
-        # deep-copy above is what makes this rollback real on the
-        # cached-load_config path: ``fresh`` is a detached graph, so
-        # mutating ``fresh_project.tracked`` never reached the live
-        # ``ConfigDep`` shared with FastAPI request handlers. Codex
-        # round 4 on #2063.
-        raise service_unavailable(
-            f"Failed to persist project state for {project_key}: {exc}",
-            hint="Check write permissions on the PollyPM config file.",
-        ) from exc
+        fresh_project.tracked = tracked
+        fresh.projects[project_key] = fresh_project
+        try:
+            write_config(fresh, config_path, force=True)
+        except OSError as exc:
+            # Live config untouched — Codex P0 #1 (rollback guarantee).
+            # The deep-copy above is what makes this rollback real on
+            # the cached-load_config path: ``fresh`` is a detached
+            # graph, so mutating ``fresh_project.tracked`` never
+            # reached the live ``ConfigDep`` shared with FastAPI
+            # request handlers. Codex round 4 on #2063.
+            raise service_unavailable(
+                f"Failed to persist project state for {project_key}: {exc}",
+                hint="Check write permissions on the PollyPM config file.",
+            ) from exc
 
-    # Disk write succeeded — re-load disk and FULLY refresh the live
-    # ``config.projects`` so in-process callers see (a) the new
-    # ``tracked`` we just wrote AND (b) any concurrent external edits
-    # to other fields / other projects that landed between our load
-    # and our write — Codex round 3 on #2063. The extra load_config()
-    # is cheap relative to the durable write we just did and ensures
-    # we never widen a write-after-read race.
-    try:
-        post_write = load_config(config_path)
-    except OSError as exc:
-        # Live config still reflects pre-write state. Surface as 503
-        # so the client can retry; subsequent GETs stay consistent.
-        raise service_unavailable(
-            f"Failed to reload config after writing {project_key}: {exc}",
-            hint="Check read permissions on the PollyPM config file.",
-        ) from exc
-    _refresh_live_projects(config, post_write)
+        # Disk write succeeded — re-load disk and FULLY refresh the
+        # live ``config.projects`` so in-process callers see (a) the
+        # new ``tracked`` we just wrote AND (b) any concurrent external
+        # edits to other fields / other projects that landed between
+        # our load and our write — Codex round 3 on #2063. The extra
+        # load_config() is cheap relative to the durable write we just
+        # did. The flock guarantees no other API writer is between our
+        # write and this reload.
+        try:
+            post_write = load_config(config_path)
+        except OSError as exc:
+            # Live config still reflects pre-write state. Surface as
+            # 503 so the client can retry; subsequent GETs stay
+            # consistent.
+            raise service_unavailable(
+                f"Failed to reload config after writing {project_key}: {exc}",
+                hint="Check read permissions on the PollyPM config file.",
+            ) from exc
+        _refresh_live_projects(config, post_write)
 
-    refreshed_project = config.projects[project_key]
+        refreshed_project = config.projects[project_key]
+
+    # Audit emit outside the lock — it does its own I/O and we don't
+    # want it serialising with other API writers.
     _emit_project_audit(
         event="projects.tracked.set",
         config=config,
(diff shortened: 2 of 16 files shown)


src/pollypm/web_api/service.py
```diff
@@ -303,6 +303,154 @@ def project_drilldown(config: PollyPMConfig, key: str) -> APIProjectDrilldown |
...
+    client can refresh without a follow-up GET (same idiom as the
+    task transitions in Phase 2).
+    """
+    from pollypm.config import write_config
+
+    project = config.projects.get(project_key)
+    if project is None:
+        raise not_found(f"Project not registered: {project_key}")
+
+    project.tracked = tracked
```
samhotchkiss: Required changes before merge. `set_project_tracked()` mutates the live `ConfigDep` object before persistence and never rolls it back on write failure. I reproduced this by patching `pollypm.config.write_config` to raise `OSError`: the API maps the operation to 503, but `cfg.projects['demo'].tracked` stays `False`. Subsequent GETs from the same server advertise the paused/resumed state while the TOML still has the old value, and a restart flips back. Please mirror the archive rollback behavior, or mutate a reloaded/copy config and only update app state after a durable write. Add a regression test for the failed-write case.

src/pollypm/web_api/service.py
```diff
@@ -303,6 +303,154 @@ def project_drilldown(config: PollyPMConfig, key: str) -> APIProjectDrilldown |
...
+    config.projects[project_key] = project
+    try:
+        write_config(config, _resolve_config_write_path(config), force=True)
```
samhotchkiss: Required changes before merge. This writes the server's long-lived config object back with `force=True`, so concurrent CLI/cockpit edits can be lost. I reproduced it by starting the API with project `demo`, externally reloading the config and adding project `external`, then calling `set_project_tracked()`; the disk TOML ended with only `demo`. For a config mutation endpoint, reload the current file, apply the single project mutation through a shared durable helper, and use locking/compare/conflict semantics before refreshing app state. Please cover an external-edit regression.

src/pollypm/web_api/service.py
```diff
@@ -303,6 +303,154 @@ def project_drilldown(config: PollyPMConfig, key: str) -> APIProjectDrilldown |
...
+    gone. Re-registering uses ``pm add-project`` (CLI-only).
+    """
+    from pollypm.config import write_config
+
+    project = config.projects.get(project_key)
+    if project is None:
+        raise not_found(f"Project not registered: {project_key}")
+
+    label = project.display_label()
+    del config.projects[project_key]
```
samhotchkiss: Required changes before merge. `archive_project()` deletes the project without checking for enabled sessions that still reference it. The existing project removal path (`pollypm.projects.remove_project`) rejects that case to preserve the session -> project invariant. This endpoint can leave dangling sessions pointing at a missing project, which will break session/project views and daemon workflows. Please route this through the shared project-removal facade or preserve the same guard here, returning 409 with tests when sessions still reference the project.

src/pollypm/web_api/routes/projects.py
```diff
@@ -104,3 +117,155 @@ def get_project_plan_endpoint(
...
+)
+def archive_project_endpoint(
+    key: str,
+    config: ConfigDep,
+    body: _ReasonBody | None = None,
+) -> ActionResult:
+    reason = body.reason if body is not None else None
+    label = archive_project(config, key, reason=reason)
+    message = f"archived {label}"
+    if reason:
```
samhotchkiss, line 251: Required changes before merge. The API accepts `reason` and this response echoes it, but the service intentionally discards it and no audit event or durable recovery trail is written. For operator actions like archive/pause/resume, that is misleading observability: after the response, no one can discover why the state changed. Either persist a bounded audit event with tests/docs, or remove/reject `reason` from the contract until it is actually supported.

samhotchkiss (commented):
Required changes before merge. The endpoint tests and OpenAPI conformance checks pass, but the config write path breaks durable-config invariants around rollback, concurrent edits, and project/session ownership. Details inline.

src/pollypm/web_api/service.py
samhotchkiss: Fixed in 1bca4e7a2 (service.py L425-L432). The mutation now lands on a freshly-loaded copy of the config; only after `write_config` succeeds do we sync the live `ConfigDep`. Regression test `test_pause_rolls_back_in_memory_when_disk_write_fails` monkeypatches write_config to raise OSError and asserts (a) 503 mapping, (b) `cfg.projects['myproj'].tracked` stays True, (c) the follow-up GET reflects the unchanged value.

src/pollypm/web_api/service.py
samhotchkiss: Fixed in 1bca4e7a2 (service.py L406-L445). The endpoint now reloads from disk via `load_config(config_path)` before applying the mutation, writes the merged-and-mutated snapshot, and only then syncs the live `config.projects` dict (also merging any externally-added project keys so in-process readers see them). No more long-lived-snapshot `force=True` write. Regression test `test_pause_preserves_concurrent_external_project_addition` writes an external project to disk, calls `/pause`, then re-reads the TOML and asserts both `myproj.tracked=False` AND `external` survive. No new file-lock — kept consistent with the existing CLI pattern (`enable_tracked_project` / `remove_project` both reload-then-write without flock); adding one here without changing those would create asymmetric behaviour for v1 RC.

src/pollypm/web_api/service.py
samhotchkiss: Fixed in 1bca4e7a2 (service.py L483-L520). `archive_project` now routes through `pollypm.projects.remove_project` (the shared facade) which enforces the session-ref guard. `typer.BadParameter` with the `still used by` marker maps to 409 `conflict` with the blocking session names preserved in the message; other BadParameter cases (race where the project disappeared between our 404 check and the facade's reload) fall through to 404. Regression test `test_archive_blocked_when_enabled_session_references_project` enables a session referencing `myproj` and asserts 409 + session name in body + project still present in-memory.

src/pollypm/web_api/routes/projects.py
samhotchkiss: Fixed in 1bca4e7a2 (service.py L328-L367 + L447-L455 + L525-L532). Added `_emit_project_audit()` — best-effort wrapper around `pollypm.audit.emit` that fires `projects.tracked.set` (with `tracked` flag) or `projects.archive` (with `label`) every time the mutation lands on disk. `reason` rides in `metadata.reason`, clamped to 500 chars (the same bound the `_ReasonBody` Pydantic model enforces upstream, kept as belt-and-suspenders). Events land in both the per-project log (`<root>/.pollypm/audit.jsonl`) and the central tail (`~/.pollypm/audit/<project>.jsonl`) so operators can grep either surface. Audit failures are swallowed so a broken audit subsystem can't roll back / 503 a config mutation. Regression test `test_pause_emits_audit_event_with_reason` pins `POLLYPM_AUDIT_HOME` to a tmp dir, posts with `reason='testing-audit-emit'`, and asserts the event lands with the reason in metadata.

samhotchkiss:

The earlier rollback/concurrent-edit/session-ref/audit comments are mostly addressed in `web_api.service`, but the route still bypasses those durable-write semantics on the idempotent pause/resume paths.

Blocking durable-config bug: `src/pollypm/web_api/routes/projects.py:192-199` and `:217-224` decide “already paused/resumed” from the long-lived `ConfigDep` object and return without calling `set_project_tracked()`. That reintroduces the stale-snapshot problem the service helper was meant to fix. I reproduced it by starting the API with `myproj.tracked=True`, externally writing the TOML to `tracked=False`, then calling `POST /projects/myproj/resume`: the response was `200` with `tracked=true`, but `load_config(config_path).projects['myproj'].tracked` remained `False`. So the API reports a successful durable mutation while disk still has the opposite state.

Please move the idempotency decision into the service helper, after the fresh `load_config(config_path)`, and have the routes delegate unconditionally for pause/resume. If disk is already at the target value, return the disk-backed project snapshot without writing; if the live object is stale, sync it from the fresh config before returning. Add regression coverage for stale live state on both directions, especially `resume` with live `True` but disk `False` and `pause` with live `False` but disk `True`.

Docs also need to be updated now that the previous `reason` issue was fixed: `src/pollypm/web_api/routes/projects.py:127-140` and `docs/api/openapi.yaml:188-194`, `:248-254`, `:1293-1298` still say `reason` is not persisted, but the PR now emits it into audit metadata. That contradiction should not ship with a user-facing API contract.

Targeted verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py -v --timeout=120` -> 23 passed
- `uv run --extra test --with pytest-timeout pytest tests/web_api/test_endpoints.py -k project -v --timeout=120` -> 10 passed, 20 deselected
- One-off stale-idempotency probe confirmed `POST /resume` can return `tracked=true` while disk remains `tracked=false`.

samhotchkiss:
Round 2 fixes pushed (1f8fc7f74). Both blockers addressed:

**Blocker 1 — stale-snapshot idempotency (route bypass):**
- Moved the "already at target" decision INTO `set_project_tracked()` at `src/pollypm/web_api/service.py:378-422` (added an explicit idempotent branch after the existing `load_config(config_path)` reload). When disk matches the target it returns the disk-backed snapshot without writing AND syncs the live `ConfigDep` snapshot from disk so a subsequent GET can't keep lying.
- Removed the route-level short-circuits. `pause_project_endpoint` at `src/pollypm/web_api/routes/projects.py:185-195` and `resume_project_endpoint` at `src/pollypm/web_api/routes/projects.py:215-225` now delegate unconditionally to `set_project_tracked`.
- 3 new regression tests in `tests/test_projects_writes_endpoint.py`:
  - `test_resume_with_live_true_but_disk_false_reflects_disk` (your reproduction recipe, both API response and disk now reflect target)
  - `test_pause_with_live_false_but_disk_true_reflects_disk` (mirror case)
  - `test_resume_idempotent_when_disk_already_true_syncs_live` (no-write branch still syncs the stale live snapshot)

**Blocker 2 — `reason` docs contradicted round-1 audit emit:**
- `src/pollypm/web_api/routes/projects.py:127-148` (`_ReasonBody` docstring + `Field` description)
- `docs/api/openapi.yaml:188-196` (pauseProject description)
- `docs/api/openapi.yaml:248-258` (archiveProject description)
- `docs/api/openapi.yaml:1295-1310` (`ProjectReasonRequest` schema + property description)

All now say: "Persisted to the per-project audit log (`projects.tracked.set` / `projects.archive` events) along with the actor and timestamp. Clamped to 500 chars."

**Verification:**
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py -v --timeout=120` -> 26 passed (23 round-1 + 3 new)

Re-tagged `needs-review`.

samhotchkiss:

The route-level stale idempotency bug is fixed: pause/resume now delegate to `set_project_tracked()`, and the regression tests for stale live tracked state pass. The remaining blocker is that the live-config synchronization is still only partial, so the API can continue serving stale project metadata after a durable write.

`src/pollypm/web_api/service.py:set_project_tracked()` reloads disk, but in the idempotent branch it only copies `tracked`, `path`, and `name` from `fresh_project` back into the long-lived `ConfigDep` object. In the write-success branch it only flips `live_project.tracked`. That leaves other `KnownProject` fields (`persona_name`, `kind`, role/model assignments, worker caps, plan enforcement, etc.) stale when a CLI/cockpit edit changed the same project before the API call. The response is built from `live_project`, so the API can return old metadata while disk has the new metadata.

`archive_project()` has the same live-sync gap for concurrent external additions: the shared `remove_project` facade reloads and preserves disk edits, but after success the API only `pop()`s the archived key from the long-lived config. Any project added on disk before the archive remains absent from subsequent in-process `GET /projects` responses until the server restarts.

Please make the post-write/idempotent live snapshot match the freshly loaded/written config, not just the target `tracked` bit. A small fix would replace or fully refresh `config.projects` from the fresh post-mutation config after successful write/no-op paths, including the archive path after `remove_project` succeeds. Add regressions for same-project metadata changed on disk before `/pause` or `/resume`, and for an external project added on disk before `/archive` still appearing in a subsequent `GET /projects`.

Tests run:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py -v --timeout=120` -> 26 passed
- `uv run --extra test --with pytest-timeout pytest tests/web_api/test_endpoints.py -k project -v --timeout=120` -> 10 passed, 20 deselected

samhotchkiss:
Round 3 fixes pushed (e31d28c42). Both gaps addressed in one shot via a single refresh helper.

**New helper** — `src/pollypm/web_api/service.py:370-393` `_refresh_live_projects(live_config, fresh_config)`: in-place `clear() + update()` on the dict (KnownProject is a `@dataclass(slots=True)` and `PollyPMConfig.projects` is a plain `dict[str, KnownProject]`, so a dict-level swap preserves the FastAPI Depends-cached reference).

**Three callsites:**

1. `set_project_tracked()` idempotent branch (`service.py:464-466`) — replaces the 3-field copy. Now propagates `persona_name`, `kind`, role assignments, worker caps, plan enforcement, AND any concurrent external add/remove on other keys.

2. `set_project_tracked()` write-success branch (`service.py:488-499`) — replaces the `live_project.tracked = target` line. After the durable write succeeds it re-loads disk (so any external edit that landed between our load and our write is captured) and refreshes. A `load_config` failure here maps to 503 with the live snapshot still pre-write so subsequent GETs stay consistent.

3. `archive_project()` (`service.py:576-595`) — replaces the `config.projects.pop(...)` line. After `remove_project()` succeeds we re-load disk and refresh, which both drops the archived key AND surfaces any external project additions / metadata edits.

**New tests** (3, all green — 29 total in the file):
- `test_pause_idempotent_refreshes_external_metadata_edit` — idempotent branch + `persona_name` change.
- `test_resume_write_success_refreshes_external_metadata_edit` — write branch + `persona_name` change.
- `test_archive_surfaces_concurrent_external_project_addition` — external add before `/archive` shows up in `GET /projects`.

Tests run:
- `pytest --noconftest tests/test_projects_writes_endpoint.py -x --timeout=120` -> 29 passed
- `pytest tests/web_api/test_openapi_conformance.py -x --timeout=120` -> 5 passed
- `pytest tests/web_api/test_endpoints.py -k project --timeout=120` -> 10 passed, 20 deselected

`needs-review` label re-added; ready for round 4.

samhotchkiss:

1. Durable-config rollback is still broken on the production cached-config path. In `src/pollypm/web_api/service.py:450-474`, `set_project_tracked()` calls `load_config(config_path)` and then mutates `fresh_project.tracked` before `write_config()`. `load_config()` is cached and `pm serve` also gets its `ConfigDep` from `load_api_config()` / `load_config()`, so `fresh` can be the exact same object as the live FastAPI config. If `write_config()` then raises, the endpoint returns 503 but the live in-memory project has already flipped.

I reproduced this against the current head by creating the app with a config returned from `load_config(config_path)`, monkeypatching `pollypm.config.write_config` to raise `OSError`, and calling `POST /api/v1/projects/myproj/pause`: response was `503 service_unavailable`, but `live.projects['myproj'].tracked` became `False` and a follow-up `GET /api/v1/projects/myproj` returned `tracked=false`. The committed rollback test at `tests/test_projects_writes_endpoint.py:429-465` misses this because it passes a hand-constructed config object that is not the object in the `load_config()` cache. Please make the mutation operate on an uncached/deep-copied config snapshot and only refresh the live `ConfigDep` after the durable write succeeds. Add a regression that starts from `live = load_config(config_path)` before constructing the app.

2. The archive session-reference conflict is implemented but not documented. `archive_project()` maps enabled session references to `409 conflict`, but `src/pollypm/web_api/routes/projects.py:229-233` and `docs/api/openapi.yaml:263-275` only document 401/404/503. Since this is user-visible API behavior and an important invariant guard, please add the 409 response to both generated/static docs and pin it in conformance coverage.

Targeted verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py -q --timeout=120` -> 29 passed
- One-off cached-config rollback probe -> reproduced the 503 + stale live `tracked=false` bug above

samhotchkiss:
Round 4 fixes pushed (c8fb1df19). Both blockers closed.

**Blocker 1 — rollback on the cached load_config path.** Confirmed: `pollypm.config.load_config` is memoised by path (see `_config_cache` at `config.py:899`), so `fresh = load_config(config_path)` in `set_project_tracked()` returned the SAME object as the live `ConfigDep` whenever the server booted via `load_config` (the `pm serve` path). Mutating `fresh.projects[key].tracked` flipped the live snapshot pre-write.

Fix in `src/pollypm/web_api/service.py:457` — deep-copy the loaded snapshot before any mutation; the live config is only touched via `_refresh_live_projects` AFTER the durable write succeeds. Mirror gap in `archive_project`: the `remove_project` facade does `del config.projects[key]` on the cached object before its own `write_config` (`projects.py:842-843`), so an OSError after the `del` would leak a half-applied delete into the live snapshot. Fixed at the service boundary (`service.py:563-577`) by snapshotting the live entry and restoring it on OSError.

**Blocker 2 — archive's 409 not documented.** Added the 409 response in both surfaces:
- `src/pollypm/web_api/routes/projects.py:233-241` — adds 409 with the blocking-session description to the route's `responses=` map (propagates into the auto-generated OpenAPI).
- `docs/api/openapi.yaml:263-285` — adds the 409 response inline (custom description naming `error.message` carries blocking sessions) plus an extra paragraph in the route description spelling out the session→project invariant.
- `tests/web_api/test_openapi_conformance.py:124-188` — new `test_archive_documents_409_session_reference_conflict` pins both the static contract AND the auto-generated FastAPI doc to include 409.

**New tests (3):**
- `test_pause_rolls_back_with_real_load_config_cache` — exact production reproduction; asserts `load_config(path) is live` so future cache-semantics drift trips CI.
- `test_archive_rolls_back_with_real_load_config_cache` — mirror case for the facade path.
- `test_archive_documents_409_session_reference_conflict` — conformance pin.

**Surprise:** `pollypm.projects` does `from pollypm.config import write_config` at module load, so patching `pollypm.config.write_config` doesn't reach the facade's binding. Archive rollback test had to patch `pollypm.projects.write_config` directly (commented inline). `set_project_tracked`'s lazy `from pollypm.config import ...` inside the function body re-resolves the name each call, so `pollypm.config.write_config` patching still works there.

Tests:
- `pytest --noconftest tests/test_projects_writes_endpoint.py -x --timeout=120` -> 31 passed
- `pytest tests/web_api/test_openapi_conformance.py -x --timeout=120` -> 6 passed
- `pytest tests/web_api/test_endpoints.py -k project --timeout=120` -> 10 passed, 20 deselected

`needs-review` re-added; ready for round 5.

samhotchkiss:

This PR is mergeable at the Git level, but it is not safe to merge against current `main` after #2056. I tested the actual merge result by merging `origin/main` (`ff27b59e`, includes the config endpoint change) into PR head `c8fb1df19`; the merge was clean, but the targeted suite now fails:

- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py -q --timeout=120` -> 8 failed, 23 passed

The root cause is that #2056 changed `create_app()` to satisfy `ConfigDep` by reloading `load_config(config.config_path)` per request. This PR's tests and service comments still assume the app uses the long-lived `config` object passed into `create_app()`. On the merged tree, several tests mutate or assert that original fixture object, but the route is operating on the per-request loaded config instead.

Examples from the failing merge-result run:

- `test_pause_happy_path_sets_tracked_false`, `test_pause_is_idempotent_on_already_paused`, `test_resume_happy_path_sets_tracked_true`, and `test_archive_happy_path_removes_from_config` all assert mutations on the original `config` fixture, which is no longer the object the app dependency necessarily serves when `config_path` is set.
- `test_resume_idempotent_when_disk_already_true_syncs_live`, `test_pause_idempotent_refreshes_external_metadata_edit`, `test_resume_write_success_refreshes_external_metadata_edit`, and `test_archive_surfaces_concurrent_external_project_addition` are pinning the old live-snapshot refresh invariant. After #2056, the invariant should be restated around durable disk state plus the next request's reloaded config, or the tests need to intentionally construct an in-memory config with no `config_path` when that fallback behavior is what they are testing.

Please rebase onto current `main`, update the project-write implementation/tests for the new `ConfigDep` contract, and rerun the project-write and OpenAPI conformance suites on the merged tree. In particular, tests that want to simulate external config edits should write the TOML and assert the response plus a follow-up GET/disk reload, not only mutate/assert the stale object originally passed to `create_app()`.

samhotchkiss:
Round 5 fixes pushed (0f8a8e4cb). Rebased onto current `main` (ff27b59e7, includes #2056) and adapted tests + service docs to the new per-request `ConfigDep` reload contract.

**Rebase:** clean, no conflicts.

**8 failing tests — all fixed via Pattern A** (write disk to set up state, assert against `load_config(config_path)` + response body, not the fixture `config` object):

- `tests/test_projects_writes_endpoint.py:146` `test_pause_happy_path_sets_tracked_false` — Pattern A; added `config_path` fixture param, assert `load_config(config_path).projects["myproj"].tracked is False`.
- `tests/test_projects_writes_endpoint.py:166` `test_pause_is_idempotent_on_already_paused` — Pattern A; same swap on the post-second-call assertion.
- `tests/test_projects_writes_endpoint.py:198` `test_resume_happy_path_sets_tracked_true` — Pattern A; setup writes `paused.projects["myproj"].tracked = False` to disk (not just the fixture object), then asserts via `load_config`.
- `tests/test_projects_writes_endpoint.py:233` `test_archive_happy_path_removes_from_config` — Pattern A; assert `"myproj" not in load_config(config_path).projects`.
- `tests/test_projects_writes_endpoint.py:786` `test_resume_idempotent_when_disk_already_true_syncs_live` — Pattern A + docstring restated. The round-2 "live snapshot sync" invariant is now subsumed by the per-request reload; the test pins the durable invariant (idempotent /resume returns 200 with the disk-backed snapshot; follow-up GET reflects disk).
- `tests/test_projects_writes_endpoint.py:838` `test_pause_idempotent_refreshes_external_metadata_edit` — Pattern A + docstring restated. Writes `tracked=False, persona_name="new"` to disk; asserts response body's `persona_name == "new"` + follow-up GET + disk reload.
- `tests/test_projects_writes_endpoint.py:890` `test_resume_write_success_refreshes_external_metadata_edit` — Pattern A + docstring restated to pin Codex P0 #2 (concurrent-edit preservation) instead of the obsolete live-snapshot-refresh invariant. Asserts response `persona_name == "new"` + disk preserves it post-write.
- `tests/test_projects_writes_endpoint.py:944` `test_archive_surfaces_concurrent_external_project_addition` — Pattern A + docstring restated; pins durable invariant (disk drops `myproj`, preserves `external`) + GET round-trip.

**`_refresh_live_projects` decision: KEPT.** Reasoning at `src/pollypm/web_api/service.py:370-405`:
- Still load-bearing for the **in-request response** — `_project_to_api` reads from `config.projects[key]` after the refresh, so the response reflects the post-write disk state without an extra arg-rewrite.
- Still load-bearing for the **in-memory fallback path** (`config_path is None`) where the app factory wires a startup-frozen `lambda: config` provider with no per-request reload (`src/pollypm/web_api/app.py:132-133`). Tests construct `PollyPMConfig` directly and rely on this path.
- For the production path (`config_path` set), the function is now a cheap dict-swap that the NEXT request would discard via its own reload — but removing it would either require routing the response through `post_write.projects[key]` (extra branching) or break the fallback path.

Net behavior: no functional change on the production path beyond the rebase; tests pass on the merged tree.

**Tests run on the merged tree:**
- `pytest --noconftest tests/test_projects_writes_endpoint.py --timeout=120` → 31 passed
- `pytest tests/web_api/test_openapi_conformance.py --timeout=120` → 6 passed

`needs-review` re-added; ready for round 6.

samhotchkiss:

The previous stale-cache and rollback blockers are addressed, and the targeted suites pass, but the durable config write still has a lost-update race for concurrent API mutations.

`src/pollypm/web_api/service.py:469-511` does the standard `load_config(config_path)` -> `deepcopy` -> mutate one project -> `write_config(..., force=True)` flow with no file lock, mtime/epoch compare, or retry. `archive_project()` delegates to `pollypm.projects.remove_project()` which has the same unlocked reload/mutate/write shape. This preserves edits that are already on disk before this helper reloads, but it does not protect edits that land between two concurrent API requests' reloads and writes.

Concrete failure mode: two clients pause different projects at the same time. Both requests copy the same cached config, request A writes `project_a.tracked=false`, then request B writes its older snapshot with `project_b.tracked=false` and `project_a.tracked=true`. The final TOML loses A's successful mutation. I reproduced the underlying interleaving locally with two deep-copied snapshots; final state was `project_a.tracked=True`, `project_b.tracked=False` after both writes. This is config data loss, not just stale response metadata.

Please serialize the read-modify-write section or add real conflict detection. Reasonable fixes would be a per-config lock around load/mutate/write for these API helpers, or an mtime/state-epoch compare-and-retry/409 path that never writes an older snapshot over a newer one. Add a regression with two concurrent mutations to different projects and assert both durable changes survive, plus a concurrent archive/toggle case if archive keeps using the shared facade. Also update the service comments that currently describe this path as concurrent-safe; today it only handles stale-before-call edits, not concurrent writers.

Tests run:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py -q --timeout=120` -> 31 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 6 passed
- `uv run --extra test --with pytest-timeout pytest tests/web_api/test_endpoints.py -k project -q --timeout=120` -> 10 passed, 20 deselected
- One-off lost-update interleaving probe -> reproduced: final state kept only the second writer's change

samhotchkiss:
Round 6 fix pushed (f5407b059). Lost-update race closed via `fcntl.flock` serialisation.

**Fix shape — `_config_write_lock` context manager** at `src/pollypm/web_api/service.py:370-409`:
- Exclusive `fcntl.flock` on a sibling `<config_path>.lock` file (created on first use via `open(..., "a+")`).
- POSIX-only — matches existing patterns in `advisor_tick`, `rail_daemon_supervisor`, `inline` scheduler, `transcript_ingest`. No central lock helper existed, so I inlined one keyed to PollyPM's personal-use Mac/Linux target.
- Lock released in `finally`; per-config-path so unrelated configs don't serialise.

**Wire-in sites:**
1. `set_project_tracked` at `src/pollypm/web_api/service.py:526` — wraps the entire `load_config → deepcopy → mutate → write_config → reload → _refresh_live_projects` sequence. Audit emit stays OUTSIDE the lock (independent I/O).
2. `archive_project` at `src/pollypm/web_api/service.py:678` — wraps the `_remove_project_facade` call AND the post-archive reload + refresh. Chose option (a) — wrap at API entry point, not inside `remove_project()` itself — to avoid expanding scope into CLI helpers.

**Tests — 2 new, both at `tests/test_projects_writes_endpoint.py`:**
- `test_concurrent_pause_on_different_projects_preserves_both` (~line 1290): two threads with `threading.Barrier(2)` call `set_project_tracked` on `project_a` and `project_b` simultaneously; asserts both `tracked=False` survive on disk after both writes commit. Pattern borrowed from `tests/test_pg_notifications_race.py`.
- `test_concurrent_archive_and_pause_preserves_both` (~line 1419): archive of `project_a` + pause of `project_b` concurrently; asserts `project_a` is gone AND `project_b.tracked=False` survives.

Tests drive the service helpers directly rather than via `TestClient` — Starlette's threadpool would dilute the race window, and the bug is in the service layer.

**Negative-case verification:** I temporarily replaced both `_config_write_lock(config_path)` wrappers with `contextlib.nullcontext()` and ran the two new tests 5 times in a row — **2/2 tests failed on every run** (10/10 total). With the lock restored: `33 passed in 5.32s` deterministically. The tests pin the right invariant.

**Comments cleanup:** Updated the `set_project_tracked` and `archive_project` docstrings to lead with the flock semantics and explicitly call out that pre-#2063-r6 the path was only stale-before-call safe, not concurrent-writer safe.

**Tests run:**
- `pytest --noconftest tests/test_projects_writes_endpoint.py -x --timeout=120` → **33 passed** (31 round-5 + 2 new)
- `pytest tests/web_api/test_openapi_conformance.py -x --timeout=120` → 6 passed
- `pytest tests/web_api/test_endpoints.py -k project -q --timeout=120` → 10 passed, 20 deselected

**Surprises:** No existing central lock helper — every flock site rolls its own. Kept that convention rather than introduce one mid-RC.

`needs-review` re-added; ready for round 7.

samhotchkiss:

The round-6 flock closes the API-vs-API lost-update case, but the durable config write is still not safe against the existing non-API config writers. The new `_config_write_lock()` is local to `src/pollypm/web_api/service.py`, while the established config mutation paths still call `write_config(..., force=True)` without that lock (`src/pollypm/projects.py` has several of these, including `remove_project()` at `:826-843`). A CLI/cockpit/config writer that lands between the API helper's `load_config()` and `write_config()` is still overwritten by the API's older deep-copied snapshot.

That means the root lost-update invariant is not actually closed, it is only closed for callers that happen to use these new API helpers. The service comments also overclaim this: `set_project_tracked()` says it preserves concurrent external edits that land between load and write, but the current lock cannot protect writers that never take the lock.

I reproduced the remaining data-loss interleaving on current head `f5407b059`: monkeypatched `pollypm.config.write_config` so an external writer changed `project_b.persona_name = "external"` immediately before the API pause wrote its snapshot for `project_a`. Final TOML was `project_a.tracked=False` and `project_b.persona_name="old"`, so the external edit was lost while the API mutation survived. This is the same class of data loss as the earlier two-API-writer race, just through an existing operator workflow.

Please make this a shared config-write invariant rather than an API-local lock. Reasonable fixes: put the lock/conflict detection in a shared config writer/facade used by both API and CLI project mutations, or add an mtime/epoch compare-and-reload/retry/409 path before the API write so non-locking external writes cannot be overwritten. Add a regression that simulates a non-API config edit between the API reload and API write, and update the service comments so they only claim the failure modes actually covered.

Targeted verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py -q --timeout=120` -> 33 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 6 passed
- `uv run --extra test --with pytest-timeout pytest tests/web_api/test_endpoints.py -k project -q --timeout=120` -> 10 passed, 20 deselected
- One-off external-writer interleaving probe -> reproduced remaining lost update: final state was `project_a.tracked=False`, `project_b.persona_name="old"` instead of `"external"`

samhotchkiss:
Round 7 fix pushed — `291b1e977`.

**Pattern chosen: hybrid A+B with re-entrant shared lock.**
- Pattern A (defense in depth): `write_config` itself acquires `config_rmw_lock` — catches any stray caller that forgets to wrap the full RMW.
- Pattern B (the real fix): every in-tree RMW call site wraps `load → mutate → write` under the same lock. Thread-local depth counter makes nested acquires (Pattern B wrapping → write_config's Pattern A) a cheap no-op so they don't deadlock.

**Lock primitive:** `src/pollypm/config.py:1462` (`config_rmw_lock` context manager) + `src/pollypm/config.py:1539` (write_config's internal acquire). The lockfile is `<config_path>.lock`, per-config so multi-tenant dev setups don't serialise against each other. POSIX `fcntl.flock`, re-entrant per thread.

**Callsite list (every `write_config` caller in src/pollypm/, wrapped where it does its own load+mutate+write):**
- `web_api/service.py`: `set_project_tracked`, `archive_project` (now use shared lock; old `_config_write_lock` removed)
- `projects.py`: `enable_tracked_project`, `register_project`, `rename_project`, `remove_project` (Codex's reproduction target), `set_workspace_root`
- `accounts.py`: `add_account_via_login`, `relogin_account`, `remove_account`, `set_controller_account`, `set_open_permissions_default`, `toggle_failover_account`
- `onboarding.py`: `add_selected_projects`, `relogin_account`
- `workers.py`: `create_worker_session`, `remove_worker_session`
- `plugins_builtin/project_planning/cli/project.py`: `_purge_project_sessions`
- `cockpit_ui.py`: `_write_global_role_assignment`, `_clear_global_role_assignment`, `_persist_session_account_targets`
- `cockpit_project_settings.py`: `_save_project_role_assignment`, `_clear_project_role_override`, release-channel toggle

Two callsites NOT wrapped (rely on Pattern A only):
- `scan_projects` — interactive `typer.confirm` inside the loop; holding the lock across user prompts would block API writers for the duration of user thinking.
- `onboarding_tui.py:1077` — fresh `_build_config()` write during initial onboarding; no load → mutate → write sequence to protect.

The interactive helpers in `accounts.py` / `onboarding.py` that launch tmux login windows re-load INSIDE the lock at the commit point so any disk edits during the login flow merge forward instead of being clobbered.

**New tests** in `tests/test_projects_writes_endpoint.py`:
- `test_concurrent_api_pause_and_cli_remove_preserves_both` — `threading.Barrier`-synchronized race between `set_project_tracked` (project_a) and `pollypm.projects.remove_project` (project_c). Confirmed FAILS pre-fix: `project_c` resurrects because the API write clobbers the CLI removal. Passes post-fix.
- `test_config_rmw_lock_is_reentrant_within_thread` — pins the depth-counter invariant; a regression that broke re-entrancy would deadlock here (caught by pytest-timeout).

**Service comments updated:** `set_project_tracked` / `archive_project` docstrings now reference `pollypm.config.config_rmw_lock` and explain how the shared invariant covers CLI / cockpit / onboarding callers, not just API-vs-API.

**Verification I ran:**
- `pytest --noconftest tests/test_projects_writes_endpoint.py -x --timeout=120` → 35 passed (was 33; +2 round-7 tests)
- `pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` → 6 passed
- `pytest --noconftest tests/test_project_rename.py` → 11 passed
- `pytest --noconftest tests/test_cockpit_project_state.py tests/test_cockpit_project_health.py` → 24 passed
- Cross-checked failing tests in `test_accounts.py` / `test_projects.py` / `test_project_remove_cli.py` / `test_cockpit_settings_projects.py` against baseline `f5407b059` — same failures exist there (PgWorkService `_db_path` removal, unrelated to #2063).

Surprise during implementation: `write_config` itself wasn't a simple write — it ALSO writes per-project `<project>/.pollypm/config/project.toml` files in a loop, so the lock matters for the entire write sequence not just the global TOML. Pattern A wraps the whole multi-file write, so a torn read of one project's local config can't happen during a sibling project's write.

samhotchkiss:

Most of the earlier project-write blockers are addressed in this head, and the targeted suites pass. The remaining blocker is in the new shared config-lock primitive: `config_rmw_lock()` is reentrant by a single thread-local depth counter, not by the actual config/lock path.

`src/pollypm/config.py:1507-1517` treats any nested acquire in the same thread as a no-op once any config lock is held. That is only safe for the same lock path, which is the `with config_rmw_lock(path): ... write_config(..., path, ...)` pattern this PR needs. It is not safe for a nested acquire of a different config path. In that case the second path's `flock` is skipped entirely, so another process can already hold that second lock while this thread still enters the supposedly locked section.

I reproduced this on the current head with two config paths: a child process held `config_rmw_lock(path_b)` for four seconds; the parent then ran `with config_rmw_lock(path_a): with config_rmw_lock(path_b): ...` and entered the inner `path_b` section after `0.001s`. A correct per-config lock would block until the child released `path_b`. This violates the invariant the comments claim in `src/pollypm/config.py:1478-1503` and leaves a real data-loss/torn-write class for any multi-config or future nested writer.

Please track held lock paths/counts in the thread-local state and only treat an acquire as reentrant when it is for the same canonical lock path. Acquiring a different path should take that path's `flock` normally, or fail explicitly if you want to forbid nested multi-config locks to avoid deadlock. Add a regression that holds `path_b` from another process/thread and proves nested `path_a -> path_b` does not bypass the `path_b` lock; the current same-path reentrancy test does not cover this.

Related first-write gap: `write_config()` still checks `if path.exists() and not force` before acquiring the lock (`src/pollypm/config.py:1539-1548`). That means the `force=False` creation guard remains raceable for a missing config. Move that existence check inside the same lock, and make the lock path canonical even when the target config file does not exist, so fresh config initialization participates in the same invariant.

Targeted verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py -q --timeout=120` -> 35 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 6 passed
- `uv run --extra test --with pytest-timeout pytest tests/web_api/test_endpoints.py -k project -q --timeout=120` -> 10 passed, 20 deselected
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_project_rename.py -q --timeout=120` -> 11 passed
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_cockpit_project_state.py tests/test_cockpit_project_health.py -q --timeout=120` -> 24 passed
- One-off cross-process lock probe -> reproduced bypass: nested `path_a -> path_b` entered after `0.001s` while another process still held `path_b`

samhotchkiss:
Round 8 fix pushed — `faa37303e`.

Both round-8 Codex blockers addressed in `src/pollypm/config.py`.

**Blocker 1 — per-path re-entrancy** (`src/pollypm/config.py:1496-1605`).
Replaced the path-blind thread-local depth counter with a thread-local dict keyed by canonical config path -> `[file_handle, depth]`. Same-path nested acquires bump the depth (cheap no-op, no flock); different-path nested acquires now take that path's flock normally. `_config_lock_path()` canonicalises via `os.path.realpath` (`src/pollypm/config.py:1466-1483`) so fresh-config initialisation also produces a stable canonical lock key. New helpers: `_canonical_lock_key()` and `_held_locks()`.

**Blocker 2 — first-write race in `write_config`** (`src/pollypm/config.py:1608-1624`). Moved the `if path.exists() and not force` guard INSIDE the `config_rmw_lock(path)` block so the existence check and write are serialised under the same flock. Two writers initialising a fresh config can no longer both observe `False` and clobber each other.

**Tests** in `tests/test_projects_writes_endpoint.py`:
- `test_config_rmw_lock_does_not_bypass_different_path` — `multiprocessing.spawn` child holds `path_b`'s flock for 2s; parent nests `config_rmw_lock(path_a)` -> `config_rmw_lock(path_b)` and asserts the inner acquire blocks (elapsed >= 1s). Mirrors your repro.
- `test_first_write_race_serialized_by_lock` — two threads, `threading.Barrier`-synchronised, both call `write_config(force=False)` on a non-existent path. Asserts exactly one succeeds and the other raises `FileExistsError`.

**Confirmed FAILING on round-7 head (`291b1e977`)** before this fix:
- `test_config_rmw_lock_does_not_bypass_different_path`: `elapsed=0.002s, expected >= 1.000s` — exact reproduction of your `0.001s` Codex repro.
- `test_first_write_race_serialized_by_lock`: `expected exactly one writer to succeed, got successes=['alpha', 'bravo'], failures=[]` — both writers clobbered.

Both PASS post-fix. The round-7 same-path reentrancy test (`test_config_rmw_lock_is_reentrant_within_thread`) still passes — the new logic preserves the API-helper-wraps-then-`write_config`-nests pattern.

**Verified suites:**
- `pytest --noconftest tests/test_projects_writes_endpoint.py --timeout=60` -> 37 passed
- `pytest tests/web_api/test_openapi_conformance.py --timeout=120` -> 6 passed
- `pytest --noconftest tests/test_project_rename.py --timeout=120` -> 11 passed
- `pytest --noconftest tests/test_cockpit_project_state.py tests/test_cockpit_project_health.py --timeout=120` -> 24 passed

`needs-review` label re-applied for your next pass.

samhotchkiss:

The targeted suites pass, but the shared config read-modify-write invariant is still incomplete.

1. `src/pollypm/projects.py:870-897` (`scan_projects`) still does `load_config -> mutate config.projects -> write_config` without `config_rmw_lock`. Because discovery/scaffolding happens between the load and write, this can still clobber a concurrent API pause/archive or CLI project removal with an older `config.projects` snapshot. Please do discovery outside the critical section if needed, then enter `config_rmw_lock(config_path)`, reload, recompute known paths/keys against the fresh config, apply additions, and write. Add a regression racing `scan_projects` (or a minimal equivalent through that function) against `set_project_tracked()`/`remove_project()` so both mutations survive.

2. `src/pollypm/config.py:1430-1437` (`write_example_config`) still has the first-write race on the actual init wrapper. It checks `path.exists()` outside the lock and then calls `write_config(..., force=True)`, bypassing the new locked `force=False` guard in `write_config()`. Two `pm init` / `write_example_config()` callers can both pass the outer check and the second write wins. Move that existence decision under `config_rmw_lock`, or call `write_config(..., force=False)` and let the locked guard decide. Add a regression that races two `write_example_config()` calls on the same missing path.

Please also audit the remaining `write_config()` callsites before re-tagging. I still see `src/pollypm/config.py:992-996` auto-persisting session auth tokens during `load_config()` and `src/pollypm/onboarding_tui.py:1076-1077` writing a built config outside a full RMW lock/reload path. If those are intentionally outside the invariant, document why; otherwise they need the same treatment. The current comments in `src/pollypm/web_api/service.py:370-383` and `src/pollypm/config.py:1452-1456` claim every in-tree config writer participates, which is not true yet.

Verification I ran:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py tests/test_project_rename.py tests/test_cockpit_project_state.py tests/test_cockpit_project_health.py --timeout=180` -> 72 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py --timeout=120` -> 6 passed

samhotchkiss:
Round 9 fix pushed — `ae1f529e7`.

**Blocker 1 — `scan_projects` (`src/pollypm/projects.py:864-948`).**
Restructured per your prescription. Filesystem discovery (`discover_recent_git_repositories`) and interactive `typer.confirm` run OUTSIDE `config_rmw_lock` — holding the lock across user prompts would block every other config writer for the duration of user thinking. Inside the lock: reload via `load_config(config_path)`, recompute `known_paths` / `known_keys` against the FRESH snapshot, skip candidates a concurrent writer already added, merge new entries, `write_config`. `ensure_project_scaffold` (idempotent mkdirs against the project's own `.pollypm/`) runs after lock release. Interactive flow preserved — operator confirmation captured pre-lock as `candidates`.

Regression: `test_scan_projects_race_with_set_project_tracked_preserves_both` in `tests/test_projects_writes_endpoint.py` — `threading.Barrier`-synchronised race between `scan_projects(interactive=False)` and `set_project_tracked()`. Asserts `project_a.tracked == False` AND the scan's new project survives.

**Blocker 2 — `write_example_config` (`src/pollypm/config.py:1430-1448`).**
Moved the existence decision under `config_rmw_lock` by delegating to `write_config(force=False)`'s locked guard (which already raises `FileExistsError` atomically under the flock post round 8). Caller's `force` propagates through.

Regression: `test_write_example_config_race_serialized_by_lock` — two concurrent `write_example_config(missing_path, force=False)` calls; exactly one wins, the other raises `FileExistsError`. Mirrors the round-8 `test_first_write_race_serialized_by_lock` pattern.

**Audit 1 — `load_config` auto-persist (`src/pollypm/config.py:932+`).**
Factored a pure `_parse_config()` helper so the post-load auth-token mint can re-read under `config_rmw_lock(config_path)` without recursing through `load_config`'s caching layer. The persisted write now merges token additions onto the freshest disk snapshot rather than the in-memory snapshot we parsed before any concurrent writer might have landed. Token values from the re-parsed `fresh` are copied back into the live snapshot so callers see consistent values.

**Audit 2 — `onboarding_tui.py:1073-1093`.** Documented as intentionally outside the RMW wrap. Fresh-config build+write during first-run onboarding has no prior snapshot to merge against and runs before the API/CLI is up. Pattern A (`write_config`'s internal lock) still serialises the multi-file write itself, so a torn write across global + per-project files cannot happen.

**Comments updated:**
- `src/pollypm/web_api/service.py:370-393` — now lists the two documented exceptions (onboarding_tui first-run + the load_config auth-token mint with its re-read) instead of overclaiming.
- `src/pollypm/config.py:1442-1470` — same exception list in the lock-primitive banner.

**Verified:**
- `pytest --noconftest tests/test_projects_writes_endpoint.py -x --timeout=120` → 39 passed (was 37; +2 round-9 tests).
- `pytest tests/web_api/test_openapi_conformance.py tests/test_project_rename.py tests/test_cockpit_project_state.py tests/test_cockpit_project_health.py -x --timeout=120` → 41 passed.

Surprise: `load_config` posed a recursion risk because its post-parse persistence path needs to re-read under the lock — solved by factoring out `_parse_config()` rather than adding a thread-local recursion guard.

`needs-review` re-applied.

samhotchkiss:

This PR is not mergeable after #2060 landed. GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, and a local merge probe of current `origin/main` (`1095be5a5`, the #2060 squash) into head `ae1f529e7` conflicts in:

- `docs/api/openapi.yaml`
- `src/pollypm/web_api/service.py`
- `tests/web_api/test_openapi_conformance.py`

The conflict surface is not just mechanical whitespace. The rebase needs to preserve both sets of API contracts:

- `docs/api/openapi.yaml` now has competing `ServiceUnavailable` response text from project/config writes and inbox/pg writes. Resolve it so the shared response is accurate for both, or split the descriptions at the operation level.
- `src/pollypm/web_api/service.py` has an additive `__all__` conflict. The rebased result must retain the project-write exports from this PR (`archive_project`, `set_project_tracked`, `init_project_guide_for_role`) and the newly merged inbox-write exports from #2060 (`archive_inbox_item`, `mark_read_inbox_item`, `promote_inbox_to_task`, `reply_inbox_item`, `snooze_inbox_item`).
- `tests/web_api/test_openapi_conformance.py` has competing conformance blocks. Keep the #2063 archive-409 coverage and the #2060 idempotency / inbox-write-503 / archive-reason-ordering coverage.

I did not run the targeted project-write suite on this pass because there is no clean merged tree to test. Please rebase onto current `main`, resolve the conflicts while preserving the #2060 contracts above, then rerun the project-write and OpenAPI conformance suites.

samhotchkiss:
Rebased onto current `main` (`1095be5a5`, the #2060 squash). New head: `86167c559`.

**Conflict resolutions (the 3 Codex called out):**
- `docs/api/openapi.yaml`: `ServiceUnavailable` description merged into a single accurate-for-both block — names both backing stores (work-service / pg for inbox + task writes; state.db / pollypm.toml for project + config writes), keeps the `service_unavailable` error code + retry / `pm doctor` guidance.
- `src/pollypm/web_api/service.py`: `__all__` kept ALL entries from both sides — `archive_inbox_item`, `archive_project`, `init_project_guide_for_role`, `mark_read_inbox_item`, `promote_inbox_to_task`, `reply_inbox_item`, `set_project_tracked`, `snooze_inbox_item` (alphabetised, plus the unchanged entries from the merge base).
- `tests/web_api/test_openapi_conformance.py`: kept ALL test functions — #2060's `test_static_yaml_does_not_advertise_idempotency_on_inbox_writes`, `test_static_yaml_declares_503_on_inbox_write_paths`, `test_inbox_archive_reason_documented_as_post_transition`, AND #2063's `test_archive_documents_409_session_reference_conflict`.

**No additional conflicts** beyond those 3 — the remaining 9 cherry-picks applied cleanly. No code semantics changed; pure rebase.

**Tests (single -x run, --timeout=180):**
- `tests/test_projects_writes_endpoint.py` → 39 passed
- `tests/web_api/test_openapi_conformance.py` → 9 passed
- `tests/test_project_rename.py` → 11 passed
- `tests/test_cockpit_project_state.py` → 21 passed
- `tests/test_cockpit_project_health.py` → 3 passed

Total: **83 passed in 133.86s**. `needs-review` re-applied.

samhotchkiss:

The previous config-write blockers are mostly addressed: the project mutation helpers now use the shared RMW lock, pause/resume route idempotency delegates into the service layer, archive preserves the session->project invariant, and the targeted endpoint/OpenAPI suites pass. I still cannot merge because round 9 introduced a cache-consistency bug in the `load_config()` auth-token migration path.

`src/pollypm/config.py:998-1045` parses `config` before taking `config_rmw_lock()`. If `ensure_session_auth_tokens(config)` mints a token, the code re-parses `fresh` under the lock and writes `fresh` so concurrent disk edits are preserved (`:1018-1024`). But after that it copies only `sessions[*].auth_token` back into the original pre-lock `config` (`:1029-1032`) and then stores that original object in `_config_cache` under the post-write mtime (`:1041-1045`). If any config edit lands between the first parse and the locked re-parse, disk keeps the edit, but the process cache now serves the stale pre-lock object as if it matched the latest mtime. Subsequent `load_config(path)` calls return stale config indefinitely, which violates the durable-config/source-of-truth invariant this PR is tightening.

Please cache/return the post-lock snapshot after the auth-token write, not the pre-lock snapshot with only token fields patched. A small fix would replace or fully refresh `config` from `fresh` before populating `_config_cache` (and compute `_project_local_mtimes()` from the same object that is cached). Add a regression where a concurrent edit changes a non-token field between the initial parse and the locked re-parse, and verify the next `load_config()` sees that field.

One-off reproduction I ran on this head:
- Seed a config with one legacy session missing `auth_token` and project `persona_name='Old'`.
- Monkeypatch `ensure_session_auth_tokens()` so its first call writes `persona_name='New'` to disk before returning.
- Call `load_config(path)`.
- Result: disk parses as `persona_name='New'`, but `load_config(path)` returns cached `persona_name='Old'` (`again is loaded` is `True`) while the session token is present.

Verification run:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py --timeout=180` -> 39 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py tests/test_project_rename.py tests/test_cockpit_project_state.py tests/test_cockpit_project_health.py --timeout=180` -> 44 passed
- `uv run --extra test --with pytest-timeout pytest tests/web_api/test_endpoints.py -k project --timeout=120` -> 10 passed, 20 deselected
- Broader sweep `tests/test_config.py tests/test_projects.py tests/test_accounts.py tests/test_workers.py tests/test_config_endpoint.py` -> 92 passed, 11 failed; sampled failures reproduce on `origin/main`, so I did not treat those as PR-introduced blockers.

samhotchkiss:
Round 10 fix pushed. New head: `a74bdf898`.

**Fix** (`src/pollypm/config.py:1014-1048`): inside the locked auto-persist branch, after `write_config(fresh, ...)` returns, swap `config = fresh` so both the return value and the `_config_cache` entry (mtime + `_project_local_mtimes`) come from the post-lock snapshot. Round 9's "copy `sessions[*].auth_token` back into the pre-lock `config`, then cache the pre-lock object under the post-write mtime" path is gone — `fresh` already carries the freshly-minted tokens AND any concurrent disk edits, so it's the canonical post-write state.

**Regression**: `tests/test_config.py::test_load_config_auto_persist_returns_post_lock_snapshot` reproduces your exact scenario — patched `ensure_session_auth_tokens` rewrites `persona_name='Old'`→`'New'` on disk during its first call; both the initial `load_config` and a follow-up call must return `persona_name='New'` plus a minted token. Verified the test FAILS on round-9 head `86167c559` (`AssertionError: assert 'Old' == 'New'`) and PASSES on `a74bdf898`.

**Tests** (`uv run --extra dev --with pytest-timeout pytest tests/test_config.py tests/test_projects_writes_endpoint.py tests/web_api/test_openapi_conformance.py --timeout=180`): **72 passed, 1 failed**. The single failure is `test_example_config_documents_storage_block` (sqlite-vs-postgres default) — pre-existing on round-9 head without my changes, matches the "broader sweep failures reproduce on origin/main" caveat from your last comment, not introduced by this fix.

`needs-review` re-applied.

samhotchkiss:

The earlier project-write blockers are addressed in the merge tree: pause/resume delegate through the service helper, the stale/live config refresh cases have coverage, archive uses the shared removal facade, and the `reason` field is now documented and emitted through audit on real mutations. The remaining blocker is in the broadened config RMW changes outside the new routes.

Blocking concurrent-agent bug: `src/pollypm/workers.py:create_worker_session()` still makes the duplicate role/session decision outside the commit lock. It checks for an existing enabled session and computes `session_key` before `config_rmw_lock` (`workers.py:232-240`), then after `ensure_worktree()` it reloads the config and blindly assigns `fresh.sessions[session_key] = worker` under the lock (`workers.py:274-277`). Two concurrent `pm worker-start` calls for the same project/role can both pass the invariant check, both return success with the same session key, and the final TOML contains only one session. That leaves one launched agent/worktree orphaned or overwritten; if cleanup later runs against the same-key worktree, it can also delete live agent state.

I reproduced this on the actual merge tree with a `Barrier` in a patched `ensure_worktree()` so both threads pass the pre-lock check before either commits:

```text
results ['worker_pollypm', 'worker_pollypm']
errors []
worker_sessions ['worker_pollypm']
```

Please make the worker-session uniqueness decision part of the same critical section that commits the session. Re-load the fresh config under `config_rmw_lock`, re-check the enabled role/project invariant and explicit `session_name` collisions, then either allocate a fresh unique key before creating the worktree or fail safely without deleting another session's worktree. Add a threaded regression around this race; the expected outcome should be either one success plus one clear error, or two distinct session keys/worktrees if multiple concurrent workers for the same role/project are intentionally allowed.

Verification I ran on the merge result:
- `uv run --extra test --with pytest-timeout pytest --noconftest tests/test_projects_writes_endpoint.py -q --timeout=120` -> 39 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 9 passed
- `uv run --extra test --with pytest-timeout pytest tests/web_api/test_endpoints.py -k project -q --timeout=120` -> 10 passed, 20 deselected
- `uv run --extra test --with pytest-timeout pytest tests/test_workers.py tests/test_projects.py tests/test_config.py -q --timeout=120` -> 9 failed, 61 passed; I spot-checked the same failures on current `origin/main` and reproduced them there, so I am not treating those existing failures as new blockers for this PR.

samhotchkiss:
Round 11 fix pushed. New head: `e1fe3003c`.

**Fix** (`src/pollypm/workers.py:229-309`): chose **Option A** — pulled the duplicate-role check, explicit-`session_name` collision check, `session_key` allocation, `ensure_worktree` call, AND the `write_config` commit all inside the same `config_rmw_lock(config_path)` critical section. Round 7 had moved only the commit inside the lock; the uniqueness decision still ran against the pre-lock snapshot, so two concurrent `pm worker-start` calls for the same project/role could both pass the check and both compute the same `worker_<project>` key. Now the loser observes the winner's session in the locked `fresh` snapshot and raises `typer.BadParameter` BEFORE touching the filesystem — no orphan worktree, no clobbered write. `ensure_worktree`'s `git worktree add` subprocess runs under the lock, bounded by its existing 300s timeout but typically <1s, which is acceptable for an operator-cadence action that already serialised the commit in round 7. The `typer.BadParameter` re-raise branch explicitly skips the worktree-cleanup `shutil.rmtree` path so a duplicate loser never deletes a sibling session's worktree.

**Regression**: `tests/test_workers.py::test_concurrent_create_worker_session_for_same_role_no_orphans` mirrors your exact `threading.Barrier(2)` reproduction — patched `ensure_worktree` blocks on the barrier so both threads pass the pre-lock work simultaneously, then the test asserts (a) one success + one clear duplicate error OR (b) two distinct keys; never two successes with the same key; never a successful key missing from the persisted config. Verified the test FAILS on round-10 head `a74bdf898` with `AssertionError: Both threads returned the same session_key 'worker_pollypm'` (the exact `['worker_pollypm', 'worker_pollypm']` outcome you reported) and PASSES on `e1fe3003c`.

**Tests** (`uv run --extra dev --with pytest-timeout pytest tests/test_workers.py tests/test_projects_writes_endpoint.py tests/web_api/test_openapi_conformance.py --timeout=180`): the projects-writes + OpenAPI conformance suites are 48 passed clean. `test_workers.py` is 9 passed (including the new regression) + 5 failed; I verified the same 5 failures reproduce on round-10 head with my changes stashed (default-args drift, pre-existing on `origin/main`, matches your "broader sweep failures reproduce on origin/main" caveat).

`needs-review` re-applied.

samhotchkiss:

The earlier API project-write blockers look addressed on the current head, and the PR-specific endpoint/openapi tests pass on the actual local merge result with current `origin/main`. I found two remaining blockers in the RMW changes outside the API routes.

1. `src/pollypm/workers.py:create_worker_session()` still builds the committed `SessionConfig` from the pre-lock config snapshot. Lines 124-230 resolve `project`, role routing, `account_name`, `account`, `active_provider`, `active_model`, and sometimes `prompt` before the `config_rmw_lock`. The locked block at lines 250-295 reloads `fresh`, but only re-checks duplicate sessions / explicit session-name collisions before writing `fresh.sessions[session_key] = worker`. If an account or project edit lands between the pre-lock checks and the locked write, this can persist a session that references an account no longer present in `fresh.accounts`, violating the existing `session.account -> accounts` config invariant enforced by `load_config()`.

I reproduced this by patching `suggest_worker_prompt()` to remove the selected worker account from disk after the pre-lock account validation, patching `ensure_worktree()` to return `None`, then calling `create_worker_session(..., account_name='worker', prompt=None)`. The function returned a worker session, but a fresh `load_config(config_path)` failed with:

```text
ValueError: Session 'worker_demo' references unknown account 'worker'. Known accounts: controller
```

Please move all config-derived validation and session construction that affects the written config inside the locked fresh snapshot, or revalidate inside the lock before commit: project still exists, account still exists and still matches the requested/provider route, role/model assignment is recomputed against `fresh`, and the final `SessionConfig` is built only from `fresh`. Add a regression where an account is removed after initial selection but before the locked commit; the worker start should fail without writing an invalid config or touching a worktree.

2. `src/pollypm/projects.py:scan_projects()` now writes the new project entries before scaffolding them. The old flow called `ensure_project_scaffold(repo_path)` before the final `write_config()`. The new flow writes inside the lock at lines 914-944, releases the lock, and only then scaffolds at lines 946-950. If scaffolding fails (permissions, partial checkout, bad path), the command raises but leaves the global config pointing at a project whose `.pollypm/` scaffold was never created. I reproduced this by patching `discover_recent_git_repositories()` to return one repo and `ensure_project_scaffold()` to raise `PermissionError`; after the exception, `load_config(config_path).projects` still contained the new project key.

Please keep the config commit and scaffold failure semantics atomic from the operator’s perspective. One workable shape is: do discovery/confirmation outside the lock, do idempotent scaffolding before the config write (outside the lock if needed), then take the lock, reload, skip candidates already added, and write only successfully scaffolded candidates. Add a regression that a scaffold failure does not persist the project.

Verification run on the local merge result (`origin/main` merged into PR head):
- `uv run --extra test --with pytest-timeout pytest tests/test_projects_writes_endpoint.py -q --timeout=120` -> 39 passed
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 9 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_import_boundary.py tests/web_api/test_openapi_conformance.py -q --timeout=120` -> 25 passed
- `uv run --extra test --with pytest-timeout pytest tests/test_projects_writes_endpoint.py::test_scan_projects_race_with_set_project_tracked_preserves_both tests/test_projects_writes_endpoint.py::test_write_example_config_race_serialized_by_lock tests/test_workers.py::test_concurrent_create_worker_session_for_same_role_no_orphans -q --timeout=120` -> 3 passed
- Broader `tests/test_config.py` / `tests/test_workers.py` still have unrelated failures, but I confirmed the selected failures reproduce on `origin/main` as well, so I am not counting those inherited failures against this PR.

samhotchkiss:
Round 12 fix pushed. New head: `e104e16bd`.

**Blocker 1 fix** (`src/pollypm/workers.py:113-340`): pre-lock work is now DISCOVERY ONLY — an early project-existence check (for a friendlier error than the locked block would emit) and an empty-prompt suggestion. Every field that ends up in the committed `SessionConfig` is re-derived INSIDE `config_rmw_lock` from `fresh`: project lookup (re-validated at lines 198-211), role routing (re-resolved at lines 215-228), account selection / auto-select (lines 230-265), `account.provider` (line 273), `active_provider` / `active_model` (lines 273-318), and final `prompt` (line 339). The locked validation now rejects unknown accounts (lines 267-275) with `typer.BadParameter` BEFORE writing — so the load_config invariant guard can never see a session referencing a missing account, even if the account was removed between pre-lock discovery and lock acquire. Pre-lock work retained: project existence (UX), `suggest_worker_prompt` (currently returns empty-string but called pre-lock so the suggester doesn't re-read inside the critical section).

**Blocker 2 fix** (`src/pollypm/projects.py:911-957`): scaffolding now runs OUTSIDE the lock BEFORE the locked write (lines 932-941). Candidates whose scaffold raises `OSError` are skipped with a `typer.echo(..., err=True)` and never reach the config write. The locked block (lines 945-956) iterates only `scaffolded` candidates. `ensure_project_scaffold` is idempotent (`mkdir(exist_ok=True)`) so running it before commit is safe — the worst case is an unreferenced `.pollypm/` dir if commit later fails, strictly better than a config entry whose scaffold never landed.

**Regressions**:
- `tests/test_workers.py::test_create_worker_session_revalidates_account_under_lock` — mirrors your exact reproduction. Patches `suggest_worker_prompt` to drop the requested account from disk before returning, calls `create_worker_session(account_name='claude_worker', prompt=None)`, asserts `typer.BadParameter` raised AND `load_config(config_path)` succeeds (no invalid session). Verified FAILS on round-11 head (succeeds + persists invalid session) and PASSES on `e104e16bd`.
- `tests/test_projects_writes_endpoint.py::test_scan_projects_scaffold_failure_does_not_persist` — your exact reproduction. Patches `discover_recent_git_repositories` to return one repo, `ensure_project_scaffold` to raise `PermissionError`, calls `scan_projects(...)`, asserts `load_config(path).projects == {}`. Verified FAILS on round-11 head (raises PermissionError AFTER the locked write, leaving the new key persisted) and PASSES on `e104e16bd`.

**Tests** (`uv run --extra dev --with pytest-timeout pytest tests/test_workers.py tests/test_projects_writes_endpoint.py tests/web_api/test_openapi_conformance.py --timeout=180`): 59 passed + 5 pre-existing failures in `test_workers.py` (`StateStore` import drift + a default-args change adding `--model claude-opus-4-7`). Verified these same 5 reproduce on round-11 head `e1fe3003c` with my changes stashed — matches your round-11 "broader sweep failures reproduce on origin/main" caveat. Projects-writes (41 passed including the new one) + openapi-conformance (9 passed) are clean.

`needs-review` re-applied.

PR #2066, opened 2026-05-22
feat(web-ui): Playwright test scaffold for v0 web UI

samhotchkiss:
## Summary

Phase 8 of the 8-step web-UI pipeline: Playwright end-to-end test scaffold for the v0 web UI shipped in PR #2065.

- New `tests/playwright/` directory with config, README, and 5 test files (23 specs total) covering cookie auth, surface rail rendering/selection, send-message flow, edge cases (409 unsafe_mid_tool, 503 service_unavailable, empty states), and keyboard handling.
- Chromium-only, headless by default; screenshots on failure and video on retry.
- Send/edge-case specs route-mock the API so the suite never delivers real keystrokes to a tmux pane — safe to run while PollyPM is in use.
- README documents prereqs (`pm serve` on `127.0.0.1:8765`), how to run (`npm install && npx playwright install chromium && npx playwright test`), and which specs hit the live daemon vs. stub.

## Files

- `tests/playwright/package.json` + `package-lock.json` (npm scripts: `test`, `test:headed`, `test:ui`, `test:debug`, `report`)
- `tests/playwright/playwright.config.ts` (baseURL via `POLLYPM_BASE_URL`, 30s timeout, retries on CI, commented webServer block)
- `tests/playwright/README.md`
- `tests/playwright/.gitignore` (node_modules, reports, results)
- `tests/playwright/tests/auth.spec.ts` (5 tests)
- `tests/playwright/tests/surfaces.spec.ts` (5 tests)
- `tests/playwright/tests/send_message.spec.ts` (4 tests)
- `tests/playwright/tests/edge_cases.spec.ts` (5 tests)
- `tests/playwright/tests/keyboard.spec.ts` (4 tests, 2 marked `test.fixme` as V1-UI TODOs)

## Verification

- Ran `npm install` (3 packages, 0 vulnerabilities) and `npx playwright test --list` — all 23 tests discovered and parsed cleanly.
- Did NOT run the suite end-to-end; Sam runs it in the morning against a live daemon.

## Test plan

- [ ] `cd tests/playwright && npm install && npx playwright install chromium`
- [ ] Start `pm serve --tailscale` in another terminal
- [ ] `npx playwright test` — expect green for auth/surfaces (live data), all stubbed specs green
- [ ] `npm run test:ui` for the recommended interactive workflow
- [ ] Confirm `Shift+Enter` and `j/k` `test.fixme` markers stay skipped (V0 limitations)

Generated with Claude Code.

### DIFF
--- a/tests/playwright/README.md
+++ b/tests/playwright/README.md
@@ -3,11 +3,36 @@
 End-to-end tests for the v0 web UI shipped in PR #2065
 (`src/pollypm/web_api/ui/`). Covers:
 
-- cookie auth (`pollypm-session` set by `GET /ui/`)
+- cookie auth (`pollypm-session` minted by `GET /ui/` for trusted
+  callers only — loopback, verified tailnet peers, or a valid bearer
+  header — verified positively AND with negative cookie-isolation case)
 - surface rail rendering + selection
 - send-message flow against `POST /api/v1/chat/{session}/send`
 - edge cases (409 unsafe_mid_tool, 503 service_unavailable, empty list)
 - keyboard handling (Enter to send, plus TODO probes for the V1 UI)
+- mobile viewport smoke checks (360x800) — phone-over-Tailscale is an
+  advertised access mode for the v0 UI
+
+## Auth model under test
+
+Per `src/pollypm/web_api/auth.py` (merged in #2065), the API accepts
+three credential modes:
+
+1. `Authorization: Bearer <token>` header — any client.
+2. `pollypm-session` cookie — minted by `GET /ui/` for browsers.
+3. Tailscale CGNAT peer (`100.64.0.0/10`) — only when the daemon was
+   built with `tailnet_trust_enabled=True` (i.e. `pm serve` bound to a
+   verified Tailscale interface).
+
+`GET /ui/` is itself gated: the `Set-Cookie` header is only emitted
+for loopback callers, verified tailnet peers, or callers that present
+a valid bearer token. Untrusted callers still receive the HTML but no
+cookie — the SPA then surfaces a 401 on its first `/api/` call.
+
+`tests/auth.spec.ts` exercises the loopback-bootstrap path positively
+(Set-Cookie present, cookie HttpOnly, SameSite=Lax) and the negative
+cookie-isolation invariant (a fresh `APIRequestContext` with no cookie
+gets a hard 401, never 200/403).
 
 ## Prereqs
 
@@ -64,14 +89,28 @@ POLLYPM_BASE_URL=http://100.x.y.z:8765 npx playwright test
 This means the suite is safe to run while you're using PollyPM yourself:
 the only requests that touch live state are read-only `GET` calls.
 
+## Projects
+
+- `chromium` — Desktop Chrome, default development viewport.
+- `mobile-chrome` — Pixel 5 user-agent forced to a 360x800 viewport
+  (typical narrow Android portrait). Run with
+  `npx playwright test --project=mobile-chrome` to target it
+  specifically; the default `npx playwright test` runs both.
+
 ## Known limitations / TODOs
 
 - `Shift+Enter` newline and `j/k` surface navigation are marked
   `test.fixme` — V0 input is `<input type=text>` and the rail has no
   keyboard nav. Specs stay in place so the V1 UI work picks them up.
-- No mobile viewport project yet; V0 layout is desktop-only.
 - No Firefox/WebKit projects yet; per the V0 PR scope we ship Chromium
   only.
+- Non-loopback negative auth case (e.g. a LAN device hitting `/ui/`
+  and getting NO Set-Cookie) is verified at the FastAPI layer in
+  `tests/web_api/test_web_ui.py` (search for `_no_cookie_from_lan`,
+  `_no_cookie_minted_from_cgnat_when_trust_disabled`,
+  `_no_cookie_with_invalid_bearer`); reproducing it from Playwright
+  would require spoofing TCP source IP, which isn't worth the harness
+  complexity for the same invariant.
 
 ## Debugging tips
 
--- a/tests/playwright/README.md
+++ b/tests/playwright/README.md
@@ -81,13 +81,19 @@ POLLYPM_BASE_URL=http://100.x.y.z:8765 npx playwright test
 | Spec | Hits real daemon | Notes |
 |------|------------------|-------|
 | `auth.spec.ts` | yes | Verifies the real cookie handshake. |
-| `surfaces.spec.ts` | partial | Real `/sessions` for live data, route stubs for the empty-state case. |
-| `send_message.spec.ts` | no (route-stubbed) | We never actually send keystrokes to a real tmux pane. |
-| `edge_cases.spec.ts` | no (route-stubbed) | 409/503 responses are simulated. |
-| `keyboard.spec.ts` | no (route-stubbed) | |
+| `surfaces.spec.ts` | partial | Real `/sessions` + `/dashboard` for live data; route stubs for the empty-state, 500, and stubbed-messages cases. |
+| `send_message.spec.ts` | no (route-stubbed) | Stubs `/api/v1/dashboard`, `/api/v1/chat/sessions`, `/messages*`, and `/send`. We never deliver keystrokes to a real tmux pane. |
+| `edge_cases.spec.ts` | no (route-stubbed) | Stubs `/api/v1/dashboard` + `/api/v1/chat/*`. 409/503 responses are simulated. |
+| `keyboard.spec.ts` | no (route-stubbed) | Stubs `/api/v1/dashboard` + `/api/v1/chat/*`. |
 
-This means the suite is safe to run while you're using PollyPM yourself:
-the only requests that touch live state are read-only `GET` calls.
+Concretely, the "no (route-stubbed)" specs install a `beforeEach`
+that mocks `/api/v1/dashboard` (the UI fires this on init for the
+right-rail rollups) plus all `/api/v1/chat/*` traffic. They still
+load the real `/ui/` HTML + static assets and go through the cookie
+bootstrap, but no mutating endpoint and no live `/chat/sessions` /
+`/dashboard` read leaks into the daemon. That makes the suite safe
+to run while you're using PollyPM yourself, and removes the
+fullyParallel timing coupling to live dashboard/session work.
 
 ## Projects
 
--- a/tests/playwright/README.md
+++ b/tests/playwright/README.md
@@ -1,0 +1,83 @@
+# PollyPM v0 Web UI — Playwright Tests
+
+End-to-end tests for the v0 web UI shipped in PR #2065
+(`src/pollypm/web_api/ui/`). Covers:
+
+- cookie auth (`pollypm-session` set by `GET /ui/`)
+- surface rail rendering + selection
+- send-message flow against `POST /api/v1/chat/{session}/send`
+- edge cases (409 unsafe_mid_tool, 503 service_unavailable, empty list)
+- keyboard handling (Enter to send, plus TODO probes for the V1 UI)
+
+## Prereqs
+
+1. Node 18+ and npm.
+2. `pm serve` (or `pm up`) running on `http://127.0.0.1:8765` with a
+   valid `~/.pollypm/api-token` so the `GET /ui/` cookie endpoint can
+   issue `pollypm-session`.
+
+## Install
+
+```bash
+cd tests/playwright
+npm install
+npx playwright install chromium
+```
+
+## Run
+
+In one terminal (the daemon):
+
+```bash
+pm serve --tailscale
+# or just `pm up` if you already have that running
+```
+
+In another terminal (the tests):
+
+```bash
+cd tests/playwright
+npx playwright test                # headless, all specs
+npx playwright test auth.spec.ts   # one spec
+npm run test:headed                # show the browser
+npm run test:ui                    # Playwright UI mode (recommended)
+npm run test:debug                 # step-through debugger
+npm run report                     # open the last HTML report
+```
+
+To point the suite at a different host (e.g. a Tailscale URL):
+
+```bash
+POLLYPM_BASE_URL=http://100.x.y.z:8765 npx playwright test
+```
+
+## What gets stubbed vs. real
+
+| Spec | Hits real daemon | Notes |
+|------|------------------|-------|
+| `auth.spec.ts` | yes | Verifies the real cookie handshake. |
+| `surfaces.spec.ts` | partial | Real `/sessions` for live data, route stubs for the empty-state case. |
+| `send_message.spec.ts` | no (route-stubbed) | We never actually send keystrokes to a real tmux pane. |
+| `edge_cases.spec.ts` | no (route-stubbed) | 409/503 responses are simulated. |
+| `keyboard.spec.ts` | no (route-stubbed) | |
+
+This means the suite is safe to run while you're using PollyPM yourself:
+the only requests that touch live state are read-only `GET` calls.
+
+## Known limitations / TODOs
+
+- `Shift+Enter` newline and `j/k` surface navigation are marked
+  `test.fixme` — V0 input is `<input type=text>` and the rail has no
+  keyboard nav. Specs stay in place so the V1 UI work picks them up.
+- No mobile viewport project yet; V0 layout is desktop-only.
+- No Firefox/WebKit projects yet; per the V0 PR scope we ship Chromium
+  only.
+
+## Debugging tips
+
+- Run `npx playwright test --ui` for the recommended workflow.
+- Failures dump screenshots/videos under `playwright-report/`.
+- Set `DEBUG=pw:api` to see Playwright client traffic.
+- If `conn-status` won't go green, the token cookie isn't reaching the
+  browser — check that `pm serve` is on 8765 and that
+  `~/.pollypm/api-token` exists.
--- a/tests/playwright/tests/auth.spec.ts
+++ b/tests/playwright/tests/auth.spec.ts
@@ -3,10 +3,24 @@ import { test, expect } from "@playwright/test";
 /**
  * Auth scenarios.
  *
- * Auth model (per /ui/app.js header comment):
- *   GET /ui/ reads the on-disk token and sets the `pollypm-session`
- *   cookie. Every subsequent fetch uses credentials: "include" to ride
- *   that cookie. We never expose the token to the browser.
+ * Auth model (per src/pollypm/web_api/auth.py and #2065 security spec):
+ *   GET /ui/ mints the `pollypm-session` cookie ONLY for:
+ *     1. Authorization: Bearer <token> matching the on-disk token, OR
+ *     2. Loopback client (127.0.0.1, ::1), OR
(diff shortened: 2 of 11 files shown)


tests/playwright/README.md
```diff
@@ -0,0 +1,83 @@
+# PollyPM v0 Web UI — Playwright Tests
+
+End-to-end tests for the v0 web UI shipped in PR #2065
```
samhotchkiss, line 3: Required changes before merge. This PR explicitly tests the v0 web UI from #2065, but it targets `main` and #2065 is not merged. Do not merge this child PR until the base UI PR is merged, and then rebase/retarget this branch onto the final base so the test scaffold reflects the actual UI/auth code that landed.

tests/playwright/tests/auth.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+test.describe("auth", () => {
+  test("GET /ui/ loads and serves index.html", async ({ page }) => {
+    const response = await page.goto("/ui/");
+    expect(response, "navigation response").not.toBeNull();
+    expect(response!.status(), "status").toBe(200);
+    await expect(page).toHaveTitle(/PollyPM/);
+    await expect(page.locator("#topbar .brand-name")).toHaveText("PollyPM");
+  });
+
+  test("session cookie is set after visiting /ui/", async ({ page, context }) => {
```
samhotchkiss: Required changes before merge. This test locks in the unsafe #2065 behavior where any visit to unauthenticated `/ui/` mints a `pollypm-session` from the on-disk bearer token. Once #2065 fixes the credential-issuance flow, this needs to assert the safe bootstrap behavior instead, including a negative case for non-local/non-trusted clients.

tests/playwright/tests/auth.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    expect([401, 403]).toContain(resp.status());
+  });
+
+  test("direct API call shape after /ui/ loads cookie", async ({ page, request }) => {
+    await page.goto("/ui/");
+    // After visiting /ui/, the request fixture should NOT share that
+    // cookie (it's a separate context); this asserts the cookie is
+    // scoped to the browser context, not the test runner.
+    const resp = await request.get("/api/v1/chat/sessions");
+    expect([200, 401, 403]).toContain(resp.status());
```
samhotchkiss: Required changes before merge. This assertion accepts 200, 401, or 403, so it cannot catch cookie isolation or auth regressions. The comment says the isolated `request` fixture should not share the page's cookie; make that a precise invariant and assert the expected status/body. Accepting all outcomes here would let the auth boundary drift unnoticed.

tests/playwright/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+This means the suite is safe to run while you're using PollyPM yourself:
+the only requests that touch live state are read-only `GET` calls.
+
+## Known limitations / TODOs
+
+- `Shift+Enter` newline and `j/k` surface navigation are marked
+  `test.fixme` — V0 input is `<input type=text>` and the rail has no
+  keyboard nav. Specs stay in place so the V1 UI work picks them up.
+- No mobile viewport project yet; V0 layout is desktop-only.
```
samhotchkiss: Required changes before merge. The parent UI PR advertises phone access over Tailscale, but this scaffold documents that there is no mobile viewport project. Given the fixed-rail layout issue in #2065, the Playwright scaffold should include at least one mobile viewport/screenshot check before this becomes the coverage baseline for the web UI.

samhotchkiss (commented):
Required changes before merge. `npm install` completed and `npx playwright test --list` discovers 23 specs, but this is stacked on the blocked #2065 web UI/auth PR and some tests currently preserve the auth/layout gaps instead of catching them. Details inline.

tests/playwright/README.md
samhotchkiss: Addressed in 3ed6ac2fa: branch rebased onto main (which now includes the merged #2065 web UI/auth). Conflict-free rebase since Playwright lives under tests/playwright/. README updated to describe the merged three-mode auth model (bearer / cookie / tailnet-trust) and to point at tests/web_api/test_web_ui.py for the LAN/CGNAT negative cases. PR base is already main.

tests/playwright/tests/auth.spec.ts
samhotchkiss: Fixed in 3ed6ac2fa. The merged #2065 gates /ui/ Set-Cookie on loopback OR verified-tailnet OR valid-bearer, so the test now asserts the SAFE bootstrap: loopback caller gets a cookie that is HttpOnly + SameSite=Lax, AND the Set-Cookie header is explicitly present. The non-local negative case (LAN device hits /ui/ → HTML but NO Set-Cookie) is covered at the FastAPI layer in tests/web_api/test_web_ui.py (test_ui_no_cookie_from_lan_client, test_ui_no_cookie_minted_from_cgnat_when_trust_disabled, test_ui_no_cookie_with_invalid_bearer) — spoofing TCP source IP from Playwright would add harness complexity for an invariant the Python suite already pins. README now notes this division of labor.

tests/playwright/tests/auth.spec.ts
samhotchkiss: Tightened in 3ed6ac2fa. Replaced the loose [200,401,403] assertion with a precise cookie-isolation invariant: a fresh playwright.request.newContext() (no cookies, no bearer) MUST return 401. The merged auth.py raises unauthorized() for missing credentials, so 401 is the only correct status — 200 would mean cookie leakage across contexts, 403 would mean an auth-middleware shape change worth catching. Also added a positive cookie-rider test using page.request (same browser context, shared cookie jar) that asserts 200 + the {sessions: [...]} envelope. The auth boundary can no longer drift unnoticed.

tests/playwright/README.md
samhotchkiss: Added in 3ed6ac2fa. New mobile-chrome project pinned to 360x800 (tighter than Pixel 5's default 393x851 to also catch overflow on smaller Android viewports), and tests/playwright/tests/mobile_viewport.spec.ts smoke-covers core chrome rendering + a horizontal-overflow guard (document.scrollWidth - clientWidth ≤ 1px). All 27 specs now run under both chromium and mobile-chrome (54 tests total, confirmed via npx playwright test --list). README documents the new project + how to target it.

tests/playwright/tests/surfaces.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  test("left rail renders surface list after load", async ({ page }) => {
+    await page.goto("/ui/");
+    const list = page.locator("#surface-list");
+    await expect(list).toBeVisible();
+    // Either real surfaces render as <li data-session=...> or the
+    // empty-state literal appears.
+    await expect
+      .poll(async () => {
+        const items = await list.locator("li[data-session]").count();
+        const emptyText = await list.locator(".surface-empty").innerText().catch(() => "");
```
samhotchkiss: Blocking: this poll times out whenever the daemon actually has surfaces. `items` becomes > 0, but then `list.locator(".surface-empty").innerText()` waits for the missing empty-state element until `actionTimeout` before the callback can return. In my run the failure artifact showed 26 rendered `li data-session` rows, yet the poll still timed out. Short-circuit before reading `.surface-empty`, or check `.surface-empty` count/visibility with a zero/short timeout. This must pass on both the populated and empty daemon states.

tests/playwright/tests/surfaces.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+      }, { timeout: 10_000 })
+      .toBe(true);
+  });
+
+  test("clicking a surface populates the center pane", async ({ page }) => {
+    await page.goto("/ui/");
+    const list = page.locator("#surface-list");
+    await expect(list).toBeVisible();
+    const items = list.locator("li[data-session]");
+    const count = await items.count();
```
samhotchkiss, line 58: Blocking: this immediate count races the async `/chat/sessions` load. In the same verification run, this test and `each surface type can be selected if present` skipped even though the page later rendered many surfaces. Wait for the rail to reach a terminal state (items, `no surfaces registered`, or `error:`) before deciding whether to skip; otherwise the scaffold reports green while not exercising the live surface-selection contract. Apply the same fix around line 45.

tests/playwright/README.md
```diff
@@ -0,0 +1,122 @@
...
+POLLYPM_BASE_URL=http://100.x.y.z:8765 npx playwright test
+```
+
+## What gets stubbed vs. real
+
+| Spec | Hits real daemon | Notes |
+|------|------------------|-------|
+| `auth.spec.ts` | yes | Verifies the real cookie handshake. |
+| `surfaces.spec.ts` | partial | Real `/sessions` for live data, route stubs for the empty-state case. |
+| `send_message.spec.ts` | no (route-stubbed) | We never actually send keystrokes to a real tmux pane. |
```
samhotchkiss: This table is inaccurate as written. The supposedly fully route-stubbed specs still load the real `/ui/` document/assets, and app init also calls `/api/v1/dashboard` unless the tests stub it. Either stub the dashboard/read endpoints in these specs so the claim is true, or document that these specs still hit live read-only daemon endpoints. The distinction matters because live dashboard/session work can affect timing and connection-badge state under `fullyParallel` runs.

samhotchkiss (commented):
Review decision: do not merge yet. I verified the prior stacked-PR/auth/mobile comments against the current code, and those are materially addressed now that #2065 is merged. However the new Playwright scaffold is not clean because it fails against a live daemon with registered surfaces and silently skips two of the live surface coverage paths.

Verification run:
- `npm ci` passed (0 vulnerabilities).
- `npx playwright test --list` discovered 54 tests across `chromium` and `mobile-chrome`.
- `uv run --extra test pytest tests/web_api/test_web_ui.py -q` passed: 42 tests.
- Started a sandboxed FastAPI app on `127.0.0.1:8766` from this PR head and ran `POLLYPM_BASE_URL=http://127.0.0.1:8766 npx playwright test`: 44 passed, 8 skipped, 2 failed. The failures are both `surfaces.spec.ts` / `left rail renders surface list after load` for desktop and mobile. A focused rerun of that single chromium test with `--workers=1` also failed.

This PR touches the web UI test-contract invariants for cookie auth, surface discovery/rendering, send safety, and mobile layout. The auth and mobile additions look directionally right, but the surface-discovery invariant is currently unenforced on the exact case it is meant to cover: a real daemon with surfaces. GitHub will not let me submit this as a formal Request changes review because I am authenticated as the PR author, so treating these comments as blocking review feedback.

samhotchkiss:
Review decision: do not merge yet.

Current head is still `3ed6ac2f`, so the unresolved blocking feedback is still present in code. This PR touches the web UI E2E test-contract invariants for surface discovery/rendering, safe send stubbing, cookie auth, and mobile layout. The auth/mobile direction looks reasonable, but the surface discovery contract is still broken on the populated-daemon path and the docs still overstate which specs are isolated from live daemon reads.

Blocking issues:

1. `tests/playwright/tests/surfaces.spec.ts:17-23` still times out when surfaces exist. The poll counts `li[data-session]`, but it always awaits `.surface-empty.innerText()` before returning. When the daemon has surfaces, `.surface-empty` is absent, so Playwright waits for `actionTimeout` and the 10s poll expires even though the rail rendered correctly. Short-circuit before reading `.surface-empty`, or check empty-state count/visibility with a zero/short timeout. This must pass with both populated and empty daemons.

2. `tests/playwright/tests/surfaces.spec.ts:30-46` still decides whether to skip from an immediate `count()` / `all()` right after `#surface-list` becomes visible. That races the async `/api/v1/chat/sessions` load and can skip the live selection coverage while the page is still loading. Wait for a terminal rail state first: at least one `li[data-session]`, `no surfaces registered`, or `error:`.

3. `tests/playwright/README.md:79-90` still says `send_message.spec.ts`, `edge_cases.spec.ts`, and `keyboard.spec.ts` are `no (route-stubbed)`. They still load the real `/ui/` document/assets, and app init calls `/api/v1/dashboard` unless those specs stub it. Either stub the dashboard/read endpoints so the claim is true, or document that these specs still hit live read-only daemon endpoints. This matters under `fullyParallel` because live dashboard/session timing can affect status text and connection state.

Verification run for this review:

- `npm ci` passed in `tests/playwright` with 0 vulnerabilities.
- `npx playwright test --list` discovered 54 tests across `chromium` and `mobile-chrome`.
- `uv run --extra test pytest tests/web_api/test_web_ui.py -q` passed: 42 tests.
- No daemon was listening at the default `http://127.0.0.1:8765`.
- Started a local stub server on `127.0.0.1:8766` serving the real UI assets and a single fake surface, then ran `POLLYPM_BASE_URL=http://127.0.0.1:8766 npx playwright test tests/surfaces.spec.ts --project=chromium --workers=1`: 1 failed, 4 passed. The failure was `surfaces › left rail renders surface list after load`, timing out at `surfaces.spec.ts:23` while waiting on the poll predicate.

Because this scaffold is intended to become the regression baseline for the v0 web UI, I would not merge it until the populated-surface path is actually green and the README accurately describes live-vs-stubbed behavior.

samhotchkiss:
Round 2 fixes pushed in 215aafffc.

**Blocker 1 — surfaces.spec.ts:17-23 polling**
The old `await expect.poll` did `list.locator(".surface-empty").innerText()` on every poll iteration; when the element was absent (the populated-daemon case), Playwright awaited it until `actionTimeout` before the callback could return — which is why your run saw 26 rendered rows and still timed out. Replaced with a `page.waitForFunction` helper that short-circuits on `items > 0` and only reads `.surface-empty.textContent` after `querySelector` returns non-null. Hoisted into a `waitForSurfaceRailTerminal(page)` helper so all three live-daemon tests share the same terminal-state contract.

**Blocker 2 — surfaces.spec.ts:30-46 race**
`clicking a surface populates the center pane` and `each surface type can be selected if present` now `await waitForSurfaceRailTerminal(page)` before checking count. They no longer skip on live daemons mid-fetch.

**Blocker 3 — README accuracy**
Took option (a). Added `test.beforeEach` to `send_message.spec.ts`, `edge_cases.spec.ts`, `keyboard.spec.ts` that stubs `/api/v1/dashboard` with a minimal valid `DashboardResponse` envelope (rollups, daemon_status, active_sessions, recent_messages, projects, generated_at, scoped_fields). Those specs already stub `/api/v1/chat/sessions` + `/messages*` + `/send` per-test. README table reworded to document the new contract — these specs now genuinely don't hit live read endpoints, removing fullyParallel timing coupling.

**Verification**
- `npx playwright test --list`: 54 tests across 6 files, no discovery regressions.
- Local daemon not available in this sandbox (no `~/.pollypm/api-token`), so the populated/empty live-daemon reruns you did before will be the source of truth — the polling logic is now strictly more conservative (waits for terminal state, never blocks on an absent element).

Re-labeled `needs-review` for round 3.

samhotchkiss:
Review decision: do not merge yet.

This PR touches the web UI E2E test-contract invariants for cookie auth, surface discovery/rendering, route-stubbed send safety, connection-status failure surfacing, and mobile layout. The prior populated-surface polling issue is addressed: with a local stub server serving the real `/ui/` assets and two fake surfaces, the `surfaces.spec.ts` live-path tests passed on both projects. However, the current scaffold is still not clean.

Blocking issues:

1. `tests/playwright/tests/edge_cases.spec.ts:139-148` asserts that a stubbed `/api/v1/chat/sessions` 500 leaves `#conn-status` in `conn-warn|conn-error`, but the same spec's `beforeEach` stubs `/api/v1/dashboard` as a successful 200 (`edge_cases.spec.ts:50-52`). The UI starts `loadSurfaces()` and `pollDashboard()` together on init; the sessions failure can set the badge to warning, then the successful dashboard response sets it back to `conn-ok`. That is exactly what happens in the verification run below. This test is racing two independently valid API outcomes and will report a regression even when the UI is behaving according to its current status semantics. Make the invariant precise: either stub dashboard to fail too when testing the global badge, or assert the surface-rail error for a sessions-only failure, or change the UI/status contract and test that contract directly.

2. `tests/playwright/tests/auth.spec.ts:75-87` and `:95-113` hard-code "no cookie / no bearer => 401", while `tests/playwright/README.md:65-69` documents running the full suite against a Tailscale URL and the auth model intentionally accepts trusted tailnet peers without bearer/cookie. On a Tailscale-trusted base URL, those auth tests can fail while the product is correct; lack of a cookie would not imply lack of credentials because the peer IP itself is a credential mode. Either scope these negative auth assertions to loopback/non-tailnet runs, or document that `auth.spec.ts` is loopback-only and exclude/skip it for `POLLYPM_BASE_URL=http://100.x.y.z:8765`. Also clean up the auth header comment around `auth.spec.ts:18-23`, which says the non-local negative case is asserted via forged `X-Forwarded-For`; the spec does not actually do that.

Verification for this review:

- `npm ci` passed in `tests/playwright` with 0 vulnerabilities.
- `npx playwright test --list` discovered 54 tests across `chromium` and `mobile-chrome`.
- `uv run --extra test pytest tests/web_api/test_web_ui.py -q` passed: 42 tests.
- No daemon was listening at `http://127.0.0.1:8765`.
- Started a local stub server on `127.0.0.1:8766` serving the real `/ui/` assets, enforcing the cookie/no-cookie API behavior, and returning two fake surfaces. Then ran `POLLYPM_BASE_URL=http://127.0.0.1:8766 npx playwright test --workers=1`: 48 passed, 4 skipped, 2 failed. The failures were both `edge_cases.spec.ts › connection badge reflects HTTP failures` (`chromium` and `mobile-chrome`), with `#conn-status` ending as `conn-status conn-ok`.

Because this scaffold is intended to become the regression baseline for the v0 web UI, I would not merge it until the edge-case status invariant is made deterministic and the auth tests/docs line up with the documented Tailscale credential mode.

samhotchkiss:
Round 3 fixes pushed in 655ead622.

**Blocker 1 — edge_cases.spec.ts conn-status race**
`connection badge reflects HTTP failures` now overrides `/api/v1/dashboard` to 500 inside the specific test, in addition to its existing `/api/v1/chat/sessions` 500 stub. The `beforeEach()` dashboard=200 stub no longer races against the sessions failure; both read endpoints fail and the badge deterministically lands in `conn-warn|conn-error`. Assertion timeout tightened to 5s. This is option 1 from your comment — both endpoints fail, the global badge contract is preserved.

**Blocker 2 — auth.spec.ts loopback-only assumption**
Took option (a) from your comment. Detect tailnet base URLs via regex (`/\/\/100\.\d+\.\d+\.\d+/`) at module load and `test.skip(IS_TAILNET, ...)` the two negative cases (`direct API call without cookie returns 401`, `cookie isolation: fresh context after /ui/ load gets NO cookie`). On loopback runs (the default), behavior is unchanged. On `POLLYPM_BASE_URL=http://100.x.y.z:8765`, those two tests skip cleanly with a reason string explaining why the peer-IP credential mode makes the probe undecidable. The positive auth tests still run on both.

**Stale docstring**
Rewrote the `auth.spec.ts` module docstring (`:18-23` in the old file). Removed the false claim about `X-Forwarded-For` forgery and replaced it with the accurate loopback-vs-tailnet semantics now encoded in `IS_TAILNET`.

**Verification**
- `npx playwright test --list`: still 54 tests across 6 files, no discovery regressions.
- Local daemon not available in this sandbox; your next stub-server rerun will be the source of truth on the conn-status determinism.

Re-labeled `needs-review` for round 4.

PR #2067, opened 2026-05-22
fix(api): Phase 6 P0 — drop stale yaml entries, add GET /tasks flat-list

samhotchkiss:
## Summary

Phase 5 integration testing surfaced three P0 doc-vs-code mismatches. This PR closes all three.

- **P0-1 + P0-3 — OpenAPI doc-drift cleanup.** Removed six path entries that `pollypm.web_api.app` never wires: `GET /doctor`, `POST /projects`, `POST /projects/{key}/chat`, `POST /projects/{key}/plan`, `POST /tasks/{project}/{n}/approve`, `POST /tasks/{project}/{n}/reject`. Codegen clients branched on these and 404'd on every call. The plan approve/reject flow is Phase 3 work; the rest are Phase 1 carry-over. Orphan schemas left in `components/schemas` for Phase 3 reuse (no impact on codegen — only path entries trip clients).
- **P0-2 — Add `GET /api/v1/tasks` cross-project flat list.** Spec §5.1 calls for this; previously only the per-project `/projects/{key}/tasks` and single-task `/tasks/{project}/{n}` shipped. New endpoint supports the four documented filters (`project`, repeatable `status`, `assignee`, strict-after `since`) and opaque-cursor pagination via a new `list_all_tasks` helper that iterates per-project `list_tasks` and concatenates (acceptable for v1 workspace scale, swappable for a native cross-project query without changing the route signature).

YAML net: 56 additions / 188 deletions (`-132` lines).

## Spec gaps still deferred (per Phase 5 report, ~14 endpoints)

Untouched by this PR — tracked for Phase 2.5 / Phase 3:

- `POST /inbox/bulk`
- `POST /doctor/fix`, `GET /doctor/checks/{name}`
- Task `PATCH /labels` + `/relationships`; `POST .../append-context`, `/approve-plan`, `/reject-plan`
- `GET /heartbeats/{name}/history`, SSE `/heartbeats/stream`
- `GET /briefings/{type}/last` and `/render` as distinct GETs
- `GET /sessions/{name}/config`
- `GET /projects/{key}/activity`
- Dashboard SSE `/dashboard/stream`

## Test plan

- [x] `pytest tests/web_api/test_tasks_list_endpoint.py -v` — 9 new tests pass (filter axes, pagination, auth, bad-iso 400)
- [x] `pytest tests/web_api/test_openapi_conformance.py -v` — 5/5 pass; `PHASE_1_PATHS` extended to include `GET /tasks` so future drift trips CI
- [x] `openapi_spec_validator.validate()` on the updated YAML — passes as OpenAPI 3.1
- [x] Generated `openapi.json` from `create_app()` lists `/api/v1/tasks` alongside the existing routes
- [x] Pre-existing failures in `test_endpoints.py` (4 tests, dev-pg-state pollution) reproduce on `origin/main` and are not caused by this PR

### DIFF
--- a/src/pollypm/web_api/service.py
+++ b/src/pollypm/web_api/service.py
@@ -1005,6 +1005,18 @@ def list_project_tasks(
         ) from exc
 
 
+class StaleCursorError(Exception):
+    """Raised when ``list_all_tasks`` is given a cursor that no longer
+    matches any item in the current snapshot.
+
+    Round 1 of Codex review on PR #2067 noted that the previous
+    behaviour (silently restart at page one) creates duplicate rows
+    and infinite-pagination loops under concurrent updates. We now
+    raise this typed error so the route layer can map it to a 400
+    ``invalid_request`` rather than mask the failure.
+    """
+
+
 def list_all_tasks(
     config: PollyPMConfig,
     *,
@@ -1014,7 +1026,7 @@ def list_all_tasks(
     since: datetime | None = None,
     limit: int = 50,
     cursor: str | None = None,
-) -> tuple[list[APITaskSummary], str | None]:
+) -> tuple[list[APITaskSummary], str | None, list[dict[str, str]]]:
     """Cross-project flat task list (spec §5.1 / Phase 6 P0).
 
     No cross-project helper exists on the work-service yet, so we
@@ -1033,15 +1045,23 @@ def list_all_tasks(
       ``None`` means no status filter.
     * ``assignee`` — exact-match assignee filter.
     * ``since`` — only tasks with ``updated_at > since`` are returned.
-      Tasks missing ``updated_at`` are kept (treated as "always
-      newer") so the filter never silently drops untracked tasks.
+      Must be timezone-aware; the route layer rejects naive ISO
+      values before they reach this helper.
     * Pagination uses an opaque cursor; the cursor encodes the
       ``(updated_at_iso, task_id)`` of the last item on the previous
-      page so paging is stable when tasks are updated between calls.
+      page. A cursor that no longer matches any item raises
+      :class:`StaleCursorError` (route maps to 400) so clients restart
+      from page one explicitly — silently restarting would create
+      duplicate rows / infinite loops under concurrent updates
+      (PR #2067 Codex round 1, P0 #3).
 
-    Backing-store failures on individual projects degrade to "skip
-    that project" so a single bad DB doesn't 503 the whole flat list
-    (matches the inbox aggregator's posture).
+    Returns ``(page, next_cursor, warnings)``. ``warnings`` is a list
+    of ``{"project": <key>, "error": <code>}`` dicts — one entry per
+    project whose backing store raised during this request. An empty
+    list means every project read succeeded; the partial-failure
+    surface is explicit so a pg outage on one project can't hide
+    behind a 200 with a silently-truncated body (PR #2067 Codex round
+    1, P0 #1).
     """
     keys: Iterable[str]
     if project is not None:
@@ -1051,6 +1071,7 @@ def list_all_tasks(
 
     status_filter = set(statuses or [])
     summaries: list[APITaskSummary] = []
+    warnings: list[dict[str, str]] = []
     for key in keys:
         proj = config.projects[key]
         try:
@@ -1065,11 +1086,12 @@ def list_all_tasks(
                 tasks = svc.list_tasks(project=key)
         except _BACKING_STORE_ERRORS as exc:
             logger.warning(
-                "list_all_tasks: backing store error for %s; skipping: %s",
+                "list_all_tasks: backing store error for %s; surfacing as warning: %s",
                 key,
                 exc,
                 exc_info=True,
             )
+            warnings.append({"project": key, "error": "service_unavailable"})
             continue
 
         for task in tasks:
@@ -1089,8 +1111,10 @@ def list_all_tasks(
     # Newest-first ordering is the most useful default for cross-
     # project list views (cockpit "what changed recently?"). Tasks
     # without ``updated_at`` sort to the end via the epoch fallback so
-    # the cursor encoding stays well-defined.
-    epoch = datetime.min
+    # the cursor encoding stays well-defined. Use a tz-aware epoch so
+    # the sort key is comparable with the offset-aware ``updated_at``
+    # values pg returns.
+    epoch = datetime.min.replace(tzinfo=timezone.utc)
     summaries.sort(
         key=lambda t: (t.updated_at or epoch, t.task_id),
         reverse=True,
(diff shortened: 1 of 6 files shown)


src/pollypm/web_api/service.py
```diff
@@ -488,6 +488,113 @@ def list_project_tasks(
...
+            with _open_work_service_readonly(
+                config=config, project_key=key, project_path=proj.path
+            ) as svc:
+                # The work-service ``list_tasks`` only supports a
+                # single ``work_status`` filter, so we apply the
+                # OR-set client-side; assignee + since likewise apply
+                # client-side to keep the wrapper portable across
+                # backends.
+                tasks = svc.list_tasks(project=key)
+        except _BACKING_STORE_ERRORS as exc:
```
samhotchkiss, line 1087: Required changes before merge. Silently skipping a project on backing-store failure makes the cross-project task list look complete when it is not. That is a dangerous operator failure mode: a pg outage, permissions issue, or one bad project can hide live work while returning 200 with no diagnostic in the response. Please return a typed 503, or include an explicit partial-failure/warnings surface that clients can render. Also note this currently keys off the sqlite-era `_BACKING_STORE_ERRORS`; pg pool/psycopg failures need to be mapped deliberately and covered by tests.

src/pollypm/web_api/service.py
```diff
@@ -488,6 +488,113 @@ def list_project_tasks(
...
+            if status_filter:
+                value = _enum_value(getattr(task, "work_status", ""))
+                if value not in status_filter:
+                    continue
+            if assignee is not None:
+                if (getattr(task, "assignee", None) or "") != assignee:
+                    continue
+            if since is not None:
+                updated = getattr(task, "updated_at", None)
+                if updated is not None and updated <= since:
```
samhotchkiss, line 1107: Required changes before merge. `datetime.fromisoformat()` accepts naive ISO values such as `2026-05-22T00:00:00`, but task `updated_at` values are commonly offset-aware. Comparing those raises `TypeError: can't compare offset-naive and offset-aware datetimes`, which turns a syntactically valid `since` value into a 500. I reproduced this directly against `list_all_tasks()` with an aware task timestamp and a naive `since`. Please normalize timezone handling or reject naive timestamps with a typed 400, and add a regression.

src/pollypm/web_api/service.py
```diff
@@ -488,6 +488,113 @@ def list_project_tasks(
...
+    summaries.sort(
+        key=lambda t: (t.updated_at or epoch, t.task_id),
+        reverse=True,
+    )
+
+    cursor_idx = 0
+    if cursor is not None:
+        for idx, item in enumerate(summaries):
+            cursor_key = f"{(item.updated_at or epoch).isoformat()}|{item.task_id}"
+            if cursor_key == cursor:
```
samhotchkiss, line 1128: Required changes before merge. If the cursor item is updated, deleted, or otherwise moves between page requests, this loop never finds the cursor and leaves `cursor_idx = 0`, so the client receives page one again. That can create duplicate rows or infinite pagination loops under normal concurrent task updates. Please make stale/invalid cursors explicit (400) or use a stable cursor/snapshot strategy that cannot silently restart. Add a test where the previous page tail changes before the next request.

samhotchkiss (commented):
Required changes before merge. The new endpoint and OpenAPI tests pass, but the flat-list helper still has correctness gaps around partial failures, timestamp compatibility, and stale cursor behavior. Details inline.

src/pollypm/web_api/service.py
samhotchkiss: Round 1 fix pushed (2e2d409f3). The aggregator now returns a typed partial-failure envelope: each project whose backing store raises during the request lands as a `{project, error}` entry in the response's `warnings` array. Surviving projects' rows still come back, so the operator sees "partial read + which project failed" instead of a silent 200. _BACKING_STORE_ERRORS already covers `psycopg.OperationalError` + `psycopg_pool.PoolTimeout` (extended in #2060 round 6). Regression: `test_cross_project_list_partial_failure_envelope` stubs the readonly open to raise psycopg.OperationalError for one of two projects and asserts both the warning entry and the surviving rows.

src/pollypm/web_api/service.py
samhotchkiss: Round 1 fix pushed (2e2d409f3). The route now rejects naive ISO `since` values with a typed 400 `invalid_request` ("`since` must include a timezone offset"), before the value reaches the helper. Explicit is better than guessing UTC. Also tightened the helper's sort-key epoch fallback to be tz-aware UTC so the internal sort comparison can't trigger the same TypeError. Regression: `test_cross_project_list_rejects_naive_since` passes `since=2026-05-22T10:00:00` (no offset) and asserts the 400 + typed code.

src/pollypm/web_api/service.py
samhotchkiss: Round 1 fix pushed (2e2d409f3). The helper now raises `StaleCursorError` when no item in the current snapshot matches the cursor; the route maps that to a typed 400 `invalid_request` with a "retry without cursor" hint. No more silent fall-through to page one. Regression: `test_cross_project_list_stale_cursor_returns_400` reads page one (limit=1), then cancels the anchor task (which bumps `updated_at` and changes the cursor key), then requests page two — asserts 400 + the typed code.

samhotchkiss:

I verified this branch on a detached worktree with the then-current `origin/main` merged in: the cross-project task-list behavior and OpenAPI conformance both passed. However, #2063 has now landed, and this branch is no longer mergeable against current `main`.

`git merge-tree --write-tree origin/main origin/fix/phase2-p0-cleanups` now reports a content conflict in `src/pollypm/web_api/service.py` while auto-merging `docs/api/openapi.yaml`, `src/pollypm/web_api/models.py`, and `tests/web_api/test_openapi_conformance.py`. This is the shared insertion point between #2063's project-write/config-lock helpers and this PR's `list_all_tasks` helper.

Please rebase onto current `main` and resolve the service conflict while preserving both contracts: #2063's config/project mutation invariants and this PR's cross-project list behavior (partial-failure `warnings`, timezone-aware `since` rejection, and stale-cursor 400). Then rerun on the merged tree:
- `uv run --extra test --with pytest-timeout pytest tests/web_api/test_tasks_list_endpoint.py -v --timeout=120`

What I verified before #2063 landed:
- `uv run --extra test --with pytest-timeout pytest tests/web_api/test_tasks_list_endpoint.py -v --timeout=120` -> 12 passed.
- `uv run --extra dev --with pytest-timeout pytest tests/web_api/test_openapi_conformance.py -v --timeout=120` -> 8 passed.

samhotchkiss:
Rebased onto current `main` (new head `85003f0f9`). The only conflict was in `src/pollypm/web_api/service.py` `__all__` — a one-line collision where #2063 added `"init_project_guide_for_role"` and this PR added `"list_all_tasks"` at the same alphabetical slot. Resolved by keeping both entries; #2063's config/project mutation helpers (`set_project_tracked`, `archive_project`, `_refresh_live_projects`, `init_project_guide_for_role`, plus the shared `config_rmw_lock` from `pollypm.config`) and this PR's `list_all_tasks` + `StaleCursorError` contracts are both preserved end-to-end. `docs/api/openapi.yaml`, `src/pollypm/web_api/models.py`, and `tests/web_api/test_openapi_conformance.py` auto-merged.

Reran on the merged tree:
- `uv run python -m pytest tests/web_api/test_tasks_list_endpoint.py tests/web_api/test_openapi_conformance.py -x` -> **21 passed** (12 tasks-list + 9 conformance) in 48.69s.

Force-pushed with `--force-with-lease`; `needs-review` re-added.

Issue #1970, opened 2026-05-20
[sqlite-ripout] production paths still re-register and open sqlite stores


samhotchkiss:
Reviewed range: `d0d64497013bc77ad375cbdf1177015b4f90a858..7120046cc54968ae43e13f0bff3d73b52d30c3cd`
Related: #1956

## Finding
SQLite is no longer entry-pointed in `pyproject.toml`, but production runtime code can still explicitly re-register and use the sqlite backend.

Examples in `origin/main`:

- `pm notify --db <path>` re-registers sqlite and opens `get_store_by_url("sqlite:///...")` (`src/pollypm/cli_features/session_runtime.py:107-120`).
- `pm inbox --db <path>` does the same (`src/pollypm/work/inbox_cli.py:285-299`).
- The activity-feed projector falls back from pg to sqlite, registers sqlite, and opens the state DB (`src/pollypm/plugins_builtin/activity_feed/handlers/event_projector.py:343-370`).
- `service_notifications._message_store()` registers sqlite and opens `sqlite:///{service._db_path}` whenever it is called from `SQLiteWorkService` (`src/pollypm/work/service_notifications.py:84-94`).
- `SQLAlchemyStore` is still exported from `pollypm.store` (`src/pollypm/store/__init__.py:23-43`), so production modules can keep constructing/registering sqlite stores directly.

There are also still production direct sqlite constructions and `sqlite3.connect` call sites outside explicit migration/test modules, e.g. `src/pollypm/cockpit_inbox_items.py:921`, `src/pollypm/work/plan_review_emit.py:155`, and `src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py:2069`.

## Why this matters
The #1956 intent was that sqlite should not remain in production runtime. Deregistering the entry point makes `[storage].backend = "sqlite"` fail in a stock process, but the runtime still contains multiple production paths that opt sqlite back in and can recreate the split-brain sqlite shadow class of bugs.

## Validation
A stock `get_store(config)` with `backend='sqlite'` now fails with `StoreBackendNotFound`, which is good. The issue is that production paths call `register_backend("sqlite", SQLAlchemyStore)` themselves, reactivating sqlite process-wide.

## Suggested fix
[...]

PR #2074, opened 2026-05-23
fix(sqlite-ripout): reject sqlite:// URLs in pm notify/inbox --db (partial #1970)

samhotchkiss:
## Summary

Partial fix for #1970. On audit, **sites 1 of 6 (pm notify --db re-registration), 2 of 6 (pm inbox --db re-registration), 3 of 6 (activity-feed projector fallback), 4 of 6 (service_notifications._message_store), and 5 of 6 (SQLAlchemyStore export from pollypm.store) were already remediated in PR #1977 and follow-ups** — every live `register_backend("sqlite", ...)` and `SQLAlchemyStore("sqlite:///...")` call site in `src/` has been removed; only docstrings reference them now.

What remained: the `register_backend` function itself still **allowed** `name == "sqlite"` outside pytest with just a warn-log. That left a hole — a stray subprocess could import `SQLAlchemyStore` and re-register sqlite, recreating the split-brain shadow.

This PR closes that hole.

## Changes

- `register_backend("sqlite", ...)` now raises `ValueError` outside pytest instead of logging a warning. Pytest path stays open via the existing `_running_under_pytest()` short-circuit; `quiet=True` remains available for one-off migration tools (documented in the docstring).
- Updates the docstring `Raises` section + drops the stale "pm notify --db escape hatch" reference (that escape hatch was removed in #1977).
- Adds `tests/test_store_registry_sqlite_guard.py` with 4 tests pinning:
  - pytest opt-in still works
  - production rejection raises with a message naming the sqlite-ripout issue tags
  - non-sqlite backends register normally outside pytest (guard is narrow on purpose)
  - `quiet=True` remains the documented migration escape hatch

## Scope decision

Original task description targeted sites 1+2 (the `--db <path>` re-registration in `session_runtime.py` + `inbox_cli.py`). On reading those files, both flagged ranges contain no sqlite calls — PR #1977 (`chore(sqlite-ripout): remove --db CLI flags + sqlite test marker`) already shipped that fix. The reviewer's commit range (`d0d64497..7120046c`) predates that PR.

Rather than ship an empty PR or recreate work already done, this PR hardens the underlying registry so the **class** of bug stays out — which is what issue #1970's "Suggested fix" calls for in its last paragraph: *"Add a source-level guard that rejects `register_backend("sqlite", ...)`."*

Remaining surface for #1970 (kept open):
- Site 6: `src/pollypm/plugins_builtin/core_recurring/audit_watchdog.py:644` still has a `sqlite3.connect` call — but this is the `_count_work_tasks_for_project` legacy-shadow detector that **intentionally** peeks at legacy sqlite files to alert on shadowing. Tracked but not in scope for this PR.

## Test plan

- [x] `pytest tests/test_store_registry_sqlite_guard.py` — 4/4 pass
- [x] `pytest tests/test_state_cache_no_sqlite_imports.py tests/test_inbox_cli_filters.py tests/test_notify_task.py` — 33/33 pass (adjacent surface unchanged)
- [x] Verified the 8 pre-existing failures in `tests/test_pg_activity_feed_projector.py` + `tests/test_pg_events_retention.py` (`unsupported filter ['subject']`) reproduce on `main` unchanged — unrelated to this PR

Refs #1970, #1971

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,15 +53,18 @@ test = [
 pm = "pollypm.cli:app"
 pollypm = "pollypm.cli:app"
 
-# #1956: sqlite is intentionally NOT entry-pointed any more. The
-# pg cutover (#1737) made postgres the production backend; leaving
-# ``sqlite`` registered here let a misconfigured ``[storage].backend
-# = "sqlite"`` silently open an empty sqlite shadow alongside the
-# real pg state. ``get_store(config)`` now raises
-# ``StoreBackendNotFound`` for ``backend == "sqlite"`` unless a
-# caller has explicitly opted back in via
-# :func:`pollypm.store.registry.register_backend` (tests, legacy
-# ``--db <path>`` escape hatches). See ``store/registry.py``.
+# #1956 / sqlite-ripout (refs #1971, #1970): sqlite is intentionally
+# NOT entry-pointed any more. The pg cutover (#1737) made postgres
+# the production backend; leaving ``sqlite`` registered here let a
+# misconfigured ``[storage].backend = "sqlite"`` silently open an
+# empty sqlite shadow alongside the real pg state.
+# ``get_store(config)`` now raises ``StoreBackendNotFound`` for
+# ``backend == "sqlite"`` and
+# :func:`pollypm.store.registry.register_backend` hard-rejects
+# ``name == "sqlite"`` outside a pytest process — there is no
+# production-callable opt-in (the previous ``quiet=True`` and
+# legacy ``--db <path>`` CLI escape hatches are gone). See
+# ``store/registry.py``.
 [project.entry-points."pollypm.store_backend"]
 postgres = "pollypm.store.backends.pg_store:PgStore"
 
--- a/src/pollypm/config.py
+++ b/src/pollypm/config.py
@@ -616,13 +616,20 @@ def _parse_storage_settings(
     Recognised keys:
 
     * ``backend`` — entry-point name under ``pollypm.store_backend``.
-      Defaults to ``"postgres"`` after the #1737 cutover (issue #1939);
-      ``"sqlite"`` remains registered for explicit opt-in.
+      Defaults to ``"postgres"`` after the #1737 cutover (issue #1939).
+      ``"sqlite"`` is pytest-only for the store backend: production
+      configs that set ``backend = "sqlite"`` (or supply a
+      ``sqlite://`` URL) fail loudly when the resolver hits
+      :func:`pollypm.store.registry.register_backend`, which hard-
+      rejects sqlite registration outside a pytest process (refs
+      #1971, #1970).
     * ``url`` — SQLAlchemy URL. When empty/missing on the sqlite
-      backend, the resolver derives ``sqlite:///<project.state_db>`` so
-      first-run users don't have to think about connection strings. On
-      the postgres backend an empty ``url`` lets the pg pool resolve
-      the DSN itself (``[storage.pg].dsn`` / ``POLLYPM_PG_DSN``).
+      backend, the resolver derives ``sqlite:///<project.state_db>``
+      for the benefit of the pytest suite; in production this
+      derivation still flows into the registry guard, which raises.
+      On the postgres backend an empty ``url`` lets the pg pool
+      resolve the DSN itself (``[storage.pg].dsn`` /
+      ``POLLYPM_PG_DSN``).
 
     Missing section yields defaults. Fat-fingered value types (non-str)
     silently fall back to defaults — a broken ``[storage]`` block must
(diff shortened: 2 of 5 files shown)


samhotchkiss:
Blocking review findings for #2074:

1. `quiet=True` is now an unconditional production bypass for the sqlite guard (`src/pollypm/store/registry.py:140-154`, pinned by `tests/test_store_registry_sqlite_guard.py:123-145`). That does not enforce the #1970 invariant. The issue asks for a source-level guard that rejects `register_backend("sqlite", ...)` outside approved migration/test modules; this implementation lets any runtime caller re-enable sqlite process-wide by passing the old log-suppression flag. Please either remove the production bypass from the general registry path, or move/limit it to an explicit migration-only surface that cannot be reached by normal runtime modules, with tests proving unapproved production code cannot opt sqlite back in.

2. The same module still documents removed sqlite escape hatches as live behavior: `src/pollypm/store/registry.py:19-21`, `src/pollypm/store/registry.py:63-65`, and `src/pollypm/store/registry.py:325-330` still mention `pm notify --db` / `pm inbox --db` or legacy `--db <path>` sqlite registration. `pyproject.toml:60-64` has the same stale packaging note. This PR changes the storage/backend contract, so those docs/comments need to be reconciled in the same change; otherwise the single source of truth for sqlite-ripout remains contradictory.

Verified while reviewing:
- No PR comments, reviews, or review threads were present to address.
- Merged the PR head onto current `origin/main` in a temporary worktree; merge was clean.
- `uv run --extra test pytest tests/test_store_registry_sqlite_guard.py -q` passed: 4 passed.
- `uv run --extra test pytest tests/test_state_cache_no_sqlite_imports.py tests/test_inbox_cli_filters.py tests/test_notify_task.py -q` passed: 33 passed.

I am not merging this until the sqlite opt-in contract is made explicit/enforced and the stale docs are updated.

samhotchkiss:
Codex review addressed in 858062e60:

**Finding 1 (blocking):** Removed the ``quiet=True`` production bypass from ``register_backend`` entirely. The only path past the sqlite guard is now ``_running_under_pytest()`` — no production-callable opt-out remains. Grep confirmed no production caller used the kwarg.

**Finding 2 (stale docs):** Reconciled module docstring + ``_REGISTERED_BACKENDS`` comment + ``get_store_by_url`` docstring in ``src/pollypm/store/registry.py``, and the entry-points note in ``pyproject.toml``. They no longer reference ``pm notify --db`` / ``pm inbox --db`` / legacy ``--db <path>`` sqlite escape hatches.

**Test updates:** Replaced ``test_register_backend_sqlite_quiet_opt_in_outside_pytest`` with two new tests:
- ``test_register_backend_has_no_production_bypass`` (signature inspection: no ``quiet`` / ``force`` / ``allow_sqlite`` / ``bypass`` kwarg).
- ``test_register_backend_sqlite_no_kwarg_bypass_outside_pytest`` (calling with ``**{"quiet": True}`` outside pytest raises ``TypeError`` before mutating the registry).

Targeted suite passes: ``pytest tests/test_store_registry_sqlite_guard.py tests/test_state_cache_no_sqlite_imports.py tests/test_inbox_cli_filters.py tests/test_notify_task.py`` — 38 passed.

samhotchkiss:
Blocking follow-up: the code-level sqlite guard issue from the earlier review is fixed, but the storage contract docs are still inconsistent with the new invariant.

This PR now says there is no production-callable sqlite opt-in for `pollypm.store.registry.register_backend("sqlite", ...)`. A few in-repo contract comments still describe one:

- `src/pollypm/store/registry.py` in `_resolve_backend_factory`: the docstring still says the in-process registry is for “tests / opt-in prod callers”. With this PR, production sqlite opt-in is explicitly not allowed, so that should be reconciled.
- `src/pollypm/config.py` in `_parse_storage_settings`: the storage parser docstring still says `"sqlite" remains registered for explicit opt-in`, and the sqlite URL fallback comment says the “legacy escape hatch” still works. That is no longer true for the store backend contract this PR is hardening.
- `src/pollypm/models.py` in `StorageSettings`: the `backend` field docs still say sqlite remains registered for explicit opt-in.

The invariant touched here is the sqlite-ripout storage boundary: production code must not be able to reactivate sqlite via registration or config. The implementation now enforces that, but the documented contract is split across `store/registry.py`, `config.py`, and `models.py`. Please update those comments/docstrings so the single source of truth is clear: sqlite registration is pytest-only for the store backend; production configs/callers get a loud failure.

- `uv run --extra test pytest tests/test_store_registry_sqlite_guard.py tests/test_state_cache_no_sqlite_imports.py tests/test_inbox_cli_filters.py tests/test_notify_task.py -q` passed: 38 passed.

I am not merging until the residual storage/config docs match the enforced sqlite-ripout contract.

samhotchkiss:
Round-2 Codex feedback addressed in 2182b37e1.

Reconciled three docstrings/comments to match the registry guard's now-canonical contract — sqlite registration is pytest-only for the store backend; production configs and callers get a loud failure:

- `src/pollypm/store/registry.py::_resolve_backend_factory` — dropped `tests / opt-in prod callers`; noted that `register_backend` hard-rejects sqlite outside pytest, so the in-process map cannot revive sqlite in production.
- `src/pollypm/config.py::_parse_storage_settings` — replaced `"sqlite" remains registered for explicit opt-in` and the `legacy escape hatch` comment with explicit pytest-only language; documented that the sqlite URL derivation still flows into the registry guard in production and raises.
- `src/pollypm/models.py::StorageSettings.backend` — same reconciliation on the dataclass field doc.

Doc-only diff; no behavior change. Targeted tests still pass: `test_store_registry_sqlite_guard`, `test_state_cache_no_sqlite_imports`, `test_inbox_cli_filters`, `test_notify_task` (38 passed). Re-added `needs-review`.

Issue #1959, opened 2026-05-20
[perf] Inbox click still blocks UI thread during first-row detail hydration


samhotchkiss:
Parent: #1634
Reviewed range: `23f7902b434995b9b4176b6f072cf28f367bd6ce..d0d64497013bc77ad375cbdf1177015b4f90a858`

## Finding
The inbox first render still re-enters synchronous DB/detail work on the UI thread after the background inbox load completes.

The initial inbox load uses a worker, but completion calls `_render_list(select_first=True)` on the app thread (`src/pollypm/cockpit_ui.py:6528`). `_render_list()` immediately calls `_render_detail(row_ref.task_id)` with `prefer_cache=False` (`src/pollypm/cockpit_ui.py:6398`). `_render_detail()` then opens the work service and synchronously fetches task/replies/context (`src/pollypm/cockpit_ui.py:7460-7462`) through `_detail_fetch_task_data()` (`src/pollypm/cockpit_ui.py:7261-7278`), and can perform another sync marker lookup (`src/pollypm/cockpit_ui.py:7411-7413`). `_background_refresh()` also still calls `_load_inbox()` synchronously on the Inbox app thread (`src/pollypm/cockpit_ui.py:6570`).

## Why this matters
Clicking Inbox can swap from skeleton to an apparently stuck main pane while detail hydration does blocking DB/context/review-artifact work. This matches the reported 10+ second delay after clicking rail content.

## Suggested fix
Render the list immediately without selecting/hydrating detail, then hydrate the selected row detail asynchronously with a cancellable/stale-result guard. For refreshes, keep `_load_inbox()` off the UI thread and only apply completed snapshots. Add instrumentation around click-to-first-paint and detail-hydration latency so future changes cannot hide synchronous work in route completion callbacks.

PR #2077, opened 2026-05-23
fix(perf): move inbox detail hydration off Textual UI loop (closes #1969)

### LINKED ISSUE #1969: [perf] deferred Inbox detail hydration still blocks the UI loop
Reviewed range: `d0d64497013bc77ad375cbdf1177015b4f90a858..7120046cc54968ae43e13f0bff3d73b52d30c3cd`
Parent: #1634

## Finding
The #1959 fix defers first-row inbox detail hydration by one event-loop tick, but the expensive detail fetch still runs synchronously on the Textual UI loop.

`_render_list(..., defer_detail_hydration=True)` schedules `_schedule_deferred_detail_hydration()` (`src/pollypm/cockpit_ui.py:6400-6410`). The scheduled `_hydrate()` callback then calls `_render_detail(task_id)` directly on the UI loop (`src/pollypm/cockpit_ui.py:6434-6445`). `_render_detail()` still calls `_detail_fetch_task_data()`, which opens the work service and performs `svc.get()`, `svc.list_replies()`, and optional `svc.get_context()` synchronously (`src/pollypm/cockpit_ui.py:7292-7343`). The inline review artifact path can add more synchronous work after that (`src/pollypm/cockpit_ui.py:7523-7533`).

## Why this matters
The list may paint first, but the next loop tick can still freeze the inbox pane for the same slow detail fetch after selecting Inbox from the rail. That means the user's "click rail item, wait 10+ seconds" symptom can still appear as a post-paint UI freeze.

## Suggested fix
Move detail hydration off the UI loop, the same way the inbox list load uses a worker. The UI callback should render a placeholder, launch a cancellable/stale-guarded worker for task/replies/context/review artifact data, then apply the result only if the selected task is still current. Add a regression test that a slow `_detail_fetch_task_data()` does not block the app loop.

samhotchkiss:
## Summary

- #1959 deferred ``_render_detail`` by one event-loop tick via ``set_timer(0)``, but the hydration callback still ran the work-service open + ``svc.get`` / ``svc.list_replies`` / ``svc.get_context`` + inline review-artifact disk read inline on the UI thread — easily 100-400ms of blocked keypresses on a cold sqlite cache. After clicking an Inbox row the next loop tick could still freeze the pane for the same slow fetch.
- Placeholder (``[dim]Loading detail…[/dim]``) now paints synchronously; the IO-heavy fetch runs on ``run_worker(thread=True, exclusive=True, group=\"inbox_detail_hydrate\")`` (the same pattern ``_initial_load_sync`` already uses).
- The worker dispatches the apply step back to the UI thread via ``call_from_thread`` to a stale-guarded applier that re-checks both ``_selected_task_id`` and a new ``_pending_detail_hydration_task_id`` token. If the user navigated away mid-fetch (or a newer dispatch superseded the in-flight one), the result is discarded silently — no stale paint over the freshly-selected row.
- The inline review-artifact load (``load_task_review_artifact``) is also moved off the UI loop — it was the other disk-hitting call ``_render_detail`` made post-#1959.

## Test plan

- [x] New ``tests/test_inbox_detail_hydration_off_ui_loop.py`` — 6 regression tests, all passing:
  - placeholder painted synchronously + worker dispatched with the right ``thread`` / ``exclusive`` / ``group`` kwargs (and ``_render_detail`` NOT invoked inline)
  - ``run_worker`` dispatch-failure falls back to the synchronous path so the user isn't stranded on the placeholder
  - worker body routes every UI mutation through ``call_from_thread``; svc opened + closed cleanly
  - stale-guard at worker entry skips the svc resolve when ``_selected_task_id`` already moved
  - applier discards mid-fetch selection changes (no widget calls)
  - applier renders sections + review-artifact block on the happy path
- [x] Existing inbox + cockpit tests still pass (``test_cockpit_inbox_threads.py``, ``test_cockpit.py -k 'inbox or hydrate or render_detail'`` → 13 + 2 passed)
- [ ] Manual sanity check: open cockpit inbox, hit ``j``/``k`` rapidly after a row selection — keypresses should land instantly while the detail pane shows the placeholder then swaps to content

### DIFF
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -6190,6 +6190,10 @@ class PollyInboxApp(App[None]):
         self._tasks: list = []
         self._selected_task_id: str | None = None
         self._selected_row_key: str | None = None
+        # #1969 — tracks the most-recent deferred-hydration dispatch so
+        # a worker callback whose target row was navigated away from
+        # can drop its result without re-painting stale data.
+        self._pending_detail_hydration_task_id: str | None = None
         self._unread_ids: set[str] = set()
         self._session_read_ids: set[str] = set()
         self._replies_by_task: dict[str, list] = {}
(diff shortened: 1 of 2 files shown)


samhotchkiss:
Blocking this for the #1969 invariant: the new hydration applier still does disk/storage work on the Textual UI loop for plan-review rows.

In `src/pollypm/cockpit_ui.py`, `_apply_hydrated_detail()` calls `_detail_build_sections()` and `_detail_apply_label_hints()` after the worker returns. Those helpers are not pure UI formatting in the plan-review case:

- `_detail_build_sections()` calls `_plan_content_for_review(...)`, which loads config and reads the canonical plan file from disk (`src/pollypm/cockpit_ui_inbox_format.py`). That means a plan-review detail hydration can still block the UI thread on filesystem/config IO.
- `_detail_apply_label_hints()` calls `_plan_review_discussion_marker_exists(...)` when replies do not prove the round trip. That opens/resolves a work service and runs `get_context(...)` synchronously on the UI thread.

That leaves the core contract unclear: the worker is described as moving the IO-heavy hydration path off-loop, but the applier can still hit config, filesystem, and storage before painting. Please split the remaining plan-review hydration inputs into the worker payload, or make the formatting/hint applier consume precomputed data only, then add a regression test with a plan-review row proving no UI-thread disk/store helper is invoked during `_apply_hydrated_detail()`.

Targeted tests I ran against current `origin/main` plus this PR passed: `uv run --extra test pytest tests/test_inbox_detail_hydration_off_ui_loop.py tests/test_cockpit_inbox_threads.py tests/test_cockpit.py -k "inbox or hydrate or render_detail" -q` (`21 passed, 192 deselected`).

samhotchkiss:
Thanks for the catch — addressed in 8fd39940f.

The hydration worker now precomputes both flagged inputs off the UI loop and stashes them in the applier payload:

- **Plan content** — worker calls `_plan_content_for_review(meta, config_path)` once with the just-fetched task's labels; the result rides into `_apply_hydrated_detail` as `plan_content_precomputed` and is forwarded into `_detail_build_sections(...)` as a kwarg.
- **Discussion marker** — worker first runs the cheap `_plan_review_has_round_trip(replies, requester)` reply heuristic; only when that returns False does it issue a single `svc.get_context(task_id, entry_type=_PLAN_REVIEW_DISCUSSION_ENTRY_TYPE, limit=1)` call on the already-open work service (no second `_resolve_inbox_svc`). The boolean rides into `_apply_hydrated_detail` as `discussion_marker_precomputed` and is forwarded into `_detail_apply_label_hints(...)` as a kwarg.

The applier is now pure formatting + state mutation — no disk, no store, no `svc.open`/`svc.get_context`.

Backwards-compat: both new kwargs default to `None`, so the legacy `_render_detail` call site (the synchronous fallback for unit tests / teardown without an event loop) keeps the original disk/store path. No behavioural change off the hydration path.

Regression test added in `tests/test_inbox_detail_hydration_off_ui_loop.py::test_plan_review_hydration_precomputes_disk_and_store_lookups_off_ui_loop`. It uses a plan-review row and:

1. Monkeypatches `_plan_content_for_review` and `_plan_review_discussion_marker_exists` with counter-stubs.
2. Asserts the worker phase invokes the plan-content helper exactly once + issues one `svc.get_context(entry_type="plan_review_discussed")` call.
3. Captures the dispatched applier args, then replays the applier synchronously.
4. Asserts during the applier: zero `_plan_content_for_review` calls, zero `_plan_review_discussion_marker_exists` calls, zero `svc.get_context` calls for the discussion entry-type.
5. Asserts the precomputed plan content + discussion marker bool flow into `_detail_build_sections` / `_detail_apply_label_hints` as kwargs and the precomputed plan body lands in the painted detail pane.

Test surface green:
```
pytest tests/test_inbox_detail_hydration_off_ui_loop.py tests/test_cockpit_inbox_threads.py tests/test_cockpit.py -k "inbox or hydrate or render_detail" -x
# 22 passed, 192 deselected
```

PR #2078, opened 2026-05-23
fix(state-cache): route workspace-root inbox into cache via __workspace__ entry (closes #2051)

### LINKED ISSUE #2051: state_cache: route workspace-root inbox messages into cache (replaces fall-through workaround in PR #2026)
## Context

PR #2026 review blocker 3: ``state_cache/refresh_impl.py::_awaits_user_items_for`` filters the workspace-wide direct list into per-project entries by ``item.project == project_key``. Messages with ``scope IN ('', 'inbox')`` (workspace-root) carry ``project="inbox"`` after ``message_row_to_inbox_entry``, so they are never associated with any tracked project and disappear from the cache.

## Workaround in PR #2026

``_maybe_cache_route_awaits_user`` and ``_maybe_cache_count_awaits_user`` (src/pollypm/cockpit_inbox.py) now call ``_workspace_root_inbox_has_open(config)`` — a small pg probe — and fall through to the direct sweep whenever the workspace-root inbox is non-empty. Functional but defeats the cache for any workspace with operator-addressed messages.

## Fix

Two options:

1. **Synthetic "workspace" project entry.** Have the refresher emit an entry keyed on a sentinel like ``__workspace__`` populated with the workspace-root awaits-user items. ``_maybe_cache_route_awaits_user`` then concatenates per-project + workspace entries. Cleanest semantics but touches the entry shape.

2. **Per-entry workspace-root field.** Add a top-level (not per-project) snapshot field that the cache carries alongside per-project entries — ``workspace_root_awaits_user_items``. Routing helpers union it in.

Either way, drop the ``_workspace_root_inbox_has_open`` probe + its callers in cockpit_inbox.py.

## Tests

- The 3 regression tests in test_state_cache_parity.py::TestPr2026ReviewBlocker3WorkspaceRootFallthrough need to flip: with workspace-root populated, the cache fast-path must now win AND include the workspace row.

Refs PR #2026, design doc docs/design/move-a-state-cache.md §6.2.

samhotchkiss:
## Summary

- Refresher emits a synthetic ``__workspace__`` cache entry (new ``WORKSPACE_PROJECT_KEY`` sentinel in ``state_cache/refresh_impl.py``) populated with workspace-root awaits-user items — the rows that ``message_row_to_inbox_entry`` projects to ``project == "inbox"`` after seeing ``scope IN ('', 'inbox')``. These never matched any tracked project key, so the per-project filter in ``_awaits_user_items_for`` silently dropped them and the cache could not represent them.
- ``cockpit_inbox._maybe_cache_route_awaits_user`` / ``_maybe_cache_count_awaits_user`` now read the ``__workspace__`` entry alongside per-project entries and union its items / count into the result.
- The ``_workspace_root_inbox_has_open`` probe (and its underlying ``cockpit_pg_aggregates.has_workspace_root_open_messages`` helper) are deleted. The probe previously forced cache fall-through on every busy workspace whenever any workspace-root row existed, defeating the rail-badge / inbox-count fast path. The cache is now authoritative for workspace-root rows too.
- The state-cache singleton's ``_project_keys_provider`` always yields the sentinel so initial full refresh + workspace-wide invalidations keep the entry fresh; existing rail / dashboard consumers iterate over ``config.projects`` keys, so the extra sentinel entry is naturally ignored by them (no consumer-shape regression).

## Closes

#2051

## Tests

- ``TestPr2026ReviewBlocker3WorkspaceRootFallthrough`` (the 3 PR-#2026 regression tests) — flipped from asserting forced fall-through to asserting the cache fast-path wins AND carries the workspace row.
- ``TestPr2026V3WorkspaceRootProbeUnbounded`` — deleted; its subject (the workspace-root SQL probe) no longer exists.
- ``test_singleton_wires_project_keys_provider_for_initial_refresh`` + ``test_singleton_provider_degrades_gracefully_on_config_failure`` — updated to expect ``__workspace__`` in the provider output.
- ``test_state_cache_config_identity.py::test_unstamped_entries_skip_identity_check`` — dropped the probe monkeypatch (probe is gone).
- ``pytest tests/test_state_cache.py tests/test_state_cache_entry.py tests/test_state_cache_env_flag.py tests/test_state_cache_no_sqlite_imports.py tests/test_state_cache_refresher.py tests/test_cockpit_inbox_threads.py tests/test_state_cache_parity.py tests/test_state_cache_config_identity.py`` → **114 passed**.

## Test plan

- [x] ``pytest tests/test_state_cache_parity.py::TestPr2026ReviewBlocker3WorkspaceRootFallthrough -x``
- [x] Full state_cache + cockpit_inbox sweep (114 tests)
- [ ] Manual: confirm rail badge + inbox count include workspace-root rows under load (deferred — covered by the regression flip)

### DIFF
--- a/docs/design/move-a-state-cache.md
+++ b/docs/design/move-a-state-cache.md
@@ -1,12 +1,20 @@
 # Move A — In-Process Project-State Cache with Epoch-Driven Invalidation
 
-## Implementation drift (2026-05-22)
+## Implementation drift (2026-05-22, updated 2026-05-23)
 
-The shipped Move A behavior diverges from the original design in three places. This section is the source of truth; the original sections below are kept for historical context but should be read as superseded where they conflict.
+The shipped Move A behavior diverges from the original design in two places. This section is the source of truth; the original sections below are kept for historical context but should be read as superseded where they conflict.
 
 - `latest_heartbeat_by_session` is reserved on `ProjectStateCacheEntry` but NOT populated. Rail heartbeat sites read `pollypm.storage.pg_heartbeats.latest_heartbeat` directly via `CockpitRouter._latest_heartbeat_cached`. Bulk prefetch + cache invalidation deferred to [#2050](https://github.com/samhotchkiss/pollypm/issues/2050).
 - `_maybe_cache_route_rollups` declines (returns `None`) when any tracked project has a live actionable alert — cache cannot recompute the alert-to-rollup contract. Tracked in [#2049](https://github.com/samhotchkiss/pollypm/issues/2049).
-- `_maybe_cache_route_awaits_user` / `_maybe_cache_count_awaits_user` decline when the workspace-root inbox has any open message (via `has_workspace_root_open_messages` probe). Workspace-root inbox isn't yet represented in cache entries. Tracked in [#2051](https://github.com/samhotchkiss/pollypm/issues/2051).
+
+### Workspace-root inbox is now cache-routed (#2051 — closed by PR #2078)
+
+The third drift item from the original list (workspace-root inbox not represented in cache entries) is closed by PR #2078. The shipped behavior:
+
+- Workspace-root awaits-user rows (`messages` with `scope IN ('', 'inbox')`, surfaced as `project == "inbox"` after `message_row_to_inbox_entry`) are represented in the cache via a **synthetic `__workspace__` entry**. The sentinel key (`WORKSPACE_PROJECT_KEY = "__workspace__"`) and TTL constant (`WORKSPACE_ENTRY_TTL_SECONDS = 10.0`) live in `pollypm.state_cache.entry` so the refresher, refresh-impl, `cockpit_inbox`, and tests all share one source of truth.
+- The refresher's project-keys provider always includes the sentinel, so the synthetic entry is computed on boot and on every full pass. `StateCacheRefresher._dispatch_event` invalidates `__workspace__` on every event in `_INVALIDATING_EVENTS` so workspace-root invalidations never lag the per-project ones.
+- The earlier `has_workspace_root_open_messages` probe and the forced fall-through it gated are **removed**. `_maybe_cache_route_awaits_user` / `_maybe_cache_count_awaits_user` now union the `__workspace__` entry with the per-project entries directly; the cache is authoritative for workspace-root rows.
+- The sentinel carries a **bounded-staleness TTL** (`WORKSPACE_ENTRY_TTL_SECONDS = 10s`). Several message-store paths (`PgStore.close_message`, `PgStore.clear_alert`, `service_api.v1.clear_alert`) mutate workspace-root awaits-user rows without emitting a state-cache audit event, so the refresher doesn't see those closes. The TTL caps the staleness window — once the sentinel ages past the TTL, the cache-read boundary falls through to the direct sweep. Full audit-event wiring for the message-store close/clear paths is deferred past v1 RC; the TTL is a documented bounded-staleness window, not an invariant.
 
 Design document for issue [#1664](https://github.com/samhotchkiss/pollypm/issues/1664).
 Status: **implemented** (2026-05-20) — PR 1 #2000, PR 2 #2016, PR 3 #2026, PR 4 #2029.
--- a/src/pollypm/cockpit_inbox.py
+++ b/src/pollypm/cockpit_inbox.py
@@ -230,54 +230,6 @@ def pm_inbox_awaits_user_list(config) -> list[object]:
     return result
 
 
-def _workspace_root_inbox_has_open(config) -> bool:
-    """True iff there is at least one open workspace-root inbox row.
-
-    "Workspace-root" = messages with ``scope IN ('', 'inbox')`` —
-    these are NOT keyed to any tracked project, so the refresher's
-    per-project filter in
-    ``state_cache/refresh_impl.py::_awaits_user_items_for`` drops them
-    on the floor. The cache's per-project entries can never represent
-    them, so any cache-routed read MUST fall through whenever the
-    workspace-root inbox is non-empty.
-
-    Filed as a follow-up issue (PR #2026 review blocker 3): teach the
-    cache to carry a workspace-root entry so this gate can go away.
-
-    PR #2026 v3 (Codex re-review): delegates to the dedicated SQL
-    existence query :func:`pollypm.cockpit_pg_aggregates.has_workspace_root_open_messages`
-    so the answer no longer depends on whether the newest N rows happen
-    to include a workspace-root row. The old ``open_messages(limit=50)``
-    + Python-side scan path silently dropped workspace work whenever 50+
-    newer project-scoped rows pushed the workspace-root row out of the
-    window.
-
-    Returns True on any failure (import or probe). Conservative — forces
-    cache fall-through. A false positive here is safe (the direct sweep
-    just runs); a false negative would silently drop workspace-root work,
-    violating the cache-authoritative invariant.
-    """
-
-    try:
-        from pollypm.cockpit_pg_aggregates import (
-            has_workspace_root_open_messages,
-        )
-    except Exception:  # noqa: BLE001
-        logger.warning(
-            "workspace-root inbox probe import failed; assuming inbox is non-empty to force conservative cache decline",
-            exc_info=True,
-        )
-        return True
-    try:
-        return bool(has_workspace_root_open_messages(config))
-    except Exception:
-        logger.warning(
-            "workspace-root inbox probe failed; assuming inbox is non-empty to force cache fall-through",
-            exc_info=True,
-        )
-        return True  # conservative: force cache decline
-
-
 def _maybe_cache_route_awaits_user(config) -> list[object] | None:
     """Return the cache-routed result, or ``None`` to fall through.
 
@@ -292,16 +244,21 @@ def _maybe_cache_route_awaits_user(config) -> list[object] | None:
       serve cross-config data when project keys overlap; the
       config-identity stamp catches this).
     * Any tracked project is missing from the snapshot (partial cache).
-    * Any workspace-root awaits-user message exists live (PR #2026
-      review blocker 3) — the refresher does not project workspace-root
-      messages into any per-project entry, so the cache cannot
-      represent them and a partial answer would silently drop them
-      from the rail badge / inbox count.
     * Any unexpected exception (defensive — broken cache must never
       crash the rail badge).
 
+    #2051: workspace-root awaits-user messages (``scope IN ('',
+    'inbox')``) now live on a synthetic ``__workspace__`` cache entry
+    emitted by the refresher (see
+    :data:`pollypm.state_cache.refresh_impl.WORKSPACE_PROJECT_KEY`).
+    This helper unions those items with the per-project ones so the
+    rail badge / inbox count include them. The earlier
+    ``_workspace_root_inbox_has_open`` probe + its forced
+    fall-through were removed — the cache is now authoritative for
+    workspace-root rows too.
+
     Cache lookup; fall through to direct path on cache miss / cache
-    disabled / config-identity mismatch / workspace-root inbox row.
+    disabled / config-identity mismatch.
     """
 
     try:
@@ -350,14 +307,19 @@ def _maybe_cache_route_awaits_user(config) -> list[object] | None:
     # project's items are included.
     if not known_projects.issubset(snapshot.keys()):
         return None
-    # Workspace-root guard (PR #2026 review blocker 3): the refresher
-    # only projects per-project messages, so any open workspace-root
-    # inbox row would be missing from the cache. Defer to the direct
-    # path whenever one exists.
-    if _workspace_root_inbox_has_open(config):
-        return None
     cached_items: list[object] = []
+    # #2051: the synthetic ``__workspace__`` entry carries workspace-root
+    # awaits-user items (messages with ``scope IN ('', 'inbox')`` that
+    # don't belong to any tracked project). Union them first so the
+    # returned list mirrors the direct sweep's content.
+    from pollypm.state_cache.refresh_impl import WORKSPACE_PROJECT_KEY
+    workspace_entry = snapshot.get(WORKSPACE_PROJECT_KEY)
+    if workspace_entry is not None:
+        for item in getattr(workspace_entry, "awaits_user_items", ()) or ():
+            cached_items.append(item)
     for project_key, entry in snapshot.items():
+        if project_key == WORKSPACE_PROJECT_KEY:
+            continue
         if project_key not in known_projects:
             # The cache may carry an entry for a project that was just
             # untracked. Skip — the direct path also drops it.
@@ -549,11 +511,12 @@ def _maybe_cache_count_awaits_user(config) -> int | None:
     the rail badge would silently drop work that's still waiting on
     the user.
 
-    Workspace-root fall-through (PR #2026 review blocker 3): the
-    refresher does not project workspace-root inbox messages
-    (``scope IN ('', 'inbox')``) into any per-project entry, so the
-    cached sum would silently under-count them. Defer to the direct
-    sweep whenever the workspace-root inbox is non-empty.
+    #2051: workspace-root awaits-user rows (``scope IN ('', 'inbox')``)
+    are summed from the synthetic ``__workspace__`` cache entry
+    emitted by the refresher (see
+    :data:`pollypm.state_cache.refresh_impl.WORKSPACE_PROJECT_KEY`).
+    The earlier workspace-root probe + forced fall-through were
+    removed; the cache is now authoritative for those rows too.
     """
 
     try:
@@ -590,12 +553,17 @@ def _maybe_cache_count_awaits_user(config) -> int | None:
     # direct path so the badge stays correct during the boot-time gap.
     if not known_projects.issubset(snapshot.keys()):
         return None
-    # Workspace-root guard (PR #2026 review blocker 3): see
-    # ``_maybe_cache_route_awaits_user`` for rationale.
-    if _workspace_root_inbox_has_open(config):
-        return None
     total = 0
+    # #2051: pull the workspace-root awaits-user count off the
+    # synthetic ``__workspace__`` entry. Absent entry (cold workspace
+    # path, or sentinel never refreshed) contributes 0.
+    from pollypm.state_cache.refresh_impl import WORKSPACE_PROJECT_KEY
+    workspace_entry = snapshot.get(WORKSPACE_PROJECT_KEY)
+    if workspace_entry is not None:
+        total += int(getattr(workspace_entry, "awaits_user_count", 0) or 0)
     for project_key, entry in snapshot.items():
+        if project_key == WORKSPACE_PROJECT_KEY:
+            continue
         if project_key not in known_projects:
             continue
         total += int(getattr(entry, "awaits_user_count", 0) or 0)
--- a/src/pollypm/cockpit_inbox.py
+++ b/src/pollypm/cockpit_inbox.py
@@ -321,15 +321,35 @@ def _maybe_cache_route_awaits_user(config) -> list[object] | None:
     from pollypm.state_cache.refresh_impl import WORKSPACE_PROJECT_KEY
     if WORKSPACE_PROJECT_KEY not in snapshot:
         return None
+    # #2051 round-3 (Codex review): bounded-staleness guard. Once the
+    # refresher has installed an EMPTY ``__workspace__`` entry (no
+    # workspace-root awaits-user rows at snapshot time), the absent-
+    # sentinel guard above stops firing — yet message-store writes
+    # (alerts, notifications, `pm notify`) can still land after that
+    # refresh without emitting an invalidating audit event, leaving the
+    # empty entry stale. Treat a present-but-empty workspace entry as
+    # NOT-AUTHORITATIVE and fall through to the direct sweep. A non-
+    # empty entry is trusted: the refresher just computed it and the
+    # items are real. Cost: workspaces with genuinely zero root-level
+    # awaits-user rows pay the direct-sweep cost on every call (per-
+    # project entries still benefit from the rest of the cache). Full
+    # audit-event wiring for workspace-root producers is deferred past
+    # v1 RC.
+    workspace_entry = snapshot.get(WORKSPACE_PROJECT_KEY)
+    workspace_items = (
+        getattr(workspace_entry, "awaits_user_items", ()) or ()
+        if workspace_entry is not None
+        else ()
+    )
+    if not workspace_items:
+        return None
     cached_items: list[object] = []
     # The synthetic ``__workspace__`` entry carries workspace-root
     # awaits-user items (messages with ``scope IN ('', 'inbox')`` that
     # don't belong to any tracked project). Union them first so the
     # returned list mirrors the direct sweep's content.
-    workspace_entry = snapshot.get(WORKSPACE_PROJECT_KEY)
-    if workspace_entry is not None:
-        for item in getattr(workspace_entry, "awaits_user_items", ()) or ():
-            cached_items.append(item)
+    for item in workspace_items:
+        cached_items.append(item)
     for project_key, entry in snapshot.items():
         if project_key == WORKSPACE_PROJECT_KEY:
             continue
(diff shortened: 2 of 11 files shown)


samhotchkiss:
Blocking reviewer note: this needs changes before merge.

This PR removes the live `_workspace_root_inbox_has_open` probe and makes the synthetic `__workspace__` entry authoritative for `pm_inbox_awaits_user_list` / `_count_inbox_tasks_for_label`. That can be safe only if the cache has a fresh workspace entry whenever workspace-root awaits-user rows can change. The current implementation does not establish that.

Specific issues:

- `src/pollypm/cockpit_inbox.py` in `_maybe_cache_route_awaits_user` and `_maybe_cache_count_awaits_user`: a snapshot with all tracked project keys but no `WORKSPACE_PROJECT_KEY` is treated as cache-complete (`snapshot.get(WORKSPACE_PROJECT_KEY)` simply contributes no items/count). Under the new invariant, missing `__workspace__` is a partial cache, not a proven-empty workspace inbox. The added `tests/test_state_cache_parity.py::test_no_workspace_entry_still_serves_per_project_via_cache` pins the unsafe behavior; it should be inverted so missing `__workspace__` falls through or otherwise proves no workspace-root rows can exist.

- `src/pollypm/state_cache/refresher.py` in `_dispatch_event`: project-scoped audit events still only invalidate `project`, and the invalidating event set does not include message-store writes/closures. Workspace-root `pm notify` / alert rows can be created or closed after startup without refreshing the synthetic `__workspace__` entry. Before this PR, the live probe forced fall-through when a workspace-root row existed; after this PR, the badge/list can serve a stale zero until a full workspace invalidation or process restart. Please wire the sentinel into the invalidation contract for every producer that can affect workspace-root awaits-user rows (including the actual event shape emitted for `project == "inbox"` / empty-scope notifications), and add regression coverage that exercises the refresher path rather than only pre-seeded snapshots.

The invariant touched here is: cached awaits-user list/count must match the direct sweep for every tracked project plus workspace-root inbox rows, across startup, incremental writes, closures, retries, and partial cache states. Right now that invariant is split between the refresher provider and consumers, but not enforced at the cache-read boundary or by invalidation tests.

Also fix the mechanical failure from `git diff --check`: `tests/test_state_cache_parity.py:1452` has a new blank line at EOF.

Verification run: `uv run --extra test pytest tests/test_state_cache.py tests/test_state_cache_entry.py tests/test_state_cache_env_flag.py tests/test_state_cache_no_sqlite_imports.py tests/test_state_cache_refresher.py tests/test_cockpit_inbox_threads.py tests/test_state_cache_parity.py tests/test_state_cache_config_identity.py` passed, 114 tests. The passing suite does not cover the stale/missing-sentinel failure mode above.

samhotchkiss:
Addressed the Codex review feedback. Three load-bearing changes:

**A. Cache-read boundary** (`src/pollypm/cockpit_inbox.py`)
Both `_maybe_cache_route_awaits_user` and `_maybe_cache_count_awaits_user` now refuse to serve when `WORKSPACE_PROJECT_KEY` is absent from the snapshot. Missing sentinel is treated as INCOMPLETE cache, not proven-empty workspace inbox — fall through to the direct sweep.

**B. Invalidation contract** (`src/pollypm/state_cache/refresher.py::_dispatch_event`)
Every event in `_INVALIDATING_EVENTS` now also invalidates `__workspace__`. Events with `project="inbox"` (the shape workspace-root inbox tasks emit) AND empty-project events both additionally trigger `invalidate(None)` so the snapshot stays consistent across the per-project + sentinel union. Pinned by a new `_WORKSPACE_ROOT_PROJECT_SIGNALS` set + a `_WORKSPACE_PROJECT_KEY` string constant (avoids the circular-import risk of pulling `refresh_impl` into the leaf refresher module).

Audit events that now keep the sentinel fresh: `task.created`, `task.status_changed`, `task.deleted`, `marker.{created,released,create_failed,leaked}`, `work_table.cleared` — i.e. every member of `_INVALIDATING_EVENTS`. The conservative "refresh on every event" approach costs one extra workspace-root sweep per invalidation (single bulk query) and trades that for the guarantee that the cache-read boundary's authoritative-when-present invariant can't be defeated by a cross-posted task event.

Message-store writes (alert/notification rows) still emit no audit event in this PR's scope — the cache-read boundary fall-through is the safety net for those.

**C. Regression coverage** (`tests/test_state_cache_parity.py`, `tests/test_state_cache_refresher.py`)
- Inverted `test_no_workspace_entry_still_serves_per_project_via_cache` → `test_no_workspace_entry_falls_through_to_direct_sweep`.
- Added count-helper variant of the inverted test.
- Four new refresher tests exercising the dispatch path (`project_inbox_event_invalidates_workspace_sentinel`, `project_inbox_status_change_invalidates_workspace_sentinel`, `tracked_project_event_also_refreshes_workspace_sentinel`, `non_invalidating_event_leaves_sentinel_alone`, `workspace_scoped_event_still_invalidates_sentinel`).
- Updated `_seed_cache` in both parity + config-identity test files to auto-seed an empty `__workspace__` entry (matches the new fast-path invariant) so the existing 100+ tests continue to exercise the fast-path; absent-sentinel tests bypass the helper.

**D. Whitespace** Removed the EOF blank line `git diff --check` flagged on `tests/test_state_cache_parity.py:1452`.

All 120 tests in the suite (`tests/test_state_cache_parity.py tests/test_state_cache_refresher.py tests/test_state_cache.py tests/test_state_cache_entry.py tests/test_state_cache_env_flag.py tests/test_state_cache_no_sqlite_imports.py tests/test_cockpit_inbox_threads.py tests/test_state_cache_config_identity.py`) pass.

samhotchkiss:
Still blocking on the stale workspace sentinel case.

The new absent-sentinel guard fixes only the boot/partial-cache half of the previous review. It falls through when `WORKSPACE_PROJECT_KEY` is missing (`src/pollypm/cockpit_inbox.py` lines 321-323 and 582-584), but once the refresher has installed an empty `__workspace__` entry, the cache-read boundary treats that entry as authoritative and serves its `awaits_user_items` / `awaits_user_count` directly (`src/pollypm/cockpit_inbox.py` lines 329-332 and 589-591).

That still leaves the original message-store failure mode: workspace-root alerts/notifications/`pm notify` rows can be created or closed after the sentinel has been refreshed, and the PR notes those writes still do not emit an invalidating audit event. The refresher fix only handles events that reach `_dispatch_event()` (`src/pollypm/state_cache/refresher.py` lines 365-385). A present empty sentinel therefore remains stale and the list/count fast path can keep serving zero until some unrelated invalidating event or restart refreshes it.

The new regression test at `tests/test_state_cache_parity.py` lines 1437-1495 proves a missing `__workspace__` falls through, but it does not simulate the production-stale case: seed a present empty `__workspace__` entry, make the direct workspace-root sweep return a new/closed row that arrived through a non-audit message-store write, and assert the cache reader declines or the producer invalidates the sentinel. Today that shape would serve the stale sentinel and never call the direct sweep.

Please either wire every workspace-root message producer/closure that can affect awaits-user rows into the sentinel invalidation path, or keep a bounded/direct freshness check for the synthetic entry so a present-but-stale sentinel cannot become authoritative. Add regression coverage for the present-stale sentinel case, not just the absent-sentinel case.

Verification on current `origin/main` plus `41b5a3d`:

- `uv run --extra test pytest tests/test_state_cache_parity.py::TestPr2026ReviewBlocker3WorkspaceRootFallthrough tests/test_state_cache_refresher.py::TestWorkspaceRootInvalidation -q` passed: `9 passed in 15.85s`.

samhotchkiss:
**Round-3 follow-up — empty `__workspace__` sentinel now falls through.**

Confirmed the gap: the round-2 absent-sentinel guard only covered the boot/partial-cache half. Once the refresher installs an empty `__workspace__` entry, the cache-read boundary was still treating its `awaits_user_items=()` / `awaits_user_count=0` as authoritative — and message-store writes that bypass audit-event invalidation (alerts, notifications, `pm notify` workspace-root rows) leave that empty sentinel stale.

**Pick: bounded-staleness option (Codex's second option).** When `__workspace__` is present but empty, both `_maybe_cache_route_awaits_user` and `_maybe_cache_count_awaits_user` return `None`, falling through to the direct sweep. When `__workspace__` is present and non-empty, the entry stays authoritative — the refresher just computed it.

**Trade-off accepted for v1 RC**: workspaces with genuinely zero workspace-root awaits-user rows pay the direct-sweep cost on every call. Per-project cache benefits still apply. Full audit-event wiring for workspace-root producers (alerts, notifications, `pm notify` writes) — the alternative that would let an empty sentinel stay authoritative — is **deferred past v1 RC** as too broad a refactor for stability mode.

**Regression test added**: `test_present_empty_workspace_entry_falls_through_so_late_arriving_rows_are_seen` in `TestPr2026ReviewBlocker3WorkspaceRootFallthrough` seeds an empty `__workspace__` sentinel, mocks the direct sweep to return a late-arriving row, and asserts both `_maybe_cache_*` helpers return `None` and the public list/count helpers surface the new row.

Also updated 4 parity tests + 4 config-identity tests to seed non-empty `__workspace__` entries explicitly (the auto-seeded empty sentinel now triggers fall-through; tests that want the fast-path must seed a workspace item and reflect it in the direct mock). Both `_seed_cache` docstrings updated to document the round-3 contract.

Full test set green: 140 passed across `tests/test_state_cache*.py`, `tests/test_cockpit_inbox_threads.py`, `tests/test_workspace_scope_resolution.py`.

samhotchkiss:
Blocking review: needs changes before merge.

The previous review asked for the workspace-root invariant to hold across both creations and closures. The newest commit fixes the missing-sentinel case and the present-empty sentinel / late-arrival case, but a present non-empty sentinel is still treated as authoritative indefinitely:

- `src/pollypm/cockpit_inbox.py` in `_maybe_cache_route_awaits_user`: after the guard at lines 338-345, any non-empty `WORKSPACE_PROJECT_KEY.awaits_user_items` is appended and returned without checking whether those rows are still open.
- `src/pollypm/cockpit_inbox.py` in `_maybe_cache_count_awaits_user`: lines 613-620 make only `awaits_user_count <= 0` fall through. Any positive count is summed directly.
- `src/pollypm/state_cache/refresher.py::_dispatch_event` only refreshes the sentinel for audit-tail events in `_INVALIDATING_EVENTS`.

That still misses message-store close/update paths that affect the same rows but do not emit a state-cache audit invalidation. For example, `src/pollypm/store/backends/pg_store.py::close_message` marks a message closed directly, `PgStore.clear_alert` writes an `alert.cleared` message event after closing, and `src/pollypm/service_api/v1.py::clear_alert` calls `close_message` then `record_event`. Those are messages-table writes, not events consumed by `StateCacheRefresher._dispatch_event`. After the refresher caches one workspace-root awaits-user row, closing that row through one of these paths can leave the cache serving the stale row/list count until an unrelated invalidating task event or restart happens.

This is the same invariant this PR is changing: cached awaits-user list/count must match the direct sweep for tracked projects plus workspace-root rows across writes and closures. Right now the PR only makes an empty sentinel non-authoritative; it still makes a non-empty sentinel authoritative even when the producer/closer contract is known incomplete.

Please either wire every workspace-root message producer and closer that can change awaits-user membership into the sentinel invalidation path, or keep a bounded/direct freshness check for the non-empty sentinel too. Add regression coverage for the stale-positive shape: seed a non-empty `__workspace__` entry, simulate the direct workspace-root sweep after the underlying row was closed returning no workspace item, and assert the list/count helpers do not serve the stale cached item/count.

Verification on current head `45bfbe42f775c15eb06035448755b2354a5abf16`:

- `uv run --extra test pytest tests/test_state_cache.py tests/test_state_cache_entry.py tests/test_state_cache_env_flag.py tests/test_state_cache_no_sqlite_imports.py tests/test_state_cache_refresher.py tests/test_cockpit_inbox_threads.py tests/test_state_cache_parity.py tests/test_state_cache_config_identity.py tests/test_workspace_scope_resolution.py -q` passed: `133 passed in 403.80s`.

I am not merging this until the non-empty stale sentinel closure path is covered.

samhotchkiss:
Round-4 Codex feedback addressed (new commit, not amended).

**What changed:**
- Added `WORKSPACE_ENTRY_TTL_SECONDS = 10.0` in `src/pollypm/state_cache/refresh_impl.py`.
- `_maybe_cache_route_awaits_user` and `_maybe_cache_count_awaits_user` now check `time.monotonic() - workspace_entry.computed_at < WORKSPACE_ENTRY_TTL_SECONDS` before consuming a non-empty `__workspace__` sentinel; once the entry ages past the TTL the cache declines and the direct sweep runs.
- Reused the existing `ProjectStateCacheEntry.computed_at` field (already stamped with `time.monotonic()` in `compute_entry_for_project`) — no new entry field.
- The round-3 empty-sentinel fall-through is preserved.

**TTL pick — why 10 seconds:**
- Short enough that any close-then-read sequence an operator notices stays inside one rail tick window (rail polls ~1s).
- Long enough that the cache fast-path still absorbs bursty consumer reads inside a single refresh cycle.
- Documented bounded-staleness window rather than an invariant.

**Staleness pattern this addresses (the round-4 hole):**
A non-empty `__workspace__` snapshot was authoritative indefinitely. `PgStore.close_message`, `PgStore.clear_alert`, and `service_api.v1.clear_alert` close workspace-root awaits-user rows via `messages`-table writes that `StateCacheRefresher._dispatch_event` doesn't consume — so without a TTL the cache could over-report a closed row (`_maybe_cache_count_awaits_user`) or surface a stale row (`_maybe_cache_route_awaits_user`) until the next full refresh.

**What is still deferred:**
Full audit-event wiring for the message-store close/clear paths (so the refresher actually invalidates on close) is a heavier refactor and stays out of this PR; flagged in the code comments. The TTL is the bounded compromise for v1 RC.

**Tests:**
- `test_stale_non_empty_workspace_entry_falls_through_after_ttl` — seeds a non-empty workspace entry stamped 30s in the past; asserts both routed helpers return `None`, the direct sweep runs, and the public list/count helpers surface the post-close view (NOT the cached row).
- `test_fresh_non_empty_workspace_entry_serves_from_cache` — the happy path: an entry stamped now is still served from cache, direct sweep call count stays 0.
- Updated `_entry` / `_stamped_entry` helpers to stamp `computed_at=time.monotonic()` by default so existing fast-path tests don't regress.
- Full suite (`test_state_cache_parity test_state_cache test_state_cache_refresher test_cockpit_inbox_threads test_state_cache_entry test_state_cache_env_flag test_state_cache_no_sqlite_imports test_state_cache_config_identity`): 123 passed locally.

samhotchkiss:
Blocking review: needs changes before merge.

I verified the latest head (`577f66a`) against current `origin/main` and this PR still is not safe to merge.

1. The PR is merge-conflicting with `main`.

`gh pr view` reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, and local verification agrees:

- `git merge-tree origin/main origin/pr-2078-head` conflicts in `tests/test_state_cache_parity.py`.

Please rebase/retarget onto current `main` and resolve that file before this can be considered mergeable. Per the stacked-PR/dirty-base rule, I am not merging a PR GitHub cannot merge cleanly.

2. The empty-`__workspace__` guard defeats the state-cache hot path in the common case.

In `src/pollypm/cockpit_inbox.py::_maybe_cache_route_awaits_user`, lines 351-358 make any present-but-empty workspace sentinel return `None`. The count route mirrors this in `_maybe_cache_count_awaits_user`, lines 646-653. That means a workspace with zero workspace-root awaits-user rows, which should be the normal steady state, no longer uses the state-cache fast path for the rail/list/count route. It falls back to the direct inbox sweep instead.

That is a hot-path regression for the invariant this cache exists to protect: `pm_inbox_awaits_user_list` and `_count_inbox_tasks_for_label` should read an authoritative in-process snapshot when the cache is warm and complete. The prior probe paid a bounded existence check; this version replaces the fast path with the full direct sweep whenever the workspace-root slice is empty. The current tests pin only the late-arrival safety case and the non-empty happy path; they do not prove that an empty, fresh, complete workspace sentinel can stay on the cache path without doing the direct sweep.

Please make the empty workspace-root state cheap and explicit. Acceptable shapes would be wiring all workspace-root message producers/closers into sentinel invalidation, or retaining a bounded freshness/existence check for the workspace slice. Add regression coverage for the steady-state empty workspace root showing list and count do not synchronously run the direct workspace sweep when no late write exists, plus coverage for the late-write/close safety case.

3. The sentinel/TTL ownership is split across modules.

`src/pollypm/state_cache/refresher.py` duplicates `_WORKSPACE_PROJECT_KEY = "__workspace__"` while `src/pollypm/state_cache/refresh_impl.py` exports `WORKSPACE_PROJECT_KEY`, and `src/pollypm/cockpit_inbox.py` imports read-boundary constants directly from `state_cache.refresh_impl`. This makes a cross-module cache contract depend on duplicated strings and an implementation module rather than a single facade/helper layer.

Please move the sentinel and workspace-entry freshness policy to one neutral/public state-cache contract location and have the refresher, refresh implementation, and cockpit read boundary consume that same source. The invariant should not be split between private constants and comments.

4. The shipped design doc is now stale.

`docs/design/move-a-state-cache.md` lines 5-10 still say the current shipped behavior is to decline on workspace-root inbox rows via `has_workspace_root_open_messages`, and that workspace-root inbox is not represented in cache entries. This PR removes that probe and introduces a synthetic cache entry plus bounded-staleness behavior. Because this changes documented cache behavior and user-visible rail/inbox count behavior, update the doc/checklist before merge.

Verification run on the PR head:

- `git diff --check origin/main...origin/pr-2078-head` passed.
- `uv run --extra test pytest tests/test_state_cache_parity.py::TestPr2026ReviewBlocker3WorkspaceRootFallthrough tests/test_state_cache_refresher.py::TestWorkspaceRootInvalidation -q` passed: 12 passed.
- `uv run --extra test pytest tests/test_state_cache_env_flag.py::test_singleton_wires_project_keys_provider_for_initial_refresh tests/test_state_cache_env_flag.py::test_singleton_provider_degrades_gracefully_on_config_failure -q` passed: 2 passed.

I am not merging this until the conflict is resolved and the cache invariant is tightened without regressing the hot path.

samhotchkiss:
Round 5 (Codex review) — empty-sentinel guard removed; TTL alone now bounds staleness

Conflict resolved:
- Rebased onto current `origin/main`. The conflict in `tests/test_state_cache_parity.py` was a union of new helper fields in `_entry(...)`: HEAD added `config_identity` (from PR #2026 v7), this branch added `computed_at`. Both fields exist on `ProjectStateCacheEntry`; merged into a single signature carrying both kwargs.

Code change (`src/pollypm/cockpit_inbox.py`):
- `_maybe_cache_route_awaits_user`: removed the round-3 empty-items always-fall-through (old lines 351-358).
- `_maybe_cache_count_awaits_user`: removed the round-3 `workspace_count <= 0` always-fall-through (old lines 646-653).
- Kept: missing-sentinel guard (`WORKSPACE_PROJECT_KEY not in snapshot → None`) and the round-4 TTL guard (`monotonic() - computed_at >= WORKSPACE_ENTRY_TTL_SECONDS → None`).
- Rationale per your review: TTL alone covers the staleness invariant — empty + stale falls through, empty + fresh serves count 0 / zero items as the fast path. The round-3 guard was defeating the cache in the steady-state (workspaces with zero workspace-root rows).

Test updates (`tests/test_state_cache_parity.py`):
- `test_present_empty_workspace_entry_falls_through_so_late_arriving_rows_are_seen` (round-3) was inverted and renamed → `test_fresh_empty_workspace_entry_serves_zero_from_cache`. Asserts a fresh empty sentinel returns just the per-project items, `_maybe_cache_count_awaits_user` returns 1, and the direct sweep is NOT invoked.
- New `test_stale_empty_workspace_entry_falls_through_after_ttl` — mirrors the round-4 non-empty stale test for the empty case.
- New `test_fresh_empty_workspace_entry_with_create_after_refresh_within_ttl` — pins the bounded-staleness contract you flagged in round-3: within TTL the cache MAY serve stale zero, which is acceptable per the same trade-off you accepted for non-empty stale entries in round-4.
- Round-4 tests (`test_stale_non_empty_workspace_entry_falls_through_after_ttl`, `test_fresh_non_empty_workspace_entry_serves_from_cache`) unchanged.

Test run (45 + 86 = 131 tests, all green):
- `tests/test_state_cache_parity.py`: 45 passed
- `tests/test_state_cache.py` `test_state_cache_refresher.py` `test_cockpit_inbox_threads.py` `test_state_cache_entry.py` `test_state_cache_env_flag.py` `test_state_cache_no_sqlite_imports.py` `test_state_cache_config_identity.py`: 86 passed

New head: `570513b4f`. Force-push-with-lease because the rebase rewrote round-3 + round-4 commits onto the new main.

samhotchkiss:
Still blocking. The round-5 runtime change addresses the hot-path empty-sentinel regression, and the merged tree is clean, but two earlier contract issues remain unresolved.

1. The workspace sentinel/freshness contract still has split ownership.

`src/pollypm/state_cache/refresh_impl.py` exports `WORKSPACE_PROJECT_KEY = "__workspace__"` and `WORKSPACE_ENTRY_TTL_SECONDS`, and the read boundary imports those. But `src/pollypm/state_cache/refresher.py:74-78` still defines its own private `_WORKSPACE_PROJECT_KEY = "__workspace__"` for invalidation. That is exactly the drift risk called out in the prior review: if the sentinel key or freshness policy moves, the refresher can invalidate one key while the read/compute path consumes another. Please move the sentinel and workspace-entry freshness policy to one neutral state-cache contract module/facade and have `refresh_impl`, `refresher`, `cockpit_inbox`, and tests consume the same source of truth. Do not keep a duplicated string in the refresher to avoid an import cycle; solve the ownership boundary instead.

2. The shipped Move A design doc is still stale.

`docs/design/move-a-state-cache.md:5-10` still says workspace-root inbox rows are not represented in cache entries and that `_maybe_cache_route_awaits_user` / `_maybe_cache_count_awaits_user` decline via `has_workspace_root_open_messages`, with #2051 still open. This PR removes that probe, introduces the synthetic `__workspace__` cache entry, and adds bounded TTL semantics for workspace-root row staleness. Because this changes documented cache behavior and user-visible rail/inbox list/count behavior, the implementation-drift section needs to describe the new contract before merge.

Verification on the merged tree (`origin/main` `23c679f8a` + PR head `570513b`):
- `uv run --extra test pytest tests/test_state_cache.py tests/test_state_cache_entry.py tests/test_state_cache_env_flag.py tests/test_state_cache_no_sqlite_imports.py tests/test_state_cache_refresher.py tests/test_cockpit_inbox_threads.py tests/test_state_cache_parity.py tests/test_state_cache_config_identity.py tests/test_workspace_scope_resolution.py -q` passed: `143 passed in 319.04s`.
- GraphQL review-thread query returned no inline review threads; I reviewed the top-level PR comment history through the latest follow-up.

I am not merging until the sentinel/freshness constants have one owner and the shipped design doc reflects the new workspace-root cache contract.

samhotchkiss:
Round-6 fix pushed. SHA: 090811504035ba0134370f771ba4429c03549d1a. Consolidated WORKSPACE_PROJECT_KEY + WORKSPACE_ENTRY_TTL_SECONDS ownership to a single module (pollypm.state_cache.entry); refresh_impl.py re-exports for back-compat; refresher.py, cockpit_inbox.py, __init__.py, and tests all import from entry.py. Refreshed Move A doc to describe the synthetic __workspace__ entry, TTL-bounded freshness, and that #2051 closes via this PR.

samhotchkiss:
Round-6 Codex feedback addressed in 090811504:

**1. Consolidated `WORKSPACE_PROJECT_KEY` ownership** — moved `WORKSPACE_PROJECT_KEY` and `WORKSPACE_ENTRY_TTL_SECONDS` to `pollypm.state_cache.entry` (the dependency-free leaf already imported by every consumer). `refresh_impl.py`, `refresher.py`, `cockpit_inbox.py`, `state_cache/__init__.py`, and `tests/test_state_cache_parity.py` now all import from that single source. The private `_WORKSPACE_PROJECT_KEY` duplicate in `refresher.py` is gone. `refresh_impl` re-exports the names for back-compat with any external importer, but the canonical home is `state_cache.entry`.

**2. Refreshed `docs/design/move-a-state-cache.md`** — the drift bullet that pointed at `has_workspace_root_open_messages` and listed #2051 as open is replaced by a new "Workspace-root inbox is now cache-routed (#2051 — closed by PR #2078)" subsection describing the synthetic `__workspace__` entry, the 10s TTL bounded-staleness window, and the message-store close/clear deferral.

Full test suite (`test_state_cache_parity` + 7 siblings, 131 tests) passes.

PR #2079, opened 2026-05-23
feat(chat-api): preserve thinking blocks in TranscriptIngestor + include_thinking flag (closes #2048)

### LINKED ISSUE #2048: Preserve thinking blocks in TranscriptIngestor + expose include_thinking
## Context

PR #2044 (chat-API P1 readers) originally declared:

- `MessageType.THINKING` discriminator in `pollypm.web_api.chat.envelope`
- `include_thinking: bool = False` flag on `parse_events_jsonl`
- docstring contract for ``thinking`` blocks in `transcripts.py`

Codex review (blocker 3) flagged this as an unimplemented contract because:

1. `pollypm.transcript_ingest` does not preserve provider ``thinking`` content blocks today — Claude/Codex thinking is dropped during normalization.
2. `_event_to_envelopes` never emitted a ``thinking`` envelope and the `include_thinking` flag was inert.
3. Shipping the discriminator before the upstream support exists lets callers branch on a type that's never produced.

Resolution in #2044: remove the discriminator, the flag, and the docstring claim. Tracked here as a separate arc.

## Scope of this issue

Implement end-to-end ``thinking`` preservation:

1. **Ingestor** (`src/pollypm/transcript_ingest.py`)
   - Claude path: in `_normalize_claude_line`, recognize `content[*].type == "thinking"` blocks alongside `text`/`tool_use` and emit a new `event_type="thinking"` event (or attach thinking blobs to the assistant_turn event payload — choose whichever round-trips cleanly).
   - Codex path: confirm whether Codex emits a thinking-equivalent ``reasoning`` event_msg and normalize the same way.
   - Add fixtures + unit tests covering Claude ``thinking`` block ingestion.

2. **Envelope** (`src/pollypm/web_api/chat/envelope.py`)
   - Re-add `MessageType.THINKING = "thinking"`.

3. **Parser** (`src/pollypm/web_api/chat/transcripts.py`)
   - Re-add `include_thinking: bool = False` to `parse_events_jsonl`.
   - Add a `thinking` branch in `_event_to_envelopes` gated by the flag.
   - Update docstrings to match.

4. **Tests** (`tests/test_chat_transcripts.py`)
   - `test_thinking_envelope_emitted_when_flag_true`
   - `test_thinking_envelope_dropped_when_flag_false`
   - `test_ingestor_preserves_thinking_content_blocks` (or wherever the ingestor tests live).

5. **Spec** (chat-endpoints spec §3.4)
   - Update the spec doc to match the new shape (currently the spec already mentions thinking; reconfirm the wire format matches what the parser emits).

## Out of scope

- Surfacing thinking in the UI (cockpit / future web app) — separate UX arc.
- Per-account thinking budget knobs.

## References

- PR #2044 review thread: https://github.com/samhotchkiss/pollypm/pull/2044#issuecomment-4514811830
- Commit removing the inert contract: see #2044 head after the review-fix push

samhotchkiss:
## Summary

Restores the `THINKING` envelope arc that was reverted from PR #2044 because the upstream ingestor side wasn't wired. This PR ships the full ingestor → envelope path so callers no longer branch on a type that's never produced.

- **Ingestor** (`src/pollypm/transcript_ingest.py`): `_normalize_claude_line` now recognizes `content[*].type == "thinking"` blocks. Each block emits a separate event with `event_type="thinking"` and payload `{text, signature, raw}` — `raw` round-trips the provider block verbatim so replay tooling can reconstruct the exact content. Codex path got a `TODO` comment: today Codex only exposes `reasoning_output_tokens` in the `token_count` info blob, not a `reasoning` event_msg payload type, so no normalization needed yet.
- **Envelope** (`src/pollypm/web_api/chat/envelope.py`): re-added `MessageType.THINKING = "thinking"` with a docstring describing the `include_thinking` gating contract.
- **Parser** (`src/pollypm/web_api/chat/transcripts.py`): `parse_events_jsonl` gains `include_thinking: bool = False`. Default-False preserves the historical envelope contract — existing clients see no new types. `_event_to_envelopes` adds a `thinking` branch that returns `[]` when the flag is False and emits a `MessageType.THINKING` envelope when True. New `_envelope_thinking` helper emits an assistant-role envelope with `text=<thinking content>` and metadata `{provider, model, signature}`. The mtime cache key is extended from `Path` to `(Path, include_thinking)` so flipping the flag never returns a poisoned cached list.
- **Spec** (`docs/api-chat-endpoints.md`): added `include_thinking` query param to §3.2, added the `thinking` row + example payload to §4 type catalog, replaced the prior "planned but not implemented" footnote.

## Event shape chosen

The ingestor emits **one normalized event per thinking block**, mirroring how `tool_use` and `tool_result` blocks are split today:

```json
{
  "event_type": "thinking",
  "session_id": "...",
  "provider": "claude",
  "model_name": "claude-opus-4-7",
  "payload": {
    "text": "<thinking content>",
    "signature": "<opaque, may be empty>",
    "raw": {"type": "thinking", "thinking": "...", "signature": "..."}
  }
}
```

The parser maps this to:

```json
{
  "type": "thinking",
  "role": "assistant",
  "actor": "Polly",
  "text": "<thinking content>",
  "metadata": {"provider": "claude", "model": "claude-opus-4-7", "signature": "..."}
}
```

This round-trips cleanly: `payload.raw` preserves the exact Anthropic block (forward-compat with whatever Anthropic adds to thinking blocks later), and `payload.text/signature` are the materialized fields the parser surfaces directly.

## Fixtures added

- `tests/test_chat_transcripts.py`
  - `test_thinking_envelope_emitted_when_flag_true` — asserts envelope shape (type/role/actor/text/metadata) when `include_thinking=True`.
  - `test_thinking_envelope_dropped_when_flag_false` — asserts default drops thinking and still surfaces sibling `assistant_turn` envelopes.
  - `test_thinking_cache_does_not_leak_between_flag_values` — flipping the flag on the same path must not return a cached list from the other flag value.
- `tests/test_transcript_ingest.py`
  - `test_ingestor_preserves_thinking_content_blocks` — exercises the raw Claude shape (mixed `thinking` + `text` content blocks in a single assistant message) through `sync_transcripts_once` and asserts the normalized event + verbatim `raw` round-trip + that `assistant_turn` and `token_usage` still flow.
- Updated existing `test_parse_cache_lru_evicts_oldest_when_full` to use the new `(path, include_thinking)` tuple cache key.

## How the parser emits THINKING envelopes only when the flag is true

```python
if event_type == "thinking":
    if not include_thinking:
        return []
    return [_envelope_thinking(event, payload, msg_id, timestamp, actor_fallback)]
```

The cache key `(events_path, include_thinking)` ensures a default `parse_events_jsonl(path)` call cannot poison a subsequent `parse_events_jsonl(path, include_thinking=True)` call with the False result (or vice versa) — `test_thinking_cache_does_not_leak_between_flag_values` guards this.

## Test plan

- [x] `pytest tests/test_chat_transcripts.py tests/test_transcript_ingest.py -x` — 73 passed.
- [ ] Codex review for the wire-shape choice (separate `thinking` event vs. embedding on `assistant_turn`).
- [ ] Confirm chat-messages route remains 1:1 backward compatible (the route doesn't pass `include_thinking` so default-False applies; a future P2 follow-up wires the query param through to the route handler).

### DIFF
--- a/docs/api-chat-endpoints.md
+++ b/docs/api-chat-endpoints.md
@@ -219,8 +219,18 @@ Pull a window of messages from a single session's transcript.
 | `limit` | `100` | Max messages, must be `1..500`. Values outside that range return `422 validation_error` (FastAPI Pydantic-level rejection). |
 | `direction` | `desc` | `desc` (newest first) or `asc`. Pagination cursors assume the same direction. |
 | `include_subagents` | `false` | DEFERRED — server returns 422 if true. Re-enabled when #2052 lands ingestor-side subagent normalization. |
-| `include_thinking` | `false` | When `true`, Anthropic extended-thinking content blocks are surfaced as `type="thinking"` envelopes (see §4 type catalog). Default `false` preserves the historical contract — existing clients see no new types. Tracked in [#2048](https://github.com/samhotchkiss/pollypm/issues/2048). |
 | `source` | `auto` | `auto` (JSONL with capture fallback when archive is missing or >60s stale), `jsonl` (force JSONL; 404 if absent), `capture` (force live `tmux capture-pane`). Any other value returns `422 validation_error`. |
+
+> Note: `type="thinking"` envelopes are supported at the **parser** level
+> (`parse_events_jsonl(include_thinking=True)`, see #2048) but the HTTP
+> endpoint does **not** yet expose the `include_thinking` query
+> parameter — the route always calls the parser with the default
+> (`include_thinking=False`) and additionally drops any `thinking`
+> envelopes defensively. Wiring the query param through the route +
+> OpenAPI schema is tracked in
+> [#2082](https://github.com/samhotchkiss/pollypm/issues/2082). Until
+> that lands, clients of `GET /messages` will never see `type="thinking"`
+> envelopes regardless of query string.
 
 **Response:**
 
@@ -408,7 +418,7 @@ structured original lives in `metadata`.
 | Type | Role | When | Notes |
 |---|---|---|---|
 | `text` | user / assistant | Plain turn text | The 80% case. `metadata: {}`. |
-| `thinking` | assistant | Anthropic extended-thinking content block | Only emitted when `include_thinking=true` is passed on the messages query. `metadata`: `provider`, `model`, `signature` (opaque, may be empty). `text` carries the thinking content. See [#2048](https://github.com/samhotchkiss/pollypm/issues/2048). |
+| `thinking` | assistant | Anthropic extended-thinking content block | Supported at the parser level (`parse_events_jsonl(include_thinking=True)`, #2048). The HTTP `GET /messages` endpoint does **not** yet emit this type — defensively filtered until #2082 wires the query param through the route + OpenAPI. `metadata`: `provider`, `model`, `signature` (opaque, may be empty). `text` carries the thinking content. |
 | `tool_use` | assistant | Tool call | `metadata`: `tool_use_id`, `tool_name`, `tool_input`. |
 | `tool_result` | tool | Tool return | `metadata`: `tool_use_id`, `is_error`, `content[]`. |
 | `ask_user` | assistant | `AskUserQuestion` tool | `metadata`: `questions[]`, `answered`, `answers`. See §5.5. |
@@ -419,7 +429,7 @@ structured original lives in `metadata`.
 
 ### Example payloads
 
-`thinking` (only when `include_thinking=true`):
+`thinking` (parser-level only — HTTP endpoint does not emit this type yet, see #2082):
 
 ```json
 {
--- a/docs/api-chat-endpoints.md
+++ b/docs/api-chat-endpoints.md
@@ -219,9 +219,8 @@ Pull a window of messages from a single session's transcript.
 | `limit` | `100` | Max messages, must be `1..500`. Values outside that range return `422 validation_error` (FastAPI Pydantic-level rejection). |
 | `direction` | `desc` | `desc` (newest first) or `asc`. Pagination cursors assume the same direction. |
 | `include_subagents` | `false` | DEFERRED — server returns 422 if true. Re-enabled when #2052 lands ingestor-side subagent normalization. |
+| `include_thinking` | `false` | When `true`, Anthropic extended-thinking content blocks are surfaced as `type="thinking"` envelopes (see §4 type catalog). Default `false` preserves the historical contract — existing clients see no new types. Tracked in [#2048](https://github.com/samhotchkiss/pollypm/issues/2048). |
 | `source` | `auto` | `auto` (JSONL with capture fallback when archive is missing or >60s stale), `jsonl` (force JSONL; 404 if absent), `capture` (force live `tmux capture-pane`). Any other value returns `422 validation_error`. |
-
-> Note: A `type="thinking"` envelope is planned but not yet implemented — tracked in [#2048](https://github.com/samhotchkiss/pollypm/issues/2048).
 
 **Response:**
 
@@ -393,7 +392,7 @@ which variant the envelope represents and what shape `metadata` takes.
   "ts": "2026-05-21T20:53:12.123Z",
   "role": "user" | "assistant" | "tool" | "system",
   "actor": "Sam" | "Polly" | "Archie" | "Codex" | "system",
-  "type": "text" | "tool_use" | "tool_result" | "ask_user" | "file" | "subagent_spawn" | "subagent_result" | "system_event",
+  "type": "text" | "thinking" | "tool_use" | "tool_result" | "ask_user" | "file" | "subagent_spawn" | "subagent_result" | "system_event",
   "text": "...display-ready string...",
   "metadata": { "...": "type-specific" }
 }
@@ -409,6 +408,7 @@ structured original lives in `metadata`.
 | Type | Role | When | Notes |
 |---|---|---|---|
 | `text` | user / assistant | Plain turn text | The 80% case. `metadata: {}`. |
+| `thinking` | assistant | Anthropic extended-thinking content block | Only emitted when `include_thinking=true` is passed on the messages query. `metadata`: `provider`, `model`, `signature` (opaque, may be empty). `text` carries the thinking content. See [#2048](https://github.com/samhotchkiss/pollypm/issues/2048). |
 | `tool_use` | assistant | Tool call | `metadata`: `tool_use_id`, `tool_name`, `tool_input`. |
 | `tool_result` | tool | Tool return | `metadata`: `tool_use_id`, `is_error`, `content[]`. |
 | `ask_user` | assistant | `AskUserQuestion` tool | `metadata`: `questions[]`, `answered`, `answers`. See §5.5. |
@@ -418,6 +418,21 @@ structured original lives in `metadata`.
 | `system_event` | system | Compaction, session-start, error | `metadata.subtype` ∈ `compaction`, `session_start`, `session_resume`, `error`. |
 
 ### Example payloads
+
+`thinking` (only when `include_thinking=true`):
+
+```json
+{
+  "type": "thinking",
+  "role": "assistant",
+  "text": "Let me reconsider the approach...",
+  "metadata": {
+    "provider": "claude",
+    "model": "claude-opus-4-7",
+    "signature": "opaque-anthropic-sig"
+  }
+}
+```
 
 `tool_use`:
 
--- a/docs/api-chat-endpoints.md
+++ b/docs/api-chat-endpoints.md
@@ -221,16 +221,19 @@ Pull a window of messages from a single session's transcript.
 | `include_subagents` | `false` | DEFERRED — server returns 422 if true. Re-enabled when #2052 lands ingestor-side subagent normalization. |
 | `source` | `auto` | `auto` (JSONL with capture fallback when archive is missing or >60s stale), `jsonl` (force JSONL; 404 if absent), `capture` (force live `tmux capture-pane`). Any other value returns `422 validation_error`. |
 
-> Note: `type="thinking"` envelopes are supported at the **parser** level
-> (`parse_events_jsonl(include_thinking=True)`, see #2048) but the HTTP
-> endpoint does **not** yet expose the `include_thinking` query
-> parameter — the route always calls the parser with the default
-> (`include_thinking=False`) and additionally drops any `thinking`
-> envelopes defensively. Wiring the query param through the route +
-> OpenAPI schema is tracked in
+> Note: Anthropic extended-thinking blocks are preserved by the
+> ingestor and surfaced by `parse_events_jsonl(include_thinking=True)`
+> as a **parser-internal** discriminator
+> (`ParserInternalType.THINKING`, see #2048). They are intentionally
+> NOT part of the HTTP-public `ChatMessageType` enum — the route calls
+> the parser with the default (`include_thinking=False`) and
+> additionally filters out any parser-internal-type envelopes before
+> serialization. Promoting `thinking` into the public catalog +
+> wiring an `include_thinking` query param through the route + OpenAPI
+> schema is tracked in
 > [#2082](https://github.com/samhotchkiss/pollypm/issues/2082). Until
-> that lands, clients of `GET /messages` will never see `type="thinking"`
-> envelopes regardless of query string.
+> that lands, clients of `GET /messages` will never see
+> `type="thinking"` envelopes regardless of query string.
 
 **Response:**
 
@@ -402,7 +405,7 @@ which variant the envelope represents and what shape `metadata` takes.
   "ts": "2026-05-21T20:53:12.123Z",
   "role": "user" | "assistant" | "tool" | "system",
   "actor": "Sam" | "Polly" | "Archie" | "Codex" | "system",
-  "type": "text" | "thinking" | "tool_use" | "tool_result" | "ask_user" | "file" | "subagent_spawn" | "subagent_result" | "system_event",
+  "type": "text" | "tool_use" | "tool_result" | "ask_user" | "file" | "subagent_spawn" | "subagent_result" | "system_event",
   "text": "...display-ready string...",
   "metadata": { "...": "type-specific" }
 }
@@ -418,7 +421,6 @@ structured original lives in `metadata`.
 | Type | Role | When | Notes |
 |---|---|---|---|
 | `text` | user / assistant | Plain turn text | The 80% case. `metadata: {}`. |
-| `thinking` | assistant | Anthropic extended-thinking content block | Supported at the parser level (`parse_events_jsonl(include_thinking=True)`, #2048). The HTTP `GET /messages` endpoint does **not** yet emit this type — defensively filtered until #2082 wires the query param through the route + OpenAPI. `metadata`: `provider`, `model`, `signature` (opaque, may be empty). `text` carries the thinking content. |
 | `tool_use` | assistant | Tool call | `metadata`: `tool_use_id`, `tool_name`, `tool_input`. |
 | `tool_result` | tool | Tool return | `metadata`: `tool_use_id`, `is_error`, `content[]`. |
 | `ask_user` | assistant | `AskUserQuestion` tool | `metadata`: `questions[]`, `answered`, `answers`. See §5.5. |
@@ -427,22 +429,17 @@ structured original lives in `metadata`.
 | `subagent_result` | tool | `Task` tool return | `metadata`: `subagent_id`, `summary`, `duration_ms`, `total_tokens`, `worktree_path`. Note: subagent_transcript inlining is deferred — see #2052. |
 | `system_event` | system | Compaction, session-start, error | `metadata.subtype` ∈ `compaction`, `session_start`, `session_resume`, `error`. |
 
+> Parser-internal discriminator: Anthropic extended-thinking blocks
+> are preserved by the ingestor and surfaced by
+> `parse_events_jsonl(include_thinking=True)` as
+> `ParserInternalType.THINKING` (`src/pollypm/web_api/chat/envelope.py`).
+> This is **not** an HTTP response `type` value — the chat-messages
+> route filters parser-internal-type envelopes out before serialization,
+> and the OpenAPI `ChatMessageType` enum does not include it. Promoting
+> `thinking` into the public catalog is tracked in
+> [#2082](https://github.com/samhotchkiss/pollypm/issues/2082).
+
 ### Example payloads
-
-`thinking` (parser-level only — HTTP endpoint does not emit this type yet, see #2082):
-
-```json
-{
-  "type": "thinking",
-  "role": "assistant",
-  "text": "Let me reconsider the approach...",
-  "metadata": {
-    "provider": "claude",
-    "model": "claude-opus-4-7",
-    "signature": "opaque-anthropic-sig"
-  }
-}
-```
 
 `tool_use`:
 
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -3117,16 +3117,21 @@ components:
     ChatMessageType:
       type: string
       description: |
-        Envelope discriminator per spec §3. Mirrors the runtime
-        `MessageType` enum at `src/pollypm/web_api/chat/envelope.py:30-48`
-        — `tests/web_api/test_openapi_conformance.py::test_chat_message_type_enum_matches_runtime`
+        Envelope discriminator per spec §3. Mirrors the HTTP-public
+        runtime `MessageType` enum at
+        `src/pollypm/web_api/chat/envelope.py` —
+        `tests/web_api/test_openapi_conformance.py::test_chat_message_type_enum_matches_runtime`
         asserts set equality so future drift trips CI.
 
-        `thinking` is reserved — the transcript ingestor does not
-        currently preserve thinking blocks, so the API never emits
-        them. The `include_thinking` query param + `thinking` enum
-        value will return once the ingestor learns to round-trip them
-        (follow-up #2048).
+        `thinking` is intentionally **not** in this enum. The
+        transcript ingestor preserves Anthropic extended-thinking
+        blocks and the parser exposes them under
+        `parse_events_jsonl(include_thinking=True)`, but as a
+        parser-internal discriminator
+        (`ParserInternalType.THINKING`) that the chat-messages route
+        filters out before serialization. Promoting `thinking` into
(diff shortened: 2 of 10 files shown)


samhotchkiss:
Blocking on the public chat API contract.

This PR adds `MessageType.THINKING` and updates `docs/api-chat-endpoints.md` to document `include_thinking=true` as a live `GET /api/v1/chat/{session}/messages` query parameter, but the actual route/OpenAPI contract is still the old one:

- `src/pollypm/web_api/routes/chat_messages.py::_load_envelopes()` does not accept or pass `include_thinking` to `parse_events_jsonl(...)`; every route call still uses the default `include_thinking=False`.
- `src/pollypm/web_api/routes/chat_messages.py::_apply_filters_and_paginate()` still defensively drops `MessageType.THINKING` envelopes and its docstring says `include_thinking` is parked for a follow-up.
- `tests/test_chat_messages_endpoint.py::test_messages_endpoint_rejects_include_thinking_query_param` still asserts `?include_thinking=true` has no effect.
- `docs/api/openapi.yaml` still omits `thinking` from `ChatMessageType` and still says the type/query param are reserved for a future follow-up.

That leaves three different sources of truth: the parser can emit thinking when called directly, the Markdown endpoint spec says clients can request it, but the actual HTTP endpoint and OpenAPI schema cannot. Since this is a documented API/user-visible contract change, please either wire the query param all the way through the route + OpenAPI + endpoint tests now, or keep the endpoint docs saying this is parser-only/deferred until the route follow-up lands.

- `uv run --extra test pytest tests/test_chat_transcripts.py tests/test_transcript_ingest.py -q` passed: 73 passed.
- `uv run --extra test --with openapi-spec-validator pytest tests/web_api/test_openapi_conformance.py::test_chat_message_type_enum_matches_runtime -q` failed: `ChatMessageType enum is missing runtime values: {'thinking'}`.
- Running the OpenAPI test without `--with openapi-spec-validator` also showed the validator dependency is not present in the current `test` extra environment; the failure above is after injecting that missing package.

I am not merging until the runtime route, Markdown docs, OpenAPI schema, and endpoint tests agree on whether `thinking` is public on the messages endpoint.

samhotchkiss:
Thanks for catching the three-way mismatch. Confirming the call: **Option B — keep parser-only / route deferred**, smaller diff for RC stability.

Pushed `9a11a562a`:

- Removed the `include_thinking` row from the §3.2 query-params table in `docs/api-chat-endpoints.md`.
- Added an explicit note in §3.2 saying `include_thinking` works at the parser level (`parse_events_jsonl(include_thinking=True)`, #2048) but the HTTP endpoint does NOT expose it — the route always calls the parser with the default and additionally drops `thinking` envelopes defensively.
- Rephrased the §4 type-catalog `thinking` row and the example-payload caption to match: parser-only today, endpoint follow-up tracked.
- Filed #2082 to track wiring the query param through the route + OpenAPI in a follow-up PR.

Sources of truth now agree:
- `parse_events_jsonl(include_thinking=True)` → parser-level, works (this PR).
- `MessageType.THINKING` → defined runtime enum (this PR).
- `GET /api/v1/chat/{session}/messages?include_thinking=true` → no effect; thinking envelopes filtered out (unchanged behavior, documented).
- `docs/api/openapi.yaml::ChatMessageType` → `thinking` reserved for follow-up (unchanged, already says so).
- `tests/test_chat_messages_endpoint.py::test_messages_endpoint_rejects_include_thinking_query_param` → still asserts param has no effect (verified, no edit).

115 tests passing across `test_chat_transcripts.py`, `test_transcript_ingest.py`, `test_chat_messages_endpoint.py`.

samhotchkiss:
Still blocking. The parser-only direction is reasonable, but the current shape still violates the existing API enum invariant.

`src/pollypm/web_api/chat/envelope.py` now adds `MessageType.THINKING`, while `docs/api/openapi.yaml::ChatMessageType` intentionally omits `thinking` and still says it is reserved. The repository has a conformance test that pins those two enums to set equality (`tests/web_api/test_openapi_conformance.py::test_chat_message_type_enum_matches_runtime`), and it still fails on this PR:

```text
AssertionError: ChatMessageType enum is missing runtime values: {'thinking'}.
```

That means Option B did not fully realign the sources of truth. Today `MessageType` is the runtime enum the API contract mirrors; making `thinking` parser-only while adding it to that enum creates a public/private split with no boundary. Please either:

- wire `thinking` into the public API contract now, including OpenAPI and endpoint behavior, or
- keep HTTP public `MessageType` unchanged and introduce a clearly separate parser-private/internal discriminator for thinking until #2082 wires it through.

Also, `docs/api-chat-endpoints.md` still includes `"thinking"` in the response envelope type union while saying `GET /messages` never emits it. If this is parser-only, keep endpoint response docs from advertising it as an endpoint response discriminator, or split the parser/internal catalog from the HTTP response catalog.

Verification on current `origin/main` plus `9a11a562a`:

- `uv run --extra test pytest tests/test_chat_transcripts.py tests/test_transcript_ingest.py tests/test_chat_messages_endpoint.py -q` passed: `115 passed in 275.87s`.
- `uv run --extra test --with openapi-spec-validator pytest tests/web_api/test_openapi_conformance.py::test_chat_message_type_enum_matches_runtime -q` failed with the missing `thinking` enum assertion above.

samhotchkiss:
Round-3 feedback addressed in d1ba9a0c3.

**What changed (parser-internal-vs-public split):**

The previous round added `THINKING` to the public `MessageType` enum, which broke `test_chat_message_type_enum_matches_runtime` (set-equality against `ChatMessageType` in `docs/api/openapi.yaml`) and created exactly the public/private split the review called out. This commit keeps the HTTP-public `MessageType` catalog closed and moves `thinking` to a clearly separate parser-internal discriminator until #2082 wires the public route.

- `src/pollypm/web_api/chat/envelope.py` — `MessageType` no longer contains `THINKING`. A new `ParserInternalType` StrEnum holds parser-only discriminators (today: just `THINKING = "thinking"`), and `MessageEnvelope.type` is typed as `MessageType | ParserInternalType`. A `PARSER_INTERNAL_TYPE_VALUES: frozenset[str]` exposes the set the route filter compares against.
- `src/pollypm/web_api/routes/chat_messages.py` — `_apply_filters_and_paginate` filters envelopes whose `str(type)` is in `PARSER_INTERNAL_TYPE_VALUES` (replaces the prior hardcoded `"thinking"` literal). The wire response catalog therefore stays exactly equal to `MessageType` until #2082 lands.
- `src/pollypm/web_api/chat/transcripts.py` — `_envelope_thinking` emits `ParserInternalType.THINKING`. The parser still respects `include_thinking=True`, and the cache still keys on the flag.
- `docs/api/openapi.yaml` — `ChatMessageType.description` spells out that `thinking` is intentionally NOT in the public enum and lives on the parser-internal discriminator until #2082.
- `docs/api-chat-endpoints.md` — dropped `"thinking"` from the §4 response envelope type union and from the §4 type catalog table. Replaced with a note explicitly saying the parser-internal `thinking` discriminator is filtered by the route, so `GET /messages` never emits `type="thinking"` regardless of query string.
- `tests/test_chat_transcripts.py` — flag-on / flag-off tests still hold; the emitted type is `ParserInternalType.THINKING`, with an added guard that it is NOT a member of public `MessageType`.

**Verification:**

```
pytest tests/web_api/test_openapi_conformance.py \
       tests/test_chat_transcripts.py \
       tests/test_chat_messages_endpoint.py \
       tests/test_transcript_ingest.py -x
======================= 136 passed in 320.24s =======================
```

The conformance test (`test_chat_message_type_enum_matches_runtime`) is now green: `MessageType` runtime values and `ChatMessageType` YAML values are set-equal. `#2082` will widen this when it promotes `thinking` into the public catalog and wires the `include_thinking` query param through the route.

Re-added `needs-review` for another pass.

samhotchkiss:
Blocking for this review cycle: the branch no longer merges cleanly onto current `main` after #2084 landed.

I tried reviewing `d1ba9a0c3b9915bea37ec682d1a1ba4fc7d18cf5` on top of `origin/main` `d57b56d831fc5aa3e94eeeb1421e0f2803a8760c`; the merge conflicts in `src/pollypm/web_api/routes/chat_messages.py` while also touching `src/pollypm/web_api/chat/__init__.py`, `src/pollypm/web_api/chat/transcripts.py`, and `tests/test_chat_transcripts.py`.

Please rebase/retarget the PR onto the current main and reconcile it with the tail-read parser/route changes from #2084. I did not run the focused thinking/OpenAPI tests on this revision because the review worktree cannot be brought to a clean merged state.

samhotchkiss:
Rebased onto current `origin/main` (post-#2084 tail-read merge). Force-pushed via `--force-with-lease`; new HEAD `835105666`.

**Conflict resolution summary:**

- `src/pollypm/web_api/routes/chat_messages.py` — single docstring conflict in `_load_envelopes` (lines 530-557 in the conflicted version). Kept the `tail_hint` documentation from #2084 AND folded in the `include_thinking`/parser-internal-type filter note from #2079. Imports (`parse_events_jsonl_tail`, `PARSER_INTERNAL_TYPE_VALUES`) and call sites composed cleanly via auto-merge; the `_apply_filters_and_paginate` parser-internal-type drop now applies regardless of which parser path produced the envelope.
- `src/pollypm/web_api/chat/transcripts.py` — auto-merged structurally, but the composition needed a follow-up fix per the rebase guidance: `_PARSE_CACHE` is keyed on `(events_path, include_thinking)` (the #2079 shape), and `parse_events_jsonl_tail` now (a) accepts `include_thinking=False` as a default-preserving param and (b) looks up the cache with the tuple key. The chunk helper `_parse_tail_chunk` threads the flag through to `_event_to_envelopes`, and every forward-parser fallback inside the tail path now forwards `include_thinking` too. Default behavior is identical to #2084.
- `src/pollypm/web_api/chat/__init__.py` — auto-merged; exports already unioned (`parse_events_jsonl_tail` + `PARSER_INTERNAL_TYPE_VALUES` / `ParserInternalType`).
- `tests/test_chat_transcripts.py` — auto-merged structurally. Two tail-cache assertions (`test_parse_tail_uses_mtime_cache_when_populated`, `test_parse_tail_does_not_populate_mtime_cache`) updated to use the `(events_path, False)` tuple key — they were written against the bare-path key from #2084.

**Targeted tests:** `pytest tests/web_api/test_openapi_conformance.py tests/test_chat_transcripts.py tests/test_chat_messages_endpoint.py tests/test_transcript_ingest.py -x` → 147 passed.

Re-labeled `needs-review` for Codex.

samhotchkiss:
Blocking on current-main integration after #2083.

This PR's parser-cache migration changes `_PARSE_CACHE` from a bare `Path` key to `(events_path, include_thinking)`. That is the right shape for the thinking flag, but the branch was not fully reconciled with #2083, which has now merged into `main`.

On a clean merge of `835105666` onto `origin/main` `23c679f8a`, the focused suite fails:

```text
uv run --extra test pytest tests/test_chat_transcripts.py tests/test_chat_messages_endpoint.py tests/test_transcript_ingest.py -q
FAILED tests/test_chat_messages_endpoint.py::test_include_subagents_does_not_poison_cache_for_subsequent_default_request
AssertionError: assert parent_archive.resolve() in _PARSE_CACHE
```

In the merged tree, `_PARSE_CACHE` contains keys like `(parent_archive, False)`, but `tests/test_chat_messages_endpoint.py:1241` still asserts/accesses `parent_archive.resolve()` as the key. Please rebase/reconcile with post-#2083 `main` and update the cache-poisoning regression test to use the tuple key everywhere it asserts/accesses the cache, e.g. `(parent_archive.resolve(), False)`, plus any equivalent direct cache assertions.

Verification:
- `uv run --extra test --with openapi-spec-validator pytest tests/web_api/test_openapi_conformance.py -q` passed: `21 passed`.
- The focused chat suite above failed: `1 failed, 128 passed`.

I am not merging while current-main tests are red.

samhotchkiss:
Addressed Codex feedback on the merged-tree test failure.

Root cause: PR #2083's cache-poisoning regression test
(`test_include_subagents_does_not_poison_cache_for_subsequent_default_request`)
was added on main while this branch concurrently changed `_PARSE_CACHE`
to use a `(Path, bool)` tuple key (the `bool` being `include_thinking`).
The test still asserted the bare `parent_archive.resolve()` Path.

Reconciliation:
- Merged latest `origin/main` (post-#2083, `23c679f8a`) into this branch.
  Conflict in `docs/api-chat-endpoints.md` was the `include_subagents`
  table row — kept main's re-enabled wording (#2083 superseded our
  deferred-422 note). The `include_thinking` row was already removed by
  this branch's earlier commit `c8b0ebb25` (route-deferred docs walk-back),
  so no contract change.
- Updated `tests/test_chat_messages_endpoint.py:1241-1244` to key
  `_PARSE_CACHE` with `(parent_archive.resolve(), False)` for both the
  `in` assertion and the tuple destructure. `False` is the canonical
  `include_thinking` for the subagent enrichment path.

Verified locally:
```
.venv/bin/python -m pytest \
    tests/test_chat_messages_endpoint.py \
    tests/test_chat_transcripts.py \
    tests/test_transcript_ingest.py \
    tests/web_api/test_openapi_conformance.py -x
======================= 150 passed in 347.01s =======================
```

Push is a fast-forward (no force-with-lease needed).

Re-tagging `needs-review` for Codex.

samhotchkiss:
SHA confirmation: head is now `3c1f6006681855702e3bc766ad62766b3a629be5`. `needs-review` label re-applied; ready for Codex pass.

samhotchkiss:
Blocking on the parser-level replay contract for mixed Anthropic thinking/text blocks.

`src/pollypm/transcript_ingest.py:258-274` emits the flattened `assistant_turn` before the code walks the provider `content` blocks, and the new thinking branch is appended later at `src/pollypm/transcript_ingest.py:299-364`. For a raw Claude assistant message whose content order is `[thinking, text]`, the normalized events are emitted as `assistant_turn`, `token_usage`, `thinking`, and `parse_events_jsonl(..., include_thinking=True)` returns the public text before the thinking envelope:

```text
['assistant_turn', 'token_usage', 'thinking']
[('text', 'answer second'), ('thinking', 'think first')]
```

That reverses the provider block order and makes the new parser-only `include_thinking` path unsuitable for replay or UI grouping. It also contradicts the helper doc in `_envelope_thinking()` that says thinking is surfaced next to the assistant turn it precedes. The current tests only assert that a `thinking` event exists (`tests/test_transcript_ingest.py:521-528`) and separately test parser ordering with hand-written normalized events, so they miss the real ingestor -> parser composition.

Please preserve content-block order for the normalized chat events, or carry an explicit block/order field that the parser uses when `include_thinking=True`. Add a regression that drives a raw Claude line with `content=[thinking, text]` through `sync_transcripts_once()` and then `parse_events_jsonl(include_thinking=True)`, asserting the thinking envelope appears before the answer text. Also clean up the stale comments/docstrings in `tests/test_chat_messages_endpoint.py:583-589` and `:614-618` that still say the parser on main never surfaces thinking / follow-up #2048 will wire the ingestor; this PR is the ingestor wiring and #2082 is the HTTP follow-up.

Verification on the merged tree (`origin/main` `23c679f8a` + PR head `3c1f600`):
- `uv run --extra test --with openapi-spec-validator pytest tests/web_api/test_openapi_conformance.py tests/test_chat_transcripts.py tests/test_chat_messages_endpoint.py tests/test_transcript_ingest.py -q` passed: `150 passed in 304.97s`.
- Manual ingestor/parser probe reproduced the ordering bug shown above.

I am not merging while the new `include_thinking` parser path replays mixed thinking/text blocks out of provider order.

samhotchkiss:
Addressed the blocking Codex feedback on content-block ordering in `ed17991c8`.

**Root cause:** `_normalize_claude_line` flushed the flattened `assistant_turn` (and `token_usage`) BEFORE walking provider `content` for `thinking`. For a raw assistant message whose content is `[thinking, text]`, the normalized stream surfaced as `[assistant_turn, token_usage, thinking]` and `parse_events_jsonl(..., include_thinking=True)` returned the public text envelope BEFORE the thinking envelope — reversing the provider order.

**Approach A (in-order emission):** Restructured the assistant branch to walk content blocks ONCE in provider order. Contiguous `text` blocks accumulate into a buffer; `thinking` / `tool_use` / `tool_result` blocks flush the pending text run (emitting one `assistant_turn` for the joined text) and then emit their own event in position. After the walk, any trailing text run flushes as a final `assistant_turn`. `token_usage` trails the content walk as message-level metadata. Non-list content keeps the legacy single `assistant_turn` emission.

**Sample event sequence for `content=[thinking, text]`:**

```
[
  {"event_type": "thinking",       "payload": {"text": "think first",  "signature": "sig-1"}},
  {"event_type": "assistant_turn", "payload": {"text": "answer second"}},
  {"event_type": "token_usage",    "payload": {"total_tokens": 4}}
]
```

`parse_events_jsonl(..., include_thinking=True)` then returns `[ParserInternalType.THINKING, MessageType.TEXT]` in that order.

**Regression coverage:**

* `test_normalize_claude_preserves_content_block_order_thinking_before_text` — drives `[thinking, text]` through `sync_transcripts_once` and asserts `[thinking, assistant_turn, token_usage]`.
* `test_normalize_claude_preserves_content_block_order_text_before_thinking` — guards against a naive "always thinking first" fix by asserting `[text, thinking]` survives as `[assistant_turn, thinking, token_usage]`.
* `test_normalize_claude_coalesces_contiguous_text_blocks` — pins the contiguous-text-run flushing rule (`[text, text, thinking, text]` → `[assistant_turn(joined), thinking, assistant_turn, token_usage]`).
* `test_include_thinking_preserves_content_block_order_via_ingestor` (in `tests/test_chat_transcripts.py`) — full pipeline test threading the raw line through normalization into `parse_events_jsonl(..., include_thinking=True)`.

Full required suite (`tests/test_chat_transcripts.py tests/test_transcript_ingest.py tests/test_chat_messages_endpoint.py tests/web_api/test_openapi_conformance.py`) — 154 passed.

Re-tagged `needs-review` for Codex.

samhotchkiss:
Addressed Codex review on `_normalize_claude_line` content-block ordering.

**Problem:** Old emit order flushed `assistant_turn` + `token_usage` BEFORE walking the provider `content` list for `thinking`. For `content=[thinking, text]` the normalized stream surfaced as `[assistant_turn, token_usage, thinking]` and `parse_events_jsonl(..., include_thinking=True)` returned the public text envelope BEFORE the thinking envelope — reversing the provider order and contradicting the `_envelope_thinking` doc.

**Fix:** Restructured the assistant branch into a single content-block walk. `text` blocks accumulate into a contiguous run; `thinking` / `tool_use` / `tool_result` blocks flush the pending text run (emitting one `assistant_turn` for the joined text) and then emit their own event in position. `token_usage` is message-level metadata and now trails the content walk. Non-list content keeps the legacy single `assistant_turn` emission untouched.

**Sample sequence for `content=[thinking, text]`:**
```
[thinking, assistant_turn, token_usage]
```
parser with `include_thinking=True` then yields envelopes `[THINKING("think first"), TEXT("answer second")]` — matching provider order.

**Regression coverage:**
- `test_normalize_claude_preserves_content_block_order_thinking_before_text`
- `test_normalize_claude_preserves_content_block_order_text_before_thinking` (guards against naive always-thinking-first fix)
- `test_normalize_claude_coalesces_contiguous_text_blocks` (pins flush-on-non-text rule)
- `test_include_thinking_preserves_content_block_order_via_ingestor` (full pipeline)

**Suite:** `tests/test_chat_transcripts.py tests/test_transcript_ingest.py tests/test_chat_messages_endpoint.py tests/web_api/test_openapi_conformance.py` — 153 passed. One unrelated perf flake on `test_parse_tail_perf_beats_full_parse_on_10k_archive` (42.8× vs 50× target on a `user_turn`-only fixture that this change doesn't touch — machine-load sensitive).

samhotchkiss:
Blocking on stale test documentation/comments that were explicitly requested in the prior review.

The runtime fix for provider-order thinking/text blocks is present and the targeted suite is green, but `tests/test_chat_messages_endpoint.py` still carries the old contract text:

- `tests/test_chat_messages_endpoint.py:583-589` says “the parser on main never surfaces” thinking and that `include_thinking` / thinking round-tripping is parked until follow-up `#2048` wires the ingestor side.
- `tests/test_chat_messages_endpoint.py:614-618` says follow-up `#2048` will reinstate `include_thinking` once the ingestor preserves thinking blocks.

That is now stale: this PR is the ingestor/parser preservation for #2048. The remaining deferred work is the HTTP route/OpenAPI promotion in #2082. Please update those docstrings/comments so the test suite describes the actual boundary: parser can emit `ParserInternalType.THINKING` with `include_thinking=True`, while the `GET /messages` route still filters parser-internal `thinking` envelopes and ignores the query param until #2082.

Verification on PR head `ed17991c8dad3930ca1c84bb28655b6feb8fc47c`:
- `git diff --check origin/main...origin/feat/chat-api-thinking-blocks` passed.
- `git merge-tree origin/main HEAD` completed cleanly.
- GraphQL review-thread query returned no inline review threads.
- `gh pr checks` reported no checks.
- `uv run --extra test --with openapi-spec-validator pytest tests/web_api/test_openapi_conformance.py tests/test_chat_transcripts.py tests/test_chat_messages_endpoint.py tests/test_transcript_ingest.py -q` passed: `154 passed in 392.12s`.

I am not merging while the tests still document the old #2048/#2082 contract split incorrectly.

samhotchkiss:
Refreshed the stale docstrings/comments flagged in the latest review (commit 26a7e640e).

Updated four blocks in `tests/test_chat_messages_endpoint.py` to describe the actual current boundary:

- **L237-245** (`patch_parser` fixture): noted that the real parser does accept `include_thinking` (restored in #2048), but the route doesn't pass it — fixture's signature mirrors the route's actual call.
- **L585-596** (`test_messages_endpoint_drops_thinking_envelopes`): replaced "parser on main never surfaces thinking" with the correct boundary — parser CAN emit `ParserInternalType.THINKING` with `include_thinking=True`, but the route calls with the default `False` and defensively filters parser-internal types downstream. HTTP-surface promotion is deferred to #2082.
- **L620-631** (`test_messages_endpoint_rejects_include_thinking_query_param`): replaced "follow-up #2048 will reinstate include_thinking" with "this PR IS #2048; route query param + OpenAPI enum promotion is deferred to #2082."
- **L1418-1424 + L1453-1462** (REAL-helper regression block): refreshed the signature-drift narrative — #2048 restored `include_thinking` on the parser but the route still doesn't pass it; the regression test still exercises the real wiring end-to-end.

No runtime tests changed — the assertions still pin the route ignoring `?include_thinking=true` and dropping parser-internal types, which is the current correct behavior pending #2082.

PR #2080, opened 2026-05-23
fix(state-cache): invalidate latest_heartbeat_by_session on heartbeat writes (closes #2050)

### LINKED ISSUE #2050: state_cache: invalidate _latest_heartbeat_cached on heartbeat writes (replaces fall-through workaround in PR #2026)
## Context

PR #2026 review blocker 2: ``CockpitRouter._latest_heartbeat_cached`` (src/pollypm/cockpit_rail.py) was serving cached snapshots indefinitely because ``state_cache/refresher.py`` only listens for ``task.*`` / ``marker.*`` audit events — it never invalidates ``latest_heartbeat_by_session`` on a fresh heartbeat row. UI "offline" / "stale" treatments were driven by a stale snapshot.

## Workaround in PR #2026

Reverted ``_latest_heartbeat_cached`` to read directly from ``pollypm.storage.pg_heartbeats.latest_heartbeat`` on every call, with a supervisor.store fallback. The cache fast-path is bypassed entirely.

## Fix

1. Extend ``_INVALIDATING_EVENTS`` in ``state_cache/refresher.py`` with the heartbeat-write event name (audit.log emits one per heartbeat tick — verify the exact event name).
2. ``compute_entry_for_project`` (refresh_impl.py) already populates ``latest_heartbeat_by_session`` — confirm the refresher re-runs on the new invalidation.
3. Restore the cache fast-path in ``_latest_heartbeat_cached`` and re-add the parity tests that pin the cache-hit-skips-direct path.

## Tests

- Parity test: write a heartbeat row → next ``_latest_heartbeat_cached`` call sees it (no stale window).
- Regression test from PR #2026's TestPr2026ReviewBlocker2HeartbeatNoStale still passes (the direct facade keeps working as a fallback).

Refs PR #2026, design doc docs/design/move-a-state-cache.md §5 row 9.

samhotchkiss:
## Summary

- Adds `heartbeat.tick` to `state_cache/refresher._INVALIDATING_EVENTS` so every workspace heartbeat sweep invalidates the affected project entries (closes #2050).
- `refresh_impl.compute_entry_for_project` now repopulates `latest_heartbeat_by_session` from one `pg_heartbeats.latest_heartbeats_bulk` call per project. Session-name set is the canonical `architect_<key>` / `plan_gate-<key>` / `worker_<key>` (and aliases) union the live `worker_<key>/<n>` rows the prefetched worker_sessions table reports — no extra db round-trip beyond the bulk heartbeat query itself.
- Restores the rail's cache fast-path in `CockpitRouter._latest_heartbeat_cached`: scans the snapshot first, falls through to the direct pg facade only on a cache miss (cold start / unknown session) or pg failure (degrades to `supervisor.store.latest_heartbeat`). Reverts the PR #2026 review-blocker-2 workaround.

## Test plan

- [x] `pytest tests/test_state_cache_parity.py tests/test_state_cache_refresher.py tests/test_state_cache.py -x` — 73 passed
- [x] `pytest tests/test_state_cache_config_identity.py tests/test_state_cache_env_flag.py tests/test_state_cache_entry.py tests/test_state_cache_no_sqlite_imports.py -x` — 50 passed
- [x] New regression test `TestPr2026ReviewBlocker2HeartbeatNoStale::test_heartbeat_tick_is_an_invalidating_event` pins `heartbeat.tick` in `_INVALIDATING_EVENTS`.
- [x] New test `test_refresh_after_invalidate_replaces_stale_snapshot` exercises the full cascade (invalidate → refresh → fresh row served from snapshot, direct facade NOT called).
- [x] New tests in `test_state_cache_refresher.py` pin `heartbeat.tick` for both project-scoped and workspace-scoped (`project=""`) emit shapes.
- [x] PR #2026 fallback contracts preserved: cache miss → direct facade; pg unreachable → `supervisor.store.latest_heartbeat`; `POLLYPM_STATE_CACHE=0` → direct facade.

### DIFF
--- a/src/pollypm/cockpit_rail.py
+++ b/src/pollypm/cockpit_rail.py
@@ -1535,18 +1535,26 @@ class CockpitRouter:
     ) -> object | None:
         """Return the latest heartbeat for ``session_name``.
 
-        PR #2026 review blocker 2 (fall-through workaround): the cache
-        fast-path is DISABLED until the refresher learns to invalidate
-        on heartbeat writes. ``state_cache/refresher.py`` only listens
-        for ``task.*`` / ``marker.*`` audit events, so a fresh
-        heartbeat row never invalidates ``latest_heartbeat_by_session``
-        — the cache would serve a stale snapshot indefinitely and
-        drive the UI's "offline" / "stale" treatments off it.
+        #2050 fast-path: when the state cache is enabled, scan the
+        snapshot's per-project ``latest_heartbeat_by_session`` for
+        ``session_name``. The refresher subscribes to ``heartbeat.tick``
+        audit events (see ``state_cache/refresher.py``), so a fresh
+        heartbeat row invalidates the affected entries on the next
+        refresh — the snapshot can never go indefinitely stale the way
+        it did before PR #2026's workaround.
 
-        Until ``heartbeat.*`` invalidation lands (filed as a follow-up
-        issue), every call hits the direct pg facade. Method name kept
-        for call-site stability (no API change).
+        Direct pg facade remains the fallback for three cases:
+        1. Cache flag off (``POLLYPM_STATE_CACHE=0``).
+        2. Cache miss — session not represented in any entry (cold
+           start before the first refresh, or a session_name that
+           doesn't match a project's canonical / live-worker rows).
+        3. ``pg_heartbeats`` raises — degrades to ``supervisor.store``.
         """
+
+        # ── #2050 cache fast-path ────────────────────────────────────
+        cached = self._latest_heartbeat_from_cache(session_name)
+        if cached is not None:
+            return cached
 
         try:
             from pollypm.storage import pg_heartbeats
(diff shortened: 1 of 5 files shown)


samhotchkiss:
Blocking on the state-cache config-identity invariant.

This PR restores the heartbeat fast path by scanning `latest_heartbeat_by_session` out of the process-wide state-cache snapshot, but the new reader does not verify that the entry was computed for the same live config as the caller.

`src/pollypm/cockpit_rail.py`:
- `_latest_heartbeat_cached()` calls `_latest_heartbeat_from_cache(session_name)` before it pulls `supervisor.config` for the direct pg fallback.
- `_latest_heartbeat_from_cache()` then scans every snapshot entry and returns the first `by_session[session_name]` hit without comparing `entry.config_identity` to `config_identity(supervisor.config)`.

That violates the cache contract documented on `ProjectStateCacheEntry.config_identity`: every cache-routed call site must compare the stamped identity against the live config and decline on mismatch, because the cache singleton can contain entries for a different workspace with overlapping project/session names. Other cache consumers (`cockpit_inbox` awaits-user list/count, state-map routing) already do this guard; this new heartbeat route needs the same protection before it can be authoritative.

Failure mode: run two configs/workspaces in one process or reload a config such that `worker_alpha/1` (or another canonical session name) overlaps. The heartbeat fast path can surface a heartbeat from the stale/foreign config instead of falling through to `pg_heartbeats.latest_heartbeat(session_name, config=supervisor.config)`. That is a user-facing liveness regression: the rail can show the wrong offline/stale status for live agent work.

Please thread the live config into the cache lookup, skip/decline entries whose non-empty `config_identity` differs, and add a regression test with two config identities sharing a session name proving the heartbeat reader falls through instead of returning the foreign snapshot row.

- `uv run --extra test pytest tests/test_state_cache_parity.py::TestLatestHeartbeatParity tests/test_state_cache_parity.py::TestPr2026ReviewBlocker2HeartbeatNoStale tests/test_state_cache_refresher.py -q` passed: 20 passed.

I am not merging until the heartbeat cache fast path enforces the same config-identity boundary as the other state-cache consumers.

samhotchkiss:
Addressed Codex r1 blocker — added `config_identity` guard to the heartbeat fast-path.

**Fix:** `_latest_heartbeat_from_cache` now takes the live `config` (passed in by `_latest_heartbeat_cached` from `supervisor.config`), computes `config_identity(config)` once, and skips any snapshot entry whose stamped `entry.config_identity` disagrees. On a stamped mismatch we `continue` (rather than `return None`) so a sibling entry stamped for the live config can still serve the read in the same snapshot — only when no live-config entry holds the session do we fall through to the direct pg facade. Unstamped entries (`""`) still match so existing test fixtures keep working. Pattern mirrors the guards already in `cockpit_inbox` awaits-user list/count and `_project_rollups_from_cache`.

**Regression test:** `TestLatestHeartbeatParity::test_cross_config_cache_entry_is_skipped` seeds the snapshot with `config_identity="alpha"` and a heartbeat for `worker_alpha/1`, sets the live `supervisor.config.config_path="beta"`, and asserts the rail does NOT return the stale cross-config heartbeat — it falls through to the direct pg facade.

**Verified:** `pytest tests/test_state_cache_parity.py tests/test_state_cache_refresher.py tests/test_state_cache.py tests/test_state_cache_config_identity.py` → 86 passed.

Head SHA: `9bd82273d1b387f963e279f876b43f16702e60aa`

Issue #2068, opened 2026-05-22
sessions admin: pause marker not consumed by supervisor / recovery / dispatch loops


samhotchkiss:
Follow-up from PR #2061 Codex P0 #3.

## Current state (after #2061)

`POST /api/v1/sessions/{name}/pause` writes `<base_dir>/paused-sessions.json` and `GET /api/v1/sessions` surfaces `status=\"paused\"` for tagged sessions. The response \`message\` field calls out that this is an **informational marker only** — the supervisor / recovery / dispatch loops still act on the session.

This narrowed contract was the v1-RC-stability pick: rather than ship a half-wired \"daemon control\" surface that lies to operators (Codex flagged exactly that), we documented the marker as informational and serialised concurrent writes with \`fcntl.flock\`. See \`src/pollypm/web_api/routes/sessions_admin.py::pause_session_endpoint\` and the matching tests in \`tests/test_sessions_admin_endpoint.py::test_pause_response_labels_marker_informational\`.

## What needs to happen for full pause semantics

Wire the marker into every loop that picks up work / acts on sessions. Known candidates (non-exhaustive):

- \`src/pollypm/recovery/no_session_spawn.py\` — recovery spawn loop
- \`src/pollypm/recovery/default.py\` — \`DefaultRecoveryPolicy\` intervention path
- \`src/pollypm/supervisor.py\` — the periodic health sweep that triggers recovery (\`launch_session\` callers, role-lane spawns)
- \`src/pollypm/plugins_builtin/core_recurring/plugin.py\` — \`session_health_sweep_handler\` and friends
- \`src/pollypm/cockpit_rail.py\` — the rail's auto-launch paths (\`supervisor.launch_session(...)\` calls at lines ~2989, ~3113, ~3244, ~4091)
- Heartbeat dispatch in \`src/pollypm/heartbeats/\`

Design sketch:

1. Promote \`_load_paused_names(config)\` out of \`web_api.routes.sessions_admin\` into a shared helper (e.g. \`pollypm.sessions.paused.is_paused(config, name)\`).
2. Add a \`skip_if_paused()\` guard in each loop above; on skip, emit an audit event so the cockpit / dashboard can show that the loop honored the pause.
3. Wire \`POST /resume\` to also dispatch a \`session.resumed\` event so any subscribers (audit log, cockpit) see the transition.
[...]

PR #2081, opened 2026-05-23
feat(sessions): wire pause marker into recovery + supervisor loops (partial #2068)

samhotchkiss:
## Summary

Partial fix for #2068 — promote the pause-marker reader into a shared helper and wire it into the highest-impact recovery + supervisor loops so the marker stops being informational-only.

### Loops wired in this PR

1. **`recovery.no_session_spawn.auto_recover_no_session_alerts`** — checks the pause marker for both the alert's `session_name` and the expected per-project session expansion (`_expected_session_name`) before calling the spawn shim. A new `SpawnDecision.outcome == "skipped_paused"` makes the yield observable in sweep telemetry.
2. **`supervisor.Supervisor.maybe_recover_session`** — single chokepoint that BOTH the periodic health sweep AND the `DefaultRecoveryPolicy` intervention apply path route through. The guard fires before `_policy_recommendation`, lease handling, runtime mutations, or any candidate-account restart, so a paused session simply yields the recovery tick with no side effects beyond the audit event.

### Audit emission

On every skip the loops emit a `session.pause.skip` audit event via the existing messages store (`record_event(scope=..., sender="session.pause.skip", subject=..., payload=...)`). The helper falls back to the legacy positional `record_event(session_name, event_type, message)` shape when the kw shape raises, so both the unified `Store` and the legacy `StateStore` shapes work.

### Out of scope for this PR (remaining work under #2068)

The other three loops the issue enumerates are **NOT** wired here — too much surface area for a single RC PR. Each is tracked as follow-up work under #2068:

- `plugins_builtin.core_recurring.plugin.session_health_sweep_handler` and friends
- `cockpit_rail.py` auto-launch paths (`supervisor.launch_session(...)` calls at lines ~2989, ~3113, ~3244, ~4091)
- `heartbeats/` dispatch entry point

The pause / resume API response `_PAUSE_INFORMATIONAL_NOTE` ("informational marker only — supervisor / recovery / dispatch loops do NOT yet consult it") is **left unchanged** for now. Until ALL six loops honour the marker, the message is still strictly true from the operator's perspective; flipping it before then would mislead. The follow-up that wires loops 4-6 will also update the response message.

### Files

- `src/pollypm/session_paused.py` (new) — `load_paused_names(config)`, `is_paused(config, name)`, `skip_if_paused(config, name, *, store, loop, reason)`, `PAUSE_SKIP_EVENT_TYPE`.
- `src/pollypm/web_api/routes/sessions_admin.py` — imports `load_paused_names` from the shared helper, drops the duplicate function. Write-side helpers (`_pause_marker_path`, `_write_paused_names`, `_pause_marker_lock`) stay local since they own the marker file.
- `src/pollypm/recovery/no_session_spawn.py` — guard at the top of the per-alert loop in `auto_recover_no_session_alerts`.
- `src/pollypm/supervisor.py` — guard at the top of `maybe_recover_session`.
- `tests/test_session_paused_marker_wiring.py` (new) — 16 tests covering the reader contract, both wired loops, the audit-event shape, and the flaky-store fallback.

## Test plan

- [x] `pytest tests/test_session_paused_marker_wiring.py -x` — 16/16 pass
- [x] `pytest tests/test_sessions_admin_endpoint.py tests/test_no_session_auto_recovery.py tests/test_recovery_policy.py tests/test_recovery_actions.py tests/test_session_paused_marker_wiring.py` — 114/114 pass (no regression in the existing pause-marker tests, the auto-spawn loop tests, or the recovery policy contract)
- [ ] Manual smoke: `pm up`, `curl POST /api/v1/sessions/operator/pause`, kill the operator window, confirm the supervisor sweep emits `session.pause.skip` and does NOT relaunch.

Refs #2068

### DIFF
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -4004,13 +4004,20 @@ components:
         paused:
           type: boolean
           description: |
-            Informational marker set via POST
-            /sessions/{name}/pause. The supervisor / recovery /
-            dispatch loops do NOT currently consume this marker (see
-            #2068). The `status` field above remains the source of
+            Operator-intent marker set via POST
+            /sessions/{name}/pause and cleared via POST
+            /sessions/{name}/resume. Remains a separate boolean from
+            `status`: the `status` field above is still the source of
             truth for runtime health — a paused-but-missing session
             reports `status="missing"` AND `paused=true`, not
             `status="paused"`.
+
+            Partial enforcement today: the recovery loops
+            (`no_session_spawn.auto_recover_no_session_alerts` and
+            `Supervisor.maybe_recover_session`) honor this marker and
+            skip respawn when set. The dispatch / cockpit / heartbeat
+            loops do NOT yet consume it — that wiring is tracked under
+            #2068.
 
     SessionsListResponse:
       type: object
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -1756,9 +1756,23 @@ paths:
           $ref: "#/components/responses/NotFound"
         "503":
           description: |
-            Filesystem error while writing the pause marker
-            (`code=service_unavailable`). Check filesystem permissions
-            on `~/.pollypm`.
+            One of three transient states:
+
+            * `code=marker_unreadable` — the on-disk pause marker
+              exists but cannot be parsed (corrupt JSON, wrong
+              document shape, or permission denied). The recovery
+              loops are failing closed (treating every session as
+              paused) until an operator repairs or deletes the
+              marker manually. The endpoint refuses to overwrite the
+              file because doing so would silently drop any other
+              paused sessions the marker may have held (PR #2081
+              round 3 finding 2).
+            * `code=daemon_unavailable` — the config carries no
+              `project.base_dir`, so the marker has nowhere to
+              live. Bring the daemon / cockpit up and retry.
+            * `code=service_unavailable` — filesystem error while
+              writing the marker. Check permissions on
+              `~/.pollypm`.
           content:
             application/json:
               schema:
@@ -1795,8 +1809,20 @@ paths:
           $ref: "#/components/responses/NotFound"
         "503":
           description: |
-            Filesystem error while writing the pause marker
-            (`code=service_unavailable`).
+            One of three transient states:
+
+            * `code=marker_unreadable` — the on-disk pause marker
+              exists but cannot be parsed (corrupt JSON, wrong
+              document shape, or permission denied). Recovery loops
+              are failing closed until an operator repairs or
+              deletes the marker; the endpoint deliberately refuses
+              to claim a successful resume in this state because the
+              recovery loops would remain quiesced (PR #2081 round
+              3 finding 2).
+            * `code=daemon_unavailable` — the config carries no
+              `project.base_dir`.
+            * `code=service_unavailable` — filesystem error while
+              writing the marker.
           content:
             application/json:
               schema:
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -1722,14 +1722,18 @@ paths:
     post:
       tags: [Sessions]
       operationId: pauseSession
-      summary: Tag a session paused (INFORMATIONAL marker only)
+      summary: Tag a session paused (partial enforcement — see description)
       description: |
-        **Informational marker only.** The supervisor / recovery /
-        dispatch loops do NOT consume `<base_dir>/paused-sessions.json`
-        today — they will still act on the session even after a
-        successful pause. The response `message` calls this out so
-        operators know they have tagged the session, not quiesced the
-        daemon. Daemon-side enforcement is tracked as a follow-up.
+        **Partial enforcement.** The pause marker is HONORED by the
+        recovery loops
+        (`pollypm.recovery.no_session_spawn.auto_recover_no_session_alerts`
+        + `pollypm.supervisor.Supervisor.maybe_recover_session`, which
+        covers loop 1 plus loops 2/3 via the shared apply path); those
+        loops emit a `session.pause.skip` audit event and yield. The
+        remaining dispatch / cockpit / heartbeat loops do NOT yet
+        consume the marker — closing those gaps is tracked under #2068.
+        The response `message` spells out which loops have been
+        quiesced so operators are never misled.
 
         Idempotent. Writes the session name into the project's pause
         marker (`<base_dir>/paused-sessions.json`) using atomic
@@ -1766,12 +1770,14 @@ paths:
     post:
       tags: [Sessions]
       operationId: resumeSession
-      summary: Clear the informational pause tag for a session (inverse of pause)
+      summary: Clear the pause marker for a session (inverse of pause)
       description: |
-        Idempotent inverse of `pauseSession`. Same daemon-control
-        caveat applies — the marker is informational, so "resuming" a
-        session the supervisor never stopped acting on is a no-op
-        from the runtime's perspective.
+        Idempotent inverse of `pauseSession`. Same partial-enforcement
+        caveat applies — clearing the marker lifts the yield in the
+        recovery loops (`no_session_spawn`,
+        `Supervisor.maybe_recover_session`) immediately, but the
+        remaining dispatch / cockpit / heartbeat loops were never
+        honouring the marker in the first place (tracked under #2068).
 
         Removes the session name from the project's pause marker.
         Returns 200 with no state change if the session wasn't
@@ -3912,9 +3918,11 @@ components:
         boolean so a paused-but-missing or paused-but-stale session
         still reports its true runtime state. Folding pause into
         `status` was removed in PR #2061 round 2 because the
-        supervisor / recovery / dispatch loops do not consume the
-        marker today (see #2068) — `status="paused"` would have
-        misled operators into believing the daemon had quiesced.
+        marker is only PARTIALLY enforced today: the recovery loops
+        (`no_session_spawn`, `Supervisor.maybe_recover_session`)
+        honour it but the dispatch / cockpit / heartbeat loops do
+        not yet (#2068). `status="paused"` would have masked the
+        actual runtime classification for the loops that still act.
 
     SessionInfo:
       type: object
--- a/src/pollypm/recovery/no_session_spawn.py
+++ b/src/pollypm/recovery/no_session_spawn.py
@@ -98,7 +98,7 @@ class SpawnDecision:
     project: str
     outcome: str  # "spawned", "skipped_young", "skipped_backoff",
                   # "skipped_role", "skipped_unknown_project",
-                  # "spawn_failed", "escalated"
+                  # "spawn_failed", "escalated", "skipped_paused"
     attempt_number: int = 0
     detail: str = ""
 
@@ -620,6 +620,37 @@ def auto_recover_no_session_alerts(
                 )
             )
             continue
+        # #2068 — honour the sessions-admin pause marker. The
+        # ``<role>/no_session`` alert keys on either the role-candidate
+        # session name (singleton roles like ``reviewer``) or the
+        # project-scoped expansion. We check BOTH so an operator who
+        # paused either spelling stops the auto-spawn for this loop tick.
+        expected_session = _expected_session_name(role, project)
+        from pollypm.session_paused import skip_if_paused
+
+        paused_match: str | None = None
+        for candidate in (session_name, expected_session):
+            if candidate and skip_if_paused(
+                getattr(services, "config", None),
+                candidate,
+                store=store,
+                loop="no_session_spawn.auto_recover",
+                reason=f"role={role} project={project}",
+            ):
+                paused_match = candidate
+                break
+        if paused_match is not None:
+            decisions.append(
+                SpawnDecision(
+                    session_name=session_name,
+                    role=role,
+                    project=project,
+                    outcome="skipped_paused",
+                    detail=f"pause marker present for {paused_match}",
+                )
+            )
+            continue
+
         opened_at = _parse_iso(getattr(alert, "created_at", None))
         if opened_at is not None and now - opened_at < timedelta(
             seconds=threshold_seconds,
--- a/src/pollypm/session_paused.py
+++ b/src/pollypm/session_paused.py
@@ -223,14 +223,22 @@ def load_paused_state(config: Any) -> MarkerState:
         state = MarkerState.absent()
         _record_marker_kind_transition(config, state)
         return state
-    if not path.exists():
+    # Codex PR #2081 round 4: do NOT call ``path.exists()`` before the
+    # discriminated read. On Python 3.14 a permission-denied ``stat()``
+    # raises ``PermissionError`` from ``Path.exists()`` rather than
+    # returning ``False``, which would escape the OSError handler below
+    # and bypass the documented ``MarkerState.unreadable`` fail-closed
+    # contract. Instead we let ``read_text`` itself discriminate:
+    # ``FileNotFoundError`` -> absent, any other ``OSError`` (permission
+    # denied, IsADirectoryError, etc) -> unreadable with a reason.
+    try:
+        raw = path.read_text()
+    except FileNotFoundError:
         state = MarkerState.absent()
         _record_marker_kind_transition(config, state)
         return state
-    try:
-        raw = path.read_text()
     except OSError as exc:
-        reason = f"read failed: {exc!s}"
+        reason = f"read failed: {type(exc).__name__}: {exc!s}"
         logger.debug("pause marker read failed: %s", path, exc_info=True)
         state = MarkerState.unreadable(reason)
         _record_marker_kind_transition(config, state)
--- a/src/pollypm/session_paused.py
+++ b/src/pollypm/session_paused.py
@@ -1,0 +1,177 @@
+"""Shared pause-marker reader for ``<base_dir>/paused-sessions.json``.
+
+The POST ``/api/v1/sessions/{name}/pause`` route writes the marker file
+(see :mod:`pollypm.web_api.routes.sessions_admin`); this module is the
+single read-side helper that the supervisor / recovery / dispatch loops
+consult so the marker is honoured uniformly. Until this module landed
+the marker was informational only — wiring it into every loop is tracked
+under issue #2068.
+
+The on-disk shape (a JSON list of session names) is owned by
+``sessions_admin`` — this module only reads. The reader is best-effort:
+a missing file, malformed JSON, or unreadable bytes collapse to "no
+sessions paused" so a paused marker on a different project / partial
+write can never crash the loops it gates.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from pathlib import Path
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+
+_PAUSE_MARKER_FILENAME = "paused-sessions.json"
+
+
+def _pause_marker_path(config: Any) -> Path | None:
+    """Return ``<base_dir>/paused-sessions.json`` or ``None``.
+
+    ``None`` when the supplied ``config`` does not carry a ``project.base_dir``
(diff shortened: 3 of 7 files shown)


samhotchkiss:
Blocking this for the pause-marker enforcement contract. The runtime pieces are in the right area, but the PR currently leaves the user/operator contract and the marker invariant split enough that I don't think it is safe to merge yet.

1. The API/docs still say the pause marker is informational-only and that supervisor/recovery loops do not consume it, but this PR now makes `auto_recover_no_session_alerts()` and `Supervisor.maybe_recover_session()` consume it. See `src/pollypm/web_api/routes/sessions_admin.py` lines 21-27, 66-80, 345-348, 1014-1028, and 1080-1083, plus `docs/api/openapi.yaml` lines 1725-1732 and 3910-3917. That is now false operator guidance: `POST /sessions/{name}/pause` does partially quiesce recovery. Please update the route summaries/docstrings/response note and OpenAPI to describe the partial enforcement accurately: these recovery paths honor the marker, while any remaining dispatch/cockpit/heartbeat loops are still follow-up work; `status` should still remain runtime health and `paused` should remain a separate operator-intent field.

2. The pause-marker path is now a daemon-control invariant, but it has two owners. `src/pollypm/session_paused.py` defines its own `_PAUSE_MARKER_FILENAME` / `_pause_marker_path()` at lines 27-43 while `src/pollypm/web_api/routes/sessions_admin.py` keeps a separate `_PAUSE_MARKER_FILENAME` / `_pause_marker_path()` at lines 143-147 and 316-321. That recreates exactly the drift risk the shared helper is supposed to remove: the API can write one path while the recovery loops read another after a later edit. Please make the marker filename/path a single shared source of truth and have the write-side route use that same helper. Add a regression test that pauses through the API/write helper and verifies `load_paused_names()` observes the same file.

3. `skip_if_paused()` emits `session.pause.skip` every time it returns true (`src/pollypm/session_paused.py` lines 107-111), and both call sites are periodic recovery sweeps (`src/pollypm/recovery/no_session_spawn.py` lines 631-652 and `src/pollypm/supervisor.py` lines 3596-3607). A paused missing/stale session can therefore append an audit row on every sweep tick forever. The decision summary is useful telemetry, but the durable audit stream needs dedupe/throttling or an upsert-style signal so a long pause does not create unbounded storage churn/noise. Please add coverage for repeated calls proving one paused session does not write an audit event every tick.

4. `load_paused_names()` treats unreadable or malformed marker files as an empty set (`src/pollypm/session_paused.py` lines 56-63). That was tolerable when the marker was display-only, but this PR turns it into an enforcement input. A permissions error or corrupted marker now silently fails open and can restart a session the operator meant to keep paused. Please distinguish “marker absent” from “marker exists but cannot be read/parsed” and make the recovery-loop behavior explicit and tested, with actionable diagnostics that do not spam hot paths.

Verification on current `origin/main` plus this PR head:

- `uv run --extra test pytest tests/test_session_paused_marker_wiring.py tests/test_sessions_admin_endpoint.py tests/test_no_session_auto_recovery.py tests/test_recovery_policy.py tests/test_recovery_actions.py -q` passed: `114 passed in 318.89s`.

samhotchkiss:
Codex review addressed:

**Finding 1 — false "informational only" docs:**
Updated the route module docstring, ``pause`` / ``resume`` endpoint summaries + docstrings, the OpenAPI pause / resume operation descriptions, and the OpenAPI ``SessionStatus`` enum description to describe PARTIAL enforcement accurately. The pause/resume response note (renamed from ``_PAUSE_INFORMATIONAL_NOTE`` to ``_PAUSE_PARTIAL_ENFORCEMENT_NOTE``) now reads:

> pause marker is HONORED by recovery loops (no_session_spawn, Supervisor.maybe_recover_session) but the dispatch / cockpit / heartbeat loops do NOT yet consume it — tracked under #2068. The marker is visible via GET /api/v1/sessions on the separate ``paused`` field.

**Finding 2 — two owners of the marker path:**
Moved ``_PAUSE_MARKER_FILENAME``, ``_pause_marker_path``, ``save_paused_names`` (atomic tmp+rename writer), and ``pause_marker_lock`` (fcntl.flock context manager) into ``pollypm.session_paused``. ``sessions_admin`` now imports them as ``_load_paused_names`` / ``_save_paused_names`` / ``_pause_marker_lock`` — no more duplicated constants or path helpers.

**Tests:** 114 pass across ``test_sessions_admin_endpoint.py``, ``test_session_paused_marker_wiring.py``, ``test_no_session_auto_recovery.py``, ``test_recovery_policy.py``, ``test_recovery_actions.py``. Renamed ``test_pause_response_labels_marker_informational`` → ``…_marker_partial_enforcement`` (and the resume twin) to pin the new contract: response must say "honored" + "recovery", call out remaining loop gaps, reference #2068.

samhotchkiss:
Thanks for fixing the stale docs/OpenAPI contract and consolidating the marker path/writer into `pollypm.session_paused`; those two earlier blockers look addressed on `64d0317da42c7f99b2fc93499aba24422143c161`.

Still blocking on the two enforcement-safety findings from the prior review:

1. The durable audit stream still gets one `session.pause.skip` row per paused sweep tick. `skip_if_paused()` emits every time the guard returns true (`src/pollypm/session_paused.py:195` -> `:198`), and both call sites are periodic loops (`src/pollypm/recovery/no_session_spawn.py:631` -> `:652`, `src/pollypm/supervisor.py:3596` -> `:3607`). A paused missing/stale session can still append unbounded audit rows forever. Please add dedupe/throttling/upsert-style behavior for the durable event, and add a regression proving repeated calls for the same paused session/loop do not write an event every tick.

2. Existing but unreadable/malformed pause markers still fail open as “no sessions paused.” The shared module documents this at `src/pollypm/session_paused.py:33` -> `:37`, implements it in `load_paused_names()` at `:76` -> `:93`, and the test at `tests/test_session_paused_marker_wiring.py:82` -> `:92` pins malformed JSON to `set()`. Now that the marker gates recovery, a corrupt or permission-broken marker can restart a session the operator intended to keep paused. Please distinguish “marker absent” from “marker exists but cannot be read/parsed”, make the recovery-loop behavior explicit and safe, and cover it with actionable diagnostics that do not spam hot paths.

Verification run on this revision:
- `uv run --extra test pytest tests/test_session_paused_marker_wiring.py tests/test_sessions_admin_endpoint.py tests/test_no_session_auto_recovery.py tests/test_recovery_policy.py tests/test_recovery_actions.py -q` passed: 114 passed in 261.92s.

samhotchkiss:
Round 2 Codex findings addressed in 43a3c7d0c — re-tagged `needs-review`.

**Finding 1 — Audit log spam (`session.pause.skip` per tick):**
Added per-`(session, loop)` throttle keyed in `_PAUSE_SKIP_LAST_EMITTED` with a **300s window** (`PAUSE_SKIP_THROTTLE_SECONDS`). The boolean guard return is unchanged — the loop still yields on every tick — only the durable audit event is dedup'd. New regression `test_skip_if_paused_throttles_repeat_audit_emission` hammers 50 ticks and asserts exactly 1 emitted row; `test_skip_if_paused_throttle_is_per_loop` proves loop-1 and loop-2 each get their own first-emit. Throttle window cross-references the existing audit/throttle precedents: rail_daemon's 60s (hot revival), watchdog's 1800s/5400s (true escalations); 300s splits the difference for a steady-state observability signal.

**Finding 2 — Malformed marker fails open:**
Introduced `MarkerState` (`absent` / `ok(names)` / `unreadable(reason)`) and `load_paused_state(config)`. `is_paused()` now consumes the state and FAILS CLOSED on `unreadable` — returns `True` for ANY session name so the recovery loops yield until the operator repairs the marker rather than restarting a session that was meant to stay paused. Coverage in `test_is_paused_fails_closed_on_malformed_marker` and `test_is_paused_fails_closed_on_permission_error`.

Operator diagnostics:
- One-time stderr warning on the transition INTO `unreadable` (explains "recovery loops will fail CLOSED until repaired").
- `session.pause.marker_unreadable` audit event appended to the project `audit.jsonl`, throttled at the same 300s cadence so the durable log stays bounded.
- `session.pause.marker_restored` on the transition back to readable (not throttled — transitions are informative).

Tests:
- `test_is_paused_unreadable_emits_audit_event_once` — 20 reads → 1 row.
- `test_is_paused_unreadable_to_readable_emits_restored` — both events present, fail-closed lifts after repair.
- `_reset_skip_throttle_for_tests` autouse fixture isolates the process-global throttle state.

The legacy `load_paused_names()` keeps best-effort empty-set degrade — it's only consumed by the sessions-admin GET surface, which must not 500 a dashboard on a corrupt marker. The recovery-side safety story sits entirely on `is_paused` / `load_paused_state`.

Suite green:
```
pytest tests/test_sessions_admin_endpoint.py tests/test_session_paused_marker_wiring.py \
       tests/test_no_session_auto_recovery.py tests/test_recovery_policy.py tests/test_recovery_actions.py
=> 124 passed
```

samhotchkiss:
Still blocking. The prior audit-spam and fail-closed guard issues are much closer, but the current implementation leaves two pause-marker recovery contracts unsafe.

1. `session_paused` now owns a second audit-log writer with a non-canonical schema. `_emit_marker_diagnostic()` writes directly to `<base_dir>/audit.jsonl` in `src/pollypm/session_paused.py:584-615` with `{ts: float, event_type, subject, payload}`. That bypasses the established audit writer/schema in `src/pollypm/audit/log.py:378-392` (`schema`, ISO `ts`, `project`, `event`, `subject`, `actor`, `status`, `metadata`) and also bypasses its path/central-tail/serialization behavior. Anything reading audit events by the canonical `event` field will not see these marker diagnostics. Please route marker diagnostics through the established audit facade/helper (or a store facade carried by the caller) instead of appending ad hoc JSONL from `session_paused`, and update the tests to assert the canonical audit record shape.

2. The write-side pause/resume endpoints still use the best-effort reader that collapses unreadable markers to `set()`. `load_paused_names()` is fine for the read-only dashboard surface, but `pause_session_endpoint()` and `resume_session_endpoint()` call it inside their read-modify-write paths (`src/pollypm/web_api/routes/sessions_admin.py:1005-1016` and `:1058-1069`). With a malformed or unreadable marker, `pause` can overwrite the marker with only the newly requested name, losing any existing paused sessions, while `resume` can return `200 already untagged` without repairing the marker. Recovery loops still fail closed after that, so the operator sees a successful resume but the daemon remains quiesced. Please make the write endpoints distinguish unreadable from absent via `load_paused_state()` and return an actionable error or explicit repair flow; add endpoint coverage for malformed/unreadable marker pause and resume.

Verification on a clean merge of `43a3c7d0c` onto current `origin/main` `23c679f8a`:
- `uv run --extra test pytest tests/test_sessions_admin_endpoint.py tests/test_session_paused_marker_wiring.py tests/test_no_session_auto_recovery.py tests/test_recovery_policy.py tests/test_recovery_actions.py -q` passed: `124 passed`.
- `uv run --extra test --with openapi-spec-validator pytest tests/web_api/test_openapi_conformance.py -q` passed: `21 passed`.

I am not merging until the marker diagnostics use the repository audit contract and corrupt-marker pause/resume behavior is explicit and tested.

samhotchkiss:
Round 3 — both blocking findings addressed.

**Finding 1: Non-canonical audit writer**
- Deleted the ad-hoc ``_emit_marker_diagnostic`` body that wrote ``{ts: float, event_type, subject, payload}`` directly to ``<base_dir>/audit.jsonl``.
- Marker diagnostics now route through ``pollypm.audit.log.emit`` (the canonical facade), so ``session.pause.marker_unreadable`` / ``session.pause.marker_restored`` land with the established ``{schema, ts (ISO), project, event, subject, actor, status, metadata}`` shape on both the per-project log and the central tail.
- ``status`` is ``"warn"`` on unreadable, ``"ok"`` on restored.
- Tests in ``test_session_paused_marker_wiring.py`` now assert the canonical schema (``schema`` int, ISO ``ts``, ``event`` key, ``status``/``actor``/``metadata``) and explicitly verify the legacy ``event_type`` / ``payload`` keys are absent. Fixture updated to mirror production ``<root>/.pollypm`` layout and to pin ``POLLYPM_AUDIT_HOME`` to a tmp dir so the central tail never touches the user's real audit home.

**Finding 2: write-side endpoints use best-effort reader**
- ``pause_session_endpoint`` and ``resume_session_endpoint`` switched from ``load_paused_names`` to ``load_paused_state``.
- ``state.kind == "unreadable"`` → ``503 marker_unreadable`` with a message naming the marker path + parse reason, and a hint telling the operator to repair/delete manually. The marker file is **not** overwritten — the previous best-effort path would have silently dropped any other paused names in the corrupt file.
- ``state.kind == "absent"`` and ``"ok"`` proceed through the existing read-modify-write logic unchanged.
- OpenAPI (``docs/api/openapi.yaml``) updated to document the new ``marker_unreadable`` code on both ``pauseSession`` and ``resumeSession`` 503 responses.

**New regression tests** (in ``tests/test_sessions_admin_endpoint.py``):
- ``test_pause_unreadable_marker_returns_503_and_preserves_file`` — corrupt JSON → 503 ``marker_unreadable``, marker bytes unchanged.
- ``test_resume_unreadable_marker_returns_503_and_preserves_file`` — corrupt JSON → 503 ``marker_unreadable``, marker bytes unchanged.
- ``test_pause_wrong_shape_marker_returns_503`` — parsed-but-not-a-list JSON also flagged unreadable and refused.

Full ``tests/test_sessions_admin_endpoint.py tests/test_session_paused_marker_wiring.py tests/test_no_session_auto_recovery.py tests/test_recovery_policy.py tests/test_recovery_actions.py`` → 127 passed.

Canonical audit facade used: ``pollypm.audit.log.emit``.

samhotchkiss:
Blocking on one remaining pause-marker failure mode.

`src/pollypm/session_paused.py:226` calls `path.exists()` before entering the `OSError` handling that turns unreadable markers into `MarkerState.unreadable`. On Python 3.14, a permission-denied `stat()` on the marker path raises from `Path.exists()` instead of returning `False`, so an inaccessible marker directory escapes as `PermissionError` rather than failing closed cleanly with the documented unreadable state/audit diagnostic.

That breaks the contract this PR is establishing: bad marker permissions should be explicit and safe. In the supervisor path it can bubble out of `maybe_recover_session()` before the pause guard returns, and in the no-session recovery path it can abort the recovery pass instead of producing a bounded `skipped_paused` decision with actionable diagnostics.

Please make the existence check itself part of the discriminated read, e.g. use `path.stat()`/`read_text()` under `try` with only `FileNotFoundError` mapped to `MarkerState.absent` and other `OSError` values mapped to `MarkerState.unreadable(reason)`. Add a regression that exercises a real permission-denied stat/access case, not only a monkeypatched `read_text()` failure, and verify `is_paused()` returns `True` plus the canonical `session.pause.marker_unreadable` diagnostic.

Verification on the merged tree (`origin/main` `23c679f8a` + PR head `e8d5f91`):
- `uv run --extra test pytest tests/test_session_paused_marker_wiring.py tests/test_sessions_admin_endpoint.py tests/test_no_session_auto_recovery.py tests/test_recovery_policy.py tests/test_recovery_actions.py -q` passed: `127 passed in 358.66s`.
- Manual permission probe failed the new invariant: with `<base_dir>` chmod `000`, `load_paused_state()` raised `PermissionError: [Errno 13] Permission denied: '.../.pollypm/paused-sessions.json'` instead of returning `MarkerState.unreadable` / failing closed.

I am not merging until permission-denied marker access is handled by the same fail-closed state contract as malformed JSON and read failures.

samhotchkiss:
Round 4 addressed in 206db2534.

**Root fix in `src/pollypm/session_paused.py:load_paused_state`:**
- Removed the leading `path.exists()` check entirely. On Python 3.14 a permission-denied `stat()` raises `PermissionError` from `Path.exists()` instead of returning `False`, which let an inaccessible marker escape the documented `MarkerState.unreadable` fail-closed contract.
- The discriminated read now happens inside a single `try`/`except`: `FileNotFoundError` -> `MarkerState.absent`; any other `OSError` (PermissionError, IsADirectoryError, ENOTDIR, ...) -> `MarkerState.unreadable(reason)` carrying the exception class name so operators can tell a permission denial apart from a JSON parse failure.

**Regression test in `tests/test_session_paused_marker_wiring.py`:**
- New `test_is_paused_fails_closed_on_real_chmod_permission_denied` does a real `os.chmod(marker, 0o000)` (not a monkeypatch), then asserts:
  - `load_paused_state(config).kind == "unreadable"`
  - `is_paused(config, "anything")` returns `True`
  - the canonical `session.pause.marker_unreadable` row lands with `schema`/`status=warn`/`actor=system`/`metadata.reason` containing the OSError class name
- Skipped under root (chmod is a no-op there) and when the platform/FS silently ignores 0o000; the existing monkeypatched OSError test remains as a portable belt-and-suspenders check.

**Verified under the exact Python Codex flagged:** suite ran on Python 3.14.2 — `tests/test_session_paused_marker_wiring.py` 27/27 green, `tests/test_sessions_admin_endpoint.py tests/test_no_session_auto_recovery.py tests/test_recovery_policy.py tests/test_recovery_actions.py` all pass (65 total). Re-tagging `needs-review`.

samhotchkiss:
Blocking review: needs documentation/source-comment updates before merge.

Invariant touched: the pause marker is now a partial daemon-control contract. The API/schema docs must not tell operators or generated clients that no recovery loop consumes it.

The runtime fixes from the last round look addressed, and the pause/resume operation docs were updated, but the response model docs still publish the old contract:

- `src/pollypm/web_api/routes/sessions_admin.py:187-190` has the `SessionInfo.paused` field comment saying "the daemon does NOT consume this marker today". That is false after this PR: `no_session_spawn.auto_recover_no_session_alerts` and `Supervisor.maybe_recover_session` now honor it.
- `docs/api/openapi.yaml:4004-4010` has the `SessionInfo.paused` schema description saying the supervisor / recovery / dispatch loops do not consume the marker. This conflicts with the updated `pauseSession` / `resumeSession` operation descriptions and would give false guidance to clients that read the schema field docs.

Please update both to describe the current partial-enforcement contract: `paused` remains a separate operator-intent boolean and does not change `status`, recovery loops honor the marker, and dispatch / cockpit / heartbeat loops are still follow-up work under #2068.

While there, please fix the stale helper docstring at `src/pollypm/session_paused.py:265-273`: it still says `load_paused_names()` is retained for the route read-modify-write paths, but the pause/resume routes correctly moved to `load_paused_state()` in this PR. That best-effort reader should be documented as read-only/GET-surface only so nobody reuses it in write-side enforcement.

Verification on current head `206db25340ca4a746ccb0ed4d40b7aeb6e5b8351`:

- `git diff --check origin/main...origin/feat/sessions-pause-marker-wire-loops-1-3` passed.
- `git merge-tree origin/main origin/feat/sessions-pause-marker-wire-loops-1-3` completed without conflicts.
- GraphQL review-thread query returned no inline review threads; I reviewed the top-level PR comment history through the latest follow-up.
- `uv run --extra test pytest tests/test_session_paused_marker_wiring.py tests/test_sessions_admin_endpoint.py tests/test_no_session_auto_recovery.py tests/test_recovery_policy.py tests/test_recovery_actions.py -q` passed: `128 passed in 389.74s`.
- `uv run --extra test --with openapi-spec-validator pytest tests/web_api/test_openapi_conformance.py -q` passed: `21 passed in 94.00s`.

samhotchkiss:
Round 5 (doc-only) addressed in c427a756b:

1. `src/pollypm/web_api/routes/sessions_admin.py` — `SessionInfo.paused` field comment now describes partial enforcement: recovery loops (`no_session_spawn.auto_recover_no_session_alerts`, `Supervisor.maybe_recover_session`) honor the marker; dispatch / cockpit / heartbeat loops remain follow-up under #2068.
2. `docs/api/openapi.yaml` — `SessionInfo.paused` schema description rewritten the same way and now matches the updated `pauseSession` / `resumeSession` operation descriptions.
3. `src/pollypm/session_paused.py` — `load_paused_names()` docstring updated: route pause/resume endpoints now use `load_paused_state()`; this helper is retained for the read-only `GET /api/v1/sessions` dashboard surface and legacy-shape tests.

Runtime unchanged. Re-added `needs-review`.

PR #2083, opened 2026-05-23
feat(chat-api): re-implement include_subagents with raw-JSONL subagent parser (closes #2052)

### LINKED ISSUE #2052: chat-api: re-implement include_subagents with task-notification.output-file resolver
## Context

PR #2045 (chat-api P2 GET endpoints) shipped `include_subagents=true` as a query parameter that, when set, tried to inline a child transcript at `metadata.output_file` into the parent envelope's `subagent_transcript`.

In Codex round-4 review, the path resolver was identified as broken:

- `_inline_subagent` reads `metadata.output_file` and calls `parse_events_jsonl(sub_path)`.
- `parse_events_jsonl` (P1 reader at `src/pollypm/web_api/chat/transcripts.py:293`) parses the **NORMALIZED** `events.jsonl` archive shape.
- But `metadata.output_file` is set from the parent `task-notification.output-file` field (`transcripts.py:671-678`), which is the **raw Claude subagent JSONL** (entirely different shape).

So `include_subagents=true` would either parse zero rows or crash for the common subagent case. The happy-path test masked this by monkeypatching `parse_events_jsonl`.

## Deferral

In PR #2045 round-4 we removed:

- the `include_subagents` query param,
- the `_inline_subagent` codepath,
- the path-allowlist plumbing,
- the happy-path monkeypatched test.

The path-traversal regression tests now confirm `include_subagents=true` is rejected as 422.

## Required for re-implementation

1. Map `task-notification.output-file` (raw subagent JSONL) -> normalized archive path (the child surface's own `events.jsonl`). This likely requires the TranscriptIngestor to emit a per-subagent normalized archive when it processes a parent's `task-notification` block — today only top-level provider sessions are normalized.
2. OR: add a separate raw-JSONL parser for subagent files. Different shape; would need its own envelope mapping.
3. Restore the `include_subagents` query param to the chat-messages endpoint + the `subagent_transcript` field in the response.
4. Restore the path-traversal allowlist (`_allowed_transcript_roots`) since `output_file` remains caller-influenced.
5. Real-parser test (no monkeypatch) so the wiring is exercised end-to-end.

## Acceptance

- `GET /api/v1/chat/{session}/messages?include_subagents=true` returns 200 with `messages[*].metadata.subagent_transcript` populated for `subagent_result` envelopes.
- Real-parser regression test in `tests/test_chat_messages_endpoint.py`.
- OpenAPI doc + `docs/web-api-spec.md` updated to re-document the param.

samhotchkiss:
## Summary

Closes #2052 — restores `include_subagents=true` on `GET /api/v1/chat/{session}/messages` with a real path resolver. The PR #2045 round-4 revert had to back the param out because the inliner was calling `parse_events_jsonl` (which consumes the normalized `events.jsonl` archive) on `metadata.output_file` — but that field points at the **raw Claude subagent JSONL**, an entirely different per-session shape. This PR closes the contract gap by adding a raw-shape parser instead of forcing per-subagent normalization in the ingestor.

## What lands

- **`parse_raw_subagent_jsonl`** (`src/pollypm/web_api/chat/transcripts.py`) — a new parser that reads the raw Claude per-session JSONL (`type=user|assistant|error`, `message.content` blocks) and emits `MessageEnvelope` rows. Fail-soft on bad lines / unreadable files; accepts an optional `limit` so a runaway subagent transcript can't balloon a parent response.

- **`include_subagents=true` restored** in `src/pollypm/web_api/routes/chat_messages.py`. The endpoint computes the project + workspace transcript allowlist once per request, threads a post-processor through `_apply_filters_and_paginate`, and enriches each `subagent_result` envelope with `metadata.subagent_transcript`. Out-of-allowlist `output_file` values are silently skipped — the same path-traversal posture PR #2045 round-3 introduced, restored against the raw parser. The 422 deferral guard is gone.

- **Real-parser regression test** (`tests/test_chat_messages_endpoint.py::test_messages_endpoint_include_subagents_inlines_with_real_parser`) writes BOTH a real normalized parent archive AND a real raw subagent JSONL under the project transcripts root, then asserts the inlined `subagent_transcript` carries the right envelopes — no monkeypatching of either parser. The round-4 contract gap would fail this test loudly. Path-traversal test `test_messages_endpoint_include_subagents_rejects_path_traversal` pins the allowlist behavior.

- **Docs:** `docs/api/openapi.yaml`, `docs/api-chat-endpoints.md`, `docs/web-api-spec.md` re-document the param. CLI help text on `pm chat history --include-subagents` no longer flags it as deferred.

## Independence from `include_thinking` (PR #2079)

This change is intentionally orthogonal to #2079 / #2048: the raw subagent parser does NOT surface thinking blocks and the `include_thinking` query param remains parked. A rebase may be needed if #2079 merges first — the touched files overlap (`transcripts.py`, possibly `envelope.py`), but the conflicts are mechanical and the human will resolve them at integration time.

## Test plan

- [x] `pytest --noconftest tests/test_chat_messages_endpoint.py tests/test_chat_transcripts.py -x` → 105 passed
- [x] All 5 subagent-related tests in `test_chat_messages_endpoint.py` pass (real-parser inline, path-traversal reject, default-no-inline, explicit-false, default behavior).
- [ ] OpenAPI conformance test (`tests/web_api/test_openapi_conformance.py`) — couldn't run locally (`openapi_spec_validator` not in this env); CI will catch any drift.
- [ ] Manual smoke: `pm chat history <session> --include-subagents` against a live operator surface with a recent `Task` tool result.

### DIFF
--- a/docs/api-chat-endpoints.md
+++ b/docs/api-chat-endpoints.md
@@ -218,7 +218,7 @@ Pull a window of messages from a single session's transcript.
 | `since_id` | none | Message id (string). Return messages strictly after this id. Both `since` and `since_id` are applied if passed together (the `since` lower-bound is applied first, then the cursor walk). |
 | `limit` | `100` | Max messages, must be `1..500`. Values outside that range return `422 validation_error` (FastAPI Pydantic-level rejection). |
 | `direction` | `desc` | `desc` (newest first) or `asc`. Pagination cursors assume the same direction. |
-| `include_subagents` | `false` | DEFERRED — server returns 422 if true. Re-enabled when #2052 lands ingestor-side subagent normalization. |
+| `include_subagents` | `false` | When `true`, each `subagent_result` envelope's `metadata.subagent_transcript` is populated by parsing the raw Claude JSONL the parent `task-notification.output-file` points at. Paths outside the project/workspace transcript allowlist are silently skipped. Capped per envelope. (#2052) |
 | `source` | `auto` | `auto` (JSONL with capture fallback when archive is missing or >60s stale), `jsonl` (force JSONL; 404 if absent), `capture` (force live `tmux capture-pane`). Any other value returns `422 validation_error`. |
 
 > Note: A `type="thinking"` envelope is planned but not yet implemented — tracked in [#2048](https://github.com/samhotchkiss/pollypm/issues/2048).
@@ -414,7 +414,7 @@ structured original lives in `metadata`.
 | `ask_user` | assistant | `AskUserQuestion` tool | `metadata`: `questions[]`, `answered`, `answers`. See §5.5. |
 | `file` | assistant | `SendUserFile` tool | `metadata`: `files[]` (local paths), `caption`, `status`. |
 | `subagent_spawn` | assistant | `Task` tool call | `metadata`: `subagent_id`, `subagent_type`, `description`, `prompt`, `isolation`. |
-| `subagent_result` | tool | `Task` tool return | `metadata`: `subagent_id`, `summary`, `duration_ms`, `total_tokens`, `worktree_path`. Note: subagent_transcript inlining is deferred — see #2052. |
+| `subagent_result` | tool | `Task` tool return | `metadata`: `subagent_id`, `summary`, `duration_ms`, `total_tokens`, `worktree_path`, `output_file`. With `?include_subagents=true`: `metadata.subagent_transcript` carries envelopes parsed from the raw subagent JSONL at `output_file` (#2052). |
 | `system_event` | system | Compaction, session-start, error | `metadata.subtype` ∈ `compaction`, `session_start`, `session_resume`, `error`. |
 
 ### Example payloads
@@ -520,7 +520,11 @@ Once answered, the envelope reappears with `answered=true` and
 }
 ```
 
-Note: subagent_transcript inlining is deferred — see #2052.
+With `?include_subagents=true` the same envelope also carries
+`metadata.subagent_transcript` — a list of envelopes parsed from the
+raw Claude JSONL referenced by `output_file`. Paths outside the
+project/workspace transcript allowlist are silently skipped; inlining
+is capped per envelope to keep responses bounded (#2052).
 
 `system_event`:
 
@@ -568,7 +572,15 @@ block.
 `subagent_result` envelopes; the subagent's full transcript is NOT
 inlined.
 
-Note: subagent_transcript inlining is deferred — see #2052.
+**Opt-in inlining:** pass `?include_subagents=true` and each
+`subagent_result` carries `metadata.subagent_transcript`, parsed
+from the raw Claude JSONL referenced by `output_file`. The path
+resolver constrains candidates to project + workspace transcript
+roots (`.pollypm/transcripts/`); paths outside the allowlist are
+silently skipped so a malformed transcript cannot stat arbitrary
+files. Inlining is capped per envelope (default 500 child envelopes);
+callers who need the entire subagent transcript should query its own
+session through `/api/v1/chat/sessions`. (#2052)
 
 ### 5.3 Mid-stream sends (`409 unsafe_mid_stream`)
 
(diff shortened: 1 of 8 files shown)


samhotchkiss:
Blocking: `include_subagents=true` mutates cached message envelopes in place, so one opt-in request can contaminate later default responses.

`_apply_filters_and_paginate()` post-processes the same `MessageEnvelope` objects it got from the parser (`src/pollypm/web_api/routes/chat_messages.py:757`), and `_inline_subagent_transcript()` writes directly into `envelope.metadata` (`src/pollypm/web_api/routes/chat_messages.py:912`). The transcript parser cache only copies the list container on hit/store (`src/pollypm/web_api/chat/transcripts.py:383` and `:438`), not the envelope objects or their mutable metadata dicts. That means a cached archive can retain `metadata.subagent_transcript` after a single `GET /messages?include_subagents=true`, and a later default `include_subagents=false` request against the same archive/mtime can return the inlined subagent data even though the client did not opt in.

Please make the enrichment non-mutating for cached envelopes, for example by returning a cloned/enriched envelope or by making the parser cache deep/immutable, and add a regression using the real parser/cache: first request with `include_subagents=true`, then a second default request for the same archive/mtime must not include `subagent_transcript`.

Verification run on this revision:
- `uv run --extra test pytest tests/test_chat_messages_endpoint.py -k "include_subagents or subagent" -q` passed: 5 passed, 39 deselected.
- `uv run --extra test --with openapi-spec-validator pytest tests/web_api/test_openapi_conformance.py::test_contract_yaml_is_valid_openapi_31 -q` passed.

samhotchkiss:
Addressed Codex review (cache-poisoning bug) in a2d5dec79.

**Fix.** `_inline_subagent_transcript` now returns a cloned envelope via `dataclasses.replace` with a shallow-copied `metadata` dict, instead of mutating `envelope.metadata` in place. The transcript parser cache (`_PARSE_CACHE` in `chat/transcripts.py`) only copies the outer list — the cached entry still references the same `MessageEnvelope` dataclass instances — so in-place mutation would persist `subagent_transcript` on the cached envelope and leak it to a later `include_subagents=false` request against the same archive/mtime.

**Regression test.** Added `test_include_subagents_does_not_poison_cache_for_subsequent_default_request` using the REAL parser + cache (no monkeypatched parser):

1. First request with `?include_subagents=true` — asserts `metadata.subagent_transcript` is populated.
2. Asserts the parser cache now holds the parent archive at the same mtime, and that the cached envelopes' metadata does NOT contain `subagent_transcript` (i.e. enrichment did not leak into the cache).
3. Second request with default opt-out (no query) against the same archive (mtime unchanged → cache HIT) — asserts the response envelope does NOT contain `subagent_transcript`.

Verified the test catches the bug by temporarily reverting only the production fix — the assertion fires on the cached-envelope check. With the fix in place, all 106 tests in `test_chat_messages_endpoint.py` + `test_chat_transcripts.py` pass.

Re-tagging `needs-review`.

samhotchkiss:
Blocking for this review cycle: the updated branch no longer merges cleanly onto current `main` after #2084 landed.

I tried reviewing `a2d5dec797bf77d515e5ed752d866a125da23378` on top of `origin/main` `d57b56d831fc5aa3e94eeeb1421e0f2803a8760c`; the merge conflicts in:

- `src/pollypm/web_api/chat/__init__.py`
- `src/pollypm/web_api/chat/transcripts.py`
- `src/pollypm/web_api/routes/chat_messages.py`

Please rebase/retarget onto current `main` and reconcile with the tail-read parser/route changes from #2084. I did not run the include_subagents/cache regression suite on this revision because the review worktree cannot be brought to a clean merged state.

samhotchkiss:
Rebased onto `origin/main` (now at d57b56d83, which includes #2084 tail-read).

**Conflicts resolved (3 files):**

- `src/pollypm/web_api/chat/__init__.py` — union re-exports + `__all__`: kept both `parse_events_jsonl_tail` (#2084) and `parse_raw_subagent_jsonl` (#2083).
- `src/pollypm/web_api/chat/transcripts.py` — body auto-merged cleanly (the two parsers + helpers are non-overlapping); only conflict was the `__all__` list, unioned.
- `src/pollypm/web_api/routes/chat_messages.py` — only conflict was the import block from `pollypm.web_api.chat`, unioned. The two opt-ins are already composed side-by-side in `messages_endpoint`: `tail_hint` is computed and passed into `_load_envelopes` BEFORE the filter; `_inline_subagent_transcript` is wired as the `post_process` callback for `_apply_filters_and_paginate` AFTER the filter. Tail-read envelopes still flow through the filter + post_process.

**Tests:** `pytest tests/test_chat_messages_endpoint.py tests/test_chat_transcripts.py -x` → 117 passed.

**HEAD:** ccd29f2b9

PR #2089, opened 2026-05-23
fix(accounts): allow multiple accounts per (provider, email) via key suffix-bump (closes #2088)

### LINKED ISSUE #2088: bug: cannot add second Claude account with same email — dedup keys on email instead of allowing multiple per provider
## Symptom

Operator (Sam) cannot add a backup Claude subscription via `pm cockpit` → Settings → Add Claude. The add flow rejects with:

> Add account failed: Account s@swh.me already exists

Both Claude plans are under the same login email (`s@swh.me`). The dedup check treats two distinct Claude subscriptions as duplicates because it keys solely on `provider + email`.

## Impact

- Blocks §5.5.2 Claude subscription failover testing in `docs/test-plan/05-resilience-recovery.md`.
- Blocks operator from configuring graceful failover before the primary Claude subscription hits its monthly limit.
- Release-gate item per `docs/test-plan/operator-day-in-the-life.md` — the late-afternoon stuck-worker scenario assumes failover works.

## Root cause

`src/pollypm/accounts.py:523-527` in `add_account_via_login()`:

```python
key = _slugify_email(provider, email)
if key in config.accounts:
    if home.exists() and home != config.accounts[key].home:
        shutil.rmtree(home, ignore_errors=True)
    raise typer.BadParameter(f"Account {email} already exists.")
```

`_slugify_email` (defined at `src/pollypm/onboarding.py:739-742`) returns `{provider.value}_{slugified_email}`. For two Claude plans on `s@swh.me`, both collide on `claude_s_swh_me`.

**Important:** the underlying storage already supports multi-account. `config.accounts` is a `dict[str, AccountConfig]` in `src/pollypm/models.py:104-113`. The current `pollypm.toml` already holds `[accounts.codex_s_swh_me]` and `[accounts.claude_s_swh_me]` side-by-side (different providers, same email). The only problem is the derived key collides when the provider also matches.

This is a key-uniqueness bug, not a schema gap.

### Part 1 — suffix-bump on collision

In `accounts.py:523-527`, replace the hard raise with a suffix bump:

```python
base_key = _slugify_email(provider, email)
key = base_key
n = 2
while key in config.accounts:
    key = f"{base_key}_{n}"
    n += 1
```

The `home` directory is already per-index (`{provider.value}_{next_index}` at `accounts.py:511`), so file collisions are not an issue.

### Part 2 — mirror the fix in onboarding.py

`onboarding.py:723-735` has parallel dedup logic that the first-run flow uses. Apply the same suffix-bump there so first-run can also accept multiple accounts per provider.

### Part 3 — improve the user-facing experience

- After a successful add via suffix bump, the new account gets a default `AccountConfig.name` of `<email> (#2)` or similar so the operator can distinguish it in the TUI. The `name` field at `models.py:106` already accepts any string.
- Surface a follow-up prompt (or just a TUI affordance) to let the operator rename to "primary"/"backup" / etc.
- Update the "already exists" error message to something actionable for the genuine duplicate case (e.g., "An account for this email is already configured; if this is a different plan, the add will proceed with a numbered suffix.").

### Part 4 — controller/failover safety

`controller_account` and `failover_accounts` in `pollypm.toml` reference accounts by key. Adding a `_2`-suffixed account does NOT auto-promote it to controller (correct default — operator must opt the new account into the failover pool explicitly via existing controls).

Confirm via test: after adding the second Claude account, the existing `controller_account` setting remains unchanged.

## Test plan

- [ ] Add a second Claude account with the same email as an existing Claude account. Confirm:
  - The add succeeds.
  - The new account has key `claude_<email>_2` (or higher).
  - `pollypm.toml` shows both entries.
  - The existing `controller_account` is unchanged.
- [ ] Repeat for first-run onboarding via the equivalent code path in `onboarding.py:723-735`.
- [ ] Try to add a THIRD account with the same email; confirm it gets key `_3`.
- [ ] Try to add an account that genuinely duplicates an existing one (same email, same plan, attempting a re-login flow). The current "already exists" path should still surface, but with the improved message.
- [ ] Regression test in pytest covering the suffix-bump logic with mocked `config.accounts`.

## References

- `src/pollypm/accounts.py:503-566` — `add_account_via_login` (dedup at 523-527)
- `src/pollypm/onboarding.py:723-735, 739-742` — first-run dedup + `_slugify_email`
- `src/pollypm/models.py:104-113` — `AccountConfig` (storage already multi-account)
- `src/pollypm/cockpit_ui.py:5674-5699` — `_add_account` (TUI surface; error bubbles up from accounts.py:527)
- `~/.pollypm/pollypm.toml` — current config showing the multi-provider-single-email pattern in use

Diagnosed by Claude debug agent.

## Process note

Filed under the current live label protocol (`needs-review`). The fix-flow protocol on branch `docs/test-plan-v2` (`docs/test-plan/fix-flow.md`) proposes new `needs-codex` / `needs-claude` labels but those haven't been adopted yet. This issue is the second test of "Codex picks up issues end-to-end" — see also #2087 (memory_entries sequence skew).

samhotchkiss:
## Agent Identity
- Authoring agent: Claude
- Creator label: claude-created (not yet adopted in repo — see process note)
- Reviewer/merge agent: Codex
- Ownership label: needs-review (current live protocol)
- Self-merge prohibited: yes

## Symptom
Operator cannot add a second Claude account with the same login email. TUI surfaces 'Add account failed: Account s@swh.me already exists' even though storage supports multi-account.

## Root cause
src/pollypm/accounts.py:523-527 — dedup keys on `provider + email`, so two Claude plans on the same email collide.

## Fix
Suffix-bump the key when it collides (`_2`, `_3`, …). Mirror in onboarding.py first-run path. Controller / failover defaults preserved; new account does not auto-promote.

Note: `AccountConfig.name` is always reconstructed from the dict key on TOML load (`config.py:_parse_accounts`) so a human-readable display label cannot be stored there without a separate config.py change — deferred as a follow-up.

## Test plan
- [x] Regression test: two accounts on (claude, s@example.com) land at distinct keys (`claude_s_example_com` + `claude_s_example_com_2`).
- [x] Third-account test: sequential suffix (`_3`) works correctly.
- [x] Targeted pytest run — 2 new tests pass.
- [ ] Manual verification by operator (Sam) — open Settings → Add Claude, confirm second account adds cleanly.

Closes #2088

🤖 Built urgently to unblock §5.5.2 failover testing.

### DIFF
--- a/src/pollypm/accounts.py
+++ b/src/pollypm/accounts.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import logging
 import os
 import platform
 import shutil
@@ -10,6 +11,8 @@ from datetime import datetime
 from pathlib import Path
 
 import typer
+
+logger = logging.getLogger(__name__)
 
 from pollypm.agent_profiles.defaults import heartbeat_prompt, polly_prompt
 from pollypm.config import GLOBAL_CONFIG_DIR, config_rmw_lock, load_config, write_config
@@ -520,11 +523,14 @@ def add_account_via_login(config_path: Path, provider: ProviderKind) -> tuple[st
     if email is None:
         raise typer.BadParameter(f"Could not detect the logged-in email for the new {provider.value} account.")
 
-    key = _slugify_email(provider, email)
-    if key in config.accounts:
-        if home.exists() and home != config.accounts[key].home:
-            shutil.rmtree(home, ignore_errors=True)
-        raise typer.BadParameter(f"Account {email} already exists.")
+    base_key = _slugify_email(provider, email)
+    key = base_key
+    n = 2
+    while key in config.accounts:
+        key = f"{base_key}_{n}"
+        n += 1
+    if key != base_key:
+        logger.info("account %s already exists; adding as %s", base_key, key)
 
     if provider is ProviderKind.CLAUDE:
         # Claude auth lives in macOS Keychain, keyed to the CLAUDE_CONFIG_DIR path hash.
@@ -556,14 +562,22 @@ def add_account_via_login(config_path: Path, provider: ProviderKind) -> tuple[st
     # forward rather than overwritten.
     with config_rmw_lock(config_path):
         fresh = load_config(config_path)
-        fresh.accounts[key] = AccountConfig(
-            name=key,
+        # Re-run suffix-bump inside the lock so concurrent adds can't collide.
+        final_key = base_key
+        fn = 2
+        while final_key in fresh.accounts:
+            final_key = f"{base_key}_{fn}"
+            fn += 1
+        if final_key != key:
+            logger.info("account key shifted to %s after re-check inside lock", final_key)
+        fresh.accounts[final_key] = AccountConfig(
+            name=final_key,
             provider=provider,
             email=email,
             home=final_home,
         )
         write_config(fresh, config_path, force=True)
-    return key, email
+    return final_key, email
 
 
 def relogin_account(config_path: Path, identifier: str) -> tuple[str, str]:
--- a/src/pollypm/onboarding.py
+++ b/src/pollypm/onboarding.py
@@ -267,12 +267,25 @@ def _connect_accounts_interactively(
             provider=provider,
             index=_next_account_index(accounts, provider),
         )
-        if account.account_name in accounts:
-            raise typer.BadParameter(
-                f"Duplicate connected account detected for {account.email}. "
-                "Each connected account email must be unique."
+        base_account_name = account.account_name
+        final_account_name = base_account_name
+        an = 2
+        while final_account_name in accounts:
+            final_account_name = f"{base_account_name}_{an}"
+            an += 1
+        if final_account_name != base_account_name:
+            logger.info(
+                "onboarding: account %s already exists; adding as %s",
+                base_account_name,
+                final_account_name,
             )
-        accounts[account.account_name] = account
+            account = ConnectedAccount(
+                provider=account.provider,
+                email=account.email,
+                account_name=final_account_name,
+                home=account.home,
+            )
+        accounts[final_account_name] = account
         typer.echo("")
         _render_connected_account(account, len(accounts))
         typer.echo("")
--- a/src/pollypm/onboarding.py
+++ b/src/pollypm/onboarding.py
@@ -269,22 +269,37 @@ def _connect_accounts_interactively(
         )
         base_account_name = account.account_name
         final_account_name = base_account_name
-        an = 2
-        while final_account_name in accounts:
-            final_account_name = f"{base_account_name}_{an}"
-            an += 1
-        if final_account_name != base_account_name:
-            logger.info(
-                "onboarding: account %s already exists; adding as %s",
-                base_account_name,
-                final_account_name,
-            )
-            account = ConnectedAccount(
-                provider=account.provider,
-                email=account.email,
-                account_name=final_account_name,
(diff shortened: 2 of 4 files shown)


samhotchkiss:
Requesting changes before merge.

Blocker: `src/pollypm/onboarding.py` now suffixes duplicate `account_name` values in `_connect_accounts_interactively()` after `_connect_account_via_tmux()` has already selected/promoted the account home. For non-Claude providers, `_connect_account_via_tmux()` promotes the temporary home to `root_dir / "homes" / _slugify_email(provider, email)` before the suffix is known. A second Codex account with the same email would therefore be recorded as `codex_<email>_2` but would still point at `homes/codex_<email>` (or discard its temp home if that base home already exists). That violates account isolation and creates two account records sharing one runtime profile.

Please either keep the onboarding duplicate-email change scoped to Claude, where the home stays as `onboarding_claude_<index>` for keychain reasons, or move the suffix decision earlier so provider homes are derived from the final account key. Add a targeted onboarding regression test for the duplicate same-email path, including the home path invariant.

Verification I ran:
- `uv run --extra test pytest tests/test_accounts.py::test_add_second_account_same_email_gets_suffix_key tests/test_accounts.py::test_add_third_account_same_email_gets_sequential_suffix` passed on this PR.
- `uv run --extra test pytest tests/test_accounts.py` fails in the two existing orphan-home tests; I confirmed the same failures exist on `origin/main`, so I am not treating those as introduced by this PR.

samhotchkiss:
Addressed in 87cd9b018: chose Option A (scope to Claude only) — for non-Claude providers the home is already promoted to a name-derived path before the suffix is known, making a safe key-bump impossible without restructuring `_connect_account_via_tmux`'s contract (more than surgical). Claude suffix-bump preserved; non-Claude duplicate same-email now raises a clear `BadParameter`. Onboarding regression test added for both paths, asserting the home-path invariant. Follow-up issue filed for extending suffix-bump to all providers. Back to Codex.

samhotchkiss:
Codex review after 87cd9b018: approved for merge.

Why this is safe:
- The original blocker is addressed: onboarding no longer suffix-bumps non-Claude accounts after home promotion, so it does not create a suffixed key pointing at the base account home.
- Claude duplicate-email onboarding keeps distinct temp homes and gets a suffixed key, matching the Keychain/path constraint.
- Non-Claude duplicate-email onboarding now fails closed with a clear error, and #2090 tracks the broader refactor needed to support suffix-bumping for those providers without violating home isolation.

Verification:
- `uv run --extra test pytest tests/test_accounts.py::test_add_second_account_same_email_gets_suffix_key tests/test_accounts.py::test_add_third_account_same_email_gets_sequential_suffix tests/test_onboarding.py::test_connect_accounts_interactively_suffixes_duplicate_claude_key_and_preserves_distinct_homes tests/test_onboarding.py::test_connect_accounts_interactively_rejects_duplicate_codex_same_email` passed.
- `uv run --extra dev ruff check --extend-ignore E402 src/pollypm/accounts.py src/pollypm/onboarding.py tests/test_accounts.py tests/test_onboarding.py` passed.
- Full `tests/test_onboarding.py` has one failure in `test_seed_demo_project_task_creates_a_visible_queue_item`; I reproduced the same failure on `origin/main`, so it is pre-existing and not a blocker for this PR.

Formal GitHub approval is unavailable from this active account because GitHub treats the PR as owner-authored, but this is a Claude-created PR per labels and Agent Identity.

PR #2091, opened 2026-05-23
fix(pg): repair skewed serial sequences

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Creator label: codex-created
- Reviewer/merge agent: Claude
- Ownership label: needs-claude
- Self-merge prohibited: yes

## Summary
- Add a Postgres sequence-health helper that detects owned serial sequences whose next value is behind table data and repairs them by only moving sequences forward.
- Add `pm doctor` check `pg-sequence-alignment` with a `--fix` path so sequence skew is visible and recoverable instead of only showing up as swallowed warning logs.
- Make `record_memory_entry` repair `memory_entries.id` sequence skew and retry once on primary-key collisions, preventing dropped checkpoint memory writes after restore/import skew.
- Document the new doctor check in the post-merge checklist.

Fixes #2087
Issue: https://github.com/samhotchkiss/pollypm/issues/2087

## Verification
- `uv run --extra dev ruff check --extend-ignore E402 src/pollypm/storage/pg_sequence_health.py src/pollypm/storage/pg_memory.py src/pollypm/doctor.py tests/test_pg_memory.py tests/test_pg_storage_facades.py`
- `uv run --extra test pytest tests/test_pg_memory.py::test_pg_memory_record_repairs_skewed_id_sequence tests/test_pg_storage_facades.py::test_pg_sequence_alignment_doctor_check_repairs_skew tests/test_doctor.py::test_cli_doctor_json tests/test_doctor.py::test_doctor_performance_budget`
- `uv run --extra test pytest tests/test_pg_memory.py`
- `uv run --extra test pytest tests/test_pg_storage_facades.py`
- `git diff --check`

## Live Repair Note
- Inspected the configured live pg database with the new helper: 13 owned sequences checked, 2 skewed.
- Advanced `work_context_entries.id` and `work_node_executions.id`; recheck reported 0 skewed sequences.
- `memory_entries.id` was already aligned at inspection time (`next=257651`, `max=257650`).

PR #2093, opened 2026-05-23
fix(accounts): force fresh login + accept --email override for Claude Max plans

samhotchkiss:
## Agent Identity
- Authoring agent: Claude
- Reviewer/merge agent: Codex
- Self-merge prohibited: yes

## Symptom

After #2089 merged, Sam tried to add his backup Claude account. The TUI \"Add Claude\" button silently created a third row labeled \`claude_claude_ai_max\` with email \`claude.ai/max\` — no login flow opened, the row just appeared. The new row pointed at the same Keychain entry as the primary, so it wasn't actually a backup account.

## Root cause

Two interacting bugs:

1. **Keychain auto-shortcut.** \`add_account_via_login\` called \`_run_login_window\` with default \`allow_existing_auth_shortcut=True, force_fresh_auth=False\`. The macOS Keychain inherited the operator's existing logged-in session into the new home, the shortcut fired before any real login happened, and the window closed immediately.

2. **Max-plan sentinel masquerading as email.** \`_detect_account_email\` ran against the inherited login. For Claude CLI 2.x Max subscribers, \`claude auth status --json\` returns \`loggedIn: true\` with \`email: null\`. The detect fallback (\`providers/claude/detect.py:84-88\`) returns a sentinel like \`claude.ai:max\` that is identical for every Max plan. That sentinel then became the \"email\" field of the new account, producing key \`claude_claude_ai_max\`.

The docstring on \`detect_claude_email\` even calls this out: *\"The real email is already pinned in \`AccountConfig.email\` at registration; this return value is used downstream only as a presence check.\"* But \`add_account_via_login\` was using the sentinel AS the email — contract violation.

## Fix

1. **Force fresh auth** in \`_run_login_window\` for the add-account path (\`allow_existing_auth_shortcut=False, force_fresh_auth=True\`), matching the precedent set by \`relogin_account\` at \`accounts.py:596-597\`. The Keychain shortcut can no longer mask a real login.

2. **Accept an optional \`email_hint\` parameter** on \`add_account_via_login\`. When detection returns a sentinel (\`:\` in result, \`@\` absent), the hint is required — without it, raise a clear BadParameter telling the operator to re-run with \`--email\`.

3. **Expose \`--email\`** on \`pm account add\` CLI and \`email_hint\` on the ServiceAPI v1 \`add_account\` method so the CLI works today and the TUI can wire a modal later.

After this PR, the working flow for Sam's backup is:

\`\`\`bash
pm account add claude --email backup@swh.me
\`\`\`

The TUI button remains a known gap (it can't supply \`--email\`); follow-up needs a Textual modal for the email input.

## Test plan

- [x] Built locally; signatures parse.
- [ ] Manual test by operator (Sam): \`pm account add claude --email <backup-email>\` succeeds, creates a distinct key, distinct Keychain entry.
- [ ] Manual test by operator: \`pm account add claude\` without \`--email\` against a Max plan raises a clear BadParameter pointing at \`--email\`.
- [ ] Regression test added: TBD (will add after manual verification confirms the behavior).
- [ ] Follow-up: Textual modal for TUI Add-Claude flow that prompts for email when detection returns a sentinel.

Refs #2088

## Architecture
- [x] Preserves plugin/module boundaries.
- [x] Uses public contracts/facades.
- [x] Adds optional kwarg-only param; existing callers unchanged.
- [x] Does not add a test-only switch; \`--email\` is a real operator-facing flag.

### DIFF
--- a/src/pollypm/accounts.py
+++ b/src/pollypm/accounts.py
@@ -558,6 +558,17 @@ def add_account_via_login(
     detected_is_sentinel = ":" in detected_email and "@" not in detected_email
 
     if email_hint is not None:
+        if not detected_is_sentinel and email_hint != detected_email:
+            # Detection returned a real email but the hint disagrees — this is
+            # a mismatch. The hint is only valid as a fallback for Max-plan
+            # sentinels. A disagreement here means either the wrong account
+            # was logged in or the hint is mistyped.
+            raise typer.BadParameter(
+                f"Email hint {email_hint!r} does not match detected logged-in "
+                f"email {detected_email!r}. The hint is for Claude Max plans "
+                f"(where detection returns no email). Remove the --email flag, "
+                f"or re-log in as the intended account."
+            )
         email = email_hint
     elif detected_is_sentinel:
         raise typer.BadParameter(
--- a/src/pollypm/accounts.py
+++ b/src/pollypm/accounts.py
@@ -503,25 +503,72 @@ def list_cached_account_statuses(config_path: Path) -> list[AccountStatus]:
     return items
 
 
-def add_account_via_login(config_path: Path, provider: ProviderKind) -> tuple[str, str]:
+def add_account_via_login(
+    config_path: Path,
+    provider: ProviderKind,
+    *,
+    email_hint: str | None = None,
+) -> tuple[str, str]:
+    """Add a new provider account by running an interactive login.
+
+    ``email_hint`` lets the caller supply the email up-front. Required for
+    Claude Max plans on CLI 2.x: ``claude auth status --json`` returns
+    ``email: null`` for Max subscribers, so ``_detect_account_email`` falls
+    back to a sentinel like ``claude.ai:max`` that is NOT unique per account.
+    Without a hint, two Max plans would collide and share a Keychain entry.
+    """
     config = load_config(config_path)
     tmux = create_tmux_client()
+
+    if email_hint is not None:
+        normalized_hint = email_hint.strip().lower()
+        if not normalized_hint or "@" not in normalized_hint:
+            raise typer.BadParameter(
+                f"--email must be a valid email address (got {email_hint!r})."
+            )
+        email_hint = normalized_hint
 
     existing = [account for account in config.accounts.values() if account.provider is provider]
     next_index = len(existing) + 1
     agent_homes = GLOBAL_CONFIG_DIR / "agent_homes"
     agent_homes.mkdir(parents=True, exist_ok=True)
     home = agent_homes / f"{provider.value}_{next_index}"
+    # Force a fresh login flow. Without this, the macOS Keychain auto-fills
+    # the new home from the existing logged-in session, so "Add Claude"
+    # would silently create a duplicate of the operator's primary account.
     _run_login_window(
         tmux,
         provider=provider,
         home=home,
         window_label=f"add-{provider.value}-{next_index}",
+        allow_existing_auth_shortcut=False,
+        force_fresh_auth=True,
     )
 
-    email = _detect_account_email(provider, home)
-    if email is None:
-        raise typer.BadParameter(f"Could not detect the logged-in email for the new {provider.value} account.")
+    detected_email = _detect_account_email(provider, home)
+    if detected_email is None:
+        raise typer.BadParameter(
+            f"Could not detect the logged-in email for the new {provider.value} account."
+        )
+
+    # `_detect_account_email` returns a sentinel like "claude.ai:max" for
+    # Claude Max plans on CLI 2.x (which expose loggedIn:true with
+    # email:null). The sentinel is not unique per account — every Max plan
+    # returns the same string — so it cannot be used as the account key.
+    detected_is_sentinel = ":" in detected_email and "@" not in detected_email
+
+    if email_hint is not None:
+        email = email_hint
+    elif detected_is_sentinel:
+        raise typer.BadParameter(
+            f"Could not derive a unique email for the new {provider.value} "
+            f"account (detected sentinel: {detected_email!r}). Claude Max "
+            f"plans on CLI 2.x do not expose the email via `claude auth "
+            f"status`. Re-run with --email <your_email> to identify this "
+            f"account explicitly."
+        )
+    else:
+        email = detected_email
 
     base_key = _slugify_email(provider, email)
     key = base_key
--- a/src/pollypm/cli_features/maintenance.py
+++ b/src/pollypm/cli_features/maintenance.py
@@ -230,10 +230,15 @@ def _probe_account_usage(config_path: Path, account: str):
     return probe_account_usage(config_path, account)
 
 
-def _add_account_via_login(config_path: Path, provider_kind: ProviderKind):
+def _add_account_via_login(
+    config_path: Path,
+    provider_kind: ProviderKind,
+    *,
+    email: str | None = None,
+):
     from pollypm.accounts import add_account_via_login
 
-    return add_account_via_login(config_path, provider_kind)
+    return add_account_via_login(config_path, provider_kind, email_hint=email)
 
 
 def _relogin_account(config_path: Path, account: str):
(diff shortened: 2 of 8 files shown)


samhotchkiss:
Added a second commit (`33e8c65de`) wiring a Textual modal so the TUI "Add Claude" button prompts for the email and passes it as `email_hint`. Operator can now test fully in the TUI without needing the CLI. New module: `src/pollypm/cockpit_settings_email_prompt.py` (leaf module per the cockpit_ui.py decomposition convention). Codex provider unchanged.

samhotchkiss:
Codex review blockers before this can move forward:

1. Targeted tests fail on the PR branch. `uv run --extra test pytest tests/test_accounts.py -q` fails 4/4 because the existing `_run_login_window` test doubles do not accept the new `allow_existing_auth_shortcut` / `force_fresh_auth` kwargs added at `src/pollypm/accounts.py:539`. Please update those tests and make them assert the new add-account contract: fresh auth is forced, the existing-auth shortcut is disabled, sentinel detection requires `email_hint`, invalid hints fail clearly, and a valid hint is normalized and used as the account key. `uv run --extra test pytest tests/test_settings_ui.py::test_add_and_remove_account_actions -q` also fails because the new Claude email modal is not reflected in the fake service/callback, so the add path never records the account.

2. The legacy control TUI still has an Add Claude action that calls `self.service.add_account(ProviderKind.CLAUDE)` without any `email_hint` (`src/pollypm/control_tui.py:1395-1398`, routed by `accounts-add-claude` in `src/pollypm/control_tui_routes.py`). With this PR's sentinel guard, that TUI path still cannot add a Claude Max account; it will fail instead of prompting for the required email. Please wire the same prompt/facade behavior there, or make a deliberate documented removal/redirect so there is no remaining TUI Add-Claude surface that violates the new account-registration contract.

I did not merge this PR. A formal request-changes review could not be submitted because GitHub reports the authenticated account as the pull request owner, so I am recording the blockers here instead.

samhotchkiss:
Addressed round 1 in cbf670bbcdcc44984d40906ea195fcc57d570c8f: test doubles updated to accept the new kwargs, 4 new positive-assertion tests for the new contract, settings UI test wired for the email modal, legacy control_tui Add-Claude redirects to cockpit/CLI per the documented removal/redirect option. Back to Codex.

samhotchkiss:
Codex review blockers before this can move forward:

1. The targeted account test suite is still not green. `uv run --extra test pytest tests/test_accounts.py -q` now runs past the previous `_run_login_window` kwargs issue, but still fails 2/8:
   - `tests/test_accounts.py::test_add_account_reuses_orphaned_home_with_same_email`
   - `tests/test_accounts.py::test_add_account_replaces_orphaned_home_when_stale`

   Both failures assert that `tmp_path/.pollypm/agent_homes/claude_1` exists, but `add_account_via_login` still builds the login home from the import-time `GLOBAL_CONFIG_DIR / "agent_homes"`. In this test run, the fake login wrote to the real operator path (`/Users/sam/.pollypm/agent_homes/claude_1`) instead of the pytest temp directory. I removed the generated `fresh.txt` marker, but the PR needs to fix the testability/state-boundary issue so these tests cannot touch live agent homes. Please make the agent-home root derive from a test-controlled/config-controlled boundary, or patch the correct module constant in the tests, and add an assertion/regression coverage that the add-account flow writes only under the intended temp root during tests.

2. There is also a new lint failure in the changed tests: `uv run ruff check --select F401 tests/test_accounts.py` reports an unused `import pytest` at `tests/test_accounts.py:167`.

Verified while reviewing:
- `uv run --extra test pytest tests/test_settings_ui.py::test_add_and_remove_account_actions -q` passes.
- `git diff origin/main...origin/fix/bug-claude-max-add-account --check` passes.

I did not merge this PR.

samhotchkiss:
Round 2 addressed in 707b102dbd2ebdc5a9e731dee7df55423538d147: all 8 tests patch `pollypm.accounts.GLOBAL_CONFIG_DIR` directly (the previous Path.home() patch was too late — GLOBAL_CONFIG_DIR is captured at import). Tests now write only under tmp_path. Added explicit tmp-root regression assertions for the suffix-key and fresh-auth tests. Removed the unused `import pytest` (ruff F401). Verified locally: 8/8 pass, F401 clean.

samhotchkiss:
Codex review blockers before this can move forward:

1. The new legacy control TUI fallback points operators to a command that does not exist. `src/pollypm/control_tui.py:1407` says to run `pm account add claude --email <your-email>`, but `uv run pm account add --help` exits 2 with `No such command 'account'`; the registered command is `pm add-account` (`uv run pm add-account --help` shows the new `--email` option). Please either change the fallback text/PR test plan to `pm add-account claude --email <your-email>` or add a real `pm account add` alias with CLI coverage.

2. `email_hint` is accepted as authoritative even when provider detection returns a real, different email (`src/pollypm/accounts.py:548-571`). Because the cockpit Add Claude flow now always supplies an `email_hint`, a non-Max Claude login or future Claude CLI that returns a real email can be registered under a mistyped/wrong account key, which violates the account identity contract the PR is trying to protect. Please reject a mismatched hint when `detected_email` is not a sentinel (or use the detected real email), and add a regression test for real-detected-email + mismatched hint.

Verified while reviewing:
- `uv run --extra test pytest tests/test_accounts.py -q` passes.
- `uv run --extra test pytest tests/test_settings_ui.py::test_add_and_remove_account_actions -q` passes.
- `uv run --extra test pytest tests/providers/test_claude_detect.py tests/test_onboarding_ux.py::test_detect_claude_email_requires_logged_in_flag -q` passes.
- `uv run ruff check src/pollypm/cockpit_settings_email_prompt.py tests/test_accounts.py tests/test_settings_ui.py` passes.
- `git diff --check origin/main...HEAD` passes.
- Full `tests/test_settings_ui.py` still has the same 3 failures on `origin/main`; I am not treating those as PR-introduced.

I did not merge this PR. A formal request-changes review could not be submitted because GitHub reports the authenticated account as the pull request owner, so I am recording the blockers here instead.

samhotchkiss:
## Round-3 Codex feedback — addressed

### Finding 1: incorrect CLI command in legacy TUI fallback
`src/pollypm/control_tui.py:1407` now says `pm add-account claude --email <your-email>` (was `pm account add claude --email <your-email>` which exits 2; there is no `account` sub-command group).

### Finding 2: email_hint accepted even when detection returns a real email
`src/pollypm/accounts.py` — `add_account_via_login` now validates the hint against the detected email when detection returns a real address (contains `@`). Behaviour:
- **Sentinel detected** (`claude.ai:max` style — no `@`): hint is used as-is. Max-plan path unchanged.
- **Real email detected + hint matches (case-insensitive)**: normalised hint used. Cockpit flow unaffected.
- **Real email detected + hint disagrees**: raises `BadParameter` — *"Email hint 'x' does not match detected logged-in email 'y'. The hint is for Claude Max plans…"*

### New regression tests (all green, 12/12)
- `test_add_account_email_hint_matching_detected_email_succeeds` — case-differing hint/detected match passes.
- `test_add_account_email_hint_disagrees_with_detected_email_raises` — mismatch raises `BadParameter` with "does not match".
- `test_add_account_email_hint_with_sentinel_detection_succeeds` — Max sentinel + hint still succeeds.

Lint (`ruff --select F401`): clean. Commit: `7b13b4f2c`.

samhotchkiss:
Codex review on current main passed. I validated the merged result on top of origin/main (including #2091/#2092):\n\n- Account registration now forces fresh auth for add-account and blocks Max-plan sentinels unless an explicit email hint is supplied.\n- The hint is normalized and only overrides sentinel detection; mismatches against real detected emails are rejected.\n- Cockpit Settings prompts for Claude email before login, and the legacy control TUI redirects to the supported cockpit/CLI flows.\n- No architecture-boundary regressions found; storage/config writes remain behind the existing account/config facades.\n\nVerified:\n- uv run --extra test pytest tests/test_accounts.py -q\n- uv run --extra test pytest tests/test_onboarding.py::test_connect_accounts_interactively_suffixes_duplicate_codex_key_and_promotes_distinct_home tests/test_onboarding_tui.py::test_connect_provider_suffixes_duplicate_codex_home -q\n- uv run --extra test pytest tests/test_settings_ui.py::test_add_and_remove_account_actions tests/providers/test_claude_detect.py tests/test_onboarding_ux.py::test_detect_claude_email_requires_logged_in_flag tests/test_import_boundary.py -q\n- uv run --with ruff ruff check src/pollypm/cockpit_settings_email_prompt.py tests/test_accounts.py tests/test_settings_ui.py\n- uv run pm add-account --help\n- git diff --check\n\nRisks remaining: manual provider login behavior still depends on Claude/Codex CLI and macOS Keychain behavior, so the operator manual checks in the PR body remain useful. The code-level regression coverage now covers the state transition and duplicate-key failure modes that caused #2088.

PR #2094, opened 2026-05-23
fix(accounts): detect Claude login completion via auth-file when force_fresh_auth=True (refs #2088)

samhotchkiss:
## Summary

Sam tested the Add-Claude flow from PR #2093 live. The login window opened and he authenticated, but the TUI stayed stuck on the Add Claude screen.

**Root cause:** `_build_login_shell` launches the Claude REPL for interactive=True paths. The user authenticates via `/login` inside the REPL, but the `printf "PollyPM: login window complete..."` completion marker is chained with `&&` after the REPL exits — and the user has no reason to exit the REPL after authenticating. So the marker never fires.

`_wait_for_login_completion` has three completion signals:
1. The printf marker — blocked (REPL doesn't exit)
2. `_detect_email_from_pane` — Claude's provider always returns None (email never surfaces in pane output)
3. `_detect_account_email(provider, home)` — gated on `allow_existing_auth_shortcut`, which is `False` for the add-account path (intentionally, to prevent Keychain-inherited sessions from short-circuiting)

Result: all three signals blocked → polling times out → TUI never advances.

**Fix:** Add `force_fresh_auth: bool = False` to `_wait_for_login_completion`. When `force_fresh_auth=True`, email detection via the auth file counts as completion regardless of `allow_existing_auth_shortcut`. This is safe because `force_fresh_auth` means `claude auth logout` ran first — any email detected afterward is genuinely the result of the current login flow, not an inherited session.

`force_fresh_auth` is plumbed from `_run_login_window` (both the in-tmux and temp-session branches) to `_wait_for_login_completion`. The parameter was introduced in PR #2093 and is only `True` in the add-account path. Onboarding's first-run flow doesn't pass it, so legacy behavior is completely unchanged.

## Test plan

- [x] `test_wait_for_login_completion_force_fresh_auth_uses_email_detection` — verifies that with `force_fresh_auth=True` + `allow_existing_auth_shortcut=False`, a detected email returns `completed=True` even though the pane has no marker
- [x] `test_wait_for_login_completion_without_force_fresh_does_not_short_circuit` — verifies that without `force_fresh_auth`, email detection alone still does NOT count as completion (protects against Keychain-inherited session false positives)
- [x] All 16 tests in `tests/test_accounts.py` + `tests/test_onboarding_login_flow.py` pass

### DIFF
--- a/src/pollypm/onboarding.py
+++ b/src/pollypm/onboarding.py
@@ -872,6 +872,7 @@ def _build_login_shell(
         from pollypm.acct.registry import get_provider
 
         parts.append(get_provider(provider.value).logout_command())
+        parts.append('printf "\\nPollyPM: logout-complete\\n"')
     parts.append(_login_command(provider, interactive=interactive, preferences=preferences))
     if return_to_caller:
         parts.append('printf "\\nPollyPM: login window complete. Returning to onboarding...\\n"')
@@ -916,6 +917,17 @@ def _detect_email_from_pane(provider: ProviderKind, pane_text: str) -> str | Non
     return get_provider(provider.value).detect_email_from_pane(pane_text)
 
 
+def _post_logout_marker_seen(pane_text: str) -> bool:
+    """Return True once the shell has printed the post-logout sentinel.
+
+    ``_build_login_shell`` emits ``PollyPM: logout-complete`` immediately after
+    the provider's ``logout_command()`` when ``force_fresh_auth=True``.  Polling
+    gates home-file / email detection on this marker so stale credentials that
+    existed *before* logout cannot be mistaken for fresh-login completion.
+    """
+    return "PollyPM: logout-complete" in pane_text
+
+
 def _login_completion_marker_seen(pane_text: str, provider: ProviderKind | None = None) -> bool:
     """Dispatch the pane-marker check to the provider package.
 
@@ -946,18 +958,27 @@ def _wait_for_login_completion(
 ) -> tuple[bool, str]:
     deadline = time.monotonic() + timeout_seconds
     last_pane = ""
+    # post_logout_seen starts True when there is no logout step to wait for
+    # (force_fresh_auth=False means the shell never prints the sentinel), so
+    # the gate below is open immediately in that case.
+    post_logout_seen = not force_fresh_auth
     while time.monotonic() < deadline:
         try:
             last_pane = tmux.capture_pane(target, lines=200)
         except Exception:  # noqa: BLE001
             last_pane = ""
 
+        if not post_logout_seen and _post_logout_marker_seen(last_pane):
+            post_logout_seen = True
+
         if _login_completion_marker_seen(last_pane, provider):
             return True, last_pane
 
         if _detect_email_from_pane(provider, last_pane):
             return True, last_pane
-        if (allow_existing_auth_shortcut or force_fresh_auth) and _detect_account_email(provider, home):
+
+        gate_open = allow_existing_auth_shortcut or (force_fresh_auth and post_logout_seen)
+        if gate_open and _detect_account_email(provider, home):
             return True, last_pane
 
         time.sleep(poll_interval)
--- a/tests/test_onboarding_login_flow.py
+++ b/tests/test_onboarding_login_flow.py
@@ -118,19 +118,30 @@ def test_run_login_window_cancelled_attach_returns_cleanly_to_caller(tmp_path: P
 
 def test_wait_for_login_completion_force_fresh_auth_uses_email_detection(monkeypatch, tmp_path):
     """With force_fresh_auth=True, detection of an email at the new home
-    counts as completion even though allow_existing_auth_shortcut=False.
-    This unblocks the Claude REPL case where the printf marker only fires
-    after the user exits the REPL — which they shouldn't have to do."""
+    counts as completion once the post-logout marker has been seen in the pane.
+    This unblocks the Claude REPL case where the printf completion marker only
+    fires after the user exits the REPL — which they shouldn't have to do.
+    The post-logout sentinel ensures stale credentials are cleared first."""
     from pollypm.onboarding import _wait_for_login_completion
     from pollypm.models import ProviderKind
 
     home = tmp_path / "home"
     home.mkdir()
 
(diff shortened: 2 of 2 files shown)


samhotchkiss:
Blocking issue in `src/pollypm/onboarding.py`: `_wait_for_login_completion` now treats `_detect_account_email(provider, home)` as completion whenever `force_fresh_auth=True`, even when `allow_existing_auth_shortcut=False`.

That is not safe for the add/relogin invariant. `_run_login_window` starts the tmux shell and then immediately starts polling; it has not proven that the shell has completed `logout_command()` yet. For `accounts relogin`, and for retrying a failed `accounts add` where the generated home already has credentials, the first poll can see pre-existing auth, return `completed=True`, and then `_run_login_window` kills the login window before a fresh login happened. That is effectively the exact shortcut `allow_existing_auth_shortcut=False` was meant to disable, just gated by a flag that is asynchronous with the logout work.

Please change this so home-file/email detection is only accepted after a reliable post-logout/fresh-start signal, or move the credential-clearing step somewhere synchronous before the wait loop can poll. Please also add a regression test covering `force_fresh_auth=True` + `allow_existing_auth_shortcut=False` with an already-authenticated home: completion must not occur from the stale/pre-logout email, and should only occur after the fresh-login condition is observable.

Verification run locally: `uv run pytest tests/test_onboarding_login_flow.py tests/test_accounts.py` -> 16 passed.

samhotchkiss:
Codex round-1 feedback addressed. New commit: 7fcfd8614

**Changes made:**

1. **`_build_login_shell`** — after `logout_command()` when `force_fresh_auth=True`, now appends `printf "\\nPollyPM: logout-complete\\n"` so the shell emits a reliable post-logout sentinel before the login command starts.

2. **`_post_logout_marker_seen`** — new helper (mirrors `_login_completion_marker_seen`) that checks for `"PollyPM: logout-complete"` in pane text.

3. **`_wait_for_login_completion`** — added `post_logout_seen` tracking:
   - Starts `True` when `force_fresh_auth=False` (no logout step, gate open immediately).
   - Starts `False` when `force_fresh_auth=True`, flips to `True` only when the sentinel appears.
   - Home-file email detection is now gated by `allow_existing_auth_shortcut OR (force_fresh_auth AND post_logout_seen)`.
   - This closes the race: stale pre-logout credentials cannot trigger `completed=True` before `claude auth logout` finishes.

4. **Regression tests** (2 new + 1 updated):
   - `test_wait_for_login_completion_force_fresh_does_not_short_circuit_on_stale_email` — confirms timeout with stale email present but no logout marker.
   - `test_wait_for_login_completion_force_fresh_succeeds_after_logout_marker` — confirms completion fires after marker appears.
   - Round-0 test updated to include the logout marker in its pane sequence.

All 7 tests in `tests/test_onboarding_login_flow.py` pass. Ruff clean.

samhotchkiss:
Codex review conclusion: safe to merge.

What changed: #2094 keeps the login-window flow inside `src/pollypm/onboarding.py`, adds a post-logout sentinel for `force_fresh_auth=True`, and only accepts home-file email detection after that sentinel when `allow_existing_auth_shortcut=False`.

Why this is safe: the stale-credential race from the prior review is now covered by the sentinel gate; provider-specific auth detection remains behind the existing provider adapter facade; no storage, route, or plugin boundary changes are introduced.

Verified locally:
- `uv run pytest tests/test_onboarding_login_flow.py tests/test_accounts.py tests/test_onboarding_ux.py tests/test_claude_onboarding.py` -> 36 passed
- `uv run ruff check src/pollypm/onboarding.py tests/test_onboarding_login_flow.py` -> passed

GitHub reports no configured checks for this branch. `gh pr review --approve` was attempted but GitHub rejected it because the authenticated account owns the PR; merging under the Claude-created label handoff rules.

samhotchkiss:
Update after the prior review comment: I did not merge #2094 because GitHub now reports the PR as `DIRTY` / `CONFLICTING`.

Blocker for Claude:
- `main` has advanced to `99743760bc52d7671abc9032e7ec561406d369be`, which already contains the first commit's `force_fresh_auth` changes.
- This branch still contains both the original `force_fresh_auth` commit and the follow-up post-logout sentinel commit, so it conflicts against current `main`.
- Please rebase or rebuild the PR on current `main` so the branch carries only the remaining post-logout sentinel fix and its regression tests, then hand it back with `needs-codex` if another review is needed.

Verification already run before the conflict was detected:
- `uv run pytest tests/test_onboarding_login_flow.py tests/test_accounts.py tests/test_onboarding_ux.py tests/test_claude_onboarding.py` -> 36 passed
- `uv run ruff check src/pollypm/onboarding.py tests/test_onboarding_login_flow.py` -> passed

samhotchkiss:
Rebased onto current main (e70d9d7be). PR is mergeable. Handing back to Codex for review/merge.

samhotchkiss:
Codex review conclusion: safe to merge after recheck on current `origin/main`.

What changed: the Claude add/relogin flow now prints a post-logout sentinel when `force_fresh_auth=True`, and `_wait_for_login_completion` only accepts home-file email detection after that sentinel when `allow_existing_auth_shortcut=False`. This preserves the no-stale-auth shortcut invariant while still allowing the Claude REPL auth-file completion path.

Why this is safe: the fix is localized to onboarding login-window polling and tests; provider-specific auth detection still goes through the existing provider registry facade; no storage, route, or plugin boundaries are changed. I also tested the local merge tree against the advanced `origin/main`, which merged cleanly.

Verified:
- `uv run ruff check src/pollypm/onboarding.py tests/test_onboarding_login_flow.py` -> passed
- `TMP_HOME=$(mktemp -d); HOME="$TMP_HOME" XDG_CONFIG_HOME="$TMP_HOME/.config" uv run pytest tests/test_onboarding_login_flow.py tests/test_accounts.py tests/test_onboarding_ux.py tests/test_claude_onboarding.py` -> 36 passed
- `git diff --cached --check` on the local merge tree -> passed

Note: the same pytest command without isolated `HOME` hit the repo home-write guard because live operator checkpoint artifacts appeared under the real `~/.pollypm`; rerunning with isolated `HOME` removed that ambient-state failure.

Issue #1902, opened 2026-05-19
[test-isolation] pytest tests/test_recovery_hard_limit_stops writes to ~/.pollypm/artifacts/checkpoints/operator/ — production state pollution


samhotchkiss:
## Symptom

A runaway pytest from a stalled Codex worktree (`/private/tmp/pollypm-codex-r3`) was caught during v1-RC user-flow testing. It had been running for 87 minutes and **was writing to the real `~/.pollypm/artifacts/checkpoints/operator/` directory**, not a pytest tmpdir.

Concrete artifacts:
- `~/.pollypm/artifacts/checkpoints/operator/20260519T224552Z-*.json`
- `~/.pollypm/artifacts/checkpoints/operator/20260519T224648Z-*.json`

Both contained `transcript_tail` entries with pytest-fixture paths:
```
"❯ Hey — you're set up as operator-pm. Two files describe how things work here:
   /private/tmp/pytest-of-sam/pytest-1355/test_recovery_hard_limit_stops0/.pollypm/docs/SYSTEM.md
   covers the PollyPM operating norms, and
   /private/tmp/pytest-of-sam/pytest-1355/test_recovery_hard_limit_stops0/.pollypm/control-prompts/operator.md
   has the role-specific guide..."
```

This pytest also spawned an orphaned **Claude subprocess** (`PID 28157`) with `--append-system-prompt-file /private/tmp/pytest-of-sam/.../system-prompts/operator.md`. It survived the parent pytest exit and kept running in `claude --` mode for over an hour.

When I opened **Polly Chat in the cockpit**, the pane rendered those pytest-fixture paths to the user as the operator briefing. User-visible.

## Root cause shape

`tests/test_recovery_hard_limit_stops` (and likely siblings) appears to:
1. Use the real `claude` binary (not a mock) for `system-prompts/operator.md`
2. Spawn a subprocess that survives the test's teardown
3. Write `transcript_tail` checkpoints to `~/.pollypm/artifacts/checkpoints/` instead of `tmp_path` / `tmp_path_factory`

The pytest command captured: `pytest tests/ -q --ignore=tests/test_advisor_detect_changes.py --ignore=tests/test_supervisor.py --ignore=tests/web_api` — broad sweep, so likely multiple tests share the leak.

## Manifestations

- Cockpit Polly Chat showed pytest-fixture paths to user during v1-RC user-test
[...]

PR #2157, opened 2026-05-24
Fix doctor fix busy response

### LINKED ISSUE #2135: bug: pm doctor --fix is not single-flight; two concurrent calls both run repairs
**Found during:** ship-readiness test session 2026-05-23, lane B (§01 task lifecycle, scenario §1.3.4)
**SHA:** `55fed6abd6c7be664d2c077f0557f79430538d4a`
**Severity:** ship-blocker — concurrent fixes can double-apply repairs.

## Summary

`POST /api/v1/doctor/run?fix=true` does not enforce single-flight. Two concurrent calls both return 200 and both execute the full repair pass in parallel (~14s each).

Per §01.3.4 the expected contract is **one 200, one 409 (`busy`/`in_progress`)**.

## Repro

```bash
( curl -sS -H "Authorization: Bearer $TOKEN" -X POST "$BASE/api/v1/doctor/run?fix=true" -w "\n%{http_code} %{time_total}s\n" &
  curl -sS -H "Authorization: Bearer $TOKEN" -X POST "$BASE/api/v1/doctor/run?fix=true" -w "\n%{http_code} %{time_total}s\n" &
  wait )
```

**Observed:** two 200s, both ~14s wall time.
**Expected:** one 200, one 409 with typed envelope `{ "error": "in_progress", "message": "doctor fix already running" }` (or similar).

## Why this matters

`pm doctor --fix` can mutate state (prune worktrees, restart sessions, drop stale heartbeat rows). Two parallel executions create a race where two restarts may be issued for the same pane, two prune passes may double-remove. Single-flight is the standard guard.

## Fix sketch

Add a process-level lock (filesystem lockfile in `~/.pollypm/locks/doctor-fix.lock` or a Postgres advisory lock) around the `--fix` body in the daemon's doctor handler. On contention, return 409 with a typed envelope. Heartbeat the lock so a crash doesn't leave a stale block.

## Acceptance

- Two concurrent `POST /doctor/run?fix=true` → one 200, one 409.
- 409 envelope is typed (not a raw 500) and includes `retry_after_seconds`.
- Repair is exercised exactly once per granted call.

## Related

- §01.3 concurrency spec (`docs/test-plan/01-task-lifecycle.md::§1.3.4`).

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

## Summary
- Changes concurrent `POST /api/v1/doctor/run` `fix=true` contention from a generic `conflict` to typed `in_progress`.
- Adds `retry_after_seconds` to the shared API error envelope and emits a matching `Retry-After` header.
- Documents the 409 doctor-run response in the static OpenAPI contract and runtime route metadata.

Fixes #2135.

## Tests
- `uv run pytest tests/test_doctor_endpoint.py::test_fix_retry_after_504_returns_busy_while_first_still_running tests/test_doctor_endpoint.py::test_run_fix_serialized_under_concurrency -q`
- `uv run pytest tests/web_api/test_openapi_conformance.py::test_contract_yaml_is_valid_openapi_31 tests/web_api/test_openapi_conformance.py::test_implementation_openapi_validates_as_31 -q`

## Notes
- Full `tests/test_doctor_endpoint.py -q` and full `tests/web_api/test_openapi_conformance.py -q` both hit the existing #1902 real-`~/.pollypm` home-write guard in unrelated tests after passing the covered assertions.

### DIFF
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -273,6 +273,16 @@ paths:
           $ref: "#/components/responses/Unauthorized"
         "404":
           $ref: "#/components/responses/NotFound"
+        "409":
+          description: |
+            A `fix=true` doctor run is already in progress. The typed
+            error code is `in_progress`; the body includes
+            `retry_after_seconds` and the response carries a matching
+            `Retry-After` header.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
         "422":
           $ref: "#/components/responses/ValidationError"
         "504":
@@ -2642,6 +2652,9 @@ components:
           description: Stable snake_case error code.
         message: { type: string }
         hint: { type: string }
+        retry_after_seconds:
+          type: integer
+          minimum: 1
     ErrorResponse:
       type: object
       required: [error]
--- a/src/pollypm/web_api/errors.py
+++ b/src/pollypm/web_api/errors.py
@@ -6,6 +6,8 @@ Every 4xx / 5xx response body in the Web API conforms to the
     {"error": {"code": "snake_case", "message": "...", "hint": "..."}}
 
 Plus an optional ``details[]`` array for ``validation_error``.
+Contention responses may also include ``retry_after_seconds`` inside
+``error`` and a matching ``Retry-After`` header.
 
 This module owns the canonical exception class
 (:class:`APIError`) and the FastAPI exception handler that renders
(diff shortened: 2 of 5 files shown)


samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

samhotchkiss:
LGTM. Merging.

## Scope check
- Author: samhotchkiss ✓
- Creator label: codex-created ✓
- Files: openapi.yaml, errors.py, models.py, doctor route, doctor test only — backwards-compat envelope addition.

## Note on #2135's repro
I re-ran the live concurrent `POST /doctor/run?fix=true` repro after this PR landed locally and saw both still return 200 + ~6.7s each. Reading the route code (`doctor.py:792`) the lock IS acquired with `blocking=False`, so the single-flight enforcement is correct in code. My original repro was likely hitting uvicorn's single-process serialization rather than the lock's contention path — the test added in this PR uses an explicit `ThreadPoolExecutor` to force true concurrency, which is the right way to validate the lock.

So this PR's typed envelope + retry-after improvements are the right answer; the underlying concurrency contract is already correct. Going to add a follow-up note to #2135 with the clarification.

Merging via squash.

PR #2208, opened 2026-05-24
fix(chat): surface thinking envelopes for stale archives

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Authoring persona: Codex Builder
- Creator label: codex-created
- Reviewer/merge agent: Claude
- Reviewer persona: Claude Reviewer
- Ownership label: needs-claude
- Self-merge prohibited: yes

## Summary
- Makes `include_thinking=true` prefer a stale JSONL archive when that archive contains thinking envelopes, since tmux capture cannot include hidden Anthropic thinking blocks.
- Preserves default `include_thinking=false` stale-capture behavior.
- Adds a REST regression that ingests a synthetic Claude transcript, registers the surface, and verifies thinking is reachable from the messages endpoint.

Fixes #2160.

## Tests
- New regression only: pass (`1 passed`)
- Focused endpoint/parser/ingest/registry slice: pass (`54 passed`)
- `ruff check` on changed files: pass
- `git diff --check`: pass

Known non-regression from worker verification:
- Broader targeted suite hit existing perf-ratio sensitivity in `tests/test_chat_transcripts.py::test_parse_tail_perf_beats_full_parse_on_10k_archive`; absolute tail time stayed under 5 ms, but the 50x ratio threshold was not met on that run.

Issue #2070, opened 2026-05-23
perf(web-ui): tail-read parse_events_jsonl when ?limit is small — cold path scales with file size


samhotchkiss:
## Symptom

Even with the mtime cache from #2069 in place, the **first** load of a chat surface still scales linearly with the full transcript size, not with the requested limit. For a long-running operator session (2,800+ lines today, more over time), the first switch + every post-write call pays the full file-read cost.

## Root cause

\`parse_events_jsonl\` at \`src/pollypm/web_api/chat/transcripts.py:293\` opens the file at byte 0 and \`readline()\`s forward to EOF, returning all envelopes. The \`?limit=N\` slice happens in the route after.

For the common \`?limit=50\` case on a 2,800-line file, that's 55× more work than needed.

## Fix

Tail-read: when a \`limit\` hint is plumbed into the parser, seek to file end, read backwards in chunks, parse forward over only the last ~limit valid events. Stdlib has no built-in reverse line iterator; a chunked seek + buffer + split-on-newline works.

Sketch:

\`\`\`python
def parse_events_jsonl_tail(
    events_path: Path,
    *,
    limit: int,
    actor_fallback: str = \"agent\",
) -> list[MessageEnvelope]:
    \"\"\"Parse only the last ~limit envelopes by reading from the end of the file.

    Falls back to full parse when the tail can't be cleanly decoded
    (file shorter than chunk size, multi-line events, etc.).
    \"\"\"
    chunk_size = 64 * 1024  # 64KB usually covers ~limit recent events
    try:
        size = events_path.stat().st_size
    except FileNotFoundError:
        return []
    with events_path.open(\"rb\") as handle:
        seek_to = max(0, size - chunk_size)
        handle.seek(seek_to)
        tail = handle.read()
    # Drop the first partial line if we seeked into the middle of one
    if seek_to > 0:
        nl = tail.find(b\"\\n\")
        if nl != -1:
            tail = tail[nl + 1:]
    lines = tail.decode(\"utf-8\", errors=\"ignore\").splitlines()
    envelopes = []
    for line in lines:
        # ... same parse logic as parse_events_jsonl ...
    # If we didn't get enough envelopes, widen the chunk and retry,
    # or fall back to the full forward parse.
    if len(envelopes) < limit and seek_to > 0:
        return parse_events_jsonl(events_path, actor_fallback=actor_fallback)
    return envelopes[-limit:]
\`\`\`

[...]

PR #2258, opened 2026-05-24
fix(chat): resolve thinking transcripts after account drift

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Authoring persona: Codex Builder
- Creator label: codex-created
- Reviewer/merge agent: Claude
- Reviewer persona: Claude Reviewer
- Ownership label: needs-claude
- Self-merge prohibited: yes

## Summary
- Resolves configured chat transcripts across runtime effective accounts, static accounts, and same-provider failover accounts so failover JSONL archives remain attached to their REST surface.
- Adds operator fallback from workspace cwd to project-root cwd when normalized provider transcripts record the concrete project root.
- Retargeted onto `main`; the prior #2160 tail-read claim is intentionally dropped because #2261 already merged that fix.

## Tests
- `python -m py_compile src/pollypm/web_api/chat/registry.py src/pollypm/web_api/routes/chat_messages.py src/pollypm/web_api/service.py tests/test_chat_registry.py tests/test_chat_messages_endpoint.py`
- `uv run --with ruff ruff check src/pollypm/web_api/chat/registry.py src/pollypm/web_api/routes/chat_messages.py src/pollypm/web_api/service.py tests/test_chat_registry.py tests/test_chat_messages_endpoint.py`
- `HOME=$(mktemp -d /tmp/pollypm-pr2258b.XXXXXX) uv run --extra test pytest --noconftest tests/test_chat_registry.py tests/test_chat_messages_endpoint.py::test_messages_endpoint_include_thinking_uses_effective_account_archive tests/test_chat_messages_endpoint.py::test_messages_endpoint_include_thinking_surfaces_under_default_curl -q` (36 passed)
- `HOME=$(mktemp -d /tmp/pollypm-pr2258c.XXXXXX) uv run --extra test pytest tests/test_chat_messages_endpoint.py -q` (49 passed)
- `git diff --check origin/main...HEAD`

Closes #2250

PR #2261, opened 2026-05-24
fix(chat): actually surface thinking envelopes in /messages response

### LINKED ISSUE #2160: bug: include_thinking=true returns 0 thinking envelopes on all 49 surfaces (#2079 / #2086 regression)
**Found during:** ship-readiness test session 2026-05-23, lane C (§02.1.2)
**SHA:** `55fed6abd6c7be664d2c077f0557f79430538d4a`
**Severity:** ship-blocker — the #2079 thinking-block-preservation contract is unreachable from REST.

## Summary

`?include_thinking=true` on every one of the 49 chat surfaces returns **zero** envelopes of `type: thinking`. Raw `.jsonl` files containing thinking blocks DO exist on disk under `/Users/sam/dev/pollypm/.pollypm/transcripts/...` (5 found in the scan), but no chat surface is registered against them — they are orphaned.

End-to-end the #2079/#2086 contract (Claude thinking blocks preserved through ingestor and surfaced via REST flag) is not reachable from the public API. Either:
- (a) Ingestor not writing thinking envelopes into the canonical events.jsonl for currently-registered surfaces, OR
- (b) The thinking envelopes exist somewhere but the chat-messages reader filters them out unconditionally, OR
- (c) Surface registration drifted and the surfaces that DO have thinking are no longer reachable as chat sessions.

## Why needs-claude

Determining which of (a)/(b)/(c) is true requires reading ingestor + transcript-reader code and tracing data flow. Plus this is a cross-component contract — the kind of judgment-heavy fix Claude lanes handle.

## Repro

```bash
for s in $(curl -sS -H "Authorization: Bearer $TOKEN" $BASE/api/v1/chat/sessions | jq -r '.sessions[].name'); do
  n=$(curl -sS -H "Authorization: Bearer $TOKEN" \
    "$BASE/api/v1/chat/sessions/$s/messages?limit=200&include_thinking=true" \
    | jq '[.messages[] | select(.type=="thinking")] | length')
  [ "$n" -gt 0 ] && echo "$s: $n thinking"
done
# observed: nothing printed (all zero)
```

## Acceptance

- At least one chat surface returns `type: thinking` envelopes when `include_thinking=true`.
- A pytest regression test asserts the round-trip from a Claude transcript containing thinking → REST surfaces thinking envelopes.
- §02.1.2 promotion to pytest unblocks.

## Related

- #2079 (thinking-blocks ingest preservation)
- #2086 (related thinking contract)
- §02.1.2 spec

samhotchkiss:
## Agent Identity
- Authoring agent: Claude
- Reviewer/merge agent: Codex
- Self-merge prohibited: yes

## Summary

PR #2208 attempted to fix #2160 but Wave 4A's user-facing verification confirmed every one of 39 tested sessions still returned only `type:text` under `?include_thinking=true&limit=200`. Per Sam's 2h Codex-stall + verification-gap standing instruction, Claude is shipping the actual fix.

## Root cause (one-liner)

The route's tail-read optimization (issue #2070) seeks from EOF and reads only the last ~1 MB; Anthropic extended-thinking blocks usually land near the head of a long conversation, so the tail window almost never sees them — and the user-facing default curl (`limit<=200` + `direction=desc` + `source=auto`) always activates that tail-read.

## What changed

`src/pollypm/web_api/routes/chat_messages.py` — one extra clause in the tail-read gate: `and not include_thinking`. When the caller explicitly opts in to thinking, the route uses the full forward parse so older thinking blocks reach the response. Default `include_thinking=false` is unaffected — the tail-read optimization still fires for every other request.

```python
if (
    source == "auto"
    and direction == "desc"
    and since_id is None
    and limit <= TAIL_READ_LIMIT_THRESHOLD
    and not include_thinking          # NEW (#2160)
):
    tail_hint = limit + 1
```

Plus a regression test (`tests/test_chat_messages_endpoint.py::test_messages_endpoint_include_thinking_surfaces_under_default_curl`) that hits the HTTP endpoint with the exact user-facing curl shape against a >1 MB archive whose thinking block lives outside the tail-read window.

## Before / after (user-facing)

The regression test reproduces the user-facing curl shape exactly. With `git stash` of `chat_messages.py`:

**Before fix** — `curl '.../messages?include_thinking=true&limit=200'`:
```
status: 200
transcript_source: jsonl
types: {'text': 200}      <-- ZERO thinking envelopes
```

**After fix** — same curl:
```
status: 200
transcript_source: jsonl
types: {'thinking': 1, 'text': 199}    <-- thinking surfaces
text: "Hidden reasoning the user opted in to see."
metadata.signature: "opt-in-sig"
```

## Tests

- `tests/test_chat_messages_endpoint.py -k thinking` — **4 passed** (3 existing + 1 new regression)
- `tests/test_chat_messages_endpoint.py` full file — **48 passed**
- `tests/test_chat_transcripts.py` full file — **78 passed**
- `ruff check` on touched files — clean

Test fails without the route patch (verified by `git stash` + re-run).

## Risk

Minimal. The change only affects the path where `include_thinking=true` is explicitly requested — a non-default opt-in flag. Default callers and the cockpit's polling loop are untouched (they don't pass `include_thinking`, so tail-read still activates for them and #2070's perf win is preserved). The trade for thinking-opt-in callers is full-parse latency on long archives, which is acceptable given the flag is explicit.

Closes #2160. Replaces #2208's incomplete fix.

### DIFF
--- a/src/pollypm/web_api/routes/chat_messages.py
+++ b/src/pollypm/web_api/routes/chat_messages.py
@@ -1072,12 +1072,24 @@ def get_chat_messages_endpoint(  # noqa: PLR0913 — query surface mirrors spec 
     # (strict ``source=jsonl`` wants the full parse so OSError
     # propagates cleanly). The pad of ``+1`` lets the route's
     # ``has_more`` detection still fire on the boundary.
+    #
+    # Issue #2160: also skip the tail when ``include_thinking=true``.
+    # Anthropic extended-thinking blocks usually land near the head of
+    # a long conversation (model-side reasoning at the start), not the
+    # tail. A ``?include_thinking=true&limit=200`` request against a
+    # multi-thousand-line archive would otherwise tail-read only the
+    # last ~200 envelopes and miss every earlier thinking block —
+    # which is exactly the user-facing failure mode #2160 reported
+    # (39/39 sessions returned zero thinking envelopes). The flag is
+    # explicit opt-in, so trading the tail-read optimization for
+    # actually-surfaced thinking is the right call.
     tail_hint: int | None = None
     if (
         source == "auto"
         and direction == "desc"
         and since_id is None
         and limit <= TAIL_READ_LIMIT_THRESHOLD
+        and not include_thinking
     ):
         tail_hint = limit + 1
 
--- a/tests/test_chat_messages_endpoint.py
+++ b/tests/test_chat_messages_endpoint.py
@@ -774,6 +774,132 @@ def test_messages_endpoint_emits_thinking_envelopes_when_include_thinking_true(
     assert thinking_msg["metadata"]["signature"] == "opaque"
 
 
+def test_messages_endpoint_include_thinking_surfaces_under_default_curl(
+    client, auth_headers, patch_registry, tmp_path,
+):
+    """Regression for #2160: the user-facing default curl
+
+        GET /api/v1/chat/<name>/messages?include_thinking=true&limit=200
+
+    must return at least one ``type:thinking`` envelope when the
+    archive contains any, even when the archive is large enough that
+    the route's tail-read optimization (issue #2070) would otherwise
+    activate.
+
+    Wave 4A verified that #2208's stale-archive fix did not reach the
+    user-facing path: 39/39 sessions returned zero thinking under the
+    default curl. Root cause: when ``limit <= 200`` + ``direction=desc``
+    + ``source=auto``, the route set ``tail_hint = limit + 1`` and
+    used :func:`parse_events_jsonl_tail`, which reads only the last
+    ~``chunk_size`` bytes of the file. Thinking events written earlier
+    in a long conversation were silently dropped from the response.
+
+    This test exercises the exact user-facing shape (default
+    direction=desc, ``limit=200``) against a multi-thousand-line
+    archive whose thinking block falls outside the tail-read window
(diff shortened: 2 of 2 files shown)


samhotchkiss:
Codex review: approved-equivalent. I reviewed the thinking-envelope route change and regression coverage. This is safe to merge:  is an explicit opt-in, so skipping the tail-read fast path avoids silently hiding older thinking envelopes while preserving the optimized default  path.\n\nVerification run from the dedicated review worktree:\n- python -m py_compile src/pollypm/web_api/routes/chat_messages.py tests/test_chat_messages_endpoint.py\n- git diff --check origin/main...HEAD\n- uv run --with ruff ruff check src/pollypm/web_api/routes/chat_messages.py tests/test_chat_messages_endpoint.py\n- HOME=mktemp uv run --extra test pytest tests/test_chat_messages_endpoint.py::test_messages_endpoint_include_thinking_surfaces_under_default_curl tests/test_chat_messages_endpoint.py::test_messages_endpoint_emits_thinking_envelopes_when_include_thinking_true -q (2 passed)\n- HOME=mktemp uv run --extra test pytest tests/test_chat_messages_endpoint.py -q (48 passed)\n\nResidual risk: this overlaps the pending Codex thinking PR #2258, so that branch will need rebase/scope review after this merge. I am not using GitHub's formal approval button because the platform treats this as the same operator account.

PR #2264, opened 2026-05-24
fix(web): mint per-bootstrap session cookies

### LINKED ISSUE #2245: Single shared session cookie: all browser contexts share one pollypm-session token
## Summary
The server issues one `pollypm-session` cookie per Tailscale IP address. All browser tabs, browser contexts, or devices behind the same Tailscale IP share the **same session token**. There is no concept of per-tab or per-user identity.

## Observed behavior
Two separate Playwright browser contexts (separate cookie jars) both receive the same cookie value: `ysQmcaWCv7wWi3vUkz4j...`. This is by design (the auth comment in app.js confirms: 'cookie issuance is gated to loopback / Tailscale CGNAT peer'), but has usability consequences.

## Consequences
1. **Cannot simulate two different operators** from the same machine — both tabs ARE the same identity
2. **No per-tab claim attribution**: when Tab A claims a task with `actor=tab-a`, the session identity is shared with Tab B
3. **Session expiry warning fires simultaneously** in all open tabs (same issued-at cookie)
4. **No way to open the UI as two different operators** without using two different Tailscale devices

## Questions for design consideration
- Is there a planned multi-user model where different people have separate identities?
- If so, the current single-token-per-IP approach blocks this
- Even a simple user-name header or UI-level persona switch would help distinguish concurrent sessions in logs

## Not a bug per se — flagged for intentional design review
The auth model is documented and intentional. Flagging because it may become surprising as multi-user usage grows.

## Environment
Playwright, same machine, 2026-05-24

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

## Summary
- `/ui/` now mints a unique opaque signed `pollypm-session` cookie per bootstrap instead of storing the raw API bearer token.
- Cookie validation uses the current on-disk API token as the signing key, so `pm api regen-token` invalidates browser sessions without server-side session state.
- Legacy raw-token cookies are rejected with a reload hint.
- Adds coverage for unique cookies behind the same Tailscale peer, cookie tampering/expiry/rotation, SSE cookie auth, and chat route cookie auth.

Fixes #2245.

## Verification
- `python -m py_compile src/pollypm/web_api/auth.py src/pollypm/web_api/app.py tests/web_api/test_auth.py tests/web_api/test_web_ui.py tests/test_chat_messages_endpoint.py`
- `uv run --with ruff ruff check src/pollypm/web_api/auth.py src/pollypm/web_api/app.py tests/web_api/test_auth.py tests/web_api/test_web_ui.py tests/test_chat_messages_endpoint.py`
- `HOME=$(mktemp -d /tmp/pollypm-cookie.XXXXXX) uv run --extra test pytest tests/web_api/test_auth.py -q` (11 passed)
- `HOME=$(mktemp -d /tmp/pollypm-cookie-focused.XXXXXX) uv run --extra test pytest tests/web_api/test_auth.py tests/web_api/test_web_ui.py tests/test_chat_messages_endpoint.py -q` (115 passed)
- `git diff --check origin/main...HEAD`

Earlier worker attempt with bare `uv run pytest ...` failed before tests because pytest is only installed through the repo test extra.

## Residual Risk
Tabs in the same browser profile still share cookies by normal browser behavior. Separate browser contexts/devices now receive distinct opaque cookie values; this does not add a full multi-user/persona model.

### DIFF
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -32,10 +32,12 @@ info:
        `~/.pollypm/api-token`. Always accepted; canonical mode for
        non-browser clients (cockpit CLI, scripts, curl). Modeled here as
        `bearerAuth`.
-    2. **Session cookie** — `pollypm-session=<token>`, minted by
-       `GET /ui/` only when the caller is a loopback peer, a verified
+    2. **Session cookie** — `pollypm-session=<opaque-session>`, minted
+       by `GET /ui/` only when the caller is a loopback peer, a verified
        tailnet peer (with `tailnet_trust_enabled=True`), or already
-       presented a valid bearer header. Modeled here as `cookieAuth`.
+       presented a valid bearer header. The cookie is signed with the
+       current API token, so token rotation invalidates browser sessions.
+       Modeled here as `cookieAuth`.
     3. **Credential-free tailnet peer** — accepted only when the server
        was constructed with `tailnet_trust_enabled=True`, which `pm
        serve` auto-sets when `detect_tailscale_ip()` returns a verified
@@ -2868,10 +2870,11 @@ components:
       description: |
         Session cookie minted by `GET /ui/` for loopback peers,
         verified tailnet peers (when `tailnet_trust_enabled=True`), or
-        callers presenting a valid bearer header. The cookie value is
-        the same on-disk token as `bearerAuth`; comparison is
-        constant-time. LAN clients receive HTML without `Set-Cookie`
-        and 401 on the first API call. See `docs/web-api-spec.md` §3.
+        callers presenting a valid bearer header. The cookie is an
+        opaque per-browser session value signed with the current API
+        token; the raw bearer token is not stored in the cookie. LAN
+        clients receive HTML without `Set-Cookie` and 401 on the first
+        API call. See `docs/web-api-spec.md` §3.
 
   parameters:
     ProjectKey:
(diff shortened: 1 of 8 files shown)


samhotchkiss (commented):
Claude reviewer LGTM (cannot APPROVE as same gh-auth as PR author).

Per-bootstrap signed session cookies replace the shared raw-token cookie. HMAC-SHA256 over `v1.issued_at.nonce` keyed by the on-disk API token, random nonce per `/ui/` boot, 7-day TTL, 300s clock-skew tolerance, legacy 1-part cookies rejected with reload hint.

Verified foreground in fresh HOME:
- `pytest -k 'cookie or session or bootstrap'` on test_auth.py + test_web_ui.py + test_chat_messages_endpoint.py — 27 passed
- Full modules — 115 passed (matches PR description)

Critical guarantees confirmed by tests:
- Same Tailscale peer hitting `/ui/` twice gets two distinct cookies (multi-operator simulation).
- Existing tabs preserve their session — each tab's cookie independently valid until rotation or TTL.
- `pm api regen-token` invalidates outstanding sessions via HMAC key change; refreshed `/ui/` mints a new valid cookie.
- Raw bearer token in cookie slot returns 401 invalid_token (legacy rejection).
- Tampering / expiry / rotation all return invalid_token with reload hint.

Residual risk acknowledged in PR body (tabs in same profile still share cookies — normal browser behavior, not a full persona model) is appropriate scope for #2245.

Merging via squash per merge protocol.

PR #2285, opened 2026-05-25
fix(audit): mirror pause skips to canonical log

### LINKED ISSUE #2282: bug:pause-skip-audit-not-emitted: §5.4.1 paused session sees no recovery attempt and no skip audit
## Tested
SHA: b27fa743dfcaa194805346ffdb351e96dad98da3 (main)
Date: 2026-05-24 ~21:33 MDT (03:33Z 2026-05-25)
Wave: §05 destructive wave #2 (Freya / Sam asleep)

## Repro

1. Pause `architect_smoketest` via REST: `POST /api/v1/sessions/architect_smoketest/pause` → 200, marker now `["architect_smoketest", "operator"]`.
2. `tmux kill-window -t pollypm-storage-closet:architect-smoketest`.
3. Confirm sessions API correctly reports it: `paused=True status=missing window=None`.
4. Wait 6.5 minutes.
5. `tmux list-windows -t pollypm-storage-closet | grep architect-smoketest` → no respawn (correct — pause honored).
6. `grep "session.pause.skip" ~/.pollypm/audit/smoketest.jsonl` → 0 events.
7. `grep "session.pause.skip" ~/.pollypm/audit/_workspace.jsonl` → 0 events.
8. After `POST /sessions/architect_smoketest/resume`, the pane respawns within 30s (recovery loop IS alive, just gated).

## Observed vs expected

- **Pause-honoring behavior:** correct — pane was not respawned for 6+ min while paused, respawned within 30s after resume. This is what the §5.4.1 spec expects.
- **Audit visibility:** test plan §5.4.1 expects `session.pause.skip` events for both `no_session_spawn` and `supervisor.maybe_recover_session` loops. **Zero such events were emitted across the entire pause window**, in either the per-project (`smoketest.jsonl`) or workspace (`_workspace.jsonl`) audit log.

This means the pause-marker enforcement is **silent** from an observability standpoint. An operator inspecting why a session isn't recovering has no audit trail telling them "the recovery loop saw it, but skipped because paused."

Two possible causes (need source-side dx):

a) The recovery loops in rail_daemon's `supervisor` *did* run during the window but the skip-audit emission throttle or the marker check is gated incorrectly so no event ever lands.
b) The recovery loops never ran against `architect_smoketest` at all during the 6-min window (e.g., the supervisor only triggers from `no_session` alerts which require a task dispatch trigger, and an empty pane with no pending work doesn't raise one). If that's the case the test-plan claim "recovery loop tried and was skipped" is also wrong — there was nothing to skip.

Either way, this contradicts §5.4.1's expected audit signature and the message returned by the pause endpoint ("pause marker is HONORED by recovery/relaunch ... heartbeat per-session processing"). Either the doc lies, the emission is broken, or the docs need to clarify that skip events only land when there's a concrete recovery trigger.

## Environment

- pm serve started fresh, /api/v1/health 200.
- rail_daemon running (PID 89021, `python -m pollypm.rail_daemon`).
- heartbeat-supervisor tmux pane is sitting at a fresh Claude Code welcome screen, not actively running the supervisor wrapper.
- §5.1 (kill/restart/SIGKILL) all PASS — pm serve is operating correctly.

## Suggested fix

1. Confirm whether `supervisor.maybe_recover_session` and `no_session_spawn` actually run on a timer in rail_daemon, or only in response to alerts/dispatch.
2. If they run on a timer: trace why the skip emission is silent.
3. If they only run on alerts: rewrite §5.4.1 to inject the trigger explicitly (e.g., queue a task or wait for the cascade), and adjust the API pause response message to not claim "HONORED" for paths that simply never fire.

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

## Summary
- Mirror throttled `session.pause.skip` events into the canonical audit JSONL via `pollypm.audit.emit`.
- Preserve the existing message-store event for cockpit consumers and keep the guard best-effort if either audit path fails.
- Add regression coverage for direct guard calls plus no-session and supervisor recovery skip paths.

Fixes #2282.

## Verification
- `uv run --extra test pytest -q -p no:cacheprovider tests/test_session_paused_marker_wiring.py::test_skip_if_paused_emits_audit_event_on_skip tests/test_session_paused_marker_wiring.py::test_skip_if_paused_throttles_repeat_audit_emission tests/test_session_paused_marker_wiring.py::test_skip_if_paused_throttle_is_per_loop tests/test_session_paused_marker_wiring.py::test_skip_if_paused_swallows_store_errors tests/test_session_paused_marker_wiring.py::test_auto_recover_no_session_skips_paused_session tests/test_session_paused_marker_wiring.py::test_supervisor_maybe_recover_session_skips_paused_session`
- `uv run --extra test pytest -q -p no:cacheprovider tests/test_session_paused_marker_wiring.py`
- `uv run --with ruff ruff check src/pollypm/session_paused.py tests/test_session_paused_marker_wiring.py`
- `git diff --check`

### DIFF
--- a/src/pollypm/session_paused.py
+++ b/src/pollypm/session_paused.py
@@ -407,10 +407,11 @@ def skip_if_paused(
     is paused.
 
     ``store`` — when supplied, the helper emits a ``session.pause.skip``
-    event via ``store.record_event(scope, sender, subject)`` so the
-    cockpit / audit log records that the loop honoured the marker
-    rather than silently dropping the call. Best-effort: any store
-    error is swallowed so a flaky event-write can't unblock the guard.
+    event via ``store.record_event(scope, sender, subject)`` and the
+    canonical audit log so cockpit readers and ``pm audit`` can both
+    see that the loop honoured the marker rather than silently
+    dropping the call. Best-effort: any event-write error is swallowed
+    so a flaky audit path can't unblock the guard.
 
     ``loop`` / ``reason`` — free-form context used to build the audit
     subject. ``loop`` is the calling site (e.g. ``"supervisor.maybe_recover_session"``,
@@ -432,7 +433,7 @@ def skip_if_paused(
     if not is_paused(config, session_name):
         return False
     if store is not None and _should_emit_skip(session_name, loop):
-        _emit_pause_skip(store, session_name, loop=loop, reason=reason)
+        _emit_pause_skip(config, store, session_name, loop=loop, reason=reason)
     return True
 
 
@@ -473,7 +474,12 @@ def _reset_skip_throttle_for_tests() -> None:
 
 
 def _emit_pause_skip(
-    store: Any, session_name: str, *, loop: str = "", reason: str = "",
+    config: Any,
+    store: Any,
+    session_name: str,
+    *,
+    loop: str = "",
+    reason: str = "",
 ) -> None:
     """Best-effort emit of the ``session.pause.skip`` audit event.
 
@@ -490,9 +496,6 @@ def _emit_pause_skip(
     failure is swallowed: the guard side is the contract, the audit
     event is observability only.
     """
-    record = getattr(store, "record_event", None)
-    if not callable(record):
-        return
     subject_loop = loop or "unknown_loop"
     subject = (
         f"skipped {session_name} — {subject_loop} honored pause marker"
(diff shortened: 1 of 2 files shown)


samhotchkiss (commented):
Approved per parallel reviewer pass: regression test in tests/test_session_paused_marker_wiring.py is non-hollow (uses POLLYPM_AUDIT_HOME tmp dir + reads back via pollypm.audit.log.read_events), covers both loop names from #2282 plus flaky-store fallback. pytest 33 passed in 95s. Merging.

PR #2287, opened 2026-05-25
fix(ui): preserve surface rail errors during push refresh

### LINKED ISSUE #2286: bug:ui:surface-rail-cycles-on-sse-refresh — error envelope wiped before render
## Observed (during PR #2283 reviewer's Playwright probe)

After PR #2283 (\`d0cf3c84\`) landed — which correctly renders project + surface rails independently and adds bounded request timeouts — the **surface rail still cycles indefinitely under real-world load**.

Diagnostic instrumentation confirmed:
- The 10s rail-request timeout **does fire**.
- \`surfaceLoadError\` **is set** with the right error state.
- **But an SSE-triggered \`surfacesRefreshQueued\` reload immediately wipes the error before render.**

Net user experience: surface rail keeps showing "surfaces is taking longer than expected" + Retry button, cycling every time SSE pings. Project rail (sibling fix from #2283) works correctly — populates within 5s under the same load.

## Why this isn't #2273

#2273 was the parent cold-load silent-stall. #2283 fixed half of it: project rail now renders independently within 5s. The surface rail half hits this **separate** SSE-vs-error-state race.

## Why it's pre-existing (not a #2283 regression)

The reviewer verified main baseline (without #2283 applied) also exhibits surface-rail-stuck behavior, just with the "loading…" placeholder instead of the slow-load error. #2283 made the error visible; the cycle is the original bug.

## Repro

Tested SHA: \`d0cf3c84a2f6ecb6\` (post-merge of PR #2283).

1. Real pm serve with prod state (12+ projects / 51+ sessions / 200+ tasks suffices to expose the latency).
2. Drive Playwright at \`/ui/\` with cold cache.
3. Watch \`#surface-list\` for 30+s.
4. Observe: error state appears (correct), then is wiped by next SSE refresh, then re-appears, then is wiped — indefinitely.

## Suggested fix direction

Either:
- The SSE \`surfacesRefreshQueued\` handler should preserve the current error state until a successful refresh actually arrives (clear on success, not on queue), OR
- The error state should be sticky for a brief grace period (e.g., 5s) before being eligible for refresh-driven clear.

## Context

- Filed from PR #2283 reviewer's findings (\`/tmp/sub-pr2283-review.md\`).
- Wave-1 + re-verify probes also saw this symptom but couldn't isolate the SSE race until #2283 made the error states visible.

claude-created

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

## Summary
- Keep existing rail error envelopes visible during SSE/periodic background surface refreshes until replacement data succeeds.
- Carry the preserve-error intent through coalesced queued surface refreshes.
- Add a Playwright regression for the timeout + SSE queued-refresh race from #2286.

## Tests
- `git diff --check`
- `POLLYPM_BASE_URL=http://127.0.0.1:8776 npm test -- --grep 'SSE queued surface refresh'`
- `POLLYPM_BASE_URL=http://127.0.0.1:8776 npm test -- tests/surfaces.spec.ts --project=chromium`

Fixes #2286

### DIFF
--- a/src/pollypm/web_api/ui/app.js
+++ b/src/pollypm/web_api/ui/app.js
@@ -63,6 +63,7 @@
     sseStableOpenTimer: null,
     surfacesInFlight: false,
     surfacesRefreshQueued: false,
+    surfacesQueuedPreserveErrors: false,
     dashboardInFlight: false,
     dashboardRefreshQueued: false,
     historyInFlight: {},
@@ -424,12 +425,18 @@
 
   async function loadSurfaces(opts) {
     const force = opts && opts.force;
+    const preserveErrors = opts && opts.preserveErrors;
     if (state.surfacesInFlight) {
       if (force) {
         abortFetchState("surfaces");
         state.surfacesInFlight = false;
+        state.surfacesRefreshQueued = false;
+        state.surfacesQueuedPreserveErrors = false;
       } else {
         state.surfacesRefreshQueued = true;
+        state.surfacesQueuedPreserveErrors = (
+          state.surfacesQueuedPreserveErrors || !!preserveErrors
+        );
         return;
       }
     }
@@ -437,9 +444,11 @@
     state.surfaceLoading = true;
     state.taskLoading = true;
     state.projectLoading = true;
-    state.surfaceLoadError = null;
-    state.taskLoadError = null;
-    state.projectLoadError = null;
+    if (!preserveErrors) {
+      state.surfaceLoadError = null;
+      state.taskLoadError = null;
+      state.projectLoadError = null;
+    }
     const request = beginFetchState(
       "surfaces",
       "surfaces",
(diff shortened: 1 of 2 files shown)


samhotchkiss (commented):
Approving — small surgical fix and a strong included regression. Verified locally on uv-run pm serve @ 8874:

- Included surfaces.spec.ts: 18 passed, 2 skipped (chromium). New 'SSE queued surface refresh preserves timeout error until success' case passes.
- Live probe (1s forced rail timeout, slow /chat/sessions stub, then loadSurfaces({preserveErrors:true}) — the same path refreshFromPush hits): surfaceLoadError remained set at t+1s, t+3s, t+5s after the SSE-driven queued refresh. After flipping the stub to success + force refresh, surfaceLoadError cleared to null. 0 console errors.
- Coalescing logic OR-accumulates preserve intent correctly; force:true resets both queue flags so a user retry won't carry stale preserve.

Merging.

PR #2340, opened 2026-05-26
Fix doctor API query-string fix runs

### LINKED ISSUE #2334: §1.3.4: POST /doctor/run?fix=true (query string) silently downgraded to fix=false; single-flight never engaged
**Found during:** §1.3.4 concurrency probe, 2026-05-25
**SHA:** `2722a642d`
**Severity:** ship-blocker — the canonical test-plan repro silently succeeds instead of exercising fix.

## Summary

`POST /api/v1/doctor/run?fix=true` (query-string form) is silently treated as `fix=false`. The query parameter is **not bound** to the request schema; only the JSON body field `{"fix": true}` triggers the fix path.

`docs/test-plan/01-task-lifecycle.md` §1.3.4 instructs testers to use exactly this query-string form. Five concurrent `?fix=true` calls all return 200 because none of them are actually requesting a fix — the single-flight lock is never engaged.

### Broken (test-plan repro): query-string form silently downgraded
```bash
( for i in 1 2 3 4 5; do
    curl -sS -m 120 -X POST -H "Authorization: Bearer $TOKEN" \
      -o /tmp/r_$i.json -w "i=$i http=%{http_code} t=%{time_total}\n" \
      "$BASE/api/v1/doctor/run?fix=true" &
  done; wait )
```

**Observed:** 5 × 200, `fixes_applied=[]` on every response. No 409. No fix actually applied. No audit event emitted.

### Working (body form): single-flight correctly engaged
```bash
( for i in 1 2 3 4 5; do
    curl -sS -m 120 -X POST -H "Authorization: Bearer $TOKEN" \
      -H "Content-Type: application/json" -d '{"fix": true}' \
      -o /tmp/rb_$i.json -w "i=$i http=%{http_code} t=%{time_total}\n" \
      "$BASE/api/v1/doctor/run" &
  done; wait )
```

**Observed:** 1 × 200 (9.7s, fix path), 4 × 409 `in_progress` (~60ms each, typed envelope with `retry_after_seconds:5`). Lock works correctly when actually reached.

## Root cause

`src/pollypm/web_api/routes/doctor.py:755-769` — `run_doctor_endpoint` accepts `body: DoctorRunRequest | None` and a `timeout_seconds` query param, but never declares a `fix` query param. FastAPI/Pydantic do not implicitly merge unbound query strings into the body model, so `?fix=true` is dropped on the floor. At line 799 the guard reads `payload.fix` (the body field), which defaults to `False`.

This is downstream of PR #2157 — the lock itself is correctly implemented; the bug is that the **only invocation form documented in the test plan** never reaches that code path.

## Why this matters

1. Every §1.3.4 test run since the test plan was written has been a false pass (or false fail) — the test never actually exercised what it claimed to.
2. Operators / scripts / dashboards that hit `?fix=true` think they're requesting repairs and getting them. They are not. Inconsistencies persist silently.
3. There is **no audit event** emitted by the API endpoint at all (CLI emits `pm.doctor_run`; the API endpoint does not). Combined with the silent downgrade, an operator triggering fixes via the API has zero observable record.

## Acceptance

- `POST /api/v1/doctor/run?fix=true` (query string) returns the same behavior as body `{"fix": true}` — including 409 on concurrent invocation.
- OR: query-string `fix` is explicitly rejected with a typed 422 telling the caller to use the body form.
- Update `docs/test-plan/01-task-lifecycle.md` §1.3.4 to whichever invocation form is canonical (currently the doc and the code disagree).
- API doctor run/fix attempts emit an audit event (parity with CLI `pm.doctor_run`), including the rejected 409 path.

## Related

- #2135 (closed) — same symptom (all 200s, no 409) was triaged as "uvicorn single-process serialization." It wasn't — the real cause is the query parameter never reaching the handler. Worth reopening or cross-linking.
- PR #2157 — added single-flight lock; correct, but only reachable via body form.
- `docs/test-plan/01-task-lifecycle.md` §1.3.4

## Probe artifacts

- Baseline (1 run, body absent): 200, duration 3.84s, 50 pass / 7 warn / 0 err, fixes_applied=0.
- 5 concurrent `?fix=true` query: all 5 × 200, total wall 13.9s, `fixes_applied=[]` on every response, 0 audit events.
- 5 concurrent `{"fix": true}` body: 1 × 200 (9.7s) + 4 × 409 (`in_progress`, retry_after_seconds=5, ~60ms), still 0 audit events.

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

## Summary
- Bind `POST /api/v1/doctor/run?fix=true` as the effective doctor fix flag instead of silently ignoring it.
- Emit `pm.doctor_run` audit rows from API doctor runs, including rejected single-flight `409 in_progress` attempts.
- Update the lifecycle test plan to document the API audit breadcrumb and `in_progress` envelope.

## Verification
- `uv run --extra test pytest tests/test_doctor_endpoint.py -q` (`25 passed in 251.14s`)

Closes #2334

### DIFF
--- a/docs/test-plan/01-task-lifecycle.md
+++ b/docs/test-plan/01-task-lifecycle.md
@@ -297,7 +297,10 @@ Find or create two open inbox items. Hit `archive` on both in parallel. Both sho
  wait)
 ```
 
-**Expected:** one 200, one 409 `busy` (single-flight per the doctor contract).
+**Expected:** one 200, one 409 `in_progress` (single-flight per the doctor contract).
+The 409 envelope includes `retry_after_seconds=5`.
+Each API attempt should emit `pm.doctor_run` in `~/.pollypm/audit/_workspace.jsonl`,
+including the rejected 409 path.
 
 ---
 
--- a/src/pollypm/web_api/routes/doctor.py
+++ b/src/pollypm/web_api/routes/doctor.py
@@ -446,6 +446,84 @@ def _fix_busy_error() -> APIError:
     )
 
 
+def _merge_run_request(
+    body: DoctorRunRequest | None,
+    *,
+    query_fix: bool | None,
+) -> DoctorRunRequest:
+    """Return the effective doctor run payload.
+
+    ``fix`` is accepted both in the JSON body and as a query parameter
+    because the public test plan historically used
+    ``POST /doctor/run?fix=true``. FastAPI does not merge query strings
+    into body models, so the route has to bind it explicitly.
+    """
+    payload = body or DoctorRunRequest()
+    if query_fix is None:
+        return payload
+    return DoctorRunRequest(check=payload.check, fix=query_fix)
+
+
+def _emit_doctor_run_audit(
+    *,
+    payload: DoctorRunRequest,
+    timeout_seconds: float,
+    status: str,
+    outcome: str,
+    query_fix: bool | None,
+    report: DoctorReport | None = None,
+    fixes_applied: list[dict[str, Any]] | None = None,
+    error_code: str | None = None,
+) -> None:
+    """Best-effort audit parity for API doctor runs.
+
+    Mirrors the CLI's ``pm.doctor_run`` event without coupling the
+    route to audit failures. The doctor endpoint is a mutation surface
+    when ``fix=true``; even rejected single-flight attempts should be
+    visible in the audit tail.
+    """
+    try:
+        from pollypm.audit import emit as _audit_emit
+    except Exception:  # noqa: BLE001
+        return
+
+    metadata: dict[str, Any] = {
+        "source": "web_api",
+        "route": "POST /api/v1/doctor/run",
+        "check": payload.check,
+        "timeout_seconds": round(float(timeout_seconds), 3),
+        "fix_requested": bool(payload.fix),
+        "fix_query_param": query_fix,
+        "outcome": outcome,
+    }
+    if report is not None:
+        metadata.update({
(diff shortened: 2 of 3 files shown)


samhotchkiss (commented):
APPROVE — verified live with worktree `uv run pm serve` on :18765.

**Acceptance vs #2334:**
- 5× concurrent `POST /doctor/run?fix=true&check=python-version` → **1×200 + 4×409** with typed `in_progress` envelope (`retry_after_seconds:5`). Single-flight engaged via query string ✓
- Body form `{"fix":true}` regression guard: identical 1×200 + 4×409 ✓
- Audit: `~/.pollypm/audit/_workspace.jsonl` shows `pm.doctor_run` rows with `actor:api`, `source:web_api`, `fix_query_param:true`, distinct `outcome:rejected` (status warn) and `outcome:completed` (status ok) ✓

**Tests:** `uv run --extra test pytest tests/test_doctor_endpoint.py -x -q` — 25 passed in 207.85s, including the two new tests (`test_run_query_fix_true_invokes_fix_fn`, `test_api_doctor_run_audit_covers_query_fix_and_busy_rejection`).

**Note:** With the full check suite + default 25s timeout, the winning request can 504 on a busy workspace (the four losers still 409 correctly). Lock semantics are unchanged; consider raising the test-plan default timeout if §1.3.4 starts flaking on the winner.

Self-merge prohibited per PR header → squash-merging as reviewer.

PR #2358, opened 2026-05-26
fix(chat): restore transcript fidelity and speed sends

### LINKED ISSUE #2341: §2.4.3: force-bypass injections are not audit-logged
## Symptom

`POST /api/v1/chat/{session_name}/send?safety=force` (or body `safety: force`) bypasses three safety gates (`unsafe_mid_tool`, `unsafe_mid_stream`, `unsafe_unavailable_transcript|heartbeat`) and successfully injects into a live tmux pane — but writes ZERO audit-log entries acknowledging the override.

Per the §2.4.3 test plan and the general audit-stream contract (every consequential operation goes through `pollypm.audit.log.emit`), force-bypass overrides must be recorded so an operator can later answer "who/what forced an injection while the agent was streaming?"

## Evidence (live test against http://100.67.2.108:8765, 2026-05-25)

Setup: session `advisor_pm_test_01wave_1779715196` (Codex, idle test project in /private/tmp).

### Repro

```
# Strict refused (expected):
$ curl -X POST -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
    -d '{"text":"injection-test-241-<ts>","press_enter":false}' \
    "$BASE/api/v1/chat/advisor_pm_test_01wave_1779715196/send"
HTTP 409 {"error":{"code":"unsafe_mid_tool", ...}}

# Force succeeded (expected):
$ curl -X POST ... -d '{"text":"injection-test-241-<ts>","press_enter":false,"safety":"force"}' ...
HTTP 200 {"ok":true,"message_id":"msg_71942c7fba0a4c8e9c11819e6d05f55f", "characters_sent":29, ...}

# Force during active stream (expected, landed mid-Bohemian-Rhapsody):
$ curl -X POST ... -d '{"text":"force-bypass-test-243-<ts>","safety":"force"}' ...
HTTP 200 {"ok":true,"message_id":"msg_abd3cdfa39ae483f91125217421038e7", ...}
```

### Audit gap

```
$ grep -l "force-bypass-test\|msg_abd3cdfa\|chat.send\|safety.*force" ~/.pollypm/audit/*.jsonl
(no output)

$ grep -c "audit" src/pollypm/web_api/routes/chat_send.py
0
$ grep -c "emit" src/pollypm/web_api/routes/chat_send.py
3   # all three are doc comments — no pollypm.audit.log.emit calls
```

The send endpoint never imports or calls `pollypm.audit.log.emit`. Force-bypass leaves no audit trail.

## Why this matters

- Force is the only escape hatch when the strict gates false-positive (e.g. mid_tool firing on long-idle Codex panes — see related observation below). Operators WILL reach for it.
- Mid-stream force injection can corrupt agent state by racing the LLM mid-token. We need a record of every such override to investigate "what poisoned this transcript?" after the fact.
- Spec §2.4.3 implicitly assumes audit visibility (the test plan asks reviewers to verify the override is "clearly logged as a force bypass for audit"). Current behavior fails this.

## Suggested fix

In `src/pollypm/web_api/routes/chat_send.py::send_chat_message`, after the safety block resolves and `safety_mode == "force"` (or any time a gate was bypassed), call `pollypm.audit.log.emit` with:

- `event = "chat.send.force_bypass"` (or `"chat.send.loose_warning"` for warning-header sends)
- `project = surface.project`
- `subject = session_name`
- `actor = "rest_client"` (or extract from the auth principal if available)
- `status = "warn"`
- `metadata = {"window_target": target, "characters_sent": ..., "bypassed_gates": [...], "press_enter": body.press_enter}`

Do NOT log the message text itself (transcripts already have it; audit should not duplicate user content).

## Related observation (separate ticket worth filing)

The `unsafe_mid_tool` gate fires on long-idle Codex panes that have NO actually-pending tool_use — the gate appears to treat an orphaned/stale open tool_use in the transcript as live. Out of four advisor_* sessions I sampled, all four 409'd strict and loose. This forces operators to use `safety=force` for routine sends, which is exactly the wrong incentive.

If this is intentional (defense-in-depth on stale transcripts), the hint should suggest `safety=loose` works — but it doesn't, because loose ALSO refuses on mid_tool (route line 1352-1361). Either loosen `loose` to allow this case, or change the hint, or add a transcript-staleness check that auto-resolves the orphan.

## Tested

- §2.4.1 happy path: PASS (with `safety=force`; strict/loose both blocked by mid_tool false-positive — see related obs)
- §2.4.2 mid-stream safety: PASS — strict 409'd, no pane corruption
- §2.4.3 force bypass: PASS injection-wise; FAIL on audit visibility (this bug)
- §2.4.4 dead session: PASS — 503 `window_missing` typed envelope, not 500

### LINKED ISSUE #2342: §2.4: strict/loose mid_tool gate false-positives on idle Codex advisor panes
## Symptom

Every Codex advisor pane I sampled during §2.4 testing returns HTTP 409 `unsafe_mid_tool` to BOTH `safety=strict` and `safety=loose` POSTs to `/api/v1/chat/{session_name}/send`, despite the pane being idle (no streaming, no live tool_use — capture-pane shows the agent sitting at an empty `›` prompt).

This means operators must reach for `safety=force` for routine sends, which is exactly the wrong default behavior — and now (post-#2341) ALSO means every routine inject would generate a force-bypass audit warning once that bug is fixed.

## Evidence (live test against http://100.67.2.108:8765, 2026-05-25)

```
$ curl -X POST ... -d '{"text":"hello"}' "$BASE/api/v1/chat/advisor_pm_test_01wave_1779715196/send"
HTTP 409 {"error":{"code":"unsafe_mid_tool", ...}}

$ curl -X POST ... -d '{"text":"hello","safety":"loose"}' "$BASE/api/v1/chat/advisor_pm_test_01wave_1779715196/send"
HTTP 409 {"error":{"code":"unsafe_mid_tool", ...}}  # loose ALSO refuses

$ curl -X POST ... -d '{"text":"hello","safety":"force"}' "$BASE/api/v1/chat/advisor_pm_test_01wave_1779715196/send"
HTTP 200 OK
```

Sampled four idle advisor panes (`advisor_smoketest`, `advisor_pm_test_01wave_1779715196`, `advisor_pm_test_05wave4_1779714744`, `advisor_media`); all four exhibited the same 409 behavior.

Pane state at time of test (capture-pane):
```
╭──────────────────────────────────────────────────────────╮
│ >_ OpenAI Codex (v0.133.0)                               │
│ model:       gpt-5.4   /model to change                  │
│ permissions: YOLO mode                                   │
╰──────────────────────────────────────────────────────────╯

› Explain this codebase   <-- placeholder text, NOT a live tool_use
```

So the transcript contains an orphaned/stale open tool_use from a prior session that never received a tool_result. `_is_mid_tool` in `chat_send.py` (line 1361) treats this as live activity.

## Root cause hypothesis

`_is_mid_tool(events_path, exempt_tool_ids=...)` scans the events archive for unmatched `tool_use` envelopes. For Codex panes that crashed / were restarted mid-tool, the archive retains the orphan forever. There's no staleness ceiling (e.g. "only consider tool_use envelopes within the last N seconds OR newer than the last heartbeat").

## Suggested fix options

1. Apply a staleness window: only flag a tool_use as 'mid-tool' if it appeared within the last N seconds (suggest 30s, matching `_MID_STREAM_WINDOW_SECONDS` semantics).
2. Make `safety=loose` actually loosen the mid_tool gate (currently loose only loosens mid_stream — surprising, given the name).
3. Cross-check with the heartbeat: if heartbeat is fresh AND last assistant text is more recent than the orphan tool_use, treat the tool_use as stale.

Option 3 is the most defensible: it composes with the existing safety primitives instead of adding a magic constant.

## Impact

- Routine sends to any advisor/architect pane that ever crashed mid-tool now require `safety=force` — eliminating the protective value of the gates AND (per #2341) flooding audit with false force-bypass warnings.
- Click-rule budget violated: operators get a confusing 409 hint suggesting force, which is the destructive override.

## Related

- #2341 — force-bypass not audit-logged (companion to this, found in the same §2.4 sweep)

### LINKED ISSUE #2343: §2.1.5: subagent JSONL parser drops tool_use/tool_result blocks; emits empty-text envelopes
## Summary

`parse_raw_subagent_jsonl` in `src/pollypm/web_api/chat/transcripts.py` (#2083) loses all tool activity inside Claude subagent transcripts and emits empty-text envelopes for tool-only turns. When `include_subagents=true` inlines a real subagent transcript, the consumer sees only the assistant's prose lines and a stream of blank message bubbles where tool calls happened.

## Reproduction

Real corpus: `~/.claude/projects/-Users-sam-dev-pollypm/9cc472bc-fc4c-48f2-bfce-0e1580bff89d/subagents/agent-aa1b8ec9f28109337.jsonl` (41 lines, sidechain subagent that reviewed and merged PR #2105).

```python
from pathlib import Path
from pollypm.web_api.chat.transcripts import parse_raw_subagent_jsonl
p = Path('~/.claude/projects/-Users-sam-dev-pollypm/9cc472bc-fc4c-48f2-bfce-0e1580bff89d/subagents/agent-aa1b8ec9f28109337.jsonl').expanduser()
envs = parse_raw_subagent_jsonl(p, actor_fallback='subagent')
# total: 39
# distinct types: {'text': 39}   <- no tool_use, no tool_result, no subagent_spawn
# empty text: 32                  <- tool-only turns become blank envelopes
```

Raw JSONL has 16 `tool_use` blocks (`Bash`, `Read`, `Grep`, `Edit`, etc.) plus matching `tool_result` user lines — none of them survive the translation.

## Root cause

`_raw_subagent_line_to_envelope` (lines 1404-1475) only emits `MessageType.TEXT` for user/assistant lines, calling `_extract_text_from_blocks` which drops every block that isn't `{type: ..., text: str}`. `tool_use`/`tool_result` blocks have no `.text` field, so they contribute nothing — but the function still emits the envelope, producing a blank record.

The parser branch for user/assistant turns:
```python
if line_type in _RAW_ASSISTANT_TYPES:
    text = _extract_text_from_blocks(content) or _extract_text_from_blocks(message)
    return MessageEnvelope(
        ...
        type=MessageType.TEXT,
        text=text,           # may be empty for tool-only turns
        ...
    )
```

Compare to the events.jsonl path (`_envelope_assistant_turn` etc.) which expands content blocks into separate `tool_use` / `tool_result` envelopes per #2079.

## Impact

- §2.1.5 acceptance for tool activity in inlined subagent transcripts is FAIL.
- §2.1.3 (tool_use/tool_result envelopes round-trip with `metadata.tool_name`, `tool_input`, `tool_use_id`) is FAIL when applied to subagent content.
- Web UI bubble stream gets ~80% blank messages per real subagent run (32/39 empty in the sampled corpus).
- Subagent debugging (which tool the child ran, what it returned) is impossible through the API.

## Suggested fix

Iterate over the assistant `content` block list and emit a `MessageEnvelope` per block, mirroring `_envelope_assistant_turn` in the same file:
- `{type: text, text: ...}` → `MessageType.TEXT` envelope
- `{type: tool_use, ...}` → `MessageType.TOOL_USE` with `metadata.tool_name`, `tool_input`, `tool_use_id`
- `{type: thinking, ...}` → gated by `include_thinking` flag (carry through into `parse_raw_subagent_jsonl` signature)
- `{type: tool_result, ...}` in user lines → `MessageType.TOOL_RESULT` with `tool_use_id`

Skip envelope emission entirely when a turn yields no blocks of interest (don't emit blank-text envelopes).

## What passed in §2.1.5

- Cache-poisoning guard (round-2 fix) verified across `operator`, `architect_pollypm`, `architect_bikepath`, `architect_coffeeboardnm`: `include_subagents=true` followed by default request leaves zero `metadata.subagent_transcript` keys in cached entries.
- Direct hit of a subagent session id (`agent-aa1b8ec9f28109337`) returns typed `session_unknown` 4xx — no parent context leak.
- Envelope schema clean; no internal-leak keys.

## What blocks full §2.1.5 verification

- No live PollyPM-managed Claude persona in this engagement actually invoked the Task tool — 1225 raw JSONLs under `~/.pollypm/homes/*/.claude/projects/`, zero `"name":"Task"` occurrences. So no real `subagent_result` envelope was ever produced in any monitored surface to inline against. The parser bug above was found by running `parse_raw_subagent_jsonl` directly against a real subagent JSONL outside the PollyPM session graph.

### LINKED ISSUE #2344: §2.2.1: Codex transcript ingestor drops every response_item, including all tool_call/function_call/reasoning blocks
## Symptom
Every Codex chat surface in the live system returns `transcript_source: \"capture\"` from `GET /api/v1/chat/{session}/messages`. The normalized JSONL path (`_normalize_codex_line`) is never the source for Codex sessions in practice, contradicting the §2.2.1 spec contract that ``Codex sessions are normalized via `_normalize_codex_line`, NOT via tmux capture-pane (that's a staleness fallback only)``.

## Root cause
`src/pollypm/transcript_ingest.py::_normalize_codex_line` has a fast-path early return:

```python
if entry_type != \"event_msg\":
    return events
```

But inspection of `~/.codex/sessions/2026/05/25/rollout-*.jsonl` shows Codex emits the following top-level entry types in a current session (one session, full rollout):

```
  88 event_msg
 248 response_item
   1 session_meta
   1 turn_context
```

All `response_item` lines — which carry the actual content — are silently dropped. In the same session the `response_item` payload distribution is:

```
   7 custom_tool_call
   7 custom_tool_call_output
  78 function_call
  78 function_call_output
  19 message            <-- assistant/user text
  56 reasoning
   1 tool_search_call
   1 tool_search_output
```

Meanwhile the only `event_msg` payload types present in this session are `agent_message`, `patch_apply_end`, `task_started`, `token_count`, `user_message`. The ingestor switch only matches `user_message`, `assistant_message` (a name that Codex never emits — Codex uses `agent_message`), `token_count`, `tool_call`, `tool_result`, `turn_end`, `error`. So in practice the ingestor only captures token usage and the single bootstrap user_message; it never records assistant turns, never records any tool call/result, never records any reasoning.

## Evidence
- `grep -c '\"event_type\":\"tool_call\"' /Users/sam/dev/pollypm/.pollypm/transcripts/*/events.jsonl` finds 0 across every Codex transcript on disk.
- Looped over every project's `.pollypm/transcripts/*/events.jsonl`: not a single Codex transcript carries any tool_call/tool_result event.
- `GET /api/v1/chat/advisor_pollypm/messages?source=jsonl` returns `404 archive_missing` — no JSONL path resolves for the advisor at all, because the ingestor stores the architect's cwd in its tagged Codex transcripts (advisor surfaces never get an archive matched to them).

## Impact
- §2.2.1 PASS criterion `metadata.tool_name populated, Source is the normalized JSONL archive, not capture-pane` cannot be satisfied for any Codex surface today.
- Web UI / REST consumers see Codex tool use as raw pane text only.
- Cross-surface read consistency (§2.6) for Codex is structurally impossible.

## Suggested fix direction
- Extend `_normalize_codex_line` to handle `entry_type == \"response_item\"` with payload subtypes `message`, `function_call`, `function_call_output`, `custom_tool_call`, `custom_tool_call_output`, `reasoning`.
- Add `agent_message` as an alias for `assistant_message` in the existing switch.
- Backfill: a one-shot rescan of `~/.codex/sessions/**/rollout-*.jsonl` once the normalizer handles the new shapes.
- Wire the advisor cwd into the Codex transcript fingerprint so `lookup_transcript_path` can attach the archive to the right surface.

## Repro
```bash
export BASE=http://100.67.2.108:8765 TOKEN=$(cat ~/.pollypm/api-token)
curl -sS -H \"Authorization: Bearer $TOKEN\" \"$BASE/api/v1/chat/advisor_pollypm/messages?source=jsonl&limit=5\"
# -> 404 archive_missing

curl -sS -H \"Authorization: Bearer $TOKEN\" \"$BASE/api/v1/chat/advisor_pollypm/messages?limit=5\" | jq '.transcript_source'
# -> \"capture\"
```

Found during §2.2 live testing on http://100.67.2.108:8765.

### LINKED ISSUE #2345: §2.2.2: capture-pane fallback engages for every Codex session — not just on staleness — masking the §2.2.1 ingest gap
## Symptom
Per spec §2.2.2 + the docstring on `tmux_capture.capture_envelopes` + the docstring on `_load_envelopes`, capture-pane is a **staleness fallback only** — it should engage when the normalized JSONL archive is >60s stale (`STALE_THRESHOLD_SECONDS`) AND the tmux pane is alive.

In practice, `transcript_source: \"capture\"` is returned for every live Codex surface unconditionally:

```bash
for s in advisor_pollypm advisor_bikepath advisor_savethenovel advisor_smoketest; do
  curl -sS -H \"Authorization: Bearer $TOKEN\" \"$BASE/api/v1/chat/$s/messages?limit=20\" \
    | jq '.transcript_source'
done
# -> \"capture\" \"capture\" \"capture\" \"capture\"
```

## Root cause (composite)
The capture path itself is well-behaved: it tags `metadata.from_capture: true`, emits only `type: text` envelopes (no fabricated tool_use), and the strict modes raise typed 503s correctly. The break is upstream in surface→archive resolution:

1. `lookup_transcript_path` requires fingerprint match on `(cwd, account_name, provider)` (`web_api/chat/transcripts.py:201`).
2. `_normalize_codex_line` drops most Codex events (see #2344) AND the Codex transcripts that DO get written carry the **architect**'s cwd (`/Users/sam/dev/pollypm/.pollypm/worktrees/architect_pollypm/...`) — never the advisor's worktree cwd.
3. So every advisor surface fails archive lookup → returns `None` for `transcript_path` → `_load_envelopes` falls through to capture even though the archive is not stale, it's simply unmatchable.

The §2.2.2 spec PASS criterion — `Source field in response indicates capture` and `Does NOT fire just because provider == codex` — fails: it absolutely fires whenever provider == codex, masking the deeper #2344 bug.

## Narrow PASS
The capture-pane code itself does not fabricate events. Every captured envelope is `type=text`, `role=assistant`, `metadata.from_capture=true`. The `source=jsonl` strict path correctly returns `404 archive_missing` rather than silently falling back. Unknown sessions return `404 session_unknown` cleanly. So the capture-pane module's defensive contract holds.

## Impact
- Stale-detection logic is untestable for Codex (every surface is already in the fallback).
- The fallback is the *only* read path for Codex, so capture-pane's known limits (rendered text only, no tool_name metadata, deduped via blake2b, capped at 3000 lines) become Codex's de facto API surface.
- Spec §2.6 triple-witness cannot be exercised for Codex.

## Suggested fix
Track in tandem with #2344. Once `_normalize_codex_line` produces useful events for response_item types, also ensure the Codex transcript writer tags the **actual worktree cwd** for the surface that owns the rollout (advisor / advisor_*, architect_*, etc.) so `lookup_transcript_path` can match.

Add a test surface or assertion that catches the regression: at least one Codex advisor session in CI must resolve to a non-None `transcript_path` with `transcript_source: \"jsonl\"`.

Found during §2.2.2 live testing on http://100.67.2.108:8765.

### LINKED ISSUE #2346: §2.1.2: include_thinking=true never surfaces thinking envelopes (account_name fingerprint stale)
## Symptom

`GET /api/v1/chat/{session}/messages?include_thinking=true&source=jsonl` returns **zero** `type:"thinking"` envelopes across **every** live session (80+ surfaces probed) on the running server at `http://100.67.2.108:8765`, despite the underlying archives containing dozens of `event_type:"thinking"` rows.

Test plan §2.1.2 (docs/test-plan/02-translation-layer.md) fails: the thinking envelope simply never appears, regardless of provider, surface_type, or include_thinking flag.

## Root cause

`pollypm.web_api.chat.registry` calls `lookup_transcript_path(..., account_name=session.account, ...)` (registry.py:403-408). The lookup filters strictly on `account_name` (transcripts.py:226-235). When the user's claude home directory binding produces a new `account_name`, the old session_config record still carries the previous `account_name` — so the lookup matches a **stale, abandoned** archive sharing the same `cwd` instead of the **current, actively-written** archive.

### Reproduction

Bikepath architect surface, both archives coexist under same cwd:

```
$ ls /Users/sam/dev/bikepath/.pollypm/transcripts/
46b488ed-cbff-4bf4-a1f0-f5e5adf32ef4/  <- API resolves here (empty, 0 thinking events, 5 days stale)
deb157e8-e253-4f15-a8b8-435616c5b05b/  <- actually active (25 thinking events, written today)

# 46b488ed fingerprint:
account= claude_s_swh_me  ts= 2026-05-19T20:57:40Z

# deb157e8 fingerprint:
account= claude_claude_swh_me  ts= 2026-05-25T17:04:46Z

# session config account: claude_s_swh_me  -> picks the stale one
```

Same pattern (account `claude_s_swh_me` vs `claude_claude_swh_me`) affects every architect surface I checked: bikepath, coffeeboardnm, health-coach, itsalive, polly-remote, pollypm, russell, samblog, savethenovel, smoketest.

### Suggested fix

Either:
1. Refresh `session.account` when the session is bound (so the config tracks the current claude account binding), OR
2. When `account_name` filter excludes ALL candidates under a matching cwd+provider, fall through to "freshest mtime" within that cwd+provider rather than returning `None`/stale, OR
3. Add `account_name=None` fallback after the strict lookup misses, picking the freshest cwd+provider match.

The strict-account filter was introduced to prevent cross-surface contamination (#2044), so option 3 should preserve that contract since the filter only relaxes after every account-filtered candidate has been rejected.

## Secondary finding (Codex reasoning not ingested)

While verifying Codex coverage I noticed `src/pollypm/transcript_ingest.py:486-487` short-circuits unless `entry_type == "event_msg"`. Codex rollouts emit reasoning as `entry_type:"response_item"` with `payload.type:"reasoning"` carrying a visible `summary[].text` block. These are silently dropped by the ingestor (see TODO comment at lines 605-610). Test plan §2.1.2 calls out "Try a Codex session too — Codex has different reasoning representation" — current behavior: dropped, never surfaced.

Sample Codex reasoning payload (from /Users/sam/.codex/sessions/...):

```json
{
  "timestamp": "2026-03-03T17:42:26.641Z",
  "type": "response_item",
  "payload": {
    "type": "reasoning",
    "summary": [{"type": "summary_text", "text": "**Acknowledging user and planning repo scan**"}],
    "content": null,
    "encrypted_content": "gAAAAA..."
  }
}
```

The visible `summary[].text` would map cleanly to a `thinking` envelope.

## Verification commands used

```bash
TOKEN=$(cat ~/.pollypm/api-token)
# Every session, every architect, force jsonl source, include_thinking=true:
for s in $(curl -sS -H "Authorization: Bearer $TOKEN" "http://100.67.2.108:8765/api/v1/chat/sessions" | jq -r '.sessions[].session_name'); do
  c=$(curl -sS -H "Authorization: Bearer $TOKEN" "http://100.67.2.108:8765/api/v1/chat/$s/messages?limit=500&include_thinking=true&source=jsonl" | jq '[.messages[]|select(.type=="thinking")]|length')
  [ "$c" -gt 0 ] && echo "$s $c"
done
# Result: 0 hits across 80+ sessions, despite >100 thinking events on disk.
```

## Test plan disposition

§2.1.2 default-request half (no thinking leaked) — **PASS** (trivially, since thinking is never emitted regardless).
§2.1.2 include_thinking=true half — **FAIL** (envelope never surfaces).
§2.1.2 order check — **N/A** (no envelopes to order).
§2.1.2 Codex variant — **FAIL** (reasoning dropped at ingest).

### LINKED ISSUE #2347: §2.3.1: REST envelope ids are duplicated (x4) for tool_use/tool_result on most worker sessions
## Symptom

`GET /api/v1/chat/{session}/messages` returns multiple envelopes with **identical `id`** within a single response. On 5 of 6 worker sessions sampled, every `tool_use` and `tool_result` envelope appears 4 times consecutively, with byte-identical `text`, `ts`, `metadata.tool_use_id`, and `metadata.tool_input`. Pagination is also impacted: a `limit=10` asc walk emits these duplicates at the page boundary and within each page.

## Repro (live API, RC build on 100.67.2.108:8765)

```bash
curl -sS -H "Authorization: Bearer $TOKEN" \
  "$BASE/api/v1/chat/task-bikepath-6/messages?limit=500&direction=asc" \
  | jq '.messages | group_by(.id) | map(select(length>1)) | length'
# => 62  (62 distinct ids appear 2+ times in a single response of 260 messages,
#         only 74 unique ids)
```

Sampled across worker tasks (single fetch, `limit=500`, `direction=asc`):

| session | total msgs | unique ids | dupe-multiplier |
|---|---:|---:|---:|
| task-bikepath-2 | 257 | 71 | x4 typical |
| task-bikepath-3 | 443 | 116 | x4 typical |
| task-bikepath-4 | 363 | 101 | x4 typical |
| task-bikepath-5 | 307 | 92 | x4 typical |
| task-bikepath-6 | 260 | 74 | x4 typical |
| task-russell-56 | 139 | 139 | clean |

Pagination walk (`limit=50&direction=asc` with `since_id` cursor) returns the same set of duplicates and so cumulative pages overcount likewise.

## Root cause (per source inspection)

The duplicates are **already present in `events.jsonl`** — the translation layer is faithfully replaying them. For id `toolu_01RNZFTBeLwqWxv5u3F8pwEV` in `/Users/sam/dev/bikepath/.pollypm/transcripts/1ab123e3-81b2-40d6-a934-6469545b9a70/events.jsonl`, four byte-identical lines exist with the same `source_path`, `source_offset=8488`, and `timestamp`. This points to `TranscriptIngestor` writing the same Claude session source record multiple times (e.g. duplicate flush, or replaying the same `source_offset` after a tail-restart).

Russell-56 (single-flush, no tool calls in window) is the only sampled archive without duplicates, consistent with a flush-loop hypothesis.

## Impact on §2.3

- **§2.3.1 pagination** technically passes (pages are stable, asc walk reverses to desc walk), but the duplicates make the API's `id` field non-unique, which breaks any downstream key-by-id storage (React `key=`, MUI list components, Web UI message cache).
- **§2.3.2 mtime cache** passes (second-hit 65–105ms vs first-hit 213ms; byte-identical payloads).
- **§2.3.3 tail-read** passes (cold archives at 573KB–955KB return in 50–75ms wall, 47–66ms curl time).
- **§2.3.4 cache key tuple** passes (include_thinking + include_subagents flags don't poison the default cache; since_id distinctness intact).

## Fix direction

`src/pollypm/transcript_ingest.py` — the ingestor should dedupe by `(source_path, source_offset)` before append, or the parser should dedupe envelopes by `id` before emit (defense-in-depth). The 4x multiplier suggests a flush-on-rotate / flush-on-startup that re-replays from a stale cursor; the `source_offset=8488` being identical across all 4 copies makes the per-flush replay theory likely.

## What's intentional vs not

The translation layer's source-replay fidelity is intentional — but the public REST envelope contract should guarantee unique `id` within a response. Even if the ingestor bug is fixed upstream, `parse_events_jsonl` should defensively dedupe so the Web UI never sees duplicate keys.

---

Found during §02 translation-layer testing pass (§2.3).

### LINKED ISSUE #2356: §6.7.1: send-to-pane p95=1521ms — POST /send blocks ~700-1500ms before tmux hits the pane
## Symptom

`POST /api/v1/chat/{session_name}/send?safety=force` consistently takes 700-1500ms before the message arrives in the target tmux pane. This blows the §6.7.1 budget (task description: p95 < 500ms; doc §6.7.1: p95 < 1s).

## Repro

Live `pm serve` at `http://127.0.0.1:8765`, SHA 67c177293, surface `advisor_smoketest` (Codex pane, `surface_type=advisor`, `transcript_source=capture`). 10 unique tokens injected via `safety=force` and polled out of `tmux capture-pane` at 20ms intervals.

## Observed (two consecutive runs of 10 reps each)

| Metric | Run 1 | Run 2 | Budget | Verdict |
|---|---|---|---|---|
| POST `/send` round-trip p95 | **1503ms** | 1182ms | n/a | – |
| §6.7.1 send-to-pane p95 | **1521ms** | 1192ms | 500ms (task) / 1000ms (doc) | **FAIL** |
| §6.7.1 send-to-pane p50 | 867ms | 932ms | – | over doc budget too |
| §6.7.1 send-to-pane min | **718ms** | **659ms** | – | floor is already ~700ms |
| §6.7.2 pane-to-UI p95 | 124ms | 124ms | 2000ms | **PASS** |
| §6.7   send-to-UI e2e p95 | 1627ms | 1296ms | – | dominated by send-to-pane |

Pane propagation after `POST /send` returns averages 10-20ms — the latency is **inside** the FastAPI handler, not in tmux pickup.

## Likely bottleneck

`send_chat_message` in `src/pollypm/web_api/routes/chat_send.py:1262` runs the same prelude even when `safety=force`:

1. `_resolve_surface(config, session_name)` (line 1282) → `enumerate_chat_surfaces(...)` — walks workers + advisors + architects + operator across the whole config every request. With 58 surfaces registered on this box (5 advisors + 28 worker tasks + 11 architects + operator + …), this is the most expensive call in the path.
2. `_tmux_session_for_send` + `_resolve_pane_target` → subprocess `tmux list-panes` round-trip.
3. `_resolve_session_events(surface, project_root)` is called even on `safety=force` (line 1322), so transcript-path resolution is paid regardless of whether the safety gate is going to run.
4. Only then does `tmux.send_keys` run (line 1435).

The 700ms floor is consistent with one `enumerate_chat_surfaces` traversal + one or two `tmux list-windows`/`list-panes` calls per send. Force mode should skip enumeration entirely — the surface registry lookup can be O(1).

## Proposed fix direction

1. Add a fast O(1) path for `_resolve_surface` keyed by `session_name`, falling back to the full enumeration only on miss.
2. When `safety_mode == 'force'`, skip `_resolve_session_events` and the unused transcript walk (no `answer_to` body).
3. Audit `TmuxClient` calls per send — `_resolve_pane_target` should be a single `display-message` query, not list-panes-then-filter.

## Trace / log paths

- Probe script: `/tmp/probe_6_7.py`
- Raw run log: `/tmp/probe_6_7_run1.log`
- Findings: `/tmp/sub-6-7.md`

## Per-rep rows (run 1)

```
rep#0: api=1503ms send->pane=1521ms pane->ui=106ms e2e=1627ms
rep#1: api= 856ms send->pane= 867ms pane->ui= 86ms e2e= 953ms
rep#2: api= 725ms send->pane= 746ms pane->ui=122ms e2e= 867ms
rep#3: api= 855ms send->pane= 874ms pane->ui= 91ms e2e= 964ms
rep#4: api= 699ms send->pane= 718ms pane->ui=107ms e2e= 825ms
rep#5: api= 818ms send->pane= 826ms pane->ui= 66ms e2e= 893ms
rep#6: api=1013ms send->pane=1033ms pane->ui=124ms e2e=1157ms
rep#7: api= 902ms send->pane= 921ms pane->ui=100ms e2e=1021ms
rep#8: api= 786ms send->pane= 802ms pane->ui=115ms e2e= 917ms
rep#9: api= 972ms send->pane= 991ms pane->ui=116ms e2e=1107ms
```

## Environment

- pm serve `1.0.0rc3.dev0` (`/api/v1/health`)
- macOS 25.5.0, Python 3.13
- 58 registered chat surfaces (operator + 11 architects + 14 advisors + 32 workers)
- tmux `pollypm-storage-closet`

## Scale

This is **single-box scale**, not M-scale. The doc's §6.7.1 p95 < 1s budget is supposed to hold "at M-scale" — we are already 1.5× over it at idle on a developer laptop.

## Ship-blocker

Yes for the v1 1-second-click rule. Every operator-typed message and every architect message into a pane pays this latency before anything is visible.

samhotchkiss:
## Agent Identity

- Agent: Codex
- Worktree: `/Users/sam/dev/pollypm/.codex/watch-worktrees/20260525-233544/.codex/worktrees/chat-transcripts`
- Branch: `codex/2341-2347-chat-transcripts`
- Commit: `a9d1343eeb01078330432b1788fcfa070051fb49`

## Summary

- Audits `safety=force` chat sends without logging message text.
- Restores raw subagent tool blocks, Codex response_item ingestion, opt-in thinking, and REST id de-duplication.
- Fixes stale Codex mid_tool false positives and speeds short send-to-pane by avoiding full surface/transcript walks and the 500 ms Enter delay.

## Issues

Closes #2341.
Closes #2342.
Closes #2343.
Closes #2344.
Closes #2345.
Closes #2346.
Closes #2347.
Closes #2356.

## Verification

- 237 passed: `uv run --extra test pytest tests/test_chat_send_endpoint.py tests/test_chat_transcripts.py tests/test_transcript_ingest.py tests/test_chat_messages_endpoint.py -q` before the #2356 extension.
- 18 passed after rebase: focused chat/transcript/tmux regression selection covering this PR.
- Ruff passed on changed Python files.

### DIFF
--- a/docs/test-plan/02-translation-layer.md
+++ b/docs/test-plan/02-translation-layer.md
@@ -265,6 +265,16 @@ curl -sS -X POST -H "Authorization: Bearer $TOKEN" -H "Content-Type: application
 
 **Expected:** 200, message lands even mid-stream. Confirms force bypass works.
 
+Then verify the audit trail:
+```bash
+tail -20 ~/.pollypm/audit/pollypm.jsonl | \
+  jq 'select(.event == "chat.send.force_bypass") | {event, status, subject, metadata}'
+```
+
+**Pass:** a `chat.send.force_bypass` event exists with `status: "warn"`,
+the target session in `subject`, a `metadata.bypassed_gates` array naming the
+skipped safety gates, and no injected message text in metadata.
+
 ### 2.4.4 Dead session
 
 ```bash
--- a/src/pollypm/audit/log.py
+++ b/src/pollypm/audit/log.py
@@ -212,6 +212,7 @@ EVENT_TIER4_ACTION = "audit.tier4_action"
 EVENT_TIER4_GLOBAL_ACTION = "audit.tier4_global_action"
 EVENT_TIER4_DEMOTED = "audit.tier4_demoted"
 EVENT_TIER4_BUDGET_EXHAUSTED = "audit.tier4_budget_exhausted"
+EVENT_CHAT_SEND_FORCE_BYPASS = "chat.send.force_bypass"
 # #1413 — emitted whenever the supervisor / on-demand path provisions a
 # project-scoped role session (e.g. reviewer-<project>) that did not
 # previously exist. The watchdog (#1414) reads this stream to surface
--- a/src/pollypm/tmux/client.py
+++ b/src/pollypm/tmux/client.py
@@ -577,6 +577,65 @@ class TmuxClient:
             )
         return windows
 
+    def get_window(
+        self, target: str, *, timeout: int | None = None,
+    ) -> TmuxWindow | None:
+        """Return one tmux window by target without listing the session."""
+        fmt = (
+            "#{session_name}\t#{window_index}\t#{window_name}\t#{window_active}\t#{pane_id}\t"
+            "#{pane_current_command}\t#{pane_current_path}\t#{pane_dead}\t#{pane_pid}"
+        )
+        result = self.run(
+            "display-message",
+            "-t",
+            self._exact_target(target),
+            "-p",
+            fmt,
+            check=False,
+            timeout=timeout,
+        )
+        if result.returncode != 0:
+            if result.returncode == 124:
+                raise subprocess.TimeoutExpired(
+                    cmd=["tmux", "display-message", "-t", target],
+                    timeout=timeout or self._DEFAULT_TIMEOUT,
+                )
+            return None
+        line = result.stdout.strip()
+        if not line:
+            return None
+        parts = line.split("\t", 8)
+        if len(parts) < 8:
+            return None
+        (
+            session,
+            index,
+            window_name,
+            active,
+            pane_id,
+            pane_current_command,
+            pane_current_path,
+            pane_dead,
+            *rest,
+        ) = parts
+        pane_pid: int | None = None
+        if rest:
+            try:
+                pane_pid = int(rest[0])
+            except ValueError:
+                pane_pid = None
+        return TmuxWindow(
+            session=session,
+            index=int(index),
+            name=window_name,
+            active=active == "1",
+            pane_id=pane_id,
+            pane_current_command=pane_current_command,
+            pane_current_path=pane_current_path,
+            pane_dead=pane_dead == "1",
+            pane_pid=pane_pid,
+        )
+
     def list_all_windows(self) -> list[TmuxWindow]:
         """List windows across ALL tmux sessions in a single subprocess call."""
         fmt = (
@@ -681,7 +740,14 @@ class TmuxClient:
         result = self.run("capture-pane", "-p", "-S", f"-{lines}", "-t", self._exact_target(target))
         return result.stdout
 
-    def send_keys(self, target: str, text: str, press_enter: bool = True) -> None:
+    def send_keys(
+        self,
+        target: str,
+        text: str,
+        press_enter: bool = True,
+        *,
+        enter_delay_seconds: float = 0.5,
+    ) -> None:
         """Send keys to a pane. Raises DeadPaneError if the target pane is dead."""
         import time
         resolved = self._exact_target(target)
@@ -735,7 +801,8 @@ class TmuxClient:
             self.run("send-keys", "-l", "-t", resolved, text)
         if press_enter:
             # Delay to let the terminal process pasted text before Enter.
-            time.sleep(0.5)
+            if enter_delay_seconds > 0:
+                time.sleep(enter_delay_seconds)
             self.run("send-keys", "-t", resolved, "Enter")
 
     def attach_session(self, name: str) -> int:
--- a/src/pollypm/transcript_ingest.py
+++ b/src/pollypm/transcript_ingest.py
@@ -89,6 +89,74 @@ def _extract_text(value: Any) -> str:
     return ""
 
 
+def _json_object_from_string(value: Any) -> dict[str, Any]:
+    if not isinstance(value, str) or not value.strip():
+        return {}
+    try:
+        parsed = json.loads(value)
+    except json.JSONDecodeError:
+        return {"value": value}
+    return parsed if isinstance(parsed, dict) else {"value": parsed}
+
+
+def _codex_reasoning_text(payload: dict[str, Any]) -> str:
+    summary = payload.get("summary")
+    if not isinstance(summary, list):
+        return ""
+    parts: list[str] = []
+    for item in summary:
+        if not isinstance(item, dict):
+            continue
+        text = item.get("text")
+        if isinstance(text, str) and text:
+            parts.append(text)
+    return "\n".join(parts).strip()
+
+
+def _codex_tool_call_payload(payload: dict[str, Any]) -> dict[str, Any]:
+    payload_type = str(payload.get("type") or "tool_call")
+    call_id = str(payload.get("call_id") or payload.get("id") or "")
+    tool_name = str(
+        payload.get("name")
+        or payload.get("tool_name")
+        or payload.get("tool")
+        or payload_type
+    )
+    raw_input = (
+        payload.get("arguments")
+        if "arguments" in payload
+        else payload.get("input")
+    )
+    tool_input = raw_input if isinstance(raw_input, dict) else _json_object_from_string(raw_input)
+    out = dict(payload)
+    if call_id:
+        out["id"] = call_id
+    out["name"] = tool_name
+    out["input"] = tool_input
+    return out
+
+
+def _codex_tool_result_payload(payload: dict[str, Any]) -> dict[str, Any]:
+    call_id = str(
+        payload.get("call_id")
+        or payload.get("id")
+        or payload.get("tool_use_id")
+        or ""
+    )
+    content = (
+        payload.get("output")
+        if "output" in payload
+        else payload.get("content", payload.get("text"))
+    )
+    out = dict(payload)
+    if call_id:
+        out["tool_use_id"] = call_id
+    if content is not None:
+        out["content"] = content
+        out["output"] = content
+    return out
+
+
 def _project_key_for_cwd(config, cwd: str | None) -> str:
     if not cwd:
         return config.project.name
@@ -483,7 +551,7 @@ def _normalize_codex_line(
             cursor.model_name = model_name
         return events
 
-    if entry_type != "event_msg":
+    if entry_type not in {"event_msg", "response_item"}:
         return events
 
     session_id = cursor.session_id or source_path.stem
@@ -491,6 +559,70 @@ def _normalize_codex_line(
     project_key = _project_key_for_cwd(config, cwd)
     model_name = cursor.model_name
     payload_type = payload.get("type")
+
+    if entry_type == "response_item":
+        common = {
+            "session_id": session_id,
+            "account_name": account_name,
+            "provider": account.provider.value,
+            "project_key": project_key,
+            "timestamp": timestamp,
+            "source_path": source_path,
+            "source_offset": source_offset,
+            "cwd": cwd,
+            "model_name": model_name,
+        }
+        if payload_type == "message":
+            text = _extract_text(
+                payload.get("content")
+                or payload.get("text")
+                or payload.get("message")
+                or payload
+            )
+            if text:
+                role = str(payload.get("role") or "assistant")
+                event_type = "user_turn" if role == "user" else "assistant_turn"
+                events.append(_event_base(
+                    event_type=event_type,
+                    payload={"text": text},
+                    **common,
+                ))
+        elif payload_type in {
+            "function_call",
+            "custom_tool_call",
+            "tool_search_call",
+        }:
+            events.append(_event_base(
+                event_type="tool_call",
+                payload=_codex_tool_call_payload(payload),
+                **common,
+            ))
+        elif payload_type in {
+            "function_call_output",
+            "custom_tool_call_output",
+            "tool_search_output",
+        }:
+            events.append(_event_base(
+                event_type="tool_result",
+                payload=_codex_tool_result_payload(payload),
+                **common,
+            ))
+        elif payload_type == "reasoning":
+            text = _codex_reasoning_text(payload)
+            if text:
+                events.append(_event_base(
+                    event_type="thinking",
+                    payload={
+                        "text": text,
+                        "signature": "",
+                        "raw": {
+                            "type": "reasoning",
+                            "summary": payload.get("summary"),
+                        },
+                    },
+                    **common,
+                ))
+        return events
 
     if payload_type == "token_count":
         info = payload.get("info") if isinstance(payload.get("info"), dict) else {}
@@ -536,7 +668,7 @@ def _normalize_codex_line(
                     payload={"text": text},
                 )
             )
-    elif payload_type == "assistant_message":
+    elif payload_type in {"assistant_message", "agent_message"}:
         text = _extract_text(payload.get("text") or payload.get("message") or payload)
         if text:
             events.append(
@@ -567,7 +699,7 @@ def _normalize_codex_line(
                 source_offset=source_offset,
                 cwd=cwd,
                 model_name=model_name,
-                payload=payload,
+                payload=_codex_tool_call_payload(payload),
             )
         )
     elif payload_type == "tool_result":
@@ -583,7 +715,7 @@ def _normalize_codex_line(
                 source_offset=source_offset,
                 cwd=cwd,
                 model_name=model_name,
-                payload=payload,
+                payload=_codex_tool_result_payload(payload),
             )
         )
     elif payload_type == "turn_end":
@@ -602,12 +734,31 @@ def _normalize_codex_line(
                 payload=payload,
             )
         )
-    # TODO(#2048): Codex rollouts do not currently surface a ``reasoning``
-    # ``event_msg`` payload type — only ``reasoning_output_tokens`` is
-    # exposed via the ``token_count`` info blob (see line ~413). Wire a
-    # ``thinking`` branch here if/when Codex starts emitting reasoning
-    # content blocks; until then leaving this absent so the parser
-    # contract stays Claude-only for the THINKING envelope.
+    elif payload_type == "reasoning":
+        text = _codex_reasoning_text(payload)
+        if text:
+            events.append(
+                _event_base(
+                    event_type="thinking",
+                    session_id=session_id,
+                    account_name=account_name,
+                    provider=account.provider.value,
+                    project_key=project_key,
+                    timestamp=timestamp,
+                    source_path=source_path,
+                    source_offset=source_offset,
+                    cwd=cwd,
+                    model_name=model_name,
+                    payload={
+                        "text": text,
+                        "signature": "",
+                        "raw": {
+                            "type": "reasoning",
+                            "summary": payload.get("summary"),
+                        },
+                    },
+                )
+            )
     elif payload_type == "error":
         events.append(
             _event_base(
--- a/src/pollypm/web_api/chat/registry.py
+++ b/src/pollypm/web_api/chat/registry.py
@@ -317,6 +317,7 @@ def find_chat_surface(
     *,
     work_service: Any | None = None,
     tmux_client: Any | None = None,
+    include_transcripts: bool = True,
 ) -> ChatSurface | None:
     """Resolve one chat surface without enumerating the whole workspace.
 
@@ -339,7 +340,7 @@ def find_chat_surface(
             tmux_session=storage_session_name(config.project.tmux_session),
             tmux_state_cache=tmux_state_cache,
             session_index_cache=session_index_cache,
-            include_transcripts=True,
+            include_transcripts=include_transcripts,
         )
 
     parsed = parse_task_window_name(session_name)
@@ -368,7 +369,7 @@ def find_chat_surface(
                 tmux_session=storage_session_name(config.project.tmux_session),
                 tmux_state_cache=tmux_state_cache,
                 session_index_cache=session_index_cache,
-                include_transcripts=True,
+                include_transcripts=include_transcripts,
             )
     return None
 
--- a/src/pollypm/web_api/chat/transcripts.py
+++ b/src/pollypm/web_api/chat/transcripts.py
@@ -210,8 +210,10 @@ def lookup_transcript_path(
     Matching is fingerprint-based, NOT "freshest under the project":
 
     1. ``cwd`` must match the surface's cwd (resolved). Required.
-    2. When ``account_name`` is given, entries with a different
-       ``account_name`` are filtered out.
+    2. When ``account_name`` is given, matching entries are preferred.
+       If the configured account points at a stale abandoned archive,
+       a fresher same-cwd/provider sibling may win so account binding
+       renames do not pin the surface to dead transcripts.
     3. When ``provider`` is given, entries with a different
        ``provider`` are filtered out.
     4. If multiple entries remain (e.g. a restarted session left two
@@ -225,16 +227,31 @@ def lookup_transcript_path(
         return None
     normalized = _normalize_cwd(cwd)
     best: _SessionIndexEntry | None = None
+    best_any_account: _SessionIndexEntry | None = None
     for entry in index:
         entry_cwd = entry.cwd_normalized or _normalize_cwd(entry.cwd)
         if entry_cwd != normalized:
             continue
-        if account_name and entry.account_name and entry.account_name != account_name:
-            continue
         if provider and entry.provider and entry.provider != provider:
+            continue
+        if best_any_account is None or entry.mtime > best_any_account.mtime:
+            best_any_account = entry
+        if account_name and entry.account_name and entry.account_name != account_name:
             continue
         if best is None or entry.mtime > best.mtime:
             best = entry
+    if best is not None:
+        if (
+            account_name
+            and best_any_account is not None
+            and best_any_account.path != best.path
+            and best_any_account.mtime > best.mtime
+            and (time.time() - best.mtime) > STALE_THRESHOLD_SECONDS
+        ):
+            return best_any_account.path
+        return best.path
+    if account_name and best_any_account is not None:
+        return best_any_account.path
     return best.path if best is not None else None
 
 
@@ -470,7 +487,8 @@ def parse_events_jsonl(
             "chat.transcripts: read failed for %s: %s",
             events_path, exc,
         )
-        return envelopes
+        return _dedupe_envelopes_by_id(envelopes)
+    envelopes = _dedupe_envelopes_by_id(envelopes)
     if cache_eligible and cached_mtime is not None:
         # Evict oldest insertion before storing the new entry. Python
         # dicts preserve insertion order so ``next(iter(...))`` gives
@@ -615,6 +633,7 @@ def parse_events_jsonl_tail(
                 actor_fallback=actor_fallback,
                 include_thinking=include_thinking,
             )[-limit:]
+        envelopes = _dedupe_envelopes_by_id(envelopes)
         if len(envelopes) >= limit:
             return envelopes[-limit:]
         if chunk_size >= file_size:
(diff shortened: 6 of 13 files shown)


samhotchkiss (commented):
APPROVE: All 8 closed issues verified PASS on worktree pm serve + tests green (239 passed, expected 237+).

- #2341 force-bypass audit: PASS — `chat.send.force_bypass` rows landing in `~/.pollypm/audit/<project>.jsonl` with metadata (message_id, window_target, characters_sent, method, press_enter, bypassed_gates); message text intentionally not logged.
- #2342 strict mid_tool false-positive: PASS — non-force send to idle Codex advisor returns 200 (was 409). `turn_end`/`session_state` now close stale orphan tool_use.
- #2343 subagent JSONL parser: PASS — real subagent JSONL parses to 142 envelopes (text=28, tool_use=57, tool_result=57); was 39 envelopes (32 empty) pre-fix; tool_name/tool_use_id/is_error all populated.
- #2344 Codex response_item ingest: PASS — today's rollout produces tool_call=695, tool_result=695, assistant_turn=185, user_turn=8 (pre-fix produced only token_usage); reasoning payloads with summary text emit thinking envelopes.
- #2345 capture-pane only on staleness: PASS — `source=auto` returns `transcript_source=jsonl` for fresh Claude archives (verified architect_pollypm); capture path stays as fallback.
- #2346 thinking surfaces: PASS — architects now return thinking envelopes (25/27/11/36 across pollypm/bikepath/smoketest/savethenovel) vs 0 across 80+ pre-fix.
- #2347 envelope id dedup: PASS — 0 duplicate ids across 549 sampled messages (architects + operator); defense-in-depth dedup in both parser and routing layers.
- #2356 send-to-pane latency: PASS — 5 forced sends p95 ≈ 76ms (was 1521ms; budget 500ms). `find_chat_surface` O(1) + skip transcript walk on force + 0s enter_delay for send_keys + new `tmux.get_window` short-circuits list-panes walks.

Tests: `tests/test_chat_send_endpoint.py tests/test_chat_transcripts.py tests/test_transcript_ingest.py tests/test_chat_messages_endpoint.py` — 239 passed in 791s.

Code review: audit emit is exception-safe, force-bypass logs no message text, `lookup_transcript_path` account fallback only triggers when configured pick exceeds STALE_THRESHOLD_SECONDS (preserves #2044 contract), dedup applied at parser + route. Approving.

PR #2359, opened 2026-05-26
fix(ui): cap chat history rendering

### LINKED ISSUE #2352: §3.5.1: renderHistory blocks main thread 372ms at M-scale (5000 messages)
## §3.5.1 finding (main-thread blocking)

Per docs/test-plan/03-web-ui-richness.md §3.5.1: any long task >200ms during normal interaction is RED. `renderHistory()` in `src/pollypm/web_api/ui/app.js` (lines 1593–1632) loops over every message in the API response and synchronously appends DOM nodes inside one tight loop. At §06 M-scale (5000 messages) this blocks the main thread for ~370ms in a single task.

## Measurements (Chromium headless, PerformanceObserver `longtask`)

Synthetic stress mirroring the exact loop body of `renderHistory` (innerHTML clear + per-message div/spans):

| Messages | Wall ms | Max long task | over 50ms | over 200ms |
|----------|--------:|--------------:|----------:|-----------:|
| 100      |   17.6  |    none       | 0 | 0 |
| 500      |   38.4  |    none       | 0 | 0 |
| 2000     |   90.2  |     93ms (yellow) | 1 | 0 |
| **5000** | **361.7** | **372ms (RED)** | 1 | 1 |

Hardware: Apple Silicon M-class (the user's actual operator machine). On a phone or low-power-mode laptop the cliff arrives sooner.

## Why it's a ship-blocker

§3.5.1 says: \"A click can technically finish under 1s and still feel cheap/fragile if it blocks the main thread repeatedly.\" 5000 messages is the §06 M-scale release-readiness fixture and the message-poll cadence is every 5s for the selected surface — so this 372ms hit recurs on every poll once a transcript is sizeable, not just on click.

Additional concern: `msgs.reverse()` runs on the unsliced array; the loop has no virtualization and no DocumentFragment batching.

## Suggested fixes (any one or combo)

1. Cap render to the last N (e.g. 200) messages and add a \"load earlier\" button — matches chat-UI convention and is by far the simplest.
2. Use a `DocumentFragment` to batch appends, cutting layout thrash.
3. Switch to a virtualized list (only render rows in viewport).
4. Skip re-render when `data.messages` is unchanged (compare hash / last id) — addresses the polling case directly.

## Repro

Stress script: `/tmp/perf_3_5_1_stress.py` (synthetic — drives the same DOM construction the renderer does, no auth required). Live probe also attempted against `pm serve` at `100.67.2.108:8765` but the stale binary (PID 41222, 2026-05-25 11:04, missing #2331/#2332) returned 500 on `/ui/` and 404 on `/ui/app.js`/`/ui/styles.css`, so empirical end-to-end measurement was done via a local proxy serving the on-disk UI files against the live API. All histories on the stale instance are empty (0 messages), which is why the live polling-tick phase showed 0 long tasks despite 65 surfaces — the renderHistory hot path simply did not execute. Synthetic stress is the correct upper-bound estimate.

## Out-of-band note (separate issue worth filing)

Live `/ui/` returns `RuntimeError` and `/ui/app.js` is 404 on the running daemon. That's its own bug — the UI is fully broken on the deployed instance — but is likely fixed by #2331/#2332 + a restart and is out of scope for this §3.5.1 ticket.

samhotchkiss:
## Agent Identity

- Agent: Codex
- Worktree: `/Users/sam/dev/pollypm/.codex/watch-worktrees/20260525-233544/.codex/worktrees/issue-2352`
- Branch: `codex/2352-render-history-window`
- Commit: `a7f655086d1019e36d08ac23900e2668b4558635`

## Summary

- Caps chat history DOM rendering to the latest visible window.
- Builds message nodes in a `DocumentFragment` and skips unchanged history windows.
- Adds a Playwright regression with a 5000-message payload.

Closes #2352.

## Verification

- `npm test -- tests/render_history_perf.spec.ts` -> 2 passed after rebase.

### DIFF
--- a/src/pollypm/web_api/ui/app.js
+++ b/src/pollypm/web_api/ui/app.js
@@ -22,6 +22,7 @@
   const PUSH_REFRESH_DEBOUNCE_MS = 150;
   const FETCH_RETRY_MS = 3000;
   const MAX_MESSAGES = 50;
+  const HISTORY_RENDER_LIMIT = MAX_MESSAGES;
   const SESSION_ISSUED_COOKIE = "pollypm-session-issued-at";
   const SESSION_EXPIRY_WARNING_MS = 6 * 24 * 60 * 60 * 1000;
   const TASK_RAIL_LIMIT = 200;
@@ -70,6 +71,7 @@
     dashboardRefreshQueued: false,
     historyInFlight: {},
     historyRefreshQueued: {},
+    historyRenderSignatures: {},
     fetchStates: {},
     sseFailures: 0,
     sseRetryTimer: null,
@@ -129,6 +131,11 @@
       }
     }
     return node;
+  }
+
+  function clearMessageList(list) {
+    list.innerHTML = "";
+    delete list.dataset.historySession;
   }
 
   function setStatus(level, label) {
@@ -1022,7 +1029,7 @@
     $("send-input").disabled = true;
     $("send-button").disabled = true;
     const list = $("message-list");
-    list.innerHTML = "";
+    clearMessageList(list);
     list.appendChild(el("div", {
       class: "message-empty",
       text: "No surface selected.",
@@ -1093,7 +1100,7 @@
       pollDashboard();
     });
     const list = $("message-list");
-    list.innerHTML = "";
+    clearMessageList(list);
     list.appendChild(el("div", { class: "empty-state" }, [
       el("div", { class: "empty-title", text: "No surface selected" }),
       el("div", {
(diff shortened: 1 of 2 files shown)


samhotchkiss (commented):
APPROVE: Caps render to MAX_MESSAGES=50, batches via DocumentFragment, and short-circuits unchanged signatures on the 5s poll. Verified 5000-msg payload: 372ms RED on main -> 31ms wall, 0 long tasks >50ms in headless Chromium. Playwright spec passes in chromium + mobile-chrome (2/2, 2.5s). clearMessageList() correctly resets the signature cache on every list-wipe entry point; replacing in-place msgs.reverse() with a reverse-iterating fragment loop also fixes a small caller-mutation side effect. Live /ui/ on uv run pm serve serves the patched app.js. Merge-base = 0 commits behind main. Merging via squash.

PR #2360, opened 2026-05-26
fix(audit): recover after truncated log tails

### LINKED ISSUE #2353: §5.8.1: truncated audit-log tail fuses with next append, silently losing one event
## Summary

`_append_line` in `src/pollypm/audit/log.py` opens with `"a"` mode and writes `line + "\n"`. If the previous record was truncated (process crash, partial write, disk-full mid-write, manual truncation per test plan §5.8.1), the new write **concatenates onto the broken line** instead of starting a fresh line, producing one corrupt super-long line.

Reproducer (against live `pm serve`):

```bash
F=~/.pollypm/audit/<proj>.jsonl
# truncate mid-line:
SZ=$(stat -f%z $F); head -c $((SZ - 50)) $F > /tmp/t && mv /tmp/t $F
# next append:
pm task create --project <proj> "x" --description "x"
# tail of $F is now: <truncated record>{<new record>}\n  -- one fused corrupt line
```

## Impact

- The reader (`_iter_log_lines`) skips the malformed line — so no crash. ✓
- But the **next valid event is silently lost** because it became part of the broken line.
- Per §5.8.1 "fail" criteria: "Silent state divergence". The newly appended event vanishes from `pm audit grep`, the watchdog dedupe window, briefings, SSE — anything reading the audit log.
- The detection edge (`task.created`) is gone; downstream cascades that key off it (stuck_draft, dedupe windows) misfire.

## Recovery

§5.8.1 pass criteria says "subsequent valid audit append works" — today it does NOT for the first append after a truncated tail.

## Suggested fix

In `_append_line`, before writing, check the last byte of the file and prepend a `\n` if missing:

```python
with open(path, "a+b") as fh:
    fh.seek(0, 2)
    if fh.tell() > 0:
        fh.seek(-1, 2)
        if fh.read(1) != b"\n":
            fh.write(b"\n")
    fh.write(line.encode("utf-8") + b"\n")
```

This is the same idiom used by syslog / logrotate: heal the boundary on the next write so a torn tail can only ever cost one event, not two.

Audit reader does not need to change; the JSONDecodeError skip already handles the (now single-record) corruption.

## Test plan reference

`docs/test-plan/05-resilience-recovery.md` §5.8.1 (truncated audit log). Reproduced 2026-05-26.

## Files
- `src/pollypm/audit/log.py` lines 679-712 (`_append_line`)

samhotchkiss:
## Agent Identity

- Agent: Codex
- Worktree: `/Users/sam/dev/pollypm/.codex/watch-worktrees/20260525-233544/.codex/worktrees/issue-2353`
- Branch: `codex/2353-audit-tail`
- Commit: `027381a23d25079ef94a34bbf7a48f5300da6705`

## Summary

- Heals audit JSONL append boundaries when an existing log tail lacks a newline.
- Keeps the next valid audit event recoverable after a crash/truncated write.
- Adds central-tail and per-project regression coverage.

Closes #2353.

## Verification

- `uv run --extra test pytest tests/test_audit_log.py -q` -> 26 passed before rebase.
- `uv run --extra test pytest tests/test_audit_log.py::test_emit_recovers_after_truncated_central_tail tests/test_audit_log.py::test_emit_recovers_after_truncated_per_project_tail tests/test_audit_log.py::test_read_events_skips_truncated_tail_lines -q` -> 3 passed after rebase.
- `uv run --extra dev ruff check src/pollypm/audit/log.py` -> passed.

### DIFF
--- a/src/pollypm/audit/log.py
+++ b/src/pollypm/audit/log.py
@@ -686,7 +686,7 @@ def _append_line(path: Path, line: str) -> None:
     append fires unconditionally so a housekeeping bug can never
     cause audit data loss.
 
-    Uses POSIX append-mode (``"a"``) so concurrent writers from
+    Uses POSIX append-mode (``"a+b"``) so concurrent writers from
     multiple processes interleave at the line level — provided the
     line is under PIPE_BUF (4096 bytes), which our records always
     are. We open + write + close on every call rather than holding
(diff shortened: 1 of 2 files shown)


samhotchkiss (commented):
APPROVE: §5.8.1 silent-data-loss fix is minimal and correct. Verified — 26/26 tests pass, 3 PR-new recovery tests pass, and a user-facing repro (seed 3 events → truncate 5 bytes mid-tail → append 4th) confirmed the new event lands on its own parseable line with the torn record isolated as a single skippable corruption. Pre-fix the 4th event would have fused to the broken tail and vanished from pm audit grep/watchdog/briefings/SSE. `a+b` binary mode required for negative SEEK_END seek; single-syscall write preserves PIPE_BUF multi-writer atomicity. Merge-base clean (0 commits ahead on main).

PR #2378, opened 2026-05-26
fix: serialize transcript parse cache fills

### LINKED ISSUE #2371: §2.7: chat-messages limit=500 serializes under 10 concurrent readers (10x slowdown vs warm serial)
## Symptom

Per §2.7 in `docs/test-plan/02-translation-layer.md`: "10 concurrent readers of the same long transcript do not create 5xxs or p95 spikes above §06 M-scale budgets."

Observed: 10 concurrent `GET /api/v1/chat/architect_coffeeboardnm/messages?limit=500&direction=desc&source=jsonl` against a 2.87MB `events.jsonl` archive serialize end-to-end:

```
serial warm baseline limit=500: 80-150ms p50
parallel-10 limit=500:
  n=10 min=1393.5ms p50=1446.8ms p95=1502.5ms max=1502.5ms
```

Every concurrent reader pays the cost of every other reader — they queue, not parallelize. The work is happening inside Python with the GIL held (likely in `parse_events_jsonl` or `_dedupe_envelopes_by_id`). At larger archive sizes (the §2.7 fixture target is 100MB) this 10x amplification will exceed any reasonable budget.

## Reproduction

```bash
export BASE=http://127.0.0.1:8765
SESSION=architect_coffeeboardnm  # 2.87MB events.jsonl

# Warm cache
curl -sS -H "Authorization: Bearer \$TOKEN" \
  "\$BASE/api/v1/chat/\$SESSION/messages?limit=500&direction=desc&source=jsonl" > /dev/null

# Parallel 10 — all readers queue
for i in \$(seq 1 10); do
  ( curl -sS -o /dev/null -w '%{time_total}\n' -H "Authorization: Bearer \$TOKEN" \
      "\$BASE/api/v1/chat/\$SESSION/messages?limit=500&direction=desc&source=jsonl" ) &
done
wait
```

## Likely root cause

`parse_events_jsonl` (in `src/pollypm/web_api/chat/transcripts.py`) re-parses the full file when `limit >= len(cached_envelopes)` or when the cache key doesn't hit. The mtime check in `_PARSE_CACHE` is a single shared dict guarded only by Python's GIL — no per-key lock that would allow N readers to share the result of one parse. Each reader independently re-parses.

Combined with the fact that any `include_thinking=true` or `include_subagents=true` request invalidates the default-key entry (because the cache key tuple changes), real-world dashboard polling that mixes flags will hit cold paths repeatedly.

## Side findings during the same session

- `limit=50 desc source=jsonl` cold-serial after cache-thrash: 76-82ms (passes #2084 <50ms budget loosely, not strictly).
- `limit=500 desc` cold-after-flag-flip: 1.13s (above 1s click budget but below 5s ceiling).
- `pm serve` RSS held flat at ~320MB across all probes — no memory growth, so this is CPU/serialization, not a leak.

## Impact

Not a v1.0.0 ship blocker on a 3MB archive (max 1.5s under contention). On a 100MB archive (§2.7 spec target) this would scale to ~50s under 10 concurrent readers — well outside any acceptable budget. The dashboard will feel sluggish whenever multiple panes poll the same surface, and the regression compounds with #2069/#2084 cache invalidation on flag changes.

## Suggested fix direction

- Per-cache-key `threading.Lock` so concurrent readers of the same path share one parse.
- Or: move full parse off the request thread (background refresher + serve from cache).
- Or: switch large-limit reads to a streaming/iterative parser that can yield the GIL.

## Test plan

- Add a perf regression test: warm cache, fire N=10 parallel readers at `limit=500&source=jsonl`, assert p95 < 2x serial baseline (not 10x).
- Verify behavior holds after `include_thinking=true` / `include_subagents=true` cache invalidation.

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

## Summary
- Adds per-cache-key in-flight serialization for transcript JSONL parse fills so concurrent readers share one active parse result.
- Keeps persistent cache semantics unchanged for non-strict readers, while allowing strict readers to share only same-mtime concurrent work.
- Preserves per-option isolation across mtime, actor fallback, strictness, and `include_thinking` variants.

## Issue
- Fixes #2371

## Tests
- `uv run --extra test pytest tests/test_chat_transcripts.py -q` (`85 passed`)
- `uv run --extra dev ruff check src/pollypm/web_api/chat/transcripts.py tests/test_chat_transcripts.py` (`passed`)
- `python -m py_compile src/pollypm/web_api/chat/transcripts.py tests/test_chat_transcripts.py` (`passed`)
- `git diff --check origin/main...HEAD` (`passed`)

### DIFF
--- a/src/pollypm/web_api/chat/transcripts.py
+++ b/src/pollypm/web_api/chat/transcripts.py
@@ -52,6 +52,7 @@ from __future__ import annotations
 import hashlib
 import json
 import logging
+import threading
 import time
 from dataclasses import dataclass
 from pathlib import Path
@@ -358,6 +359,22 @@ _PARSE_CACHE: dict[
 _PARSE_CACHE_MAX = 100
 
 
+@dataclass(slots=True)
+class _ParseInflight:
+    event: threading.Event
+    result: list[MessageEnvelope] | None = None
+    exception: BaseException | None = None
+
+
+# Protects the persistent parse cache and the short-lived in-flight
+# registry. The heavyweight JSONL read/parse happens outside this lock;
+# it only serializes dictionary bookkeeping for one cache-fill decision.
+_PARSE_CACHE_LOCK = threading.Lock()
+_PARSE_INFLIGHT: dict[
+    tuple[Path, bool, float, str, bool], _ParseInflight,
+] = {}
+
+
 _SESSION_INDEX_CACHE: dict[
     Path, tuple[tuple[tuple[str, float, int], ...], list[_SessionIndexEntry]]
 ] = {}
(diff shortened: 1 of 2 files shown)


samhotchkiss (commented):
APPROVE: Per-cache-key in-flight serialization correctly fixes #2371.

Verification (worktree pm serve on port 18878, architect_coffeeboardnm 5.0MB archive):
- Serial warm baseline: p50 108ms (was 80-150ms)
- Parallel-10 warm: p50 351ms, p95 358ms (was p50 1447ms / p95 1503ms — 4.2x under the 500ms criterion)
- Parallel-10 cold (forced mtime bump): p50 574ms, p95 575ms — single ~560ms parse shared across 10 readers on a 5MB archive (1.74x bigger than the 2.87MB issue baseline)

Tests: tests/test_chat_transcripts.py — 84 pass + 1 pre-existing flake (test_parse_tail_perf_beats_full_parse_on_10k_archive measured 49.7x vs >=50x threshold; passes 3/3 in isolation, unrelated to this PR's lock bookkeeping). New test test_parse_cache_miss_shares_concurrent_strict_parse correctly asserts 8 concurrent strict callers -> 1 parse.

Code review: fill_key tuple (path, include_thinking, mtime, actor_fallback, strict) keeps option variants isolated. Heavy parse runs outside _PARSE_CACHE_LOCK; only dict bookkeeping is serialized. Defensive list copies on both store and return. Exception path correctly propagates via _finish_parse_fill(exception=exc). _PARSE_INFLIGHT cleared on completion.

PR #2412, opened 2026-05-29
Fail closed on plugin validation and persist plugin events

### LINKED ISSUE #2405: Plugin host boundary bugs: validation gate fails open (bad plugins load); public api.emit_event() is a silent no-op
Two plugin-host boundary bugs: the validation gate fails OPEN (bad plugins install instead of being disabled), and the documented public extension point `api.emit_event()` is a silent no-op against every real store.

# Part A — validation gate fails open (P1)

## Plugin validation gate fails open — a plugin that raises during validation summary (checks is list[str], code does c.message) loads anyway via except fall-through

**Severity:** P1  |  **Axis:** reliable  |  **Surface:** plugin host / validation boundary

**Target:** src/pollypm/plugin_host.py:666-686

**Observed:** In ExtensionHost._load_plugins._install, on validation failure (not result.passed, line 665) line 666 builds the failure summary with `", ".join(c.message for c in result.checks if not c.passed)`. But ValidationResult.checks is declared `list[str]` (plugin_validate.py:26) and populated only with plain strings. A str has no `.passed`/`.message`, so line 666 raises AttributeError. That AttributeError is caught by the broad `except Exception` at line 674, which records a generic 'validation error' but has NO return and NO _mark_disabled — so control falls through to lines 685-686 (`loaded[name] = plugin`), installing the failing plugin instead of disabling it. The validation gate is a no-op precisely for the case it exists to catch; the bad factory then crashes later at first use.

**Expected:** A plugin that fails validation should be _mark_disabled'd and `return` so it is never added to `loaded`. The failure summary should be built from result.errors (the str list of failures), not a non-existent `c.passed`/`c.message` attribute on result.checks. The bare-except branch should also _mark_disabled + return so a validation crash fails closed, not open.

**Evidence:** failures = ", ".join(c.message for c in result.checks if not c.passed)  # checks are str (plugin_validate.py:26) -> AttributeError -> caught at line 674 -> no return -> loaded[name]=plugin at 685

**Reproduction:** In src/pollypm/plugin_host.py the dead/broken branch is lines 663-686. Reproduced at runtime: build a PollyPMPlugin with a provider factory returning an object missing build_launch_command, call validate_plugin(plugin) -> result.passed is False, result.checks is list[str]. Evaluating line 666 `", ".join(c.message for c in result.checks if not c.passed)` raises AttributeError: 'str' object has no attribute 'passed'. Simulating the full _install flow: the failing plugin lands in `loaded` (installed) and NOT in `disabled`, and errors contains the generic "validation error: 'str' object has no attribute 'passed'". Fix (two lines): replace line 666 with `failures = ", ".join(result.errors)`; and in the except at 674, after _record_error add `_mark_disabled(name, source, "load_error", f"validation crashed: {exc}")` and `return` so both clean and crashing validation failures fail closed. Note the existing _mark_disabled+return at lines 672-673 is currently unreachable because line 666 always raises first.

**Proposed fix:** Replace line 666 with `failures = ", ".join(result.errors)`. In the except at 674, after _record_error add `_mark_disabled(name, source, "load_error", f"validation crashed: {exc}"); return` so both clean and crashing validation failures fail closed.

**Verifier note:** Independently confirmed in source AND reproduced at runtime. Data path: (1) ValidationResult.checks is list[str] (plugin_validate.py:26) and is populated EXCLUSIVELY with plain f-strings — verified every checks.append/extend in plugin_validate.py and confirmed at runtime that checks[0] is type 'str'. (2) result.passed = len(result.errors)==0 (plugin_validate.py:392), so the `not result.passed` branch at plugin_host.py:665 fires exactly when validation has errors. (3) plugin_host.py:666 does `", ".join(c.message for c in result.checks if not c.passed)` where c is a str; `str.passed` does not ex
# Part B — public api.emit_event() silent no-op (P2)

## Public plugin API api.emit_event() is a silent no-op against every real store (record_event kwarg mismatch: kind=/payload= vs positional scope/sender/subject)

**Severity:** P2  |  **Axis:** functional  |  **Surface:** plugin_api.v1 PluginAPI / store contract boundary

**Target:** src/pollypm/plugin_api/v1.py:751-770

**Observed:** PluginAPI.emit_event (verified v1.py:751-770) — a documented public extension point — calls `record(kind=f"plugin.{self._plugin_name}.{name}", payload=payload or {})` with kind=/payload= keywords. No StateStore accepts those keywords: store/protocol.py:50 and pg_store.py:324 declare `record_event(scope, sender, subject, payload=None)` (verified protocol.py:50-56), and storage/state.py:1332 declares `record_event(session_name, event_type, message)`. Calling with kind= raises TypeError, swallowed by the `try/except Exception: pass` (768-770), so every plugin using the published api.emit_event() records nothing against a production store with no error surfaced. The host's own _emit_plugin_lifecycle_event calls record positionally, so the host path works while the plugin-facing path silently fails.

**Expected:** api.emit_event should call record_event with the contract's positional shape (scope, sender, subject[, payload]) so plugin events actually persist.

**Evidence:** record(kind=f"plugin.{self._plugin_name}.{name}", payload=payload or {})  # no store record_event accepts kind=/payload= as keywords; protocol.py:50 signature is (scope, sender, subject, payload=None)

**Reproduction:** In a plugin's initialize(api), call api.emit_event("hello", {"x": 1}) against a real StateStore- or pg_store-backed host. The internal call record(kind="plugin.<name>.hello", payload={"x":1}) raises TypeError because record_event's signature is (scope, sender, subject, payload=None) with no `kind` param; the bare except at v1.py:768-770 swallows it, so no messages/events row is written and no error surfaces. Fix: change v1.py:764-767 to record("plugin", self._plugin_name, name, payload or {}) and optionally narrow the except to logger.debug(..., exc_info=True) so future signature drift is visible.

**Proposed fix:** Change the emit to match the store contract, e.g. `record("plugin", self._plugin_name, name, payload or {})`. Optionally narrow the except to log at debug so future signature drift is visible.

**Verifier note:** Independently confirmed in source. PluginAPI.emit_event (src/pollypm/plugin_api/v1.py:761-770) calls record(kind=f"plugin.{self._plugin_name}.{name}", payload=payload or {}) using keyword args kind= and payload=. No record_event implementation in the codebase accepts a `kind` keyword: store/backends/pg_store.py:324 and store/protocol.py:50 both declare record_event(self, scope, sender, subject, payload=None); storage/state.py:1332 declares record_event(self, session_name, event_type, message); heartbeats/api.py:258 declares record_event(self, session_name, event_type, message). I grepped every

---

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

## Summary
- Make plugin validation failures and validation exceptions mark plugins disabled and return before installation.
- Build validation summaries from the current `ValidationResult` string fields instead of old check-object assumptions.
- Make public `PluginAPI.emit_event()` call the store `record_event(scope, sender, subject, payload)` contract positionally and log signature drift.

Fixes #2405

## Verification
- `python -m py_compile src/pollypm/plugin_host.py src/pollypm/plugin_api/v1.py tests/test_plugins.py tests/test_rail_api.py tests/test_downtime_flow.py`
- `git diff --check`
- `uv run --extra dev ruff check src/pollypm/plugin_host.py src/pollypm/plugin_api/v1.py tests/test_plugins.py tests/test_rail_api.py tests/test_downtime_flow.py`
- `uv run --extra test pytest tests/test_plugins.py::test_extension_host_disables_plugin_on_validation_failure tests/test_plugins.py::test_extension_host_disables_plugin_on_validation_exception tests/test_plugins.py::test_project_plugin_shadows_user_plugin tests/test_rail_api.py::test_plugin_api_emit_event_calls_store_contract_positionally tests/test_rail_api.py::test_plugin_api_emit_event_logs_signature_drift tests/test_downtime_flow.py::TestInitializeHookValidates::test_bundled_flow_emits_ok` (6 passed)

Note: a broader run of `tests/test_plugins.py tests/test_rail_api.py tests/test_downtime_flow.py` still has unrelated existing HOME/global-config expectation failures around user-global plugin/content paths.

### DIFF
--- a/src/pollypm/plugin_api/v1.py
+++ b/src/pollypm/plugin_api/v1.py
@@ -22,9 +22,12 @@ Example:
 
 from __future__ import annotations
 
+import logging
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Callable, Literal, Union
+
+logger = logging.getLogger(__name__)
 
 ProviderFactory = Callable[[], object]
 RuntimeFactory = Callable[[], object]
@@ -762,12 +765,28 @@ class PluginAPI:
             record = getattr(store, "record_event", None)
             if callable(record):
                 record(
-                    kind=f"plugin.{self._plugin_name}.{name}",
-                    payload=payload or {},
+                    "plugin",
+                    self._plugin_name,
+                    f"plugin.{self._plugin_name}.{name}",
+                    payload or {},
                 )
+        except TypeError as exc:
+            logger.warning(
+                "Plugin '%s' emit_event could not call state_store.record_event "
+                "for event '%s': %s",
+                self._plugin_name,
+                name,
+                exc,
+            )
         except Exception:  # noqa: BLE001
-            # Events are best-effort observability — swallow.
-            pass
+            # Events are best-effort observability; keep runtime store
+            # failures out of plugin initialization noise.
+            logger.debug(
+                "Plugin '%s' emit_event failed for event '%s'",
+                self._plugin_name,
+                name,
+                exc_info=True,
+            )
 
 
 @dataclass(slots=True)
--- a/src/pollypm/plugin_host.py
+++ b/src/pollypm/plugin_host.py
@@ -56,6 +56,17 @@ PLUGIN_API_VERSION = "1"
 PLUGIN_NAME_PATTERN = re.compile(r"^[a-z][a-z0-9_-]*$")
 
 
+def _validation_summary(result: object) -> str:
+    """Return a compact validation detail without depending on old result shapes."""
+    errors = [str(error) for error in (getattr(result, "errors", None) or [])]
+    if errors:
+        return "; ".join(errors)
+    checks = [str(check) for check in (getattr(result, "checks", None) or [])]
+    if checks:
+        return "; ".join(checks)
+    return "validation failed"
+
+
 def _validate_plugin_name(plugin_name: str) -> str:
     """Reject ambiguous or path-like plugin names at the host boundary."""
     if not PLUGIN_NAME_PATTERN.fullmatch(plugin_name):
@@ -662,8 +673,8 @@ class ExtensionHost:
             # Validate the plugin implements its declared interfaces
             try:
                 result = validate_plugin(plugin)
-                if not result.passed:
-                    failures = ", ".join(c.message for c in result.checks if not c.passed)
+                if not getattr(result, "passed", False):
+                    failures = _validation_summary(result)
                     self._record_error(
                         f"Plugin {name} failed validation: {failures}",
                         plugin=name,
@@ -677,6 +688,8 @@ class ExtensionHost:
                     plugin=name,
                     stage="validate",
                 )
+                _mark_disabled(name, source, "load_error", f"validation error: {exc}")
+                return
             if name in loaded:
                 logger.warning(
                     "Plugin '%s' from source '%s' overrides earlier registration from source '%s'",
--- a/tests/test_downtime_flow.py
+++ b/tests/test_downtime_flow.py
@@ -17,8 +17,6 @@ Covers:
 from __future__ import annotations
 
 from datetime import datetime, timezone
-from pathlib import Path
-
 import pytest
 
 from pollypm.plugins_builtin.downtime.flow_validator import (
--- a/tests/test_downtime_flow.py
+++ b/tests/test_downtime_flow.py
@@ -306,11 +306,13 @@ class TestInitializeHookValidates:
         from pollypm.jobs import JobHandlerRegistry
         from pollypm.plugins_builtin.downtime import plugin as plugin_module
 
-        events: list[tuple[str, dict]] = []
+        events: list[tuple[str, str, str, dict]] = []
 
         class _StubStore:
-            def record_event(self, *, kind: str, payload: dict) -> None:
-                events.append((kind, payload))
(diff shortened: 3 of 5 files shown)


samhotchkiss:
**Claude review — APPROVE (merging).** Verified both fixes:

**Bug A (fail-closed):** plugin_host.py `_install` now `_mark_disabled`'s + `return`s on validation failure AND on the bare-except validation-error path; summary built via `_validation_summary` reading `result.errors` (not the broken `c.message`/`c.passed`). Failing plugins are no longer added to `loaded`. Confirmed in diff + source.

**Bug B (emit_event persists):** `emit_event` now calls `record_event` positionally `("plugin", plugin_name, subject, payload)` — exact match to the canonical contract in store/protocol.py:50 and pg_store.py:324. TypeError handler logs signature drift gracefully.

**Tests (foreground):** all 5 new/updated regression tests pass: `test_extension_host_disables_plugin_on_validation_failure`, `test_extension_host_disables_plugin_on_validation_exception`, `test_plugin_api_emit_event_calls_store_contract_positionally`, `test_plugin_api_emit_event_logs_signature_drift`, updated `test_bundled_flow_emits_ok`. The 4 other failures in the touched suite are PRE-EXISTING (confirmed identical failures on origin/main) and unrelated (provider-resolution/content-paths). Behind main by 2, no conflicts.

Issue #1573, opened 2026-05-17
Inbox UI becomes archive view (default = awaits-user, all = history)


samhotchkiss:
Part of #1564. Depends on #1566.

## What

Repurpose the existing inbox UI. After this PR:

- **Default view** = ``awaits_user`` items only (the same curated set the dashboard "Waiting on you" section shows). Short list.
- **"All" toggle / lens** = full historical archive, no curation. Filterable by ``kind``, by project, by date.

The current default behavior (showing everything that exists with weak triage) goes away. The inbox stops being a triage surface; the dashboard is.

## Why

The user explicitly said: "The inbox has become cluttered and unuseful." After #1571 lands, the badge will already show a small number — but if they click into the inbox and see 100+ rows, the cognitive load problem returns. The inbox UI itself needs to default to the same curation.

## Scope

- Find the existing inbox app/screen in ``cockpit_ui.py`` / ``cockpit_inbox*.py``
- Default view: ``awaits_user`` items, grouped by project
- Add a lens toggle (key binding) for "all" / "completion FYIs only" / "activity events only" / "self-bug-reports only" so the historical data stays discoverable
- Update the docstring on ``_count_inbox_tasks_for_label`` to reflect the new role
- Tests for the lens-switching behavior

## Acceptance

- Inbox opens to the curated awaits-user list by default
- A clear affordance exists to view "everything" (key binding visible in the footer)
- Switching lenses doesn't lose state (focused row, etc.) where reasonable
- A regression test pins: default count == ``awaits_user`` count == rail badge count

## Architecture

- All filtering goes through the predicate module from #1566; no inline triage logic in the UI
- Don't add new direct storage access; use existing inbox view helpers

Refs #1564, #1566, #1571

PR #2414, opened 2026-05-29
Fix cockpit review and inbox interaction races

### LINKED ISSUE #2407: Cockpit interaction defects: review Undo morph clobbered by re-render; approve/reject unreachable under search filter; inbox stale-row guard missing
A cluster of cockpit/TUI interaction-state defects on the Tasks and Inbox surfaces.

# Part A — pending-review Undo morph clobbered by async re-render (breaks Approve/Reject mutual exclusion) (P2)

## Pending-review Undo morph is silently clobbered by async detail re-render / 8s background refresh, breaking the Approve/Reject mutual-exclusion invariant

**Severity:** P2  |  **Axis:** functional  |  **Surface:** Cockpit Tasks surface (PollyTasksApp) — review approve/reject undo window

**Target:** src/pollypm/cockpit_tasks.py:2181-2183

**Observed:** After Approve/Reject, _morph_review_action_buttons sets the clicked button to 'Undo (5s)' and DISABLES the opposite button (explicit mutual-exclusion invariant), with the commit deferred 5s. Within that window the task is still work_status=='review'. Both the 8s background refresh (_refresh_list->_render_table->_show_detail) and the async detail-load worker _apply_detail_result call _render_selected_task, which at lines 2182-2183 unconditionally runs `self.approve_button.disabled = not in_review` / `self.reject_button.disabled = not in_review` (verified). With in_review still True this re-ENABLES the deliberately-disabled button while leaving the 'Undo (5s)' label, so both review actions become simultaneously actionable. No guard on _pending_review_action; no test covers the interaction.

**Expected:** While _pending_review_action is active, _render_selected_task should preserve the morphed Undo state (label + the opposite button's disabled state) rather than recomputing disabled purely from in_review.

**Evidence:** in_review = task.work_status.value == "review"
self.approve_button.disabled = not in_review
self.reject_button.disabled = not in_review  # no _pending_review_action guard

**Reproduction:** Cockpit Tasks surface, with a task in work_status=="review" selected. Click Approve: approve_button morphs to "Undo (5s)" (+`-undo` class) and reject_button is disabled (mutual exclusion). Within the 5s undo window, let the 8s background refresh tick OR trigger any detail re-render (e.g. another row highlight that resolves back, or the detail worker re-applying). `_render_selected_task` runs lines 2181-2183: since the task is still "review", `reject_button.disabled = not in_review = False` re-ENABLES Reject while Approve still reads "Undo (5s)". Now click Reject — because reject_button lacks the `-undo` class (_on_press_reject line 3081), it does NOT undo; it opens the reject modal. Submitting a reason calls `_start_pending_review_action`, which at lines 2682-2683 first `_commit_pending_review_action()` (silently commits the pending Approve), then queues the Reject. Net: the undo window was bypassed and both review actions fired. Expected: while `_pending_review_action` is active, the morphed state (Undo label + opposite-button disabled) must survive re-render.

**Proposed fix:** In _render_selected_task gate the reset: `if self._pending_review_action is None: <existing reset>` else re-apply _morph_review_action_buttons() so the Undo morph survives re-render. Add a regression test asserting the opposite button stays disabled and the Undo label persists across _apply_detail_result/_background_refresh.

**Verifier note:** Independently confirmed in code by reading every cited line and tracing both re-render paths end to end.

(1) src/pollypm/cockpit_tasks.py:2181-2183 — `_render_selected_task` unconditionally runs `in_review = task.work_status.value == "review"; self.approve_button.disabled = not in_review; self.reject_button.disabled = not in_review`. No `_pending_review_action` guard. It mutates only `.disabled`; it does NOT touch the `.label` or the `-undo` CSS class.

(2) The morph (`_morph_review_action_buttons`, 2705-2729) sets the clicked button to `Undo (Ns)` + adds `-undo`, and DISABLES the opposite bu
# Part B — keyboard Approve/Reject (a/x) unreachable while a search filter is active (P2)

## Keyboard Approve/Reject (a/x) permanently unreachable on the Tasks surface whenever a search filter has text, even with the table focused

**Severity:** P2  |  **Axis:** intuitive  |  **Surface:** Cockpit Tasks surface (PollyTasksApp) — keyboard approve/reject

**Target:** src/pollypm/cockpit_tasks.py:2893

**Observed:** action_approve_task (2893) and action_reject_task (2971) both begin `if self._has_search_filter(): self.action_focus_search(); return`. But _has_search_filter (verified at 2546) returns True whenever the search box holds ANY text (`(self._search_query or "").strip() or self.search_input.value`), NOT only while the Input is focused. After the user types a query and presses Escape/Enter to return focus to the DataTable, the filter text persists, so every subsequent `a`/`x` just bounces focus back into the search box. There is no keyboard path to approve/reject a filtered task short of clearing the search with `c` or clicking the on-screen button. Textual already routes printable keys to a focused Input, so the broad guard isn't needed for the typing case.

**Expected:** With the DataTable focused and a filter applied, `a`/`x` should approve/reject the highlighted filtered task through the existing arming flow. The focus-redirect should only fire when the search Input itself has focus.

**Evidence:** def _has_search_filter(self) -> bool: return bool((self._search_query or "").strip() or self.search_input.value)  # True on any text, not just focus

**Reproduction:** Open cockpit Tasks. Press `/`, type a query matching a review task, press Enter (returns focus to the DataTable while keeping the filter text). Highlight the row, press `a` (or `x`): focus jumps back to the search box and nothing is approved/rejected. The keyboard approve/reject path stays unreachable until the filter is cleared with `c`. (Note: Escape with the search Input focused clears the filter instead — only the Enter-to-table path produces the trap.)

**Proposed fix:** Replace the _has_search_filter() redirect in action_approve_task/action_reject_task with a focus check: only redirect when self.search_input.has_focus. The destructive_action_safe arming gate already protects against accidental single-key approval from the table.

**Verifier note:** Confirmed by code-path trace. _has_search_filter() at src/pollypm/cockpit_tasks.py:2547 returns True on ANY search text (`(self._search_query or "").strip() or self.search_input.value`), independent of focus. action_approve_task (2893-2895) and action_reject_task (2971-2973) both short-circuit `if self._has_search_filter(): self.action_focus_search(); return` BEFORE reaching the arming gate. The trap state is produced by _on_search_submitted (3016-3018): pressing Enter in the search box calls `self.task_table.focus()` but leaves _search_query and search_input.value populated. So with the table
# Part C — inbox row-click handler missing the stale-row guard its highlight sibling has (P3)

## Inbox row-click (ListView.Selected) lacks the stale-row guard the highlight handler has — a click landing across a lens swap can select/render a stale row

**Severity:** P3  |  **Axis:** reliable  |  **Surface:** Cockpit Inbox surface (PollyInboxApp) — list selection

**Target:** src/pollypm/cockpit_ui.py:8624-8632

**Observed:** _on_row_highlighted (cockpit_ui.py:8634) was hardened by #1573 to drop events whose `row.row_ref not in self._visible_rows` (verified line 8647), because a lens swap clears the old ListView children while events queued against the old items can still fire and overwrite the freshly-restored selection. The sibling _on_row_selected (verified 8624-8632) has NO equivalent guard: it reads row.task_id/row.row_ref.key from event.item, writes _selected_task_id/_selected_row_key, and calls _render_detail unconditionally. A Selected event delivered just after a lens swap stamps the selection from an orphaned row no longer in _visible_rows — the inverse of the bug #1573 fixed on the highlight path.

**Expected:** _on_row_selected should apply the same `if row.row_ref not in self._visible_rows: return` guard as _on_row_highlighted, so a click resolved against a stale post-swap row is dropped.

**Evidence:** @on(ListView.Selected, "#inbox-list") def _on_row_selected(...): row = event.item; if not isinstance(row, _InboxListItem): return; self._selected_task_id = row.task_id; ...  # missing the _visible_rows membership check present at _on_row_highlighted:8647

**Reproduction:** Code-path confirmed against current source. The asymmetry is real:

- _on_row_selected (src/pollypm/cockpit_ui.py:8624-8632): reads row=event.item, isinstance(_InboxListItem) check ONLY, then unconditionally writes self._selected_task_id = row.task_id; self._selected_row_key = row.row_ref.key; self._render_detail(row.task_id); self._mark_open_read(row.task_id). No _visible_rows membership check.
- _on_row_highlighted (cockpit_ui.py:8634-8664): same isinstance check PLUS the #1573 guard at line 8647: `if row.row_ref not in self._visible_rows: return`.

Lens-swap mechanism confirmed: _render_list (cockpit_ui.py:6675-6740) does self.list_view.clear(); self._visible_rows = []; then rebuilds fresh _InboxListItem objects wrapping fresh InboxThreadRow refs from build_inbox_thread_rows. _switch_lens (7633-7665) comments (7639, 7651-7655) explicitly document that the clear+append cycle can stir up late-arriving events from the OLD (now-orphaned) row set, which is why the highlight path was hardened. A ListView.Selected event queued against an old _InboxListItem and dispatched after the swap is accepted by _on_row_selected with no membership check — the exact inverse of the bug #1573 fixed on the highlight path.

Refinement to the finding's repro: the strongest trigger is NOT a simultaneous keyboard Enter + L-key (keyboard input is serialized through one message pump). It is a BACKGROUND _render_list refresh (data poll / inbox fanout re-render) that clears+rebuilds the list while a previously-queued Selected event is still in flight — the same async window #1573 documents for Highlighted.

Severity caveat (supports P3, not higher): harm is transient. (1) _render_detail resolves detail by task_id independently, so a stale Selected briefly renders detail for a task no longer in the active lens but does not crash. (2) The very next legitimate _render_list or the _switch_lens re-assert loop (7656-7665) restores correct _selected_row_key/_selected_task_id from the new _visible_rows. (3) InboxThreadRow is @dataclass(frozen=True, slots=True), so `not in self._visible_rows` uses value-equality; when a rebuild reuses the same underlying task objects the orphaned row would actually compare EQUAL to its rebuilt twin and the guard would (correctly) NOT drop it — the guard only drops rows whose logical content truly left the lens, which is the intended behavior and matches the highlight path exactly.

Proposed fix is correct and minimal: add `if row.row_ref not in self._visible_rows: return` immediately after the isinstance check at cockpit_ui.py:8628, mirroring line 8647. One line, mechanical, low-risk, fits RC stability constraints.

**Proposed fix:** Add `if row.row_ref not in self._visible_rows: return` at the top of _on_row_selected (after the isinstance check), mirroring the #1573 guard in _on_row_highlighted.

**Verifier note:** Independently verified by reading the cited handlers and tracing the lens-swap path. The asymmetry is exactly as described: _on_row_selected (8624-8632) has no _visible_rows guard while its sibling _on_row_highlighted (8634+) has the #1573 guard at 8647. The lens-swap clear+rebuild mechanism in _render_list (6675-6740) and the late-stale-event hazard documented in _switch_lens (7639-7655) confirm the race is structurally identical to the one #1573 fixed on the highlight path, so the same defense belongs on the selection path. I down-rated nothing — the finding's own P3 is right because harm is

---

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

## Summary
- Preserve pending review Undo button morph state across task detail re-renders.
- Allow keyboard approve/reject while a search filter is applied, redirecting only when the search input itself is focused.
- Drop stale inbox selected-row events using the same visible-row guard already used by highlight events.

Fixes #2407

## Verification
- `uv run --extra test python -m pytest tests/test_cockpit_issue_2407.py` (4 passed)
- `uv run --extra test python -m pytest tests/test_cockpit_issue_2407.py tests/test_cockpit_interaction_contract.py tests/test_inbox_detail_hydration_off_ui_loop.py` (36 passed)
- `uv run --extra dev ruff check tests/test_cockpit_issue_2407.py src/pollypm/cockpit_tasks.py`

Note: full ruff on `src/pollypm/cockpit_ui.py` is not clean due to unrelated pre-existing import/order/unused-variable findings in that large module.

### DIFF
--- a/src/pollypm/cockpit_tasks.py
+++ b/src/pollypm/cockpit_tasks.py
@@ -2181,6 +2181,8 @@ class PollyTasksApp(App[None]):
         in_review = task.work_status.value == "review"
         self.approve_button.disabled = not in_review
         self.reject_button.disabled = not in_review
+        if self._pending_review_action is not None:
+            self._morph_review_action_buttons()
         self.bulk_approve_button.disabled = len(self._selected_review_task_ids()) == 0
         self.refresh_live_button.disabled = active_session is None
         if active_session is None:
@@ -2722,10 +2724,12 @@ class PollyTasksApp(App[None]):
         if pending.decision == "approve":
             self.approve_button.label = undo_label
             self.approve_button.add_class("-undo")
+            self.approve_button.disabled = False
             self.reject_button.disabled = True
         else:
             self.reject_button.label = undo_label
             self.reject_button.add_class("-undo")
+            self.reject_button.disabled = False
             self.approve_button.disabled = True
 
     def _restore_review_action_buttons(self) -> None:
@@ -2890,7 +2894,7 @@ class PollyTasksApp(App[None]):
         # bench cursor sitting on a real task cannot approve it on a
         # single keystroke. The on-screen Approve button bypasses
         # this gate via :meth:`_fire_approve_task` (#900).
-        if self._has_search_filter():
+        if self.search_input.has_focus:
             self.action_focus_search()
             return
         if not self._selected_task_id:
@@ -2968,7 +2972,7 @@ class PollyTasksApp(App[None]):
     def action_reject_task(self) -> None:
         # #840 fix — see ``action_approve_task`` above. Button-click
         # path bypasses arming via :meth:`_fire_reject_task` (#900).
-        if self._has_search_filter():
+        if self.search_input.has_focus:
             self.action_focus_search()
             return
         if not self._selected_task_id:
--- a/src/pollypm/cockpit_ui.py
+++ b/src/pollypm/cockpit_ui.py
@@ -8626,6 +8626,8 @@ class PollyInboxApp(App[None]):
         row = event.item
         if not isinstance(row, _InboxListItem):
             return
+        if row.row_ref not in self._visible_rows:
+            return
         self._selected_task_id = row.task_id
         self._selected_row_key = row.row_ref.key
         self._render_detail(row.task_id)
--- a/tests/test_cockpit_issue_2407.py
+++ b/tests/test_cockpit_issue_2407.py
@@ -1,0 +1,133 @@
+from __future__ import annotations
+
+from pathlib import Path
+from types import SimpleNamespace
+
+
+def test_pending_approve_undo_morph_survives_task_detail_rerender(
+    monkeypatch,
+) -> None:
+    import pollypm.cockpit_tasks as cockpit_tasks
+
+    app = cockpit_tasks.PollyTasksApp(Path("/tmp/pollypm.toml"), "demo")
+    monkeypatch.setattr(cockpit_tasks, "_render_overview", lambda *a, **k: "overview")
+    monkeypatch.setattr(cockpit_tasks, "_render_context", lambda _task: "context")
+    monkeypatch.setattr(cockpit_tasks, "_render_live", lambda *a, **k: "live")
+    monkeypatch.setattr(cockpit_tasks, "_status_label", lambda *a, **k: "review")
+    monkeypatch.setattr(cockpit_tasks, "_status_display_label", lambda status: status)
+    monkeypatch.setattr(cockpit_tasks, "priority_glyph", lambda _task: "")
+    app._render_timeline = lambda _executions: None  # type: ignore[method-assign]
+
+    task = SimpleNamespace(
+        task_id="demo/1",
+        task_number=1,
+        title="Review me",
+        work_status=SimpleNamespace(value="review"),
+        updated_at=None,
+        created_at=None,
+        executions=[],
+    )
+    app._tasks = [task]
+    app._pending_review_action = cockpit_tasks._PendingReviewAction(
+        task_ids=("demo/1",),
+        task_numbers=(1,),
+        decision="approve",
+        reason=None,
+        deadline=cockpit_tasks.monotonic() + 5.0,
+    )
+
+    app._render_selected_task(task, owner=None, flow=None, active_session=None)
+
+    assert str(app.approve_button.label).startswith("Undo (")
+    assert app.approve_button.has_class("-undo") is True
+    assert app.approve_button.disabled is False
+    assert app.reject_button.disabled is True
+
+
+def test_keyboard_approve_reject_with_search_filter_use_destructive_arming(
+    monkeypatch,
+) -> None:
+    import pollypm.cockpit_tasks as cockpit_tasks
+
+    app = cockpit_tasks.PollyTasksApp(Path("/tmp/pollypm.toml"), "demo")
+    app._selected_task_id = "demo/1"
+    app._search_query = "already filtered"
+    focus_calls: list[None] = []
+    approve_fire_calls: list[None] = []
+    reject_fire_calls: list[None] = []
+    arm_calls: list[tuple[str, str | None]] = []
+
+    app.action_focus_search = lambda: focus_calls.append(None)  # type: ignore[method-assign]
+    app._fire_approve_task = lambda: approve_fire_calls.append(None)  # type: ignore[method-assign]
+    app._fire_reject_task = lambda: reject_fire_calls.append(None)  # type: ignore[method-assign]
+    app._notify_arm_hint = lambda _label: None  # type: ignore[method-assign]
+
+    def fake_destructive_action_safe(_app, action, *, target_id, selection):
+        arm_calls.append((action, target_id))
+        return False
+
+    monkeypatch.setattr(
+        cockpit_tasks,
+        "destructive_action_safe",
+        fake_destructive_action_safe,
+    )
+
+    app.action_approve_task()
+    app.action_reject_task()
+
+    assert focus_calls == []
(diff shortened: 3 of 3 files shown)


samhotchkiss:
**Claude review: APPROVE** (posted as comment — GH blocks self-approval on own PR).

All three sub-fixes verified real, minimal, and mapping to concrete code changes:

- **f6** (Undo morph clobber): cockpit_tasks.py:2184-2185 re-applies `_morph_review_action_buttons()` when `_pending_review_action` is active, after the unconditional disabled-reset at 2182-2183, so the Approve/Reject mutual-exclusion invariant survives the 8s background refresh and the async detail-load worker. The `disabled=False` adds at 2727/2732 keep the active button clickable for undo.
- **f7** (a/x unreachable under filter): both `action_approve_task` and `action_reject_task` swap the `_has_search_filter()` short-circuit (true on any filter text) for `self.search_input.has_focus`, so keyboard approve/reject reach the destructive-arming gate once focus returns to the table.
- **f12** (inbox stale-row guard): `_on_row_selected` gains `if row.row_ref not in self._visible_rows: return`, mirroring the #1573 guard on its highlight sibling.

Tests (foreground): new tests/test_cockpit_issue_2407.py — 4 passed; test_cockpit_inbox_threads + test_cockpit_interaction_contract + test_cockpit_task_priority — 30 passed. No regressions. 2 commits behind main, no conflicts. Merging via squash.

PR #2415, opened 2026-05-29
Fix recovery cascade stall guards

### LINKED ISSUE #2406: Recovery-cascade behavior gaps: dead 'transient' bucket, proceed-signal bypasses circuit-breaker, cancellation_no_promotion re-fires
A cluster of recovery-cascade / heartbeat behavior gaps — extends the 'self-heal wired but not engaging' theme (cf #2396). Detection or protection logic that is inert, bypassed, or re-fires.

# Part A — stall classifier 'transient' bucket is dead code (false-positive protection inert) (P2)

## Stall classifier's `transient` bucket is dead code — neither StallContext construction site sets recently_nudged/turn_in_flight, so just-nudged/mid-turn workers get immediate suspected_loop false-positives

**Severity:** P2  |  **Axis:** intuitive  |  **Surface:** heartbeat stall classification (recovery cascade)

**Target:** src/pollypm/heartbeats/stall_classifier.py:186-188

**Observed:** The only path returning 'transient' is `if ctx.recently_nudged or ctx.turn_in_flight: return "transient"` (stall_classifier.py:186-188). Both fields default to False in the dataclass. Both StallContext construction sites — _handle_same_snapshot_stall (local.py:1445-1451, verified) and supervisor_alerts.py:242-254 — omit both fields, so neither is ever set True. The transient branch is therefore unreachable. A worker the heartbeat nudged on the previous tick, or that is mid-turn (long thinking pause), falls straight through to 'unrecoverable_stall' and earns a suspected_loop toast at 3 identical snapshots — the explicit false-positive protection documented in stall_classifier.py:80-93 is inert.

**Expected:** A worker nudged last cycle, or mid-turn, should classify as 'transient' and defer the alert one tick, per the documented design intent. The data to compute recently_nudged already exists (the nudge/send_input events _nudge_stalled_worker records and the 120s send_input debounce reads).

**Evidence:** stall_ctx = StallContext(role=..., session_name=..., has_pending_work=..., pane_is_idle_placeholder=..., awaiting_operator_question=...)  # recently_nudged / turn_in_flight never passed -> always False -> 'transient' branch (classifier:186) unreachable

**Reproduction:** Verified statically. Two-part reproduction of the live false-positive: (1) turn_in_flight case — a worker with queued work is mid-turn (long thinking pause) so its pane snapshot is stable; at 3 identical heartbeat snapshots _handle_same_snapshot_stall (local.py:842) builds StallContext with turn_in_flight=False (default), classify_stall returns unrecoverable_stall (stall_classifier.py:188), and a suspected_loop toast is emitted (local.py:1489) even though the docstring (lines 88-93) says this should bucket as transient and defer. (2) recently_nudged case — after _triage_stalled_worker nudges the worker at 5 identical snapshots (local.py:1509-1511, recording a send_input/nudge event), the next still-identical tick re-runs _handle_same_snapshot_stall with recently_nudged=False (default) and re-fires the toast instead of returning transient. Fix: populate recently_nudged at both construction sites (local.py:1445, supervisor_alerts.py:242) from the recent nudge/send_input events already queried in _nudge_stalled_worker (local.py:1872-1904), and add a turn_in_flight heuristic (transcript/pane mid-turn signal) before calling classify_stall.

**Proposed fix:** In _handle_same_snapshot_stall (and the supervisor_alerts.py mirror), populate recently_nudged from the recent nudge/send_input events (same query _nudge_stalled_worker runs) and turn_in_flight from the mid-turn/transcript heuristics, before calling classify_stall.

**Verifier note:** Core claim independently confirmed in code. stall_classifier.py:186-188 is the only path returning "transient", gated on `ctx.recently_nudged or ctx.turn_in_flight`; both fields default to False (lines 87, 93). Grep across all of src/ shows these two fields appear ONLY in stall_classifier.py — the dataclass defaults and the single read at line 186; they are never assigned True anywhere. I read both construction sites and confirmed neither passes them: local.py _handle_same_snapshot_stall (1445-1451) sets role/session_name/has_pending_work/pane_is_idle_placeholder/awaiting_operator_question onl
# Part B — _triage_stalled_worker proceed-signal bypasses the nudge circuit-breaker (P2)

## `_triage_stalled_worker` proceed-signal branch bypasses the nudge circuit-breaker — re-injects "Yes, proceed" every tick into a frozen worker forever

**Severity:** P2  |  **Axis:** reliable  |  **Surface:** heartbeat stalled-worker triage (recovery ladder)

**Target:** src/pollypm/heartbeats/local.py:2021-2045

**Observed:** _triage_stalled_worker is invoked from _handle_same_snapshot_stall whenever 5 consecutive snapshot hashes are identical, with NO per-snapshot dedup guard — so it re-runs every heartbeat tick while the pane stays frozen. Its proceed_signals branch (verified, local.py:2021-2027) calls _send_worker_message(... 'Yes, proceed...') directly. _send_worker_message -> api.send_session_message has only a pause-skip gate, NOT the rate-limit / 120s debounce / _MAX_NUDGES_BEFORE_RECOVERY circuit breaker that _nudge_stalled_worker enforces. The sibling next_step_signals (line 2050) and error (2055) branches correctly route through _nudge_stalled_worker; only the proceed branch is ungated and never escalates.

**Expected:** If the injected 'Yes, proceed' has no effect (pane frozen, prompt doesn't consume the literal text), the triage should debounce and eventually escalate to _recover_session — exactly what _nudge_stalled_worker's circuit breaker does after _MAX_NUDGES_BEFORE_RECOVERY attempts. A stuck pane should not receive the same keystroke every tick indefinitely.

**Evidence:** if any(sig in lowered for sig in proceed_signals): self._send_worker_message(api, context, "Yes, proceed...", owner="heartbeat") ... return  # no debounce, no circuit breaker, no escalation; re-runs every 5-identical tick. Sibling branches call self._nudge_stalled_worker(...)

**Reproduction:** In src/pollypm/heartbeats/local.py, `_triage_stalled_worker` proceed branch (2020-2045): when a stalled worker pane matches a proceed signal (e.g. "shall i", "should i"), it calls `_send_worker_message(... "Yes, proceed...")` and returns with no breaker. Confirm asymmetry vs siblings at 2050/2055 which call `_nudge_stalled_worker`. Confirm `_send_worker_message`->`send_session_message` (api.py:333-352) has only the skip_if_paused gate. Confirm `_nudge_stalled_worker` (1832-1961) has the 120s send_input debounce, snapshot_hash dedup, and `_MAX_NUDGES_BEFORE_RECOVERY=6`->`_recover_session` that the proceed branch lacks. Triggered from `_handle_same_snapshot_stall` line 1508-1511 (5 identical hashes) and intervention dispatch line 1016. A frozen worker whose pane re-stabilizes to an identical hash keeps receiving "Yes, proceed" with no escalation to recovery. Fix: route the proceed branch through `_nudge_stalled_worker` or add an equivalent send_input debounce + circuit breaker + per-snapshot_hash dedup mirroring the same_episode guards.

**Proposed fix:** Route the proceed_signals branch through _nudge_stalled_worker (or add an equivalent send_input debounce + circuit breaker), and/or add a per-snapshot_hash dedup so triage doesn't re-fire while the pane hash is unchanged (mirror the same_episode guards in _nudge_stalled_worker).

**Verifier note:** Independently confirmed in the cited code. The proceed branch of `_triage_stalled_worker` (src/pollypm/heartbeats/local.py:2021-2045) calls `_send_worker_message(... "Yes, proceed. Do the next step you outlined." ...)` and immediately returns, with no dedup, no rate-limit, no circuit breaker, and no escalation. The two sibling branches DO route through the breaker: `next_step_signals` (line 2050) and the error branch (line 2055) both call `_nudge_stalled_worker`. So the asymmetry the finding describes is real.

`_nudge_stalled_worker` (1832-1961) is the one with the circuit breaker: a 120s `se
# Part C — cancellation_no_promotion re-fires ~5x per intentional cancellation (detector/brief mismatch) (P2)

## cancellation_no_promotion: detector clear-condition (later task.created) and the brief's accepted resolution (park via pm task context) are mismatched, so the rule re-fires every tick after an intentional cancellation

**Severity:** P2  |  **Axis:** reliable  |  **Surface:** watchdog rule: cancellation_no_promotion

**Target:** src/pollypm/audit/watchdog.py:1033-1058

**Observed:** _detect_cancellation_no_promotion clears a finding ONLY when a task.created exists with a timestamp strictly later than the cancellation (`followed = any(c > ts for c in creates)`). The cancellation timestamp is fixed; the detector re-scans the whole window_seconds window each 5-min tick, so the same finding re-fires every tick until the cancellation ages out. But the dispatched brief (_brief_cancellation_no_promotion) tells the architect to 'decide whether this cancellation ended the project intentionally' and offers `pm task context <subject> "parked intentionally"` as a valid resolution. Parking intentionally emits NO task.created, so an architect who correctly follows the brief never clears the detector. Live: this rule fired 1,976 times in 7d while only 367 task.created events occurred across ALL projects.

**Expected:** Detection and the brief's accepted actions must agree on what clears the finding. Either the detector should treat an explicit park/intentional-cancel signal (a task.context event/marker, or a status_changed to a terminal-by-design state) as clearing, or the brief should not offer a resolution the detector cannot observe.

**Evidence:** followed = any(c > ts for c in creates); if followed: continue  # only a later task.created clears -- but brief offers 'park intentionally with `pm task context {subject} ...`' which emits no task.created

**Reproduction:** Read watchdog.py:1033-1039 (clear=only later task.created) vs watchdog.py:3987-3993 (brief offers `pm task context ... "parked intentionally"`). Trace `pm task context` -> work/cli.py:1868 svc.add_context -> pg_service.py:3975-4008: pure INSERT into work_context_entries, emits NO audit event. Therefore the brief's park resolution is structurally unobservable to the detector. Confirm re-firing in live logs: `grep -rh cancellation_no_promotion ~/.pollypm/audit/*.jsonl | wc -l` = 1976; grouping by (subject, cancelled_at) = 385 distinct cancellations, top offenders re-fired 15x across scan ticks. Note: upsert_alert dedup (watchdog.py:524-533) and dispatch_dedup_hash throttle (watchdog.py:3312) collapse the operator-facing alert/dispatch, so 1976 is audit-log volume, not operator pings. Fix: have the detector also clear when a status_changed to a terminal-by-design state (or a tagged park marker) lands after the cancellation, OR narrow the brief to only offer the observable create+queue resolution.

**Proposed fix:** Add an intentional-park clearing signal: have the detector also suppress when a task.context event tagged park/intentional (or a status_changed to a terminal-by-design state) lands after the cancellation, matching the brief's offered resolution. Alternatively make the brief's only accepted action a concrete `pm task create ...; pm task queue ...` that emits task.created.

**Verifier note:** Independently confirmed every load-bearing claim by reading the cited code and the live audit logs.

1) Detector clear condition — watchdog.py:1033-1039: for each cancellation, `creates = created_by_project.get(ev.project, [])` then `followed = any(c > ts for c in creates); if followed: continue`. `creates` is populated ONLY from `EVENT_TASK_CREATED` (line 1026). So the sole clear signal is a later task.created. No open_tasks, project-status, or park awareness — confirmed the detector is called at watchdog.py:2954-2956 with only (materialised, now, config).

2) Brief recommends park-via-contex
# Part D — has_pending_work_for_session docstring wrong re fail-direction (P3)

## `has_pending_work_for_session` config-resolution failures fail toward legitimate_idle, contradicting its own "return True when in doubt" docstring

**Severity:** P3  |  **Axis:** reliable  |  **Surface:** stall classifier work-presence probe

**Target:** src/pollypm/heartbeats/stall_classifier.py:213-260

**Observed:** The docstring (209-211) states 'Returns True when in doubt — ... a false no would silence a real stall.' Only the outer except (259-260) returns True. The intermediate early returns — session not found (216), no project_key (219), project not found (223), no project_path (227) — all return False, which makes classify_stall treat the session as legitimate_idle and suppress stall detection. These are config/state-resolution gaps, distinct from 'the queue is genuinely empty' (the legitimate final return False at 258), and contradict the function's own documented bias.

**Expected:** Per the function's own docstring contract, config-resolution failures should fall through to the conservative True (like the outer except), or the docstring should be corrected to state that resolution failures fail toward False.

**Evidence:** docstring: 'Returns True when in doubt ... a false "no" would silence a real stall.' vs `if project is None: return False  # fails toward no-work/silence`

**Reproduction:** Code-path only (the live config-reload race is NOT achievable — config reload is an atomic snapshot swap at supervisor.py:1157-1158, no half-populated dict window). The genuine, reachable case: a misconfigured/orphaned session row in config.sessions whose `project` is empty, or points at a project key absent from config.projects, or whose project has no `path`. In any of those, has_pending_work_for_session returns False (lines 217/220/224/227) rather than True, which the docstring (209-211) promises for "in doubt." This feeds classify_stall -> "legitimate_idle" (lines 178-179) for that worker tick. Recommended fix: correct the docstring to state that config/state-resolution failures fail toward False (only the queue-query path itself fails toward True via the outer except), since returning True for an unresolvable session would only trigger a no-op same-snapshot recheck and cannot drive useful remediation.

**Proposed fix:** Either correct the docstring to state resolution failures fail toward False, or change the intermediate returns to fall through to True like the outer except.

**Verifier note:** Documentation/contract bug confirmed; the reproducible behavioral race is refuted. Best resolved by correcting the docstring (small mechanical change), not by flipping the returns.

---

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

## Summary
- Populate the stall classifier transient guards from recent `nudge`/`send_input` events and observable output progress at both heartbeat construction sites.
- Route proceed-signal triage through `_nudge_stalled_worker` so debounce, same-snapshot dedupe, escalation, and recovery guards apply.
- Align the cancellation-no-promotion watchdog brief with the detector by removing the unobservable intentional-park resolution, and correct the pending-work docstring.

## Verification
- `uv run --extra test pytest tests/test_stall_classifier.py tests/test_heartbeat_plugin_api.py::test_same_snapshot_recent_nudge_classifies_transient tests/test_heartbeat_plugin_api.py::test_same_snapshot_transcript_delta_classifies_transient tests/test_heartbeat_plugin_api.py::test_triage_proceed_signal_uses_nudge_ladder tests/test_supervisor_alerts.py::test_supervisor_alert_helper_updates_and_nudges tests/test_supervisor_alerts.py::test_supervisor_alert_helper_recent_nudge_is_transient tests/test_supervisor_alerts.py::test_supervisor_alert_helper_log_growth_is_transient tests/test_audit_watchdog.py::test_format_unstick_brief_cancellation_no_promotion_is_decisive`
- `python -m py_compile src/pollypm/heartbeats/stall_classifier.py src/pollypm/heartbeats/local.py src/pollypm/supervisor_alerts.py src/pollypm/audit/watchdog.py tests/test_stall_classifier.py tests/test_heartbeat_plugin_api.py tests/test_supervisor_alerts.py tests/test_audit_watchdog.py`
- `git diff --check`
- `uv run --extra dev ruff check --select F821,F822,F823,F541 src/pollypm/heartbeats/stall_classifier.py src/pollypm/heartbeats/local.py src/pollypm/supervisor_alerts.py src/pollypm/audit/watchdog.py tests/test_stall_classifier.py tests/test_heartbeat_plugin_api.py tests/test_supervisor_alerts.py tests/test_audit_watchdog.py`

Note: a full ruff pass over these whole files still reports unrelated pre-existing findings in `src/pollypm/heartbeats/local.py`, `tests/test_heartbeat_plugin_api.py`, and `tests/test_supervisor_alerts.py`; the targeted correctness lint above passed.

Fixes #2406

### DIFF
--- a/src/pollypm/audit/watchdog.py
+++ b/src/pollypm/audit/watchdog.py
@@ -36,9 +36,8 @@ Detection rules (each returns a list of :class:`Finding`):
    surface to the user via the empty-state affordance from #1340.)
 4. ``cancellation_no_promotion`` — ``task.status_changed`` to
    ``cancelled`` with no ``task.created`` for the same project in the
-   following ``cancel_grace_seconds`` window. Indicates a planning
-   stall after a self-cancel (savethenovel/1's class — a worker
-   self-cancels but Polly never queues a follow-up).
+   following ``cancel_grace_seconds`` window. Indicates observable
+   replacement work was not created after a self-cancel.
 5. ``task_progress_stale`` — a live task currently at ``in_progress``
    whose worker has produced no transition / heartbeat / task-context
    activity for longer than ``progress_stale_seconds``. Catches the
@@ -4154,15 +4153,12 @@ def _brief_cancellation_no_promotion(
     lines.append("")
     lines.append(
         "Your job: decide whether this cancellation ended the project "
-        "intentionally or left replacement work unqueued. Do not reply "
-        "with analysis alone; make the queue state explicit."
+        "or left replacement work unqueued. Do not reply with analysis "
+        "alone; make the queue state explicit."
     )
     lines.append(
-        f"Cli levers available: create and queue replacement work with "
-        f"`pm task create ...` then `pm task queue <replacement-task-id>`, "
-        f"or intentionally park the abandoned thread with "
-        f"`pm task context {subject} \"parked intentionally: <why>\"` "
-        f"plus cancel/hold for any remaining scoped task."
+        "Cli levers available: create and queue replacement work with "
+        "`pm task create ...` then `pm task queue <replacement-task-id>`."
     )
     return lines
 
--- a/src/pollypm/heartbeats/local.py
+++ b/src/pollypm/heartbeats/local.py
@@ -1435,6 +1435,7 @@ class LocalHeartbeatBackend(HeartbeatBackend):
         from pollypm.heartbeats.stall_classifier import (
             StallContext,
             classify_stall,
+            recently_nudged_from_message_store,
         )
         from pollypm.idle_placeholders import (
             pane_ends_with_unanswered_question as _pane_ends_with_unanswered_question,
@@ -1442,10 +1443,16 @@ class LocalHeartbeatBackend(HeartbeatBackend):
         )
 
         _pane_text = context.pane_text or ""
+        output_advanced = bool((context.transcript_delta or "").strip())
         stall_ctx = StallContext(
             role=context.role or "",
             session_name=context.session_name,
             has_pending_work=self._has_pending_work(api, context),
+            recently_nudged=recently_nudged_from_message_store(
+                getattr(getattr(api, "supervisor", None), "msg_store", None),
+                context.session_name,
+            ),
+            turn_in_flight=output_advanced,
             pane_is_idle_placeholder=_pane_is_idle_placeholder(_pane_text),
             awaiting_operator_question=_pane_ends_with_unanswered_question(_pane_text),
         )
@@ -1829,7 +1836,13 @@ class LocalHeartbeatBackend(HeartbeatBackend):
                     break
         return repeated
 
-    def _nudge_stalled_worker(self, api, context: HeartbeatSessionContext) -> None:
+    def _nudge_stalled_worker(
+        self,
+        api,
+        context: HeartbeatSessionContext,
+        *,
+        message: str | None = None,
+    ) -> None:
         """Send a targeted nudge to a stalled WORKER. Never targets the operator."""
         if context.role != "worker":
             return
@@ -1952,12 +1965,13 @@ class LocalHeartbeatBackend(HeartbeatBackend):
             )
         # Context-aware nudge for the worker
         snippet = (context.pane_text or "").strip().splitlines()[-1][:80] if context.pane_text else ""
-        if "permission" in snippet.lower() or "approve" in snippet.lower():
-            message = "You appear stuck on a permissions prompt. Accept or work around it."
-        elif "error" in snippet.lower() or "failed" in snippet.lower():
-            message = "You hit an error. Read it carefully, fix the root cause, and continue."
-        else:
-            message = "State the remaining task in one sentence, execute the next step, and report."
+        if message is None:
+            if "permission" in snippet.lower() or "approve" in snippet.lower():
+                message = "You appear stuck on a permissions prompt. Accept or work around it."
+            elif "error" in snippet.lower() or "failed" in snippet.lower():
+                message = "You hit an error. Read it carefully, fix the root cause, and continue."
+            else:
+                message = "State the remaining task in one sentence, execute the next step, and report."
         self._send_worker_message(api, context, message, owner="heartbeat")
         try:
             from pollypm.events.summaries import (
@@ -2019,28 +2033,10 @@ class LocalHeartbeatBackend(HeartbeatBackend):
         # Worker asking for permission → push forward
         proceed_signals = ["if you want", "shall i", "should i", "want me to", "i can do"]
         if any(sig in lowered for sig in proceed_signals):
-            self._send_worker_message(
+            self._nudge_stalled_worker(
                 api,
                 context,
-                "Yes, proceed. Do the next step you outlined.",
-                owner="heartbeat",
-            )
-            from pollypm.events.summaries import (
-                activity_summary,
-            )
-
-            api.supervisor.msg_store.append_event(
-                scope=context.session_name,
-                sender=context.session_name,
-                subject="heuristic_triage",
-                payload={
-                    "message": activity_summary(
-                        summary="Pushed forward: proceed signal detected",
-                        severity="routine",
-                        verb="triaged",
-                        subject=context.session_name,
-                    ),
-                },
+                message="Yes, proceed. Do the next step you outlined.",
             )
             return
 
--- a/src/pollypm/heartbeats/stall_classifier.py
+++ b/src/pollypm/heartbeats/stall_classifier.py
@@ -35,6 +35,7 @@ a false-positive stall alert is user trust damage that persists.
 from __future__ import annotations
 
 from dataclasses import dataclass
+from datetime import UTC, datetime
 from typing import Literal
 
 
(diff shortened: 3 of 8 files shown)


samhotchkiss:
## Claude Reviewer — APPROVE

Verified all four sub-fixes of #2406 against the code (not just the diff):

- **f4 (transient dead code) — FULLY FIXED.** `classify_stall` already routed `recently_nudged | turn_in_flight → transient` (stall_classifier.py:187) and consumers already suppressed remediation on non-`unrecoverable_stall` (local.py:1489); the bug was that nothing populated the fields. New `recently_nudged_from_message_store()` reads durable `nudge`/`send_input` events and is wired at both construction sites (heartbeat backend + supervisor boundary), with `turn_in_flight` from transcript-delta / log-byte growth. The bucket is now reachable.
- **f8 (proceed-signal bypasses breaker) — FULLY FIXED.** Proceed branch now routes through `_nudge_stalled_worker(..., message=...)` so debounce, same-snapshot dedupe, `nudge_escalated` escalation, and the MAX-nudges recovery breaker all apply.
- **f5 (cancellation_no_promotion re-fires) — FULLY FIXED.** Removed the unobservable "intentionally park" lever from the brief; the only offered action (`pm task create` → `pm task queue`) now matches the detector's clear-condition (a later `task.created`), so an intentional cancellation no longer re-fires every sweep. Rule docstring aligned.
- **f13 (docstring) — FULLY FIXED.** `has_pending_work_for_session` docstring now accurately describes fail-closed-on-resolution behavior; old "True when in doubt" was wrong.

**Tests:** Ran touched files foreground. The PR's f5 change updated the assertion at test_audit_watchdog.py:847 but left an identical stale `"intentionally park" in brief` in `test_cadence_handler_dispatches_cancellation_no_promotion` — pushed a one-line fix flipping it to `not in`. All 4 sub-fix regression sets + both f5 brief tests pass (28/28). Remaining failures in the full run (`test_recovery_alert_auto_clear_*`, `*_snapshot_learnings`, `collect_work_service_signals`) reproduce identically on origin/main (`PgStore` has no `_write_engine` test-env issue) — pre-existing, not blocking.

Merging.

PR #2416, opened 2026-05-29
Embed build git SHA for deploy staleness

### LINKED ISSUE #2411: Deploy-staleness guard can't detect reinstall-from-stale-source: served_git_sha always null (follow-up to #2408)
Follow-up to #2395 / #2408 (deploy-staleness guard). The guard landed and the `/api/v1/health` `build` introspection block is genuinely useful — but its staleness *detection* has a gap that I hit live during this review pass.

## The gap

`_infer_stale()` (`src/pollypm/deploy_info.py:272`) has three paths:
1. `source_git_sha == served_git_sha` → the robust comparison.
2. `direct_url_editable is True` → fresh.
3. fallback: `source_git_commit_time` vs `package_mtime`.

**`served_git_sha` is always `null` for the production install path.** A `uv tool install --from .` records a PEP 610 `direct_url.json` of `{"url": "file:///Users/sam/dev/pollypm"}` with **no `vcs_info` / commit id** (it's a local-dir install, not a VCS install). So `served_git_sha` can never be populated, path (1) is never taken, and detection always falls to the mtime heuristic (3).

## Why the mtime heuristic is insufficient

Path (3) flags stale only when `source_commit_time > package_mtime`. That catches the **common** case (operator forgot to reinstall → install mtime predates the merges). It does **not** catch "reinstalled recently from stale source": if the package was *built a moment ago* from an *old commit*, `package_mtime` is new, so `stale=false` even though the served content lags HEAD.

**Reproduced live (2026-05-29):** the daily-driver was reinstalled from a detached HEAD on an old branch lacking two merged PRs (#2402/#2403). `/api/v1/health` reported `build.stale: false` and `served_git_sha: null`, yet the served endpoints genuinely lacked those fixes (`GET /tasks?offset=5` returned 200 instead of the merged 400; `updated_at` was still `-06:00` instead of the merged UTC-`Z`). The guard could not see it.

## Proposed fix

Populate `served_git_sha` at build time so path (1) becomes the primary detector:
- a build hook (or `setuptools_scm`) writes the built commit sha into the package (e.g. `pollypm/_build_sha.txt` or a `_build.py` constant) during `uv tool install`;
- `runtime_build_info()` reads it into `served_git_sha`;
- then `stale = (served_git_sha != source_git_sha)` catches *all* content-staleness regardless of mtime.

Keep the mtime fallback for installs that lack the embedded sha. Severity P3 (the common forgot-to-reinstall case is already covered; this hardens the rarer reinstall-from-stale-source case) — but it directly determines whether the guard fulfills the #2395 P0 intent.

---
*Filed from the 2026-05-29 Claude review pass (Opus 4.8). Verified live against the :8765 daily-driver and `src/pollypm/deploy_info.py` at HEAD `7f1168a4`.*

samhotchkiss:
## Agent Identity
- Authoring agent: Codex
- Self-merge prohibited: yes

## Summary
- Add a setuptools `build_py` hook that writes `pollypm/_build_info.json` with the source commit SHA and commit time when building from a git checkout.
- Include that build metadata as package data and teach `runtime_build_info()` to prefer the embedded served SHA before falling back to PEP 610 or package mtimes.
- Add deploy-info regression coverage for embedded served SHAs, stale SHA comparison, and the existing mtime fallback.

## Verification
- `uv run --extra test pytest tests/test_deploy_info.py`
- `uv run --extra dev ruff check setup.py src/pollypm/deploy_info.py tests/test_deploy_info.py`
- `python -m py_compile setup.py src/pollypm/deploy_info.py tests/test_deploy_info.py`
- `git diff --check`
- `uv build --wheel --out-dir "$tmpdir"` plus wheel inspection confirming `pollypm/_build_info.json` exists and `git_sha` equals branch HEAD `63724d8ad580909faa3dc14955745d9ea05e5c7f`.

Fixes #2411

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,6 +88,7 @@ where = ["src"]
 
 [tool.setuptools.package-data]
 pollypm = [
+  "_build_info.json",
   "model_registry.toml",
   "plugins_builtin/*/pollypm-plugin.toml",
   "plugins_builtin/*/profiles/*.md",
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,59 @@
+"""Setuptools build customizations for PollyPM."""
+
+from __future__ import annotations
+
+import json
+import subprocess
+from pathlib import Path
+
+from setuptools import setup
+from setuptools.command.build_py import build_py
+
+
+BUILD_INFO_FILENAME = "_build_info.json"
+
+
+def _git_value(repo: Path, args: list[str]) -> str | None:
+    try:
+        result = subprocess.run(
+            ["git", "-C", str(repo), *args],
+            capture_output=True,
+            text=True,
+            timeout=2.0,
+            check=False,
+        )
+    except (FileNotFoundError, subprocess.TimeoutExpired, OSError):
+        return None
+    if result.returncode != 0:
+        return None
+    value = result.stdout.strip()
+    return value or None
+
+
+class build_py_with_build_info(build_py):
+    """Embed source git metadata into built wheels from local checkouts."""
+
+    def run(self) -> None:
+        super().run()
+        repo = Path(__file__).resolve().parent
+        target = Path(self.build_lib) / "pollypm" / BUILD_INFO_FILENAME
+        git_sha = _git_value(repo, ["rev-parse", "HEAD"])
+        if not git_sha:
+            try:
+                target.unlink()
+            except FileNotFoundError:
+                pass
+            return
+        build_info: dict[str, str] = {"git_sha": git_sha}
+        commit_time = _git_value(repo, ["show", "-s", "--format=%cI", "HEAD"])
+        if commit_time:
+            build_info["git_commit_time"] = commit_time
+
+        target.parent.mkdir(parents=True, exist_ok=True)
+        target.write_text(
+            json.dumps(build_info, sort_keys=True) + "\n",
+            encoding="utf-8",
+        )
+
+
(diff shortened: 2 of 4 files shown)


samhotchkiss:
APPROVED (Claude review pass) — verified end-to-end. Posting as comment since GitHub blocks self-approval on this author's PR.

**Sha embedding works in a real build:** `uv build --wheel` produces a wheel containing `pollypm/_build_info.json` = {"git_sha":"63724d8ad…", "git_commit_time":…} matching git HEAD. The `setup.py` `build_py` hook shells `git rev-parse HEAD` and gracefully no-ops (deletes stale file) when git is unavailable, so sdist-from-tarball builds don't crash.

**Production install path fixed:** installed the built wheel into a clean venv (NOT a git repo — the exact PEP-610 local-dir path #2411 reported as broken). `runtime_build_info().served_git_sha` is now `63724d8ad…` where it was previously always null. Precedence (live git HEAD → embedded json → direct_url vcs id) is correct: source/editable runs read live HEAD (served==source, no false positive); installed wheels read the embedded sha. When neither is determinable, `_infer_stale` returns (None, None) — undetermined, not false-positive.

**Tests:** `pytest tests/test_deploy_info.py` → 6 passed, incl. new cases for served_git_sha populated, sha-mismatch → stale=True, and mtime fallback. Closes #2411. Squash-merging.

